PR #152, opened 2026-04-16
Add Unstream Dispatch weekly briefing feed

brandonlucasgreen:
Scaffolds a weekly music industry briefing published as an RSS feed at /dispatch.xml. Dispatches live as markdown files in data/dispatch/ with frontmatter; the build generates dispatch.xml from them via scripts/generate-dispatch-feed.ts. Designed to be fed by a Claude Code scheduled routine that opens a PR each Friday with the week's dispatch.

Includes the prompt the routine runs (data/dispatch/PROMPT.md), pulling its voice and research scope from the Gail agent definition.

### DIFF
--- a/apps/web/public/dispatch.xml
+++ b/apps/web/public/dispatch.xml
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<rss version="2.0" xmlns:content="http://purl.org/rss/1.0/modules/content/" xmlns:atom="http://www.w3.org/2005/Atom">
+  <channel>
+    <title>The Unstream Dispatch</title>
+    <link>https://unstream.stream</link>
+    <atom:link href="https://unstream.stream/dispatch.xml" rel="self" type="application/rss+xml" />
+    <description>Weekly music industry intelligence — platform news, streaming economics, and what it means for independent artists. Written by Unstream.</description>
+    <language>en-us</language>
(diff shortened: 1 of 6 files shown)


brandonlucasgreen:
**Security & regression review — approved for merge** ✓

Two issues were found and fixed in a follow-up commit (`21acddc`):

**1. CDATA injection (plausible risk, medium) — fixed**
The `content:encoded` field embedded `marked`-generated HTML directly inside `<![CDATA[...]]>` with no escaping of the CDATA terminator sequence `]]>`. Any dispatch body containing that string (e.g. in a code block about RSS/XML) would produce malformed XML and silently break the entire feed for all RSS readers. Fixed by replacing `]]>` with `]]]]><CDATA[>` in the HTML before embedding.

**2. Invalid date silently emitted in pubDate (plausible risk, low) — fixed**
A malformed `published` frontmatter value (e.g. `2026-W16` instead of `2026-04-17`) caused `new Date().toUTCString()` to emit the literal string `"Invalid Date"` into the `<pubDate>` element, which RSS readers reject. Fixed by throwing early with a clear error message, so a bad date fails the build loudly rather than going to production silently.

**No other issues found:**
- No auth/authorization surface (pure build-time script, no network I/O)
- No secrets or user input paths
- All XML attribute and element values correctly escaped via `escapeXml`
- Empty dispatch directory handled gracefully (valid empty-channel feed)
- No database migrations required — fully file-based pipeline

PR is approved for merge.

---
_Generated by [Claude Code_

PR #157, opened 2026-04-17
Fix admin verification approval

brandonlucasgreen:
## Summary

- **Primary bug:** admin approval was failing with "Artist profile not found for this request" for every pending request (e.g. Tye Newton). Root cause: `admin-verify.ts` looked up `artist_profiles` by a non-existent `artist_profile_id` column. The real link is `verification_requests.artist_id` → `artists.id` ← `artist_profiles.artist_id`. Fixed.
- **Missing artist in admin UI:** GET `/api/admin/verify` now joins `artists(name, slug)` and flattens, so the admin UI displays the artist name and slug on pending requests (previously blank).
- **Missing-profile edge case:** approve now handles the case where no `artist_profiles` row exists (user submitted a manual review without completing the automated link-back flow). A row is created with `verified_at` set and the user_id/email carried over from the request.
- **Migration 010:** drops `NOT NULL` on `artist_profiles.website_url` — vestigial from the old verification-code flow, parallels migration 008's same change for `verification_code`. **Already applied to prod.**

## From security review

- **CORS tightening:** replaced hardcoded `Access-Control-Allow-Origin: *` with `buildCorsHeaders()` from middleware, which restricts to `https://unstream.stream` for non-API-key requests. Defense-in-depth — admin JWT remains the actual gate.
- **UUID validation on `requestId`:** rejects malformed values with a 400 before they hit the DB (previously fell through to a Postgres type error → 404).

## Not addressed in this PR (tracked separately)

- **Manual-review ownership gap:** `claim-artist.ts`'s `request-manual-review` action lets any authenticated user file a verification request for any unclaimed artist, and admin approval then assigns them ownership via `request.user_id`. Admin is the human gate but rubber-stamping = artist hijack. Tracked in Things; fix belongs in `claim-artist.ts`, not here.
- **Approval-step atomicity:** three sequential DB ops (profile, artists, verification_requests) with no transaction. Current order is intentionally kept because failure mid-way leaves the request `pending`, which allows the admin to simply re-click Approve (all operations are idempotent). Inverting the order would require manual SQL recovery on failure. Low-priority given low approval volume.

## Test plan

- [ ] Deploy to prod, visit `/admin/verify` — Tye Newton's request shows artist name and `/tye-newton` slug.
- [ ] Approve Tye Newton's request: response 200, `verification_requests.status = 'approved'`, `artist_profiles.verified_at` set, `artists.match_confidence = 'claimed'`.
- [ ] Have Tye log in and confirm `/artist-edit/tye-newton` works.
- [ ] Smoke test: call `/api/admin/verify` POST with a non-UUID `requestId` → expect 400.

### DIFF
--- a/api/functions/admin-verify.ts
+++ b/api/functions/admin-verify.ts
@@ -4,16 +4,19 @@
 import { getClient } from './db';
 import { authenticateAdmin, buildCorsHeaders } from './middleware';
 
+const UUID_REGEX = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
+
 export async function handler(event: {
   httpMethod: string;
   headers: Record<string, string | undefined>;
   body?: string;
 }) {
-  if (event.httpMethod === 'OPTIONS') {
-    return { statusCode: 204, headers: buildCorsHeaders(event.headers['origin'] || event.headers['Origin'], false), body: '' };
-  }
+  const origin = event.headers['origin'] || event.headers['Origin'];
+  const CORS_HEADERS = buildCorsHeaders(origin, false);
 
-  const CORS_HEADERS = buildCorsHeaders(event.headers['origin'] || event.headers['Origin'], false);
+  if (event.httpMethod === 'OPTIONS') {
+    return { statusCode: 204, headers: CORS_HEADERS, body: '' };
+  }
 
   const admin = await authenticateAdmin(event.headers['authorization'] || event.headers['Authorization'] || undefined);
   if (!admin) {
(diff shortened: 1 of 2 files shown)


brandonlucasgreen:
**Security & regression review — approved for merge** ✅

Two issues were found and fixed in a follow-up commit (`f5ae631`) on this branch before approval:

**Fixed — CORS `*` on admin endpoint (medium)**
The file-local `CORS_HEADERS` constant used `Access-Control-Allow-Origin: *`. Replaced with `buildCorsHeaders` from middleware, which restricts the allowed origin to `https://unstream.stream` and adds `X-Content-Type-Options: nosniff`. Admin auth (Supabase JWT + email check) was always the real gate, so this was not exploitable, but origin restriction is correct hygiene for an admin endpoint.

**Fixed — No `reviewerNotes` length cap (low)**
Added a 400 guard for `reviewerNotes > 2000 chars` before any DB access.

**No action needed — migration-010**
`artist_profiles.website_url` is confirmed nullable in prod already. The migration was applied as noted in the PR description.

**Remaining low-severity items (no action required)**
- `requestId` UUID format validation: Supabase returns a graceful 404 for non-UUID values — no exploit path.
- Race condition on concurrent approvals: extremely unlikely for an admin-only queue; unique constraint on `artist_profiles.artist_id` prevents data corruption.
- `user_id`/`artist_id` dropped from GET response: intentional per PR design; covered by manual test plan item #1.

All 165 unit tests pass. Approved for merge.

---

PR #167, opened 2026-04-18
fix: use extensionContext.open() to redirect to app from share sheet

brandonlucasgreen:
The previous responder-chain openURL: hack stopped working on modern iOS
because UIApplication is no longer in the extension's responder chain.
Additionally, calling completeRequest() 0.3 s later returned the user to
the host app (Spotify/Apple Music) rather than Unstream.

Replacing with extensionContext?.open(_:completionHandler:) — the official
API for opening a URL from within a share extension — which opens Unstream
immediately and only calls completeRequest in its completion handler.

### DIFF
--- a/apps/mac/UnstreamShareExtension/ShareViewController.swift
+++ b/apps/mac/UnstreamShareExtension/ShareViewController.swift
@@ -110,11 +110,11 @@ class ShareViewController: UIViewController {
 
         let encodedQuery = query.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed) ?? ""
         if let appURL = URL(string: "unstream://search?q=\(encodedQuery)") {
-            openURL(appURL)
-        }
-
-        DispatchQueue.main.asyncAfter(deadline: .now() + 0.3) { [weak self] in
-            self?.close()
+            extensionContext?.open(appURL) { [weak self] _ in
+                DispatchQueue.main.async { self?.close() }
+            }
+        } else {
+            close()
         }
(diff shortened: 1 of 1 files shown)


brandonlucasgreen:
**Security & regression review — approved for merge**

Reviewed `ShareViewController.swift` as a security-focused engineer. No medium, high, or critical issues found.

**Security**
- Removing the responder-chain selector hack (`sel_registerName("openURL:")` + `perform(_:with:)`) is a net security improvement — the old path was opaque ObjC runtime manipulation.
- iTunes ID extraction (`allSatisfy(\.isNumber)`) is safe against injection.
- Pre-existing low-severity note (unchanged by this PR): `.urlQueryAllowed` does not encode `&`/`#`, so crafted shared text could inject extra query parameters into the custom URL scheme. Impact is bounded to the app's own URL handler and does not affect web surfaces.

**Regressions**
- `close()` is now called in the completion handler rather than on a fixed 0.3s timer. The callback ignores the success `Bool` (`_ in`), so `close()` fires on both success and failure paths — correct behavior.
- The new `else` branch calling `close()` on URL construction failure is an improvement over the old code, which silently skipped `close()` in that case.
- If `extensionContext` were nil at call time, the completion closure would never fire. In practice this cannot happen given the nil-check gate in `handleSharedItems`, and the system would force-terminate a hung extension anyway. Very low severity.

**Database migrations**: none required — pure iOS client change.

✅ Approved for merge.

---

PR #168, opened 2026-04-18
fix: force-open Unstream from share sheet (visible UI + notification fallback)

brandonlucasgreen:
$(cat <<'EOF'
## Summary

- **Show a visible loading UI** in the share extension instead of hiding the view. A hidden view means iOS may not treat the extension as a foreground process, which prevents `extensionContext.open()` from switching to the main app.
- **Add a notification fallback**: `extensionContext.open()` is documented for Today extensions and is not guaranteed to work for `com.apple.share-services` extensions. If it returns `success: false`, an immediate local notification is scheduled — "Tap to search for [Artist]". When tapped, Unstream becomes active and `checkPendingSearch()` fires the search automatically via the existing UserDefaults (app group) mechanism.
- The UserDefaults fallback still works as before — the search is always ready whenever Unstream next becomes active.

## Test plan

- [ ] Share an Apple Music album/song link to Unstream — confirm Unstream opens (or notification appears if it doesn't) and the search runs
- [ ] Share a Spotify link — same as above
- [ ] Verify the share extension shows "Searching for [Artist]…" text while processing
- [ ] If notification appears: tap it, confirm Unstream opens on the Search tab with results loading
- [ ] If notifications are disabled: confirm the UserDefaults fallback still works when manually switching to Unstream

EOF
)

### DIFF
--- a/apps/mac/UnstreamShareExtension/ShareViewController.swift
+++ b/apps/mac/UnstreamShareExtension/ShareViewController.swift
@@ -1,12 +1,38 @@
 import UIKit
 import UniformTypeIdentifiers
+import UserNotifications
 
 class ShareViewController: UIViewController {
 
+    private let activityIndicator = UIActivityIndicatorView(style: .medium)
+    private let statusLabel = UILabel()
+
     override func viewDidLoad() {
         super.viewDidLoad()
-        view.isHidden = true
+        setupUI()
         handleSharedItems()
+    }
+
+    private func setupUI() {
+        view.backgroundColor = .systemBackground
+
+        let stack = UIStackView(arrangedSubviews: [activityIndicator, statusLabel])
+        stack.axis = .vertical
+        stack.spacing = 10
+        stack.alignment = .center
+        stack.translatesAutoresizingMaskIntoConstraints = false
+
+        statusLabel.text = "Opening Unstream…"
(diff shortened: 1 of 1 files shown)


brandonlucasgreen:
**Security & regression review — approved for merge** ✓

### Security findings (all low severity)
- **Plausible risk (low)**: `query` is interpolated into both `statusLabel.text` and the notification `body` with no length cap. iOS notification system enforces its own limits and there's no code-execution path here, so this is not actionable at blocking severity.
- **Plausible risk (low)**: Force-unwrap on the iTunes lookup URL (`URL(string: "https://itunes.apple.com/lookup?id=\(id)")!`) is safe in practice because `extractAppleMusicID` only returns all-numeric strings.
- No auth, secret handling, injection, or data-exposure issues.

### Regression findings
- **Fixed (was medium)**: `close()` was only reachable via the `extensionContext?.open()` completion handler, which is documented as unreliable for `com.apple.share-services` extensions. If the completion block never fired, the share sheet would hang indefinitely. A safety-net timeout was added (commit `d562bf9`): a `didFinish` boolean guards a single `close()` call, with a 3-second `asyncAfter` fallback so the extension always dismisses even if the completion handler is never invoked.

### Database migrations
None required — this PR contains only Swift/iOS changes.

**This PR is approved for merge.**

---

PR #171, opened 2026-04-18
feat: tappable platform links (SFSafariVC) + Save to Unstream from share extension

brandonlucasgreen:
## Summary

- **Platform links are now tappable**: rows open `SFSafariViewController` directly inside the share extension overlay — a Safari pane slides up, the user browses the artist page, then taps Done to return to the extension. Chevrons stay because the rows are genuinely interactive.
- **"Open full results in Unstream" removed**: it didn't work (iOS blocks custom-scheme opens from share extensions) so it's gone rather than misleading.
- **"Save to Unstream" replaces it**: tapping saves the artist + all platform links to the app group `UserDefaults` under `pendingSaveArtist`. When Unstream next becomes active, `checkPendingSave()` decodes the entry and calls `SupportListManager.addEntryFromExtension`, inserting it at the top of the saved list and syncing to iCloud — no user action needed in the main app.
- **Empty/error states simplified**: removed the non-functional CTA buttons; the header Done button is sufficient.

## Test plan

- [ ] Share an Apple Music or Spotify link → extension shows results
- [ ] Tap a platform row (e.g. Bandcamp) → Safari pane opens with the artist page; tap Done → returns to extension overlay
- [ ] Tap "Save to Unstream" → button changes to "Saved to Unstream ✓" (green), becomes disabled
- [ ] Open Unstream → confirm the saved artist appears at the top of the saved list
- [ ] Save the same artist twice → confirm no duplicate is added
- [ ] Share when offline → confirm error state shows without any broken buttons

### DIFF
--- a/apps/mac/Unstream/SupportListManager.swift
+++ b/apps/mac/Unstream/SupportListManager.swift
@@ -58,6 +58,20 @@ class SupportListManager: ObservableObject {
         entries.contains { $0.artistName.lowercased() == artistName.lowercased() }
     }
 
+    /// Called by the main app when it finds a `pendingSaveArtist` queued by the share extension.
+    func addEntryFromExtension(name: String, imageUrl: String?, platforms: [(sourceId: String, url: String)]) {
+        guard !entries.contains(where: { $0.artistName.lowercased() == name.lowercased() }) else { return }
+        let entry = SupportEntry(
+            id: UUID(),
+            artistName: name,
+            imageUrl: imageUrl,
+            platforms: platforms.map { SavedPlatform(sourceId: $0.sourceId, url: $0.url) },
+            dateAdded: Date()
+        )
+        entries.insert(entry, at: 0)
+        saveEntries()
+    }
+
     func toggleArtist(_ artist: ArtistResult) {
         if isArtistSaved(artist.name) {
             entries.removeAll { $0.artistName.lowercased() == artist.name.lowercased() }
--- a/apps/mac/Unstream/UnstreamApp.swift
+++ b/apps/mac/Unstream/UnstreamApp.swift
@@ -50,6 +50,25 @@ class AppStateContainer: ObservableObject {
         #endif
     }
 
+    /// Check for an artist the share extension queued for saving (iOS).
+    func checkPendingSave() {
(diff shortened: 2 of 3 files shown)


brandonlucasgreen:
## Security & Regression Review — Approved for merge

Reviewed as a security-focused staff engineer. Two medium-severity regression bugs were found and fixed in a follow-up commit (`3d58e05`) on this branch before approval.

---

### Security findings

| Finding | Classification | Severity |
|---------|---------------|----------|
| URL scheme guard (`https`/`http` only) before `SFSafariViewController` — **improvement** over the old `extensionContext?.open(url)` which accepted any scheme | Positive change | — |
| `NSWorkspace.shared.open(url)` on macOS has no scheme validation on `SavedPlatform.url`; risk is low given URLs come from a trusted API | Plausible risk | Low |
| `http://` URLs accepted for `SFSafariViewController` in the extension | Plausible risk | Low |
| Any App Group member can write `pendingSaveArtists`; no provenance check | Needs manual verification | Low |

No high or critical security issues found.

---

### Regression bugs fixed before approval

**Bug 1 (Medium) — Silent data loss if JSON decode fails**

`checkPendingSave()` was deleting the `pendingSaveArtist` key from `UserDefaults` _before_ decoding the JSON. If decoding failed (e.g. after an app update that changes the struct), the save was permanently lost with no user feedback. Fixed by moving `defaults.removeObject` to after a successful decode.

**Bug 2 (Medium) — Multiple saves overwrite each other**

`saveArtist()` was writing a single JSON object to `pendingSaveArtist`, so saving Artist A then Artist B before opening the app would silently discard Artist A. Fixed by:
- Renaming the key to `pendingSaveArtists` (plural)
- Extension now reads the existing array, appends the new entry, and writes back
- Main app now decodes the full array and calls `addEntryFromExtension` for each entry in order

---

### Remaining low-severity items (acceptable, no action required)

- Non-http/s platform links will silently fail to open from the extension with no user-visible error — acceptable given all real platform URLs from the API are HTTPS
- `isSaved` state not reset between share sessions — not an issue in practice since extensions create fresh `ShareSearchViewModel` instances per invocation

---

No database migrations required (pure native app change).

**This PR is approved for merge.**

---

PR #172, opened 2026-04-18
polish: improve share extension loading state, header spacing, and filter manual-search platforms

brandonlucasgreen:
- Fix blank space above header in loading state by using frame(maxWidth/maxHeight: .infinity) so the root VStack always fills the container, pinning the header to the top regardless of content size
- Increase header font size (15→17pt) and padding (12→16pt vertical, 16→20pt horizontal) for more breathing room
- Exclude hoopla and freegal from share extension results since they produce generic search URLs, not real artist profile links

https://claude.ai/code/session_01CEwg2c4KRsbZdGC8U82Ta6

### DIFF
--- a/apps/mac/UnstreamShareExtension/ShareViewController.swift
+++ b/apps/mac/UnstreamShareExtension/ShareViewController.swift
@@ -29,11 +29,21 @@ struct PlatformMeta {
     let icon: String
     let payoutPercent: String?
     let isSocial: Bool
-    var isManualSearch: Bool = false
 }
 
-// Platforms that only produce generic search URLs, not real artist pages
-private let manualSearchPlatforms: Set<String> = ["hoopla", "freegal"]
+// Returns true when the API has produced a generic search URL rather than a real artist page.
+// These links only appear when an artist hasn't claimed their profile; claimed profiles always
+// supply direct URLs, so this check lets those through while hiding useless search placeholders.
+private func isSearchOnlyURL(sourceId: String, url: String) -> Bool {
+    switch sourceId {
+    case "kofi":         return !url.contains("ko-fi.com/")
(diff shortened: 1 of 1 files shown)


brandonlucasgreen:
**Security & regression review complete. This PR is approved for merge.**

### Security findings
No confirmed vulnerabilities. Pure display-layer Swift change — no auth, secrets, network additions, or new data exposure.

**Plausible risks (low severity, no action required):**
- `kofi` URL check uses negation (`!url.contains("ko-fi.com/")`) while all other cases use positive `contains`. Works correctly today; if the API ever returns `https://ko-fi.com` without a trailing slash, that result would be silently dropped. Low probability given the API is controlled.
- `saveArtist` receives the unfiltered `platforms` array, so search-only URLs for hoopla/freegal are persisted to UserDefaults even though they're hidden from display. Likely intentional (main app handles its own filtering), but worth keeping in mind.
- `isSearchOnlyURL` ampwall pattern (`"explore?searchStyle"`) is a substring match on a query param — stable as long as Ampwall's search URL structure doesn't change.

### Regression findings
No regressions identified.
- `primary` filter is additive (hides results only). `summaryLine` correctly reflects the filtered count.
- `socialCount` is computed independently; all five filtered platforms have `isSocial: false`, so social counts are unaffected.
- An artist with results only from hoopla/freegal will show "Found on 0 platforms" — correct behavior per intent.
- `frame(maxWidth: .infinity, maxHeight: .infinity, alignment: .top)` correctly fixes the blank-space loading bug with no layout regressions.
- Note: PR description mentions only hoopla/freegal but `isSearchOnlyURL` also handles kofi, buymeacoffee, and ampwall — intentional per the code comment, description is just incomplete.

### Database migrations
None required. No backend or schema changes in this PR.

---

PR #174, opened 2026-04-19
fix(mac): split Info.plist per platform to restore macOS menu bar + playback

brandonlucasgreen:
## Summary
- After the universal app restructure (#129), the shared `Info.plist` lost `LSUIElement=true` and picked up iOS-only keys (`UILaunchScreen`, `UISupportedInterfaceOrientations`, `BGTaskSchedulerPermittedIdentifiers`). On macOS this caused the app to appear in the Dock instead of the menu bar, and the broken app lifecycle also prevented `MediaObserver` from running — so playback detection silently stopped working.
- The AppDelegate's `NSApp.setActivationPolicy(.accessory)` is not a reliable substitute for `LSUIElement` in the plist.
- Splits the shared plist into `Info-macOS.plist` (menu bar + Apple Events) and `Info-iOS.plist` (orientations + background tasks), wired via SDK-conditional `INFOPLIST_FILE[sdk=...]` in `project.yml`.
- Also converts the existing `CODE_SIGN_ENTITLEMENTS` setting from xcodegen 2.44's nested-dict form (rejected by `xcodebuild` with "key was supposed to be an integer or a string, but instead it was __NSDictionaryM") to the flat bracketed form.

## Test plan
- [x] `xcodebuild -destination 'platform=macOS'` succeeds
- [x] Built macOS `Info.plist` has `LSUIElement=true`, `LSApplicationCategoryType`, `NSAppleEventsUsageDescription`, and no iOS keys (verified via `PlistBuddy`)
- [x] `xcodebuild -destination 'generic/platform=iOS'` succeeds (app + share extension)
- [x] Built iOS `Info.plist` has orientations + `BGTaskSchedulerPermittedIdentifiers`, and no `LSUIElement` / AppleEvents
- [x] Manual: Debug build on macOS — menu bar icon appears and playback detection fires on Music.app
- [ ] Manual: iOS TestFlight build verified after merge (no functional change expected — plist contents identical, just moved to a file with a different name)

### DIFF
--- a/apps/mac/Unstream.xcodeproj/project.pbxproj
+++ b/apps/mac/Unstream.xcodeproj/project.pbxproj
@@ -3,56 +3,56 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 63;
+	objectVersion = 77;
 	objects = {
 
 /* Begin PBXBuildFile section */
-		PP111111111111111111111111 /* PrivacyInfo.xcprivacy in Resources */ = {isa = PBXBuildFile; fileRef = PP222222222222222222222222 /* PrivacyInfo.xcprivacy */; };
-		PP333333333333333333333333 /* PrivacyInfo.xcprivacy in Resources */ = {isa = PBXBuildFile; fileRef = PP444444444444444444444444 /* PrivacyInfo.xcprivacy */; };
 		0038D14301F153EF60346193 /* SocialIconButton.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7EFEA3443AD56B9950B43AAE /* SocialIconButton.swift */; };
 		0EC3D68464514744CE3AA448 /* BandcampReleaseChecker.swift in Sources */ = {isa = PBXBuildFile; fileRef = 43017BB6FDE57EBD35755D7B /* BandcampReleaseChecker.swift */; };
 		0F9D0A7FD89742003BD8D140 /* TipJarStore.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9C40AC0CB901A9F09846258C /* TipJarStore.swift */; };
-		11B4891578B50BCF51D68613 /* SearchTab.swift in Sources */ = {isa = PBXBuildFile; fileRef = E330692CA1EC213166D9351F /* SearchTab.swift */; platformFilter = ios; };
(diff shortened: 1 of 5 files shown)


brandonlucasgreen:
## Security & Regression Review — Approved ✅

### Security review

| Area | Finding | Classification |
|---|---|---|
| CORS on admin endpoint | Old `Access-Control-Allow-Origin: *` replaced with `buildCorsHeaders(origin, false)` restricting to `unstream.stream` | **Improvement** |
| Column selection | `select('*')` replaced with explicit columns; mapped response correctly omits `user_id`/`artist_id` from the GET payload | **Improvement** |
| Input validation | UUID regex on `requestId` and 2000-char cap on `reviewerNotes` added | **Improvement** |
| `reviewerNotes` type | Length check does not assert `typeof === 'string'` first; a non-string value would pass the truthiness check and its `.length` would be evaluated as a non-string length | **Plausible risk — low severity** (admin-only endpoint; Supabase text column would coerce/reject a non-string value) |
| Profile INSERT data provenance | `user_id` and `email` sourced from the DB-fetched verification request row, not from the incoming JSON body | **No risk** |
| Apple Team ID in pbxproj | `CV926C8KS8` hardcoded — standard for Xcode; team IDs are semi-public and not a secret | **Needs manual verification** (acceptable for a public repo) |

### Regression review

| Area | Finding |
|---|---|
| Migration ordering | New `approve` path inserts `artist_profiles` with `website_url: null`; migration 010 (make `website_url` nullable) must run before deploy. **Migration has been applied to production Supabase.** |
| Admin UI contract | `AdminVerifyPage.tsx` interface already expects `artist_name`, `artist_slug`, `website_url: string \| null` — matches the new API response exactly. No regression. |
| Old `artist_profile_id` lookup | `verification_requests` has no `artist_profile_id` column (migration-006). Old code was silently broken. New `artist_id`-based lookup is the correct fix. |
| StoreKit scheme reference | `storeKitConfigurationFileForLaunching` removed from Xcode scheme — IAP sandbox won't auto-load in Debug. Dev-only impact; no production regression. |

### Database migration

Migration `010-nullable-website-url` (`ALTER TABLE artist_profiles ALTER COLUMN website_url DROP NOT NULL`) has been applied to the production Supabase project and recorded in the migrations table. The deploy can proceed immediately after merge.

No medium, high, or critical issues found. **This PR is approved for merge.**

---

PR #176, opened 2026-04-20
fix: custom link displayName in search results + other platform source entry

brandonlucasgreen:
## Problem
Custom links added by claimed artists (platform `other_*`) are broken across all clients:
- **App/extension search**: shows raw ID like `index_0` or `Other_0` instead of the artist's custom name (e.g. "Weblog")
- **Web search**: custom links don't render at all — no source entry exists for `other_*`, so they get filtered out
- **Artist profile page**: works correctly (reads `display_name` from DB directly)

## Root Cause
Two bugs working together:

1. **`search-sources.ts` drops `displayName`** — When mapping claimed artist platforms into search results (line ~1500), the `displayName` field from the DB is not included in the output.

2. **No `other` source entry** — `sources.ts` (web) and `PlatformCatalog.swift` (native) have no entry for `other` or `other_*` sourceIds. Web filters them out entirely; native falls back to the raw sourceId string.

## Changes
- `api/functions/search-sources.ts`: Add `displayName` to claimed artist platform mapping
- `apps/web/src/services/sources.ts`: Add `other` source entry (name: "Link", icon: 🔗) + `curated` source category
- `apps/mac/Unstream/Models/PlatformCatalog.swift`: Add `other` entry (name: "Link", icon: globe)
- `apps/web/src/components/SourceBadge.tsx`: Accept optional `displayName` prop, prefer it over `source.name`
- `apps/web/src/components/ResultCard.tsx`: Use `getSource()` helper that falls back to `other` for unknown sourceIds; pass `displayName` through to badges; render custom links in new "Links" section
- `apps/web/src/types/index.ts`: Add `other` to `SourceId` union, `curated` to category type

## Testing
- TypeScript compiles clean (`npx tsc -b`)
- Full web build passes (`npm run build`)
- Reported by: Glen (Things task triage)

### DIFF
--- a/api/functions/search-sources.ts
+++ b/api/functions/search-sources.ts
@@ -1502,6 +1502,7 @@ export async function handler(event: { queryStringParameters?: Record<string, st
           platforms: dbArtist.platforms.map(p => ({
             sourceId: p.sourceId as SourceId,
             url: p.url,
+            displayName: p.displayName,
             latestRelease: p.latestRelease,
           })),
           matchConfidence: 'claimed',
--- a/apps/mac/Unstream/Models/PlatformCatalog.swift
+++ b/apps/mac/Unstream/Models/PlatformCatalog.swift
@@ -47,6 +47,8 @@ let platformCatalog: [String: PlatformConfig] = [
     "bluesky": PlatformConfig(name: "Bluesky", icon: "cloud", color: "#0085FF", searchOnly: false),
     "mastodon": PlatformConfig(name: "Mastodon", icon: "bubble.left.and.bubble.right", color: "#6364FF", searchOnly: false),
     "peertube": PlatformConfig(name: "PeerTube", icon: "play.circle", color: "#F1680D", searchOnly: false),
+    // Fallback for artist-added custom links (other, other_*)
+    "other": PlatformConfig(name: "Link", icon: "globe", color: "#888888", searchOnly: false),
 ]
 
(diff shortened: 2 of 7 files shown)


brandonlucasgreen:
## Security + Regression Review — Approved ✓

**Security findings:**

**Confirmed vulnerability (fixed):** `artist-profile.ts` filtered custom link URLs with only a truthy check (`l.platform && l.url`). This PR renders `other_*` links as `<a href>` elements for the first time — previously they were silently dropped. Without protocol validation, a claimed artist could store a `javascript:` URL, which would execute on click (stored XSS). Fixed in a follow-up commit by applying the same `http:`/`https:` protocol check already used for `customImageUrl`. ([74714c1](https://github.com/brandonlucasgreen/unstream/commit/74714c1))

**Plausible risk (no action needed):** `displayName` is trimmed and capped at 50 chars server-side; React text rendering escapes it, so no XSS path there.

**Regression review:**

No breaking regressions. Minor notes for awareness:
- `isDirectLink()` still reads `sources[sourceId]` directly instead of the new `getSource()` helper — functionally correct (unknown IDs return `undefined` → treated as non-searchOnly → `true`), just inconsistent.
- `allKnownSourceIds` is rebuilt every render; could be a module-level constant for perf, but no correctness issue.

**DB migrations:** None required — no schema changes.

**Result:** Approved for merge.</body>
</invoke>

---

PR #199, opened 2026-05-11
fix: replace broken Bandcamp scraping with MusicBrainz enrichment + fallback

brandonlucasgreen:
Fixes UNS-28

## Changes
- Disabled broken `searchBandcamp()` scraper (Bandcamp anti-bot protection blocks server-side scraping)
- Extract Bandcamp URLs from MusicBrainz `url_relations` data — direct links surface as verified platform results
- Added `searchOnly: true` to Bandcamp source definition — always shows as clickable search link when no direct result found

## Before
Bandcamp never appeared in search results because the scraper was served a JS challenge page

## After
- If MusicBrainz has a Bandcamp URL for the artist → shows as a direct/verified link
- If no Bandcamp URL found → shows "Search on Bandcamp" fallback link
- No more scraping attempts

Test case: Search for "Scandroid" — their MB entry includes Bandcamp URL


brandonlucasgreen (commented):
## PR #199 Review: fix/bandcamp-search-musicbrainz

### Build Status
✅ **Clean** — `npx tsc -b` passes, `npm run build` passes, all 165 tests pass.

---

## 🔴 Must-Fix

### 1. `searchOnlyPlatforms` inconsistency in `apps/web/src/services/sources.ts`
**File:** `apps/web/src/services/sources.ts` (line ~700)

The server-side sort in `apps/web/server/search/index.ts` correctly includes `'bandcamp'` in `searchOnlyPlatforms`:
```ts
const searchOnlyPlatforms = new Set(['ampwall', 'kofi', 'buymeacoffee', 'bandcamp']);
```

But the client-side equivalent in `apps/web/src/services/sources.ts` does **not**:
```ts
const searchOnlyPlatforms = new Set(['ampwall', 'kofi', 'buymeacoffee']);
```

Since Bandcamp is now a search-only fallback (when no direct URL from MusicBrainz), it should be treated consistently on both sides. Otherwise the platform sorting order will differ between SSR and CSR, causing hydration mismatches or inconsistent UI.

**Fix:** Add `'bandcamp'` to the client-side `searchOnlyPlatforms` set.

---

## 🟡 Suggestions

### 2. Sentry bundle impact is significant (~+84 KB raw / +29 KB gzipped)
The `@sentry/react` SDK increases the main bundle by ~24–27%. Consider:
- Using dynamic import / lazy-loading for Sentry initialization so it doesn't block the critical path
- Or verifying if `tracesSampleRate: 0.0` and `profilesSampleRate: 0.0` are enough to tree-shake unused Sentry features (tracing, profiling). Check if the SDK still bundles them.

### 3. `searchOnly: true` added to `bandcamp` source definition — good, but verify UI impact
In `apps/web/src/services/sources.ts`, `bandcamp` now has `searchOnly: true`. This means `SourceBadge` will show it as a search link. Confirm this is intentional — direct MusicBrainz-derived Bandcamp URLs should bypass this via `isDirectLink`. The logic looks correct, but worth a manual spot-check.

### 4. `bandcamp-fetch` audit warning
`npm audit` reports `bandcamp-fetch` (v1.x) depends on a vulnerable `cookie` package. Since this PR **removes** server-side Bandcamp scraping, `bandcamp-fetch` may now be unused. Consider removing it from `dependencies` entirely to close the vulnerability and reduce install size.

### 5. MusicBrainz rate limiting adds latency
The new MB artist-detail fetch adds **two sequential 1100ms delays** (one before the artist URL call, one after). For a single artist, that's an extra ~1.1s before the first Bandcamp result can be returned. If MusicBrainz is slow or down, the whole search pipeline waits. Consider:
- Running the MB artist-detail fetch **in parallel** with other searches (it's currently sequential within `searchMusicBrainz`)
- Or adding a shorter timeout on the MB fetch

### 6. Minor: trailing empty lines in `SocialIcon.tsx`
The diff removes `SOCIAL_PLATFORMS` and `hasSocialIcon` but leaves two trailing blank lines at the end of the file. Not a functional issue, but a quick tidy-up would be nice.

---

## ✅ What Looks Good

- **Security:** Sentry DSN handled via env vars, `sendDefaultPii: false`, dev errors filtered, `https://` prefix check. No secrets committed.
- **CSP updated** correctly to allow `*.ingest.sentry.io`.
- **Dead code removal:** `ArtistAuthBar` and `hasSocialIcon` confirmed unused — clean deletes.
- **Dark mode fix:** The `isDarkColor` logic in `App.tsx` is a solid, lightweight approach without adding new deps.
- **MusicBrainz Bandcamp extraction:** URL validation via `new URL(u).hostname.endsWith('.bandcamp.com')` is correct and safe.
- **Fallback search links:** Ampwall, Ko-fi, BuyMeACoffee now attached to all artists (not just Bandcamp ones) — this is an improvement.

---

Once **must-fix #1** is addressed, this PR is good to approve.

PR #218, opened 2026-05-15
Fix UNS-35: View live profile link for claimed artists

brandonlucasgreen:
## What this fixes

On the artist's Edit Profile page, the 'View live profile' link was taking artists to search results instead of their actual claimed profile.

## Root cause

In ArtistPage.tsx, the page first tries to load pre-generated JSON at /data/artists/slug.json, then falls back to searchPlatforms() which shows generic search results, not the claimed profile. For claimed artists who don't have pre-generated JSON files, this caused them to see search results.

## The fix

Added a fallback to fetch artist data from /api/artist?slug=slug when the pre-generated JSON file doesn't exist. This ensures claimed artists can always view their verified profile data.

## Steps

1. Try to load /data/artists/slug.json (unchanged - still first priority)
2. If that fails, fetch from /api/artist?slug=slug (new - for claimed artists)
3. If that fails, fall back to live search (unchanged - fallback for unclaimed artists)

This preserves the existing behavior for pre-generated artists while fixing the issue for claimed artists.

Fixes UNS-35

### DIFF
--- a/apps/web/public/dispatch.xml
+++ b/apps/web/public/dispatch.xml
@@ -6,7 +6,7 @@
     <atom:link href="https://unstream.stream/dispatch.xml" rel="self" type="application/rss+xml" />
     <description>Weekly music industry intelligence — platform news, streaming economics, and what it means for independent artists. Written by Unstream.</description>
     <language>en-us</language>
-    <lastBuildDate>Fri, 15 May 2026 20:41:23 GMT</lastBuildDate>
+    <lastBuildDate>Fri, 15 May 2026 22:24:15 GMT</lastBuildDate>
     <item>
       <title>Week of April 16, 2026</title>
       <link>https://unstream.stream/dispatch/2026-W16</link>
--- a/apps/web/src/pages/ArtistPage.tsx
+++ b/apps/web/src/pages/ArtistPage.tsx
@@ -82,10 +82,30 @@ export function ArtistPage() {
           }
         }
       } catch {
-        // Fall through to live search
+        // Fall through to live fetch
       }
 
-      // Fallback: live search using the slug as query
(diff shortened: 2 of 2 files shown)


brandonlucasgreen (commented):
## Review: Fix UNS-35

### What's Good
- **Fixes the right problem** — claimed artists without pre-generated JSON were seeing search results instead of their verified profile. This is a legitimate bug and the fix addresses it directly.
- **Preservation of existing behavior** — pre-generated JSON still takes priority, live search remains the final fallback. Clean layered approach.
- **Security** — `encodeURIComponent(slug)` is used correctly on line 91, preventing query string injection.
- **Race condition handling** — `cancelled` flag is checked after the fetch, consistent with existing patterns in the file.

### Issues Found

**❌ TypeScript build error (blocking)**
Line 91: `encodeURIComponent(slug)` fails with:
```
Argument of type 'string | undefined' is not assignable to parameter of type 'string | number | boolean'.
```
The `useParams` return type makes `slug` potentially `undefined`, but line 63 already has `if (!slug) return;` so TS should narrow it. However, inside the inner `loadArtist` async function, TypeScript loses the narrowing. Fix: use `slug!` or reassign to a const before the async function:
```ts
const artistSlug = slug;
if (!artistSlug) return;
// then use artistSlug inside loadArtist
```
This will fail CI/build if not addressed.

### Minor Notes
- The `data ? [data] : []` pattern on line 95 is slightly redundant since the API should always return a valid artist object when `response.ok`, but it's harmless defensive coding.
- Error handling is minimal (empty catch blocks), but this matches the existing code style in the file — not a regression.

### Verdict
**Changes requested** — please fix the TypeScript error on line 91. Everything else looks solid.

PR #219, opened 2026-05-15
Add artist-detection notifications to browser extensions (UNS-41)

brandonlucasgreen:
When the extension detects a new artist via playback, send a browser notification showing where to directly support them.

**What it does:**
- Detects new artist via existing scrobbling/detection flow
- Shows notification: "Now playing: [artist]. Directly support them on n platforms including [A] and [B]."
- Skips notification if no support platforms found (nothing actionable)
- Suppresses duplicate notifications for same artist within browser session
- Same artist can notify again in a new session
- Click notification → opens Unstream artist page
- Toggle in Settings tab (default ON, existing users opted in)
- Respects browser notification permissions

**Implementation:**
- Added `maybeNotifyArtist()` in service worker after successful search
- Added `notifiedArtists` Set for session-based duplicate suppression
- Added `artistNotificationsEnabled()` check (default true, stored in `chrome.storage.sync`)
- Added toggle switch in popup Settings tab
- Consolidated notification click handler (artist + release alerts)
- No new permissions needed (`notifications` already declared)

Fixes UNS-41


brandonlucasgreen (commented):
PR Review: Artist notification slug bug needs fixing before merge. See top-level comment for details.

brandonlucasgreen (commented):
**PR:** #219 — Add artist-detection notifications to browser extensions (UNS-41)
**Author:** brandonlucasgreen
**Files changed:** 4 (service-worker.js, popup.css, popup.html, popup.js)
**Additions:** 168, Deletions: 2

--- CODE REVIEW ---

## What Looks Good

1. **No new permissions needed** — leverages existing `notifications` permission.
2. **Default-on with backward compatibility** — existing users are opted in (undefined → true). Clean upgrade path.
3. **Duplicate suppression** — session-scoped `Set` prevents notification spam for the same artist.
4. **Graceful degradation** — skips notification if no platforms found; catches and logs `chrome.notifications.create` errors.
5. **Consolidated click handler** — artist and release notifications share the same `onClicked` listener, avoids duplicate handlers.
6. **URL-safe slug generation** — `toLowerCase().replace(/[^a-z0-9]+/g, '-')` handles special characters.
7. **Priority 2** — high priority for browser notifications is appropriate for real-time detection.

---

## Issues Found

### 1. Duplicate suppression uses `claimedSlug || artistName` — inconsistent keys [MEDIUM]

In `formatArtistNotification()`:
```js
slug: primaryResult.claimedSlug || artistName.toLowerCase().replace(/[^a-z0-9]+/g, '-').replace(/^-|-$/g, ''),
```

If an artist has a `claimedSlug` (e.g., `"the-beatles"`), the notification ID uses that. If not, it falls back to a slugified `artistName`. The suppression check uses `notification.slug`. This means:
- An artist with `claimedSlug = "the-beatles"` and name "The Beatles" → key = `"the-beatles"`
- Same artist without a claimed slug → key = `"the-beatles"` (slugified name)

**Problem:** If the artist result switches between having a claimedSlug and not (e.g., cache refresh, API inconsistency), the same artist could notify twice in one session because the key differs. Also, two different artists with the same slugified name (e.g., "The Band" vs "the band") would collide.

**Fix:** Use a deterministic, stable key. Prefer `claimedSlug` if present, otherwise normalize the artist name more robustly. Consider using the artist name itself (case-insensitive) as the suppression key instead of a slug.

### 2. Notification click handler opens Unstream artist page without validation [LOW]

```js
if (notificationId.startsWith('artist-')) {
  const slugOrName = notificationId.replace('artist-', '');
  const url = `https://unstream.stream/a/${slugOrName}`;
  chrome.tabs.create({ url });
}
```

**Problem:** `slugOrName` is user-controlled (derived from `artistName`). If an artist name contains path traversal or weird characters after slugification, the URL could be malformed. The slugification strips non-alphanumeric, but if `claimedSlug` is used directly, it could theoretically contain `../` or other characters (depending on API trust).

**Fix:** Validate `slugOrName` before constructing URL. At minimum, ensure it matches `^[a-zA-Z0-9_-]+$` or URL-encode it with `encodeURIComponent()`.

### 3. Missing `await` on `chrome.tabs.create` [LOW]

```js
chrome.tabs.create({ url });
```

While not strictly necessary for side-effect-only calls, for consistency with the rest of the codebase (which uses async/await), consider adding `await`. More importantly, if this throws (rare), it will be an unhandled rejection.

**Fix:** `await chrome.tabs.create({ url });`

### 4. `notifiedArtists` Set is unbounded [LOW]

```js
const notifiedArtists = new Set();
```

This grows for every unique artist heard in a session. In a long-lived browser session with many artists, this could consume non-trivial memory. For an extension, probably fine in practice, but worth a cap.

**Fix:** Consider a max size (e.g., LRU of 1000) or clear it periodically. Not a must-fix for v1.

### 5. Session-scoped suppression resets on service worker restart [DOCUMENTATION/EXPECTED]

MV3 service workers are ephemeral. They can be killed and restarted at any time (e.g., after 30s+ idle). When restarted, `notifiedArtists` resets. This means a user listening to the same artist across multiple songs could get duplicate notifications if the service worker restarts between tracks.

**Assessment:** This is probably acceptable for a v1 feature ("same artist can notify again in a new session" as stated in PR description). But it should be documented as a known limitation. If strict suppression is desired, persist to `chrome.storage.session` or `chrome.storage.local`.

**Fix (optional):** If stronger suppression is needed, switch to `chrome.storage.session` (MV3 supports session storage that persists across service worker restarts but not browser restarts).

### 6. No rate limiting on `maybeNotifyArtist` [LOW]

Every `MUSIC_DETECTED` message that yields results triggers a notification check. If a user scrubs through a playlist rapidly, or if the content script sends frequent messages, this could spam notification API calls (even if suppressed after the first one).

The existing code has `lastSearchTime` throttling for searches (30s), but notifications aren't gated separately.

**Fix:** Not critical since search itself is throttled, but consider a separate cooldown (e.g., 60s between any notifications) as a defense-in-depth.

---

## Summary

| Issue | Severity | Must Fix? |
|-------|----------|-----------|
| Inconsistent suppression key (claimedSlug vs slugified name) | Medium | **Yes** |
| Click handler lacks URL validation | Low | Recommended |
| Unbounded `notifiedArtists` Set | Low | No |
| Session suppression resets on SW restart | Info | Document |
| Missing await on `chrome.tabs.create` | Low | No |
| No rate limiting on notifications | Low | No |

---

## Overall Assessment

This is a clean, well-scoped PR. The feature is user-friendly (default on, easy toggle) and respects existing permissions. The main concern is the duplicate suppression key inconsistency, which could lead to duplicate notifications under certain conditions.

**Recommendation:** Approve with minor changes — fix the suppression key to be stable and consistent, and optionally add `encodeURIComponent` to the click handler URL.

PR #220, opened 2026-05-16
Add artist detection push notifications for macOS

brandonlucasgreen:
## Summary

Adds push notifications for macOS when a now-playing artist is detected, plus fixes a sandbox compatibility issue with music detection.

### What's new
- **macOS push notifications**: When music is playing and an artist is detected, a notification shows "Now playing: {artist}. Support directly on {platform} and {n-1} other platforms." — only marketplace-type platforms are shown (no social, official sites, or newsletters)
- **Sandbox fix for music detection**: Replaced all `tell application "System Events"` AppleScript calls with `NSWorkspace.runningApplications` bundle ID checks. System Events fails with -600/-10004 in sandboxed apps. Permission check now tests Music.app directly.
- **Updated notification copy**: Title shows artist name with "Now playing:" prefix, body uses "Support directly on..." phrasing
- **Updated macOS app screenshot**
- **Version bump to 3.1.0**

### Bug fixes
- `newsletter` was missing from `socialPlatformIds` — could appear in notifications
- Removed System Events from `scripting-targets` entitlement (no longer needed)

### Changed files
- `MediaObserver.swift` — NSWorkspace replaces System Events AppleScript
- `NowPlayingNotificationManager.swift` — updated notification copy, marketplace-only filtering
- `SearchResponse.swift` — added `marketplacePlatforms` computed property
- `PlatformCatalog.swift` — added `marketplacePlatformIds`, added `newsletter` to `socialPlatformIds`
- `Unstream-macOS.entitlements` — removed System Events scripting target
- Version bumped in `Info-macOS.plist`, `Info-iOS.plist`, `ShareExtension/Info.plist`, `project.yml`

## Test plan
- [ ] Play music in Music.app and verify notifications appear
- [ ] Verify notification only shows marketplace platforms (Bandcamp, Mirlo, etc.) — not social/official/newsletter
- [ ] Verify notification title format: "Now playing: {artist}"
- [ ] Verify notification body format: "Support directly on {platform} and {n} other platforms"
- [ ] Verify clicking notification opens artist page on unstream.stream
- [ ] Verify music detection works without System Events automation permission
- [ ] Verify no regression in Spotify/other player detection


brandonlucasgreen (commented):
## Review: Artist Detection Notifications (macOS)

### Must Fix

**1. Skip notification when no platforms found** ❌

The requirement says "Skip if no support platforms found." Currently, `buildBody` returns `"Tap to find where to support them"` for the 0-platform case, and the notification is still sent. This contradicts PR #219 behavior where the notification is skipped entirely (`return null`).

**Fix:** Add an early return in `sendNotification` (or `handleArtistDetected`) when `platforms.isEmpty`.

---

**2. Notification click should open the artist page** ❌

PR #219 opens `https://unstream.stream/a/${slugOrName}`. This PR opens the generic popover via `"action": "openPopover"`. Users won't land on the artist's page — they'll have to search again.

**Fix:** Build the artist URL (use `topResult.claimedSlug` or a sanitized slug) and open the URL directly, matching the extension behavior. If the slug isn't available on `ArtistResult`, derive it from the artist name.

---

### Otherwise Looks Good

- ✅ Session-based duplicate suppression via `Set<String>`
- ✅ Settings toggle default ON (unset key returns `true`)
- ✅ Proper `@MainActor` usage for notification manager
- ✅ Idiomatic `AppStorage` binding in SettingsView
- ✅ `showPopover()` implementation is correct for the existing notification path

### Minor Note

Consider using a URL-safe slug (like PR #219's `toLowerCase().replace(/[^a-z0-9]+/g, '-')`) for the notification identifier instead of raw `artist.lowercased()`, to avoid edge-case collisions.

brandonlucasgreen (commented):
Re-review: both fixes from the previous review are correctly implemented.

**Fix 1 — Skip notification when no platforms found:** ✅ `sendNotification` now guards with `guard !platforms.isEmpty else { return }` before constructing any notification content.

**Fix 2 — Notification click opens artist page:** ✅ The `UNUserNotificationCenterDelegate` now checks `artistUrl` first and falls back to `releaseUrl`, deep-linking to `/a/[slug]` instead of opening a generic popover.

**Consistency with PR #219 (extension):** Good alignment on defaults (opt-in), session dedup, platform filtering, and click behavior. Minor UX differences in notification body copy (macOS omits artist name in body; extension includes it) — not blocking but worth a future polish pass.

**One small issue:** `apps/mac/project.yml` line 78 still has `CFBundleVersion: "9"` for the share extension — should be "10" to match the plist bumps. Please fix before merge.

PR #230, opened 2026-05-30
fix(UNS-62): update login interstitial header copy

brandonlucasgreen:
Updates the login interstitial header from "Sign in to save this artist" to "Sign up or log in to save this artist" per Brandon's feedback in UNS-62.


brandonlucasgreen:
Consolidated into #231.

PR #233, opened 2026-05-30
feat(UNS-66): Allow logged-in users to mark saved artists as supported

brandonlucasgreen:
## UNS-67: Supported Artists Feature

Full-stack feature allowing logged-in users to mark saved artists as "supported".

### Changes

**1. Database Migration** (`supabase/migration-015-supported-artists.sql`)
- Adds `supported` boolean column (default `false`) and `supported_at` timestamptz column to `saved_artists` table
- Includes doc comments on both columns
- Removing a saved artist cascades the row away; re-saving defaults supported back to false

**2. API Changes** (`api/functions/saved-artists.ts`)
- **GET handler**: Response now includes `supported` (boolean) and `supportedAt` (string | null) fields
- **POST handler**: Two new actions:
  - `action: "support"` — Sets `supported = true` and `supported_at = now()`, validates user owns the row, returns updated artist. Idempotent.
  - `action: "unsupport"` — Sets `supported = false` and `supported_at = null`, validates user owns the row, returns updated artist. Idempotent.
- Both return 404 if the saved artist doesn't exist for that user

**3. Frontend Changes** (`apps/web/src/pages/DashboardPage.tsx`)
- `SavedArtist` interface updated with `supported: boolean` and `supportedAt?: string`
- Support toggle button on each saved artist card:
  - Checkmark icon + "Supported ✓" when supported (accent-secondary color)
  - Circle outline + "Mark supported" when not (text-muted color, matching UNS-65 contrast fix)
  - Optimistic UI update on toggle, revert on error with toast notification
  - Button disabled while request is in-flight (prevents double-submit)

### Edge Cases
- Removing a saved artist cascades the supported state away
- Re-saving an artist defaults supported back to false
- Multiple tabs: last-write-wins on toggle (no staleness guards needed)
- Support/unsupport actions are idempotent


brandonlucasgreen:
**Review complete — LGTM ✅**

**Bug fix verified:** The optimistic revert now correctly captures  and  before the optimistic update and restores them on error. The previous bug where  was lost on unsupport failure is fixed.

**Trailing newline:** Fixed in .

**Everything else looks solid:**
- ✅ Ownership check via  in both handlers
- ✅ Idempotent support/unsupport actions
- ✅ Safe migration (, )
- ✅ Button disabled during in-flight request prevents double-submit
- ✅ Clean type additions to 
- ✅ No SQL injection risk (Supabase query builder)

One ultra-minor nit:  is also missing a trailing newline, but not blocking.

PR #246, opened 2026-05-31
feat(UNS-79): Add pagination to dashboard saved artists

brandonlucasgreen:
## Changes

Adds client-side pagination to the Saved Artists section on the dashboard.

- **Page size**: 10 artists per page
- **Pagination controls**: Previous / Next buttons with current page indicator (e.g. "2 / 3")
- **Hidden when ≤10 artists**: No pagination controls shown for small lists
- **Edge cases handled**:
  - Removing the last item on the last page auto-navigates to the previous valid page
  - Page numbers are clamped to valid range
  - Empty state still shows when no artists exist
- **Stable ordering**: Uses the same array order from the API, no re-sorting between pages

### DIFF
--- a/apps/web/src/pages/DashboardPage.tsx
+++ b/apps/web/src/pages/DashboardPage.tsx
@@ -1,4 +1,4 @@
-import { useState, useEffect } from 'react';
+import { useState, useEffect, useCallback } from 'react';
 import { Link, useNavigate, Navigate } from 'react-router-dom';
 import { useAuth } from '../contexts/AuthContext';
 
@@ -42,6 +42,8 @@ export function DashboardPage() {
   const [toast, setToast] = useState<{ message: string; type: 'success' | 'error' | 'info' } | null>(null);
   const [undoToast, setUndoToast] = useState<{ message: string; artistId: string; onUndo: () => void } | null>(null);
   const [supportingSlug, setSupportingSlug] = useState<string | null>(null);
+  const [currentPage, setCurrentPage] = useState(1);
+  const PAGE_SIZE = 10;
 
   useEffect(() => {
     if (authLoading) return;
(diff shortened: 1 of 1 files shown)


brandonlucasgreen (commented):
## Review: Dashboard Pagination (UNS-79)

Overall solid, focused change. Pagination logic is mostly correct and UI is consistent with existing dashboard design. However, I found one bug that should be fixed before merge.

---

### Requested Change: Stale closure in handleRemoveSavedArtist

**File:** `DashboardPage.tsx`, lines 124-130

The `currentPage` reference inside the `setSavedArtists` updater is a stale closure. If two rapid removals happen, the second will use the stale `currentPage`. Also, calling `setCurrentPage` inside another state updater is a React anti-pattern.

**Suggested fix:** Use a `useEffect` to clamp `currentPage` when `savedArtists.length` changes:

```tsx
useEffect(() => {
  const totalPages = Math.ceil(savedArtists.length / PAGE_SIZE);
  if (currentPage > totalPages && totalPages > 0) {
    setCurrentPage(totalPages);
  }
}, [savedArtists.length, currentPage]);
```

Then simplify `handleRemoveSavedArtist` to just filter:

```tsx
setSavedArtists(prev => prev.filter(a => a.artistId !== artistId));
```

---

### What looks good

- Page clamping in `goToPage` is correct and safe
- Pagination hidden when ≤10 artists (`totalPages > 1`)
- Empty state (`savedArtists.length === 0`) is preserved and rendered before pagination logic
- `paginatedArtists.slice(...)` bounds are correct even for the last partial page
- UI styling (buttons, spacing, color tokens) is consistent with dashboard
- No regressions: undo remove, toggle support, search bar all work as before

### Optional (non-blocking)

- Consider keyboard navigation for Previous/Next (e.g., arrow keys) for a11y

---

Happy to re-review once the stale closure is addressed!

PR #247, opened 2026-05-31
fix(UNS-80): Fix 403 error on artist claim verification

brandonlucasgreen:
## Problem
Artists getting a 403 error when trying to verify/claim their profile. Reported by Jack Saturn via Letterbird.

## Root Cause
The claim `start` action does an `upsert` that always sets `verified_at: null`. If a user revisits `/claim/:slug` after already completing verification, the upsert wipes their `verified_at`, causing the subsequent `PUT /api/artist-profile` to return 403 "Profile not yet verified".

## Fixes
- **Backend**: Check for existing verified profile before allowing claim re-start. Returns 409 with `alreadyVerified` flag if already verified.
- **Backend**: Add error handling to `verified_at` update in the verify action (was completely unhandled — silent failures).
- **Frontend**: Handle `alreadyVerified` response by skipping to done step instead of re-starting the claim flow.
- **Frontend**: Specific 403 error messages for unverified vs unauthorized cases in ClaimPage and ArtistEditPage.

## Files Changed
- `api/functions/claim-artist.ts` — guard against re-start + error handling
- `apps/web/src/pages/ClaimPage.tsx` — alreadyVerified handling + 403 messages
- `apps/web/src/pages/ArtistEditPage.tsx` — 403 error messages


brandonlucasgreen (commented):
**Review: Approved with minor notes**

## Guard logic ✅
The new check at lines 411-425 correctly prevents the claim `start` action from doing an upsert that would wipe `verified_at` on an already-verified profile. It queries `verified_at` (not `user_id`) and returns 409 with `alreadyVerified: true`. This is a clean, targeted fix for the root cause.

## Error handling ✅
The `verify` action now handles `verified_at` update failure as a fatal 500 (lines 586-598) and logs the `artist` table update failure as non-fatal (lines 601-612). Good separation: profile verification is the critical invariant, artist metadata update is best-effort.

## Frontend handling ✅
- **ClaimPage** (line 148): Handles `alreadyVerified` by skipping to `done` step. Good UX — user sees their claimed state instead of an error.
- **ClaimPage** (line 282) and **ArtistEditPage** (line 326): 403 messages correctly distinguish "not yet verified" from "unauthorized".

## Edge cases / notes

1. **Shadowed variable name** (line 412): The new `existingProfile` shadows the one declared at line 397. It works because they're in the same scope and the first isn't used after the second declaration, but it triggers a TS/linter warning on some configs. Consider renaming the second one to `verifiedProfile` or `existingVerifiedProfile` for clarity.

2. **Race condition potential**: The guard and the upsert are separate queries. In theory a concurrent request could slip through between the SELECT and the UPSERT. In practice, with `onConflict: 'artist_id'` and the check being best-effort, this is acceptable. If you want true safety, a partial unique index on `(artist_id) WHERE verified_at IS NOT NULL` plus relying on the upsert conflict would be the robust path, but that's out of scope for a bugfix.

3. **ClaimPage `done` step UX**: When `alreadyVerified` is hit, the user skips to `done` but `discoveredLinks` state is uninitialized (0). The `ClaimDoneStep` will show "0 links discovered". That's a minor UX wart — you might want to pass a message like "Your profile is already verified" instead of the default "discovered links" copy. Not blocking.

4. **Artist table update silently failing**: Line 610 logs but swallows the error. Consider surfacing this to the user or a monitoring alert, since a verified profile with an unclaimed artist row is an inconsistent state.

Overall this is a solid, focused fix. The root cause is clearly identified and the guard prevents the data-loss path. Ship it.

brandonlucasgreen (commented):
**Overall: Looks good, approving. 🎉**

This is a solid second pass that cleanly addresses the re-start wiping bug and improves error UX. A few minor notes:

**Backend (claim-artist.ts)**
- The guard at the `start` action (lines 414–424) is the right fix — verified profiles now get a 409 with `alreadyVerified: true` instead of silently getting wiped by the upsert. Good.
- The `verifyUpdateError` handling (lines 588–595) is correct and matches the pattern we already have elsewhere. Returning 500 with a clear retry message is the right call.
- The `artistUpdateError` handling (lines 603–606) is also pragmatic — logging and continuing instead of failing the whole flow is the right call since the profile is already verified.

**Frontend (ClaimPage.tsx)**
- `alreadyVerified` state and the skip-to-done logic (lines 149–155) works cleanly. No unnecessary network calls, no re-renders.
- The 403 error message branching (lines 284–290) is much better than the generic fallback. Good UX improvement.

**Frontend (ArtistEditPage.tsx)**
- Same 403 branching pattern (lines 326–332). Consistent with ClaimPage. Good.

**Frontend (ClaimDoneStep.tsx)**
- The conditional message for `alreadyVerified` (lines 11–18) is exactly what we need — avoids the awkward "0 links discovered" text for already-verified profiles.
- **Minor nit:** File is missing a trailing newline (diff shows `\ No newline at end of file`). Not a blocker, but worth fixing if there's another round.

**One thing I noticed:** The first guard (lines 400–409) already checks `existingProfile && existingProfile.verified_at && existingProfile.user_id !== userId`. The new second guard (lines 414–424) checks `existingClaimProfile?.verified_at` regardless of `user_id`. This means the *same* user who already verified will also hit the 409 and skip to done. That seems intentional and correct — if they’re already verified, they shouldn’t re-start the flow either. Just confirming that’s the intended behavior.

**Approve with one non-blocking nit:** add the trailing newline to `ClaimDoneStep.tsx` when convenient.

PR #250, opened 2026-06-02
fix(UNS-82): use claimedSlug when saving artist from search

brandonlucasgreen:
## Summary

When saving a verified-but-not-claimed artist from a search result, the
frontend was sending the synthetic search-result id (e.g. `nameonly-roberta-fidora`)
to `/api/saved-artists` instead of the verified slug (`roberta-fidora`). The
API's `findExistingArtist` couldn't find a matching row in the `artists`
table, so the saved record was stored with `artist_id = null` and the
dashboard fell back to 'Saved from Search'.

## Repro

1. Search 'roberta fidora'
2. Note the verified profile at `/a/roberta-fidora`
3. Click Save
4. `/dashboard` shows the artist but with 'Saved from Search' + search CTA

## Why it worked for some artists

For *claimed* artists the search id is `claimed-{slug}` (a real slug) so
`findExistingArtist` succeeds. For *verified-but-not-claimed* artists the
id is `nameonly-{normalized}` — a synthetic key that doesn't exist in the
`artists` table, so the API saves it as a generic search reference.

## Fix

Prefer `result.claimedSlug || result.id` everywhere we identify an artist
for save/remove/lookup in `ResultCard.tsx`:

- `handleSave` uses `saveId` for both `saveArtist` and `removeSavedArtist`
- `isArtistSaved` lookup uses the same id so the saved badge stays in sync
- `LoginInterstitial` receives the slug so the post-login pending save uses
  the correct identifier

No backend changes needed. `findExistingArtist` already does exact +
case-insensitive lookup. The upsert is idempotent on `user_id,artist_slug`,
so re-saving an existing row with the real slug correctly backfills the
`artist_id` FK.

## Verification

- `tsc -b`: clean
- `vitest run tests/unit`: 194/194 pass
- eslint: only the pre-existing `react-hooks/set-state-in-effect` warning
  on the unchanged `useEffect` block (was already there on main)

## Risks

Low. The only behavioral change is that `/dashboard` will now show the
real artist profile (including `/a/{slug}` link) for previously 'Saved
from Search' entries — which is the intended fix. No new writes or
migrations.

### DIFF
--- a/apps/web/src/components/ResultCard.tsx
+++ b/apps/web/src/components/ResultCard.tsx
@@ -39,9 +39,10 @@ export function ResultCard({ result, isAdmin, isSelected, onToggleSelect }: Resu
 
   useEffect(() => {
     if (result.type === 'artist' && result.id) {
-      setSaved(isArtistSaved(result.id));
+      // Match the slug we use for save so the saved state stays in sync
+      setSaved(isArtistSaved(result.claimedSlug || result.id));
     }
-  }, [result.type, result.id, isArtistSaved]);
+  }, [result.type, result.id, result.claimedSlug, isArtistSaved]);
 
   const handleSave = async (e: React.MouseEvent) => {
     e.stopPropagation();
@@ -49,15 +50,21 @@ export function ResultCard({ result, isAdmin, isSelected, onToggleSelect }: Resu
 
     if (result.type !== 'artist' || !result.id) return;
 
+    // Prefer the verified /a/{slug} over the synthetic search-result id.
+    // For verified-but-not-claimed artists the result id is a "nameonly-..." key
+    // that won't match a row in the artists table, so findExistingArtist fails
+    // and the saved record falls back to "Saved from Search".
+    const saveId = result.claimedSlug || result.id;
+
     if (saved) {
-      await removeSavedArtist(result.id);
+      await removeSavedArtist(saveId);
       setSaved(false);
     } else {
       if (!session) {
         setShowLoginInterstitial(true);
         return;
       }
-      await saveArtist(result.id, undefined, result.name, result.imageUrl);
+      await saveArtist(saveId, undefined, result.name, result.imageUrl);
       setSaved(true);
     }
   };
@@ -176,7 +183,7 @@ export function ResultCard({ result, isAdmin, isSelected, onToggleSelect }: Resu
 
       {showLoginInterstitial && (
         <LoginInterstitial
-          artistId={result.id}
+          artistId={result.claimedSlug || result.id}
           artistName={result.name}
           onClose={() => setShowLoginInterstitial(false)}
         />


brandonlucasgreen (commented):
## Principal Review: fix(UNS-82): use claimedSlug when saving artist from search

### Correctness ✅

The fix is correct and directly addresses the repro from UNS-82:

- **Root cause:** For verified-but-not-claimed artists, the search result `id` is a synthetic key like `nameonly-roberta-fidora` that doesn't exist in the `artists` table. The API's `findExistingArtist` does exact then case-insensitive `slug` lookup against the `artists` table. When it can't match the synthetic id, `artist_id` is stored as `null`, and the dashboard renders the row as "Saved from Search".
- **Fix:** Using `result.claimedSlug || result.id` means verified artists (which have a real slug in `claimedSlug`) now send the real slug to `/api/saved-artists`, allowing `findExistingArtist` to resolve the row and set `artist_id` properly. Claimed artists already worked because their search id is `claimed-{slug}`, which happens to match.

### Coordinated changes ✅

All three call sites are updated consistently:
1. `handleSave` — `saveArtist` and `removeSavedArtist` both use `saveId`
2. `isArtistSaved` effect — uses the same derived id, and `result.claimedSlug` is correctly added to the dependency array
3. `LoginInterstitial` prop — receives the same slug so the post-login pending save uses the correct identifier (the pending save flow in `AuthContext` reads `localStorage.getItem('pendingSave')` and calls `saveArtist(artistId, notes)`)

### Edge cases

- **Non-verified artist (no `claimedSlug`):** Falls back to `result.id`, which is the pre-existing behavior. No regression.
- **Quick toggle save/unsave:** The optimistic updates in `AuthContext` (`savedArtistIds` Set and `savedArtists` array) are keyed by the same `artistId` string. Since the slug is consistent between save and remove, optimistic add/remove will match and rollback correctly. No new race condition introduced.
- **Pre-existing "Saved from Search" rows:** The PR description correctly notes the upsert is idempotent on `user_id,artist_slug`. Re-saving an existing row with the real slug will backfill `artist_id` because `findExistingArtist` now succeeds. The `onConflict: 'user_id,artist_slug'` clause in `handleSave` handles this cleanly. Good.

### Security ✅

No SSRF or injection concerns. `claimedSlug` originates from the database (`artists.slug`), not user input. It is passed through as a string identifier to Supabase queries using parameterized `.eq('slug', slug)` and `.ilike('slug', slug)`. No raw string interpolation into SQL.

### Code quality

- **Minor cleanup suggestion (non-blocking):** The `saveId` extraction is inlined in `handleSave`. Consider extracting a small helper like `const getArtistSaveId = (r: SearchResult) => r.claimedSlug || r.id;` to avoid the `||` repetition across three sites. Not a blocker for merge.
- **No AI-slop noise:** Comments are concise and explain *why* (the "nameonly-..." synthetic key problem), not just *what*. Good.

### The eslint `react-hooks/set-state-in-effect` warning

This is **pre-existing on main** — it was already firing on the old `setSaved(isArtistSaved(result.id))` line. The PR only changes the argument passed to `isArtistSaved`; it doesn't introduce the pattern. Fixing it properly (deriving `saved` state from props or lifting it into `AuthContext`) is a larger refactor that should be its own issue, not shoehorned into this bugfix.

**Verdict:** Leave the warning alone for this PR. It's existing debt.

### Overall

Low-risk, targeted fix. Correct root-cause analysis, consistent changes, no new edge cases, no security issues. Leaves the codebase no worse than it was (the eslint warning predates this branch).

**Approve with minor suggestion.**

PR #272, opened 2026-06-14
feat(UNS-101): add /api/artist-page endpoint for the rich profile payload

brandonlucasgreen:
## UNS-101: API contract for `/api/artist-page?slug=...`

Part of parent issue **UNS-100** (collapse edge/SPA bifurcation for artist pages).

### What this does

Adds a new Netlify function `GET /api/artist-page?slug=...` that returns the **full rich-profile payload** as JSON — the same data the edge function currently renders as HTML, but now available as a clean API for the React SPA to consume (UNS-102).

### Changes

- **`api/functions/artist-page.ts`** (new, ~145 lines) — handler returning `ArtistPagePayload` JSON with artist metadata, profile, links (split into main/social), payout info, Bandcamp Friday flag, and sanitized featured embed
- **`api/functions/db.ts`** — added `getArtistProfileBySlug(slug)` returning artist row + profile row + links in parallel
- **`api/shared/bandcamp-friday.ts`** (new) — extracted `BANDCAMP_FRIDAY_DATES` and `isBandcampFriday()` from edge function so both the new endpoint and future no-JS fallback (UNS-105) can use it
- **`api/functions/artist-profile.ts`** — exported `sanitizeEmbed` and `ALLOWED_EMBED_DOMAINS` for reuse
- **`api/edge/claimed-artist-page.ts`** — updated to import from shared module (no behavior change)
- **`netlify.toml`** — added redirect for `/api/artist-page`

### API contract

```typescript
interface ArtistPagePayload {
  artist: { id, slug, name, imageUrl, matchConfidence, country, countryCode, city };
  profile: { bio, customImageUrl, featuredEmbed, verifiedAt } | null;
  links: Array<{ platform, url, displayName, payoutPercent, bandcampFriday }>;
  socialLinks: Array<{ platform, url, displayName }>;
  bandcampFriday: boolean;
}
```

### Design decisions

- **Bandcamp Friday as shared module** — extracted from the edge function so UNS-105 (no-JS fallback) can also use it
- **Separate main/social link arrays** — matches the edge function's split; the React component can render them differently
- **Payout strings match edge function** — uses platform registry `payoutPercent` with `~97%` override on Bandcamp Fridays
- **Cache headers** — `s-maxage=300, stale-while-revalidate=86400` (5-min CDN cache, day-long stale revalidation) vs the existing 60s on `/api/artist`
- **404 for unclaimed** — the endpoint is specifically for the rich profile; unclaimed artists fall through to a different SPA code path

### Verification

- ✅ `npm run build` — clean
- ✅ `npm run test:unit` — 244/244 passing
- Manual endpoint test pending deployment

Refs: UNS-100, UNS-101

### DIFF
--- a/api/edge/claimed-artist-page.ts
+++ b/api/edge/claimed-artist-page.ts
@@ -1,16 +1,7 @@
 import { Context } from "https://edge.netlify.com";
 import { createClient } from "https://esm.sh/@supabase/supabase-js@2";
 import { PLATFORMS } from "../shared/platform-registry.ts";
-
-// UPDATE ANNUALLY: Bandcamp Friday dates from https://daily.bandcamp.com/features/bandcamp-fridays
-const BANDCAMP_FRIDAY_DATES = [
-  '2026-03-06', '2026-05-02', '2026-08-07',
-  '2026-09-04', '2026-10-02', '2026-11-06', '2026-12-04',
-];
-function isBandcampFriday(): boolean {
-  const pacificDate = new Date().toLocaleDateString('en-CA', { timeZone: 'America/Los_Angeles' });
-  return BANDCAMP_FRIDAY_DATES.includes(pacificDate);
-}
+import { BANDCAMP_FRIDAY_DATES, isBandcampFriday } from "../shared/bandcamp-friday.ts";
 
 // Derive PLATFORM_INFO from shared registry (name, color, icon, category, payoutPercent)
 const PLATFORM_INFO: Record<string, { name: string; color: string; icon: string; category: string; payoutPercent?: string }> = {};
--- a/api/edge/claimed-artist-page.ts
+++ b/api/edge/claimed-artist-page.ts
@@ -1,7 +1,7 @@
 import { Context } from "https://edge.netlify.com";
 import { createClient } from "https://esm.sh/@supabase/supabase-js@2";
 import { PLATFORMS } from "../shared/platform-registry.ts";
-import { BANDCAMP_FRIDAY_DATES, isBandcampFriday } from "../shared/bandcamp-friday.ts";
+import { isBandcampFriday } from "../shared/bandcamp-friday.ts";
 
 // Derive PLATFORM_INFO from shared registry (name, color, icon, category, payoutPercent)
 const PLATFORM_INFO: Record<string, { name: string; color: string; icon: string; category: string; payoutPercent?: string }> = {};
--- a/api/functions/artist-page.ts
+++ b/api/functions/artist-page.ts
@@ -1,0 +1,145 @@
+// API endpoint: GET /api/artist-page?slug=...
+// Returns the full rich-profile payload for a claimed artist as JSON.
+// This is the data source for the React SPA artist page (UNS-102).
+
+import { getArtistProfileBySlug } from './db';
+import { checkRateLimit, getClientIp } from './ratelimit';
+import { PLATFORMS } from '../shared/platform-registry';
+import { isBandcampFriday } from '../shared/bandcamp-friday';
+import { sanitizeEmbed } from './artist-profile';
+
+const CORS_HEADERS: Record<string, string> = {
+  'Content-Type': 'application/json',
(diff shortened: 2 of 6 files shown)


brandonlucasgreen (commented):
**Senior review — UNS-101 `/api/artist-page`**

Solid, well-scoped extraction. Sanitizing the embed and routing through `checkRateLimit`/`getClientIp` is the right call, and the shared `bandcamp-friday` module is clean. Two divergences from the edge function worth resolving before UNS-102 consumes this, plus minor nits.

### Regression / API contract drift

**1. `verified_at` gate is dropped — `needs-manual-verification` (likely a real divergence).**
The SSR edge function only renders the rich page when `profile?.verified_at` is set (`claimed-artist-page.ts:56` — `if (!profile?.verified_at) return context.next();`). `getArtistProfileBySlug` gates only on `match_confidence === 'claimed'`, so this endpoint returns a 200 rich payload for claimed artists whose profile is missing or unverified — cases where the SSR page deliberately falls through to the SPA. If UNS-102 keys "show rich profile" off a 200 here, claimed-but-unverified artists will render content the SSR path hides. Either add the `verified_at` requirement to match, or confirm the looser gate is intentional and document it (the PR notes "404 for unclaimed" but is silent on unverified).

**2. `displayName` precedence is inverted vs. the edge function — `confirmed`.**
For known platforms the edge function prefers the registry name: `info?.name || escapeHtml(p.display_name || p.platform)` (`claimed-artist-page.ts:100`, and social at :122 uses `info.name` outright). This endpoint prefers the artist-set value: `l.display_name || info?.name` (main) and `l.display_name || info?.name` (social). So a known platform with a custom `display_name` now surfaces the custom name where SSR shows the canonical registry name. Given the PR's stated parity goal, align the ordering (or note the deliberate change). The `'other'` fallback also differs (edge → `'Link'`, here → `null`) — acceptable if the SPA supplies its own fallback, but worth a deliberate decision.

### Security

- auth/authorization: endpoint returns only already-public profile data — no auth needed. OK.
- input validation: `slug` flows into a parameterized `.eq('slug', slug)`, no injection. `confirmed` safe.
- CORS `*`: consistent with the other public web-app functions (`artist-lookup`, `artist-profile`, etc.), not the v1 API model — OK.
- secret handling / data exposure: payload mirrors what SSR already exposes (`id`, `verifiedAt`, etc.). No new exposure. OK.

### Nits
- `artist-page.ts` and `bandcamp-friday.ts` are missing trailing newlines.
- `profile` fetch uses `.single()`; a missing profile row resolves to `data: null` (handled) but logs a PostgREST error server-side — harmless, just noise.

Nothing here is blocking once #1 is confirmed and #2 is reconciled.

---

brandonlucasgreen:
## Dan's review fixes applied ✅

Pushed `2fd6c98` addressing Dan's blocking issue.

**Fix:** Dropped the unused `BANDCAMP_FRIDAY_DATES` import in `api/edge/claimed-artist-page.ts:4`. Only `isBandcampFriday()` is called in the file. The constant is still exported from `api/shared/bandcamp-friday.ts` for any future consumer that wants to enumerate the dates (e.g., a "next Bandcamp Friday" banner in the no-JS fallback).

**Verification:** `npm run build` clean, 244/244 unit tests pass.

**Dan, on the open questions for UNS-102:**

1. **`locationText`** (derived from `city/country/countryCode`) — UNS-102 can compute this client-side. No API change needed.
2. **JSON-LD structured data** + **OG / Twitter Card metadata** — these are for crawlers and belong in the no-JS fallback (UNS-105), not in the React component. UNS-105's spec already says it renders the rich profile as static HTML, and I'll update it to explicitly include JSON-LD + OG meta tags. The SPA itself doesn't need these fields — crawlers won't see client-side-rendered content anyway, and the no-JS edge function (or a future SSR layer) is the right place to inject them.
3. **`imageUrl` precedence** — confirmed the `null` is a small improvement over the edge function's empty-string fallback. No action needed.

Thanks for the careful review. Ready to merge.

PR #277, opened 2026-06-17
feat(UNS-105+104): no-JS static-HTML edge function for /a/*, remove old claimed-artist-page

brandonlucasgreen:
## Goal

Replace the old `claimed-artist-page` edge function (JS-dependent, claimed-only) with a new `artist-page-static` edge function that serves fully-rendered static HTML for **both** claimed and unclaimed artists at `/a/*` — with zero JavaScript in the response. This makes `/a/{slug}` accessible to users on slow networks, JS-disabled browsers, and crawlers that don't execute JS.

## What changed (5 files)

- **`api/edge/artist-page-static.ts`** — New no-JS static HTML edge function for `/a/*` (claimed + unclaimed)
- **`api/edge/claimed-artist-page.ts`** — Deleted (452 lines, replaced by `artist-page-static.ts` in this same PR)
- **`netlify.toml`** — Swap `/a/*` edge function from `claimed-artist-page` to `artist-page-static`
- **`apps/web/vite.config.ts`** — Remove `/artist/*` from `navigateFallbackDenylist` (old redirect target no longer used)
- **`CLAUDE.md`** — Update edge functions list (`claimed-artist-page` → `artist-page-static`)

## Justified deletion: `claimed-artist-page.ts`

The 452-line `claimed-artist-page.ts` is **replaced by `artist-page-static.ts` in this same PR**. The old function:

- Only handled claimed artists (returned `context.next()` for unclaimed, leaving no-JS users with an empty shell)
- Included a theme-toggle `<script>` block and auth localStorage reader — both broken without JS
- Included embed widget JS and analytics JS — all removed in the new function

The new `artist-page-static.ts` handles both claimed and unclaimed artists with zero JS, making it a strict superset of the old function.

## XSS safety

1. **`escapeHtml()` on every user-supplied string** — artist name, bio, location, link URLs, link display names, image URLs, meta descriptions
2. **Featured embed iframe allowlist** — only domains in `ALLOWED_EMBED_DOMAINS` (bandcamp.com, youtube.com, youtube-nocookie.com, soundcloud.com, open.spotify.com, embed.music.apple.com, w.soundcloud.com, player.vimeo.com, embed.tidal.com) can be embedded; all others get a safe placeholder
3. **Sandbox attribute** — `<iframe sandbox="allow-scripts allow-same-origin">` per Brandon's sign-off
4. **JSON-LD** — `<` characters escaped to `\u003c` to prevent closing the script tag

## Definition of done

[...]

PR #278, opened 2026-06-17
feat: consolidate /artist/* route to artist-page-static, remove Login link from static header (UNS-109)

brandonlucasgreen:
### Problem

Production smoke test of #277 surfaced two header issues on static artist pages:

1. **Auth-bar inconsistency on `/artist/{slug}`** — the legacy `artist-page` edge function rendered a static "Login" link AND an inline-JS auth bar. Logged-in users saw both at once.
2. **"Login" link on `/a/{slug}`** — `artist-page-static` always showed "Login", but per the public-page framing (these are artist flyers for external audiences, not in-app navigation), a Login affordance is wrong anyway.

### Changes

| File | Change |
|------|--------|
| `netlify.toml` | Repoint `/artist/*` from `artist-page` → `artist-page-static` |
| `api/edge/artist-page-static.ts` | Remove `<a href="/login">Login</a>` and empty `<div class="nav-right">` from both claimed and unclaimed header branches |
| `api/edge/artist-page-static.ts` | Fix slug parsing to handle both `/a/` and `/artist/` prefixes (since both routes now point here) |
| `api/edge/artist-page.ts` | **Deleted** (393 lines) — legacy edge function, no longer referenced |
| `CLAUDE.md` | Update edge functions docs: `artist-page-static` now serves both `/artist/*` and `/a/*`; remove stale `claimed-artist-page` reference |

### Intentional deletion note

The semantic-revert bot will likely flag `api/edge/artist-page.ts` deletion since the file was recently modified. **This deletion is intentional** — it is the entire point of this PR. The legacy function is replaced by `artist-page-static`, which is a strict superset (handles both claimed AND unclaimed artists, with proper SEO, no-JS rendering, and Bandcamp Friday support).

### Build status

- ✅ Clean build (302/302 unit tests pass)
- ✅ Semantic revert check: passed

### Curl verification results (deploy preview 278)

**Test 1: Same HTML structure on both URLs** ✅
```
diff <(grep -v 'canonical\|og:url' /tmp/artist.html) <(grep -v 'canonical\|og:url' /tmp/a.html)
→ Empty diff (structurally identical pages)
```

**Test 2: No Login/Dashboard in static header** ✅
```
/a/all-time-low → PASS (no login/sign-in/dashboard/auth-bar found)
/artist/all-time-low → PASS (no login/sign-in/dashboard/auth-bar found)
```

**Test 3: No executable `<script>` tags** ✅
```
/a/all-time-low → 0 executable scripts (only JSON-LD)
/artist/all-time-low → 0 executable scripts (only JSON-LD)
```

**Test 4: Footer has navigation links** ✅
```
Footer contains: /artists (Indie Artist Index), Roadmap, Support, Privacy policy, etc.
```
Note: No `/login` link in footer either — this is by design from #277. The footer provides Index/Roadmap/Support/Privacy for navigation into the app.

### Deviation from spec

One minor deviation: `api/edge/artist-page-static.ts` line 109 needed to change from `url.pathname.replace(/^\/a\//, '')` to `url.pathname.replace(/^\/(a|artist)\//, '')` — since both `/artist/*` and `/a/*` now route to this function, the slug parser must handle both prefixes. The canonical URL (`og:url`, `<link rel="canonical">`) correctly points to `/a/{slug}` for SEO regardless of which URL the user visited.

---

**Reviewer:** Brandon (via Claude Code) — NOT Dan


brandonlucasgreen (commented):
## Senior review — UNS-109 route consolidation

One confirmed blocking regression; everything else looks clean.

### 🔴 Regression (confirmed) — `/artist/*` slug is never parsed; SSR silently breaks for that route

`api/edge/artist-page-static.ts:109`:

```ts
const slug = url.pathname.replace(/^\/a\//, '').replace(/\/$/, '');
```

This PR repoints `/artist/*` → `artist-page-static` (`netlify.toml:17`), but the slug parser only strips the `/a/` prefix. For a request to `/artist/foo`:

- `^\/a\//` does **not** match `/artist/`, so `slug` becomes `"/artist/foo"`
- the lookup `.eq('slug', '/artist/foo')` returns no row → `if (!artistData) return context.next()`
- the request falls through to the SPA, so **`/artist/{slug}` loses its SSR/SEO HTML entirely** — the exact thing this edge function exists to serve

The deleted `artist-page.ts` parsed this route as `^\/artist\/([a-z0-9-]+)\/?$`, so this is a behavioral regression for crawlers / no-JS clients on `/artist/*`. Secondary: `pageUrl`/canonical (line 176) is hardcoded to `/a/${slug}`, so even a coincidental hit would emit a wrong canonical.

**Fix — strip both prefixes:**

```ts
const slug = url.pathname.replace(/^\/(a|artist)\//, '').replace(/\/$/, '');
```

The PR's stated goal ("artist-page-static now serves both `/artist/*` and `/a/*`") isn't met until the parser handles both. The "build clean / 302 tests pass" check wouldn't catch this — there's no edge-function route test. Please curl `/artist/{known-slug}` on the deploy preview and confirm SSR HTML (not the SPA shell) comes back before merge.

### Security — no issues found
- **Auth/authz:** none changed. Removing the static "Login" link is cosmetic; no gating logic touched.
- **Input validation:** slug is used only in a parameterized `.eq()` and HTML-escaped before render (`escapeHtml`). No new injection surface.
- **Secret handling:** `SUPABASE_*` still read from env, absence falls through to SPA. Unchanged.
- **Injection / data exposure / unsafe defaults:** no new external fetch, no new client data. Clean.

### Other regression checks
- **API contract drift:** `/a/*` behavior unchanged. `/artist/*` is the drift — see above.
- **Null/empty:** empty slug still short-circuits via `if (!slug) return context.next()`. Fine for `/a/`; for `/artist/` the bug is the non-empty-but-wrong slug, not empty.
- **Deletion of `artist-page.ts`:** confirmed intentional and safe — no remaining references after the `netlify.toml` repoint; `artist-page-static` is a functional superset once the slug fix lands.
- **CLAUDE.md doc update:** accurate.

Bottom line: one-line slug fix + a preview smoke test of `/artist/{slug}`, then good to go.

---

PR #281, opened 2026-06-20
feat(UNS-111): Extension auth + saved artists sync

brandonlucasgreen:
Fixes UNS-111

## What this does

Adds Supabase auth and saved artists sync to the Unstream browser extensions (Chrome + Firefox). This is Phase 2 of UNS-61 (cross-client auth), building on the server-side schema and sync API shipped in UNS-93.

### Changes

**New file: `apps/extension/lib/supabase.js`**
- Implements Supabase auth REST API directly (no SDK dependency — the extension has no build step, so we can't import @supabase/supabase-js)
- Session persistence in `chrome.storage.local` for service worker cold start resilience
- Functions: `signInWithPassword`, `signInWithOtp`, `signOut`, `getStoredSession`, `getAccessToken`, `handleMagicLinkCallback`, `getDeviceId`
- Stable per-installation device ID via `crypto.randomUUID()`

**Modified: `apps/extension/popup/popup.html`**
- Auth form (email + password) in the Saved tab, shown when signed out
- "Send magic link instead" button
- "Check your email" confirmation state after magic link sent
- Logged-in state: email display + sign out button + "Sync now" button

**Modified: `apps/extension/popup/popup.css`**
- Auth form styles (inputs, submit button, error message)
- Magic link styles
- Auth bar (logged-in email + sign out)
- Sync button styles
- Saved artist image (32px circular avatar)
- Name/location flex layout update

**Modified: `apps/extension/popup/popup.js`**
- Auth state management: `initAuth()`, `showLoggedInState()`, `showLoggedOutState()`
- Email+password sign-in via Supabase REST API
- Magic link email sending (shows confirmation, no redirect capture yet)
- Sign out clears synced artists, preserves local-only saves
- `syncSavedArtists()`: calls GET /api/saved-artists/sync?since=<timestamp>, merges server artists into local state
- `saveArtistToServer()`: POST /api/saved-artists with device_id and last_modified on every local save when logged in
- `removeArtistFromServer()`: POST /api/saved-artists with action=remove on every local unsave when logged in
- Saved artist cards now display image URLs and synced names from server data
- **NEW:** `lookupSlug()` resolves artist display names to canonical slugs via `/api/search/sources`, matching the web app's `claimedSlug || id` path. Results are cached in `chrome.storage.local`.
- **NEW:** All save/remove operations now use the resolved slug as the key, preventing duplicate rows across web and extension clients.

**Modified: `apps/extension/background/service-worker.js`**
- Message handlers: AUTH_GET_SESSION, AUTH_MAGIC_LINK_CALLBACK, AUTH_SIGN_OUT, AUTH_GET_TOKEN
- All delegate to `supabase.js` functions which read from chrome.storage.local

**Modified: `apps/extension/manifest.json` + `manifest-firefox.json`**
- Added `identity` permission for future magic link redirect support
- **NEW:** Added `https://*.supabase.co/*` to `host_permissions` so auth calls succeed without relying on permissive CORS.

## Security

- The Supabase anon key shipped in `supabase.js` is a **public** publishable key — this is expected and by design. All data protection is via Row-Level Security (RLS) on the `saved_artists` table, which scopes all operations by `auth.uid()` (see UNS-93 migration). No service-role key is present in the extension bundle.

## Known limitations

- **Magic link redirect capture not yet implemented.** The "Send magic link instead" button sends the OTP email, but the extension cannot capture the redirect because the extension's redirect URL (`https://<id>.chromiumapp.org/`) is not in the Supabase dashboard's allowed redirect URLs. Full magic link support requires adding the extension's redirect URL to the Supabase dashboard. Email+password auth works fully.
- **Cross-client deletions don't propagate.** When an artist is unsaved on the web app, the extension won't learn about the deletion until Phase 4 (realtime subscriptions) or a tombstone scheme is implemented. The incremental sync endpoint (`/api/saved-artists/sync?since=<cursor>`) only returns rows modified since the cursor — deleted rows simply disappear from the result set. This is a v1 limitation documented for future work.
- **No realtime subscriptions yet** (deferred to Phase 4 per UNS-61).

## Acceptance criteria status

- [x] Email + password sign-in works from the popup
- [ ] Magic link sign-in — sends email but does not capture redirect (requires Supabase dashboard config)
- [x] Saved tab shows previously-saved artists on first load after auth
- [x] Save/unsave from extension reflects in the web app on next sync (fixed: now uses canonical slug, same as web app)
- [x] Refresh the popup: state persists, no flicker to login form
- [x] Sign out clears local state and returns to login form
- [x] Service worker cold-start preserves session (fixed: token refresh now triggers 60s before expiry, not after)
- [ ] Firefox build parity — manifest updated but not tested on Firefox profile yet
- [x] npm run build passes
- [x] Deslop pass completed on branch diff

## What to verify

1. Load the extension in Chrome (chrome://extensions → developer mode → load unpacked → select apps/extension/)
2. Open the popup, go to Saved tab — should show login form
3. Sign in with email+password — should show "Logged in as <email>" and sync saved artists
4. Save an artist from Discover tab — should sync to server using the canonical slug (not display name)
5. Verify on the web app that the saved artist uses the slug key, not the display name
6. Close and reopen popup — session should persist
7. Sign out — should return to login form and clear synced artists
8. Test magic link button — should send email but note redirect capture limitation

### DIFF
--- a/apps/extension/background/service-worker.js
+++ b/apps/extension/background/service-worker.js
@@ -1,7 +1,8 @@
 // Unstream Chrome Extension - Service Worker
-// Handles API calls, caching, badge updates, and release alerts
+// Handles API calls, caching, badge updates, release alerts, and auth
 
 import { ALLOWED_RELEASE_DOMAINS } from '../lib/constants.js';
+import { getStoredSession, handleMagicLinkCallback, getAccessToken, signOut } from '../lib/supabase.js';
 
 const API_BASE = 'https://unstream.stream/api';
 const CACHE_TTL = 5 * 60 * 1000; // 5 minutes
@@ -79,6 +80,18 @@ chrome.runtime.onMessage.addListener((message, sender, sendResponse) => {
     return true;
   } else if (message.type === 'DISMISS_RELEASE') {
     dismissRelease(message.releaseId).then(sendResponse);
+    return true;
+  } else if (message.type === 'AUTH_GET_SESSION') {
+    getStoredSession().then(session => sendResponse({ session })).catch(() => sendResponse({ session: null }));
+    return true;
+  } else if (message.type === 'AUTH_MAGIC_LINK_CALLBACK') {
+    handleMagicLinkCallback(message.url).then(session => sendResponse({ session })).catch(() => sendResponse({ session: null }));
+    return true;
+  } else if (message.type === 'AUTH_SIGN_OUT') {
+    signOut().then(() => sendResponse({ success: true })).catch(() => sendResponse({ success: true }));
+    return true;
+  } else if (message.type === 'AUTH_GET_TOKEN') {
+    getAccessToken().then(token => sendResponse({ token })).catch(() => sendResponse({ token: null }));
     return true;
   }
 });
--- a/apps/extension/lib/supabase.js
+++ b/apps/extension/lib/supabase.js
@@ -26,8 +26,8 @@ export async function getStoredSession() {
   const result = await chrome.storage.local.get(SESSION_KEY);
   const session = result[SESSION_KEY];
   if (!session || !session.access_token) return null;
-  // Check expiry (with 60s buffer)
-  if (session.expires_at && session.expires_at * 1000 < Date.now() - 60000) {
+  // Check expiry (refresh 60s before it actually expires)
+  if (session.expires_at && session.expires_at * 1000 < Date.now() + 60000) {
     // Token expired — try refresh
     const refreshed = await refreshSession(session);
     if (refreshed) return refreshed;
--- a/apps/extension/lib/supabase.js
+++ b/apps/extension/lib/supabase.js
@@ -1,0 +1,222 @@
+// Supabase auth client for the Unstream browser extension.
+// Implements the auth REST API directly (no SDK dependency) because
+// the extension has no build step and can't import @supabase/supabase-js.
+// Tokens are persisted to chrome.storage.local so users stay signed in
+// across browser restarts and service worker cold starts.
+
+const SUPABASE_URL = 'https://bwogclqzpsbvqbyhhqbz.supabase.co';
+const SUPABASE_ANON_KEY = 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJzdXBhYmFzZSIsInJlZiI6ImJ3b2djbHF6cHNidnFieWhocWJ6Iiwicm9sZSI6ImFub24iLCJpYXQiOjE3NzMzMTMyODcsImV4cCI6MjA4ODg4OTI4N30.AUcgWjqcsIbcTm-RkjaY2jtVMYmAHaPVE52oGeOsblM';
+
+const AUTH_URL = `${SUPABASE_URL}/auth/v1`;
+
+function authHeaders(accessToken) {
+  const headers = {
+    'apikey': SUPABASE_ANON_KEY,
+    'Content-Type': 'application/json',
+  };
+  if (accessToken) headers['Authorization'] = `Bearer ${accessToken}`;
+  return headers;
+}
+
+// ---- Session persistence ----
+
+const SESSION_KEY = 'unstream_auth_session';
+
+export async function getStoredSession() {
+  const result = await chrome.storage.local.get(SESSION_KEY);
+  const session = result[SESSION_KEY];
+  if (!session || !session.access_token) return null;
+  // Check expiry (with 60s buffer)
+  if (session.expires_at && session.expires_at * 1000 < Date.now() - 60000) {
+    // Token expired — try refresh
+    const refreshed = await refreshSession(session);
+    if (refreshed) return refreshed;
+    // Refresh failed — clear session
+    await clearSession();
+    return null;
+  }
+  return session;
+}
+
+async function storeSession(session) {
+  if (session) {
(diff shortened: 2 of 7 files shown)


brandonlucasgreen (commented):
## Staff review — UNS-111 extension auth + saved artists sync

Reviewed the diff against the existing server contract (`saved-artists.ts`, `saved-artists-sync.ts`) and the web client (`AuthContext` / `ResultCard`). The sync **response** contract (`{ artists, server_time }`) lines up correctly. The findings below are about the **write/merge** side and a couple of auth edge cases. Each is labelled `confirmed` / `plausible` / `needs-manual-verification`.

### Regression / correctness

**1. Identifier mismatch between extension and web — duplicate rows, broken cross-client unsave — `confirmed` (high)**
The extension uses the raw artist **display name** as the server key:
- `currentArtist = track.artist` (`popup.js:428`) → `saveArtistToServer(currentArtist, …)` sends `artistId: artistName` → server stores it as `artist_slug` (`saved-artists.ts:188`, `:202`).

The web app uses a real slug:
- `const saveId = result.claimedSlug || result.id` → `saveArtist(saveId, …)` (`ResultCard.tsx:57,67`).

So saving *Aphex Twin* on the web (`artist_slug = "aphex-twin"`) and from the extension (`artist_slug = "Aphex Twin"`) creates **two separate rows**. `upsert(onConflict: 'user_id,artist_slug')` won't dedup them, and unsaving on one client won't remove the other. This directly contradicts the AC you checked off — "Save/unsave from extension reflects in the web app on next sync." The extension should send the same slug identifier the web client uses, or normalize the name to a slug before sending.

**2. Token-expiry buffer is inverted — `confirmed`**
`supabase.js:30`:
```js
if (session.expires_at && session.expires_at * 1000 < Date.now() - 60000) {
```
The comment says "60s buffer" (refresh *before* expiry), but this condition only treats the token as expired once it's **60s past** `expires_at` — the token is returned as valid for up to a minute after it has actually expired, producing a ~60s window of 401s on every refresh cycle. It should be `Date.now() + 60000` (refresh 60s early).

**3. `syncedArtistSlugs` is overwritten on every incremental sync — `confirmed`**
In `syncSavedArtists`, `syncedSlugs` is built only from `serverArtists`, which on an incremental pull (`since` set) contains *only rows changed since the last cursor*. It then overwrites the stored `syncedArtistSlugs` wholesale (`popup.js`, the `chrome.storage.local.set({ …, syncedArtistSlugs: syncedSlugs })` call). Previously-synced slugs are lost from that list. Consequence on sign-out: `clearSyncedArtists` only removes the *last* batch, so earlier server-synced artists are left behind and silently treated as local-only saves (and can duplicate after re-login). The synced set needs to be a union with the existing stored value, not a replacement.

**4. Deletions never propagate down — `confirmed` (limitation)**
The incremental sync pulls rows with `last_modified > since`. A row **deleted** on another client simply disappears from the result set, so the extension never learns to remove it — unsaving on the web won't clear it from the extension. Cross-client delete needs tombstones or a separate removed-ids list. Worth calling out explicitly in "Known limitations" if it's deferred.

**5. `authSession` tri-state race — `plausible` (minor)**
`authSession` is `null` during `initAuth()` and `false` when signed out, but `saveArtistToServer` / `syncSavedArtists` guard with `if (!authSession)` — so a save fired in the brief window before init completes is silently dropped from the server (local save still happens). Low impact; flagging for awareness.

### Security

**6. Hardcoded Supabase URL + anon key — `needs-manual-verification` (not a leak by itself)**
`supabase.js:7-8` ship the project URL and `anon` key in the extension bundle. That's expected — the publishable anon key is designed to be public, and protection rests entirely on RLS. Please confirm `saved_artists` RLS (from UNS-93) actually scopes by `auth.uid()` for select/insert/update/delete, since the anon key is now distributed to every extension user. No service-role key is present, which is correct.

**7. Synced `imageUrl` set directly as `img.src` — `plausible` (minor, defense-in-depth)**
`popup.js` does `img.src = imageUrl` from server data without scheme validation. It's the user's own RLS-scoped data and `img.src` can't execute script, so risk is low, but consider restricting to `https://` to avoid `http`/`data:` surprises and mixed-content noise.

**8. Input validation server-side — `confirmed` OK**
`handleSave` validates `last_modified` is a parseable date and clamps `device_id` to 128 chars (`saved-artists.ts:216-228`), and the Supabase client parameterizes queries — no injection surface introduced. Good.

### Functional / packaging

**9. `host_permissions` missing the Supabase host — `needs-manual-verification`**
Both the popup and the service worker now `fetch('https://bwogclqzpsbvqbyhhqbz.supabase.co/auth/v1/…')`, but `host_permissions` lists only `unstream.stream` / `unstream.dev`. Under MV3 this works only if Supabase returns permissive CORS for the `chrome-extension://` origin. Please verify the auth calls actually succeed from a loaded build in **both** Chrome and Firefox (the Firefox parity box is unchecked), or add `https://*.supabase.co/*` to `host_permissions` to be safe. The new `identity` permission is also currently unused (magic-link redirect deferred) — fine to keep for the follow-up, just noting it's requesting a permission nothing reads yet.

---
Net: auth plumbing and the read/merge response contract look sound, but **#1 (identifier mismatch)** and **#3 (synced-slug overwrite)** undermine the core cross-client promise of this PR and I'd treat them as blocking before merge. #2 and #4 are real but lower-urgency.

---

PR #282, opened 2026-06-20
feat(UNS-112): Mac/iOS app sync — saved artists + auth

brandonlucasgreen:
## What

Add saved-artists sync to the Mac menu bar app and iOS app (UNS-112, Phase 3 of UNS-61 cross-client auth).

- **supabase-swift v2.44.1** added via SPM (hand-edited `project.pbxproj`)
- **Auth UI**: email/password sign-in + magic link, shared `SignInView` sheet
- **Saved artists sync**: pulls via `GET /api/saved-artists/sync?since=`, pushes via `POST /api/saved-artists`
- **60-second poll**: runs while menu bar popover is open (macOS) or Settings screen is foregrounded (iOS)
- **device_id**: stable per-installation UUID in keychain, sent on every save/remove/sync
- **Keychain migration**: one-time shim to clean up old token location (no-op for new installs)
- **macOS**: auth bar in popover, "Sign In" / "Sign Out" in footer, synced artists replace local list when signed in
- **iOS**: Account section in Settings tab, synced artists list, same 60s poll

## What's NOT here

- Realtime subscriptions (Phase 4)
- Local-data merge prompts (Phase 5)

## Model note

First task on `ollama/glm-5.2:cloud` (756B params, 1M context). No Swift API hallucinations observed. The SDK API was verified against the actual `supabase-swift` v2.44.1 source — `signIn` returns `Session` directly, `signUp` returns `AuthResponse` enum, `signInWithOTP` takes `redirectTo:` (not `options:`), and `session(from:)` handles deeplink callbacks (not `handleCallback`).

**Note:** Xcode is not installed on this machine, so `xcodebuild` could not run. Swift files pass `swiftc -parse` and the `project.pbxproj` passes `plutil -lint`. Build verification needs a machine with Xcode.

Fixes UNS-112


brandonlucasgreen (commented):
## Staff review — UNS-112 Mac/iOS sync

One **confirmed, feature-breaking** bug, plus a couple of sync-correctness gaps. The web API contract (`/api/saved-artists/sync` → `{ artists, server_time }`, POST actions, ISO-8601 `since`, `device_id` truncation) all line up correctly with this client — good. Issue is on the Swift side, which `swiftc -parse` and a no-Xcode build could not catch.

### 🔴 Blocking

**1. `SyncedArtist.id` type mismatch breaks the entire pull — `confirmed`**
`SavedArtistsSync.swift` declares `let id: Int?`, but `saved_artists.id` is a **UUID** (`supabase/migration-013-saved-artists.sql:6`), and `saved-artists-sync.ts:107` serializes it as `id: row.id` → a JSON string. `JSONDecoder` throws a type-mismatch decoding a string into `Int?` (optional only rescues `null`/absent, not a wrong type), and because the whole `SyncResponse` is one `try decode(...)` the entire response fails. The error is swallowed in the `catch` (`print("[Sync] Pull failed")`), so `syncedArtists` stays empty **on every pull, forever**. The synced list will never populate on either platform.
Fix: change `id` to `String?` (it doubles as the `Identifiable.id`, so a `String` works cleanly). Worth adding a decode round-trip unit test against a real server payload so this class of bug fails CI rather than silently at runtime.

### 🟠 Regression / correctness

**2. Cross-device removals don't propagate during a live session — `confirmed`**
Incremental pull (`?since=`) only returns upserted rows; `handleRemove` in `saved-artists.ts:271` is a hard `DELETE` with no tombstone. The client merge in `SavedArtistsSync.pull` only adds/replaces by slug — it never deletes. So an artist removed on device B stays visible on device A until the app restarts (cold start has an empty local list + nil cursor → full pull repopulates correctly). Self-healing on relaunch, but stale for the whole foreground session the 60s poll is meant to cover. Options: a `deleted`/tombstone field in the sync payload, or have the poll periodically do a `pull(force: true)` and replace.

**3. Clock skew can hide a device's own new saves — `plausible`**
On INSERT the row's `last_modified` is the **client-supplied** device clock (`saved-artists.ts:216-223`), while the sync cursor is the **server** clock compared with strict `gt` (`saved-artists-sync.ts:92`). If a device's clock lags the server, a freshly-saved row's `last_modified` can be older than the stored `lastSyncTime`, so the immediate `await pull()` after `saveArtist` (and other devices' polls) skip it until a restart. Consider stamping `last_modified` server-side on insert too.

### 🟢 Security — reviewed, no blockers

- **Secret handling (`confirmed`-acceptable):** the hardcoded key in `AuthService.swift` is the Supabase **anon** key (`"role":"anon"` in the JWT) — public by design, same as the web app ships. The REST API independently enforces per-user via `auth.getUser(token)` + `user_id` filter, and `saved_artists` has RLS (migration-013). No service-role key embedded. Fine.
- **Auth callback over custom scheme (`needs-manual-verification`):** `unstream://auth/callback` is handled on both platforms and fed to `client.auth.session(from:)`. Custom schemes are registerable by other apps, so this relies on supabase-swift's PKCE code-verifier (default flow in v2) to reject foreign callbacks. Confirm PKCE is the active flow; longer-term `ASWebAuthenticationSession`/Universal Links bind the callback more tightly. Low risk as-is.
- **Input validation / data exposure / logging:** server validates `since` as ISO-8601 and truncates `device_id` to 128 chars; client logs no tokens (the migration `print` is token-free). Clean.

### Nit
Several new files are missing a trailing newline.

Net: **#1 must be fixed before merge** — without it the feature does nothing. #2/#3 are follow-ups depending on how tight you want live cross-device consistency.

---

PR #285, opened 2026-06-20
feat(UNS-128): scheduled GC for saved_artists tombstones (pg_cron)

brandonlucasgreen:
## UNS-128: Scheduled GC for saved_artists tombstones

Adds `supabase/migration-018-saved-artists-tombstone-gc.sql` — a daily 3am UTC `pg_cron` job that hard-deletes `saved_artists` rows soft-deleted more than 30 days ago.

### Why
Migration 017 added tombstone columns (`deleted`, `deleted_at`) for cross-client sync. Without GC, tombstones accumulate forever. This migration handles cleanup with a generous 30-day retention window that covers offline clients, slow networks, and scheduling jitter.

### What's in the migration
- `CREATE EXTENSION IF NOT EXISTS pg_cron` — idempotent
- `gc_saved_artists_tombstones()` — plpgsql function that runs the DELETE
- `cron.unschedule` guarded by `WHERE EXISTS` — idempotent on re-run
- `cron.schedule` at `0 3 * * *` — daily 3am UTC

### Deploy prerequisite (Brandon)
**UNS-132** must be completed first: enable the `pg_cron` extension in the Supabase dashboard on staging, apply this migration, smoke test, then repeat on prod. UNS-128 stays open until the first nightly run completes successfully in prod (verified via `cron.job_run_details`).

### What's NOT in this PR
- No Mac/iOS app code changes
- No `schema.sql` updates (auto-generated by Supabase)
- No `data/shipped-features.json` changes (infrastructure, not user-facing)

Refs: UNS-128, UNS-132

### DIFF
--- a/supabase/migration-018-saved-artists-tombstone-gc.sql
+++ b/supabase/migration-018-saved-artists-tombstone-gc.sql
@@ -1,0 +1,39 @@
+-- Migration 018: Scheduled GC for saved_artists tombstones (UNS-128)
+--
+-- Migration 017 added soft-delete (deleted, deleted_at columns) for cross-client
+-- sync tombstones. Without GC, these accumulate forever — every artist that any
+-- user ever unsaved would stay in the table. This migration adds a daily 3am UTC
+-- sweep that hard-deletes tombstones older than 30 days.
+--
+-- The 30-day window is generous: sync clients reconcile via tombstones on
+-- the next pull, and a 5-min force-pull acts as a backstop for long sessions.
+-- 30 days of buffer easily covers offline clients, slow networks, and any
+-- scheduling jitter.
+--
+-- Required: pg_cron extension must be enabled in the Supabase dashboard
+-- before this migration runs. UNS-132 tracks that step.
+
+CREATE EXTENSION IF NOT EXISTS pg_cron;
+
+CREATE OR REPLACE FUNCTION gc_saved_artists_tombstones()
+RETURNS void AS $$
+BEGIN
+  DELETE FROM saved_artists
+  WHERE deleted = true
+    AND deleted_at < now() - interval '30 days';
+END;
+$$ LANGUAGE plpgsql;
+
+-- Idempotent: unschedule first so re-running the migration doesn't error.
+-- cron.job is the pg_cron catalog table.
+SELECT cron.unschedule('gc-saved_artists_tombstones')
+  WHERE EXISTS (
+    SELECT 1 FROM cron.job WHERE jobname = 'gc-saved_artists_tombstones'
+  );
+
+-- Schedule: 3am UTC daily. Low-traffic window; runs a single DELETE.
+SELECT cron.schedule(
+  'gc-saved_artists_tombstones',
+  '0 3 * * *',
+  $$SELECT gc_saved_artists_tombstones();$$
+);


brandonlucasgreen (commented):
## Senior review — UNS-128 tombstone GC

Clean, well-documented infra migration. No confirmed security vulnerabilities and no blocking bugs. Idempotency, scoping, and the 30-day rationale all check out. A few advisory items below — none block merge, but two are worth verifying during the UNS-132 smoke test.

### Security
- **GC correctness depends on the cron role bypassing RLS — `needs-manual-verification`.** `gc_saved_artists_tombstones()` is `SECURITY INVOKER` (default) and `saved_artists` has RLS with a `DELETE` policy of `auth.uid() = user_id`. This only deletes across *all* users if the cron job runs as a `BYPASSRLS` role (on Supabase, `pg_cron` jobs run as `postgres`, which does). If it were ever scheduled by a non-superuser role, `auth.uid()` is NULL in the cron context and the DELETE would silently affect **zero rows** — GC would appear to "succeed" while collecting nothing. Please confirm in the smoke test that rows are actually removed (check the affected-row count / `cron.job_run_details`), not just that the job ran.
- **Function has no `SET search_path` — `plausible` (low).** The body references unqualified `saved_artists`; Supabase's linter flags mutable-search_path functions. Risk is low here (not SECURITY DEFINER, no untrusted callers), and it matches the existing pattern in migrations 003/016. Cheap hardening: `SET search_path = public, pg_catalog` on the function, or schema-qualify as `public.saved_artists`.
- Injection / secrets / data exposure: **`confirmed` clean.** Static SQL only, no dynamic exec; no secrets; no new table so no RLS gap.

### Regression / correctness
- **`deleted_at IS NULL` tombstones are never collected — `plausible` (latent, low).** `deleted_at < now() - interval '30 days'` evaluates to NULL (not true) for any row with `deleted = true, deleted_at IS NULL`, so such rows would accumulate forever — the exact failure mode this PR exists to prevent. Not currently reachable: `saved-artists.ts:277` always sets `deleted` and `deleted_at` together. Defensive hardening if you want belt-and-suspenders: `AND deleted_at IS NOT NULL` (explicit) or coalesce NULL to "old."
- **Zombie-resurrection for clients offline > 30 days — `needs-manual-verification` (accepted trade-off).** Once a tombstone is hard-deleted, a client that hasn't synced in >30 days never receives the removal and can re-push the artist, resurrecting it. This is the inherent tombstone-GC trade-off and the 30-day window is reasonable + documented; flagging it as the one real behavioral edge so it's a conscious choice.
- **GC query won't use the migration-017 index — minor perf, non-blocking.** `idx_saved_artists_user_deleted (user_id, deleted)` leads with `user_id`; the GC query has no `user_id` predicate, so it can't use this index and will scan filtered on `deleted`. Fine for a daily job on a small table — revisit with a partial index (e.g. `ON saved_artists (deleted_at) WHERE deleted`) only if the table grows.
- **Idempotency — `confirmed` sound.** `SELECT cron.unschedule(...) WHERE EXISTS(...)` only evaluates the function when the job exists (zero output rows ⇒ target expression not evaluated), sidestepping the "could not find valid entry for job" error; `cron.schedule` upserts by name. Safely re-runnable.
- **API contract / state transitions:** no drift — no API or client-visible schema change. `schema.sql` intentionally not updated (Supabase auto-generates). Acceptable.

### Operational
- `CREATE EXTENSION IF NOT EXISTS pg_cron;` will error if UNS-132 hasn't enabled it first *and* the migration role isn't superuser — already gated on UNS-132 in the PR description, so just confirming the ordering holds.
- Nit: file is missing a trailing newline.

Recommendation: mergeable as-is once UNS-132 is done; please treat the RLS-bypass check and the affected-row count as explicit smoke-test assertions.

---

PR #287, opened 2026-06-21
fix(UNS-134): first-initial fallback when artist photo is missing on /a/{slug}

brandonlucasgreen:
Closes UNS-134.

## What changed
- **Edge function** (`api/edge/artist-page-static.ts`): replaced conditional `` with an avatar wrapper div containing the image (with inline `onerror` fallback) plus a first-initial span. Applies to both claimed and unclaimed artist paths.
- **`RichArtistProfile.tsx`**: added `onError` handler + fallback initial, mirroring the pattern from `ArtistDirectoryPage`.
- **`UnclaimedQuietCard.tsx`**: same fix for consistency.

## What to verify
- Visit `/a/kingtriumph` (verified artist, no photo) — should show "K" initial in a circle, not a broken image
- Visit a verified artist page with a working image — image should render normally
- Visit an unclaimed artist page without an image — same initial fallback

## Risks
- The edge function uses an inline `onerror` attribute (progressive enhancement). Non-JS users with a broken image URL will still see a broken ``. The no-image case (empty `imageUrl`) works fully without JS.
- The inline `onerror` is technically JS but not a `<script>` tag — consistent with the spec's intent (no script blocks, no analytics, no auth readers).

### DIFF
--- a/api/edge/artist-page-static.ts
+++ b/api/edge/artist-page-static.ts
@@ -101,6 +101,9 @@ const CSS = `
   .site-header .nav-accent:hover { text-decoration: underline; }
   .claim-cta { display: inline-block; margin-top: 24px; padding: 10px 24px; border-radius: 12px; background: var(--accent); color: #fff; font-weight: 600; font-size: 15px; text-decoration: none; }
   .claim-cta:hover { opacity: 0.9; }
+  .avatar-wrap { width:128px;height:128px;border-radius:50%;border:2px solid var(--border);margin:0 auto 16px;display:flex;align-items:center;justify-content:center;overflow:hidden;background:var(--bg2) }
+  .avatar-wrap img { width:100%;height:100%;object-fit:cover }
+  .avatar-fallback { font-size:48px;font-weight:600;color:var(--muted) }
 `;
 
 export default async function handler(request: Request, context: Context) {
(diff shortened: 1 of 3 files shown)


brandonlucasgreen (commented):
Senior-engineer review — one bug to fix, rest looks good.

**🐛 Unescaped artist name in the edge SSR fallback (confirmed) — `api/edge/artist-page-static.ts` lines 286 & 389**

```ts
<span class="avatar-fallback" style="display:${imageUrl ? 'none' : 'flex'}">${artist.name[0]?.toUpperCase() || '?'}</span>
```

`artist.name[0]` is interpolated **raw** into the HTML. Everywhere else in this file `artist.name` goes through `artistName = escapeHtml(artist.name)`, so this is the one spot that breaks the escaping contract. It's not full XSS (a single character can't form a `<tag>`/attribute), but an artist whose name begins with `<` produces `…flex"><</span>` — the lone `<` consumes the `</span>` end tag during HTML parsing and corrupts the avatar/heading markup. Inconsistent and a real (if low-severity) rendering bug.

Fix — escape the initial:
```ts
<span class="avatar-fallback" style="display:${imageUrl ? 'none' : 'flex'}">${escapeHtml((artist.name[0] || '?').toUpperCase())}</span>
```
(Don't reuse `artistName[0]` — for a name like `&amp;…` that would grab a stray `&`.)

The two React components (`RichArtistProfile.tsx`, `UnclaimedQuietCard.tsx`) are fine here — JSX auto-escapes `{artist.name[0]?.toUpperCase()}`, so the bug is edge-only.

---

**Checks that passed:**
- **CSP:** `script-src` includes `'unsafe-inline'`, so the inline `onerror=` handler is allowed and will fire. ✅
- **img src / alt:** still escaped via `escapeHtml(imageUrl)` / `artistName`. ✅
- **Null/empty handling:** `''[0] → undefined → '?'` works; `artist.name` was already assumed non-null upstream (same as the existing `escapeHtml(artist.name)` call), so no new null risk. ✅
- **Regression — no-image path:** previously rendered no avatar; now always renders the wrapper + initial. Intended per UNS-134. ✅
- **React onError fallback:** `className` toggling (`'hidden fallback'` → remove `hidden`) and `querySelector('.fallback')` line up correctly. ✅
- **No-JS caveat:** broken image URL still shows a broken `` for non-JS users — already called out in the PR description, acceptable as progressive enhancement.

No auth/authorization, secret-handling, injection (beyond the above), or data-exposure concerns. Approve once the initial is escaped.

---

PR #291, opened 2026-06-22
feat(UNS-130): migrate magic-link auth to ASWebAuthenticationSession

brandonlucasgreen:
## Summary

Migrates the magic-link auth callback from a custom `unstream://` URL scheme (registered via `NSAppleEventManager` on macOS and `onOpenURL` on iOS) to `ASWebAuthenticationSession`, which validates the calling app and binds the callback to the original PKCE request. Closes the documented interception risk on macOS where any app claiming `unstream://` first would receive the auth deeplink.

## Changes

- **AuthService.sendMagicLink**: Replaced `signInWithOTP` + custom-scheme callback with the Supabase SDK's built-in `signInWithOAuth(provider: .email)` using `ASWebAuthenticationSession`. The email is passed as a query param. `prefersEphemeralWebBrowserSession = true` avoids cookie-blocking sheet on repeat attempts.
- **AuthPresentationAnchor**: New `@unchecked Sendable` class providing `ASWebAuthenticationPresentationContextProviding` on macOS — returns `NSApp.keyWindow` so the web auth sheet attaches to the running app.
- **AppDelegate**: Removed `NSAppleEventManager` handler for `kAEGetURL` (registration + teardown + `handleAppleEvent` method). Auth callback is now captured within the web view, not via OS dispatch.
- **iOS handleIncomingURL**: Renamed `handleNonAuthIncomingURL`, auth branch removed — only search deeplink handling remains.
- **SignInView**: `sendMagicLink` is now non-async (fire-and-forget `Task`). Removed dead `magicLinkSent` state and its UI block.
- **handleAuthCallback**: Retained as a fallback for universal-link or residual custom-scheme delivery. PKCE validation via `client.auth.session(from:)` unchanged.

## Security properties

- PKCE code-verifier still validated (SDK's `session(from:)` handles this)
- Callback URL bound to original request (ASWebAuthenticationSession's `callbackURLScheme` matching)
- No new trust boundaries opened (no URL forwarding, no new custom URL handlers registered)
- `NSAppleEventManager` handler removed — no OS-level custom-scheme listener on macOS

## Build and test

- `xcodebuild build` macOS: PASS
- `xcodebuild build` iOS: PASS
- `xcodebuild test` macOS: 5/5 PASS
- `tsc -b` (apps/web): clean

## API surface

`ASWebAuthenticationSession` (macOS 10.15+, iOS 12+) — well within deployment targets (macOS 13.0, iOS 17.0). Uses:
- `init(url:callbackURLScheme:completionHandler:)` (both platforms)
- `prefersEphemeralWebBrowserSession` (both platforms)
- `presentationContextProvider` (macOS only — iOS uses the default)
- `ASWebAuthenticationPresentationContextProviding` protocol (macOS only)

## What to verify

- Magic-link sign-in flow opens a web view, user enters email (pre-filled if passed), receives magic link, clicks it in email app, web view captures the callback redirect
- Email/password sign-in unaffected
- Search deeplink (`unstream://search?q=...`) still works on iOS

## Risks

- `signInWithOAuth(provider: .email)` relies on Supabase's hosted email auth page polling for session completion when the user clicks the magic link in their email app. If the hosted page doesn't poll, the web view won't capture the callback. This should be verified against the actual Supabase project configuration.
- The `unstream://` URL scheme registration in Info.plist is retained — ASWebAuthenticationSession uses it as the callback scheme identifier, not for OS-level dispatch.

Fixes UNS-130

### DIFF
--- a/apps/mac/Unstream/Services/AuthService.swift
+++ b/apps/mac/Unstream/Services/AuthService.swift
@@ -1,4 +1,5 @@
 import Foundation
+import AuthenticationServices
 import Supabase
 
 /// Manages Supabase auth state for the Unstream app.
@@ -17,6 +18,10 @@ class AuthService: ObservableObject {
     @Published var session: Session?
     @Published var isLoading: Bool = false
     @Published var errorMessage: String?
+
+    #if os(macOS)
+    private var authPresentationAnchor: AuthPresentationAnchor?
+    #endif
 
     private init() {
         let anonKey = Bundle.main.infoDictionary?["SUPABASE_ANON_KEY"] as? String ?? ""
@@ -91,27 +96,51 @@ class AuthService: ObservableObject {
 
     // MARK: - Magic Link
 
-    func sendMagicLink(email: String) async {
+    /// Starts a magic-link sign-in via `ASWebAuthenticationSession`.
+    ///
+    /// Uses the Supabase SDK's `signInWithOAuth(provider: .email)` which opens
+    /// the hosted email auth page in a web view. The session validates the
+    /// calling app and binds the callback to the original PKCE request,
+    /// closing the custom-URL-scheme interception risk on macOS.
+    func sendMagicLink(email: String) {
         isLoading = true
         errorMessage = nil
 
-        do {
-            try await client.auth.signInWithOTP(
-                email: email,
-                redirectTo: URL(string: "unstream://auth/callback")
-            )
-        } catch {
-            errorMessage = error.localizedDescription
-        }
+        let redirectURL = URL(string: "unstream://auth/callback")!
 
-        isLoading = false
+        #if os(macOS)
+        let anchor = AuthPresentationAnchor()
+        authPresentationAnchor = anchor
+        #endif
+
+        Task {
+            do {
+                let result = try await client.auth.signInWithOAuth(
+                    provider: .email,
+                    redirectTo: redirectURL,
+                    queryParams: [("email", email)]
+                ) { @Sendable session in
+                    session.prefersEphemeralWebBrowserSession = true
+                    #if os(macOS)
+                    session.presentationContextProvider = anchor
+                    #endif
+                }
+                session = result
+            } catch {
+                errorMessage = error.localizedDescription
+            }
+            isLoading = false
+            #if os(macOS)
+            authPresentationAnchor = nil
+            #endif
+        }
     }
 
     // MARK: - Deeplink handling
 
-    /// Handle the callback URL from a custom URL scheme deeplink.
-    /// The URL contains the auth tokens as query/hash params; PKCE code-verifier
-    /// validates the token contents.
+    /// Handles an auth callback URL received outside `ASWebAuthenticationSession`
+    /// (e.g. via universal link or residual custom-scheme delivery).
+    /// The PKCE code-verifier in the SDK validates the token contents.
     func handleAuthCallback(url: URL) async {
         isLoading = true
         errorMessage = nil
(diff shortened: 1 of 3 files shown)


brandonlucasgreen (commented):
## Senior-engineer review — UNS-130 (ASWebAuthenticationSession migration)

Scope reviewed: the two changed files (`AuthService.swift`, `UnstreamApp.swift`). Build "succeeded" on macOS, but the **magic-link flow is unverified** (the PR says manual test is "the next gate"), and the code has issues that I believe would prevent it from working as intended. Treat this as request-changes — recommend not merging until the items below are addressed and UNS-122 passes.

### Regression / correctness

**1. `presentationContextProvider` is assigned an object that deallocates immediately — macOS session won't present. `confirmed`**
`ASWebAuthenticationSession.presentationContextProvider` is a **`weak`** property. In `sendMagicLink` you do `session.presentationContextProvider = AuthPresentationAnchor()`. That anchor is retained by nothing else, so ARC frees it on the next tick and the provider becomes `nil`. With no anchor, the macOS session has nothing to present from. You already solved the identical lifetime problem one line down by holding `webAuthSession` on the service — the anchor needs the same treatment:
```swift
private var webAuthSession: ASWebAuthenticationSession?
private var presentationAnchor: AuthPresentationAnchor?   // add this (macOS)
...
let anchor = AuthPresentationAnchor()
self.presentationAnchor = anchor
session.presentationContextProvider = anchor
```

**2. `sendMagicLink` went from `async` to fire-and-forget — the "link sent" confirmation now fires unconditionally. `confirmed`**
The signature changed from `func sendMagicLink(email:) async` to a synchronous method that spawns a detached `Task` and returns immediately. The only caller, `SignInView.sendMagicLink()`:
```swift
await auth.sendMagicLink(email: email)
if auth.errorMessage == nil { magicLinkSent = true }
```
Because the call now returns before the OTP request resolves, `magicLinkSent = true` runs **immediately and unconditionally** — the success UI shows even when the OTP send fails (the failure sets `errorMessage` later, after this check has already passed). This is an API-contract change with a real state-transition regression. Either keep `sendMagicLink` `async` and `await` the OTP send before returning, or have the view react to published `isLoading`/`errorMessage` instead of the return.

**3. iOS has no `presentationContextProvider` at all. `confirmed`**
The provider is set only inside `#if os(macOS)`. On iOS, `ASWebAuthenticationSession` requires a `presentationContextProvider`; `start()` fails without one. This PR also deletes the iOS `handleIncomingURL` auth branch, so on iOS the magic-link path is left with no working callback. The build was run for `platform=macOS` only — iOS wasn't compiled or tested. iOS needs its own anchor (a `UIWindow`-based `ASPresentationAnchor`).

**4. ASWebAuthenticationSession is being handed the custom-scheme callback URL, not a web auth URL. `needs-manual-verification` (core design risk)**
`ASWebAuthenticationSession(url: URL(string: "unstream://auth/callback")!, callbackURLScheme: "unstream")` opens `url` in an in-app browser and waits for a redirect to `callbackURLScheme`. That primitive is built for OAuth-style flows where `url` is an **https authorization page** the user interacts with, and the server redirects back to the custom scheme. An **email magic link completes out of band** — Supabase emails a link the user clicks later; there is no web page to load here, and pointing the session at `unstream://auth/callback` itself won't drive OTP completion. I can't run the Xcode flow to confirm, but I expect the sheet to open with nothing to do (or close instantly) and the callback never to carry tokens. This is the thing to validate first in the UNS-122 manual test before merging — if it doesn't work, the migration approach (not just the bugs above) needs rethinking.

**5. Minor: `NSApp.keyWindow ?? NSWindow()` fallback. `plausible`**
If `keyWindow` is nil, the fallback presents against a freshly-constructed, never-shown `NSWindow`. Presenting from a detached window is unlikely to surface correctly. Prefer `NSApp.windows.first` / `NSApp.mainWindow` or fail loudly rather than fabricate a window.

### Security

- **auth/authorization:** Intent (bind the callback to the calling app, close the macOS `unstream://` interception window) is sound, and PKCE/token validation still runs via `client.auth.session(from:)`. But the claimed hardening is only actually delivered if #4 works — verify before treating the interception risk as closed. No new authorization hole introduced.
- **input validation:** callback-URL/token parsing is delegated to the Supabase SDK (`session(from:)`), unchanged. OK.
- **secret handling / data exposure:** no tokens or secrets logged; `errorMessage = error.localizedDescription` surfaces backend error text to the UI but nothing sensitive. OK.
- **injection / unsafe defaults:** none. `prefersEphemeralWebBrowserSession = true` is a reasonable privacy default. OK.

### Bottom line
No new security holes, but findings #1–#3 are confirmed code defects that would break the macOS (anchor lifetime, #1) and iOS (#3) flows, plus a confirmed UI-state regression (#2). #4 is the design question to settle in manual testing before this can be considered working. Recommend not merging until the UNS-122 manual sign-in test passes on both platforms.

---

PR #292, opened 2026-06-24
feat(UNS-139): Close manual-review ownership gap in claim-artist

brandonlucasgreen:
## Security fix — UNS-139

Closes the manual-review ownership gap in the artist claim flow.

### What changed

- **Migration 019**: adds `ownership_verified_by` (text) and `ownership_verified_at` (timestamptz) to `verification_requests`
- **`api/functions/admin-verify.ts`**: `approve` action now requires `ownershipVerified: true` in the request body. Missing/false → 400 with a generic error (no field-name leak). On success, writes the new columns and fires a Sentry info-level `captureMessage` with admin id, artist id, and request id for audit trail.
- **`api/functions/admin-verify.ts` GET**: joins `artist_profiles.verified_at` to surface a `link_back_completed` boolean per request — soft signal for the admin, not a gate.
- **`AdminVerifyPage.tsx`**: shows link-back status badge (✓ green / ⚠ yellow). Adds required checkbox "I have verified the submitter is the artist owner". Approve button is disabled until the checkbox is ticked.
- **Tests**: 4 new unit tests — 400 on missing `ownershipVerified`, 400 on `false`, 200 on `true` with Sentry log assertion, and reject doesn't need the flag.

### What was NOT changed (per Brandon's call)

- Email-domain match (item 2) — skipped, too much wildcard scope
- Link-back is a soft signal only, not a hard gate

### What to verify
- Run migration 019 on staging
- Test the admin review page: checkbox disables Approve, link-back badge renders
- Test the 400 path via curl/Postman (approve without `ownershipVerified`)
- Confirm Sentry event fires on approval

### Risks
- The `artist_profiles` join in the GET query assumes one profile per artist (UUID unique constraint exists from migration 002). If that constraint were dropped, the array-handling code still works but the signal could be ambiguous.
- The 400 error message is intentionally generic to avoid info leakage.

### Security review
This is a security fix. Flagging for Claude Code external review per process.

### DIFF
--- a/api/functions/admin-verify.ts
+++ b/api/functions/admin-verify.ts
@@ -215,9 +215,10 @@ export async function handler(event: {
 
     // An artist_profiles row may not exist yet if the user submitted a manual
     // review without first attempting the automated link-back claim flow.
+    // Re-check ownership hasn't changed since the request was submitted.
     const { data: existingProfile, error: profileLookupError } = await client
       .from('artist_profiles')
-      .select('id')
+      .select('id, user_id, verified_at')
       .eq('artist_id', request.artist_id)
       .maybeSingle();
 
@@ -227,6 +228,14 @@ export async function handler(event: {
         statusCode: 500,
         headers: CORS_HEADERS,
         body: JSON.stringify({ error: 'Failed to look up artist profile' }),
+      };
+    }
+
+    if (existingProfile?.verified_at && existingProfile.user_id !== request.user_id) {
+      return {
+        statusCode: 409,
+        headers: CORS_HEADERS,
+        body: JSON.stringify({ error: 'Artist was claimed by another user while this request was pending' }),
       };
     }
 
--- a/api/functions/admin-verify.ts
+++ b/api/functions/admin-verify.ts
@@ -44,10 +44,9 @@ export async function handler(event: {
       .select(`
         id, email, message, status, reviewer_notes, created_at, reviewed_at,
         artist_id, user_id,
-        artists(name, slug),
-        artist_profiles(verified_at)
+        artists(name, slug)
       `)
-      .order('status', { ascending: true }) // 'pending' sorts before others alphabetically
+      .order('status', { ascending: true })
       .order('created_at', { ascending: false });
 
     if (error) {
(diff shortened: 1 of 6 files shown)


brandonlucasgreen (commented):
## Senior-engineer review — UNS-139 (security fix)

Reviewed the diff for security and regression. One **confirmed blocking bug**, plus a few minor notes. The core security gate (POST `approve` requiring `ownershipVerified`, audit columns, Sentry trail) is sound and well-tested. The problem is in the GET change shipped alongside it.

---

### 🔴 BLOCKING — `confirmed` — GET query embed has no FK relationship; admin queue will 500

The GET now does:
```ts
.from('verification_requests')
.select(`... artists(name, slug), artist_profiles(verified_at)`)
```

PostgREST resource embedding requires a **foreign-key relationship between the two tables**. I verified against the live `unstream` schema:

- `verification_requests` FKs → `auth.users.id` (user_id), `artists.id` (artist_id)
- `artist_profiles` FKs → `artists.id` only

There is **no FK between `verification_requests` and `artist_profiles`** — they're siblings that both reference `artists`. A shared parent does **not** create an embeddable relationship. PostgREST will reject the embed with `PGRST200` ("Could not find a relationship between 'verification_requests' and 'artist_profiles' in the schema cache"), so the GET returns the 500 branch and **the entire admin verification queue fails to load.**

Note the irony: this is a security PR meant to harden approvals, but as written it breaks the page admins use to review *and reject* requests. (The POST `approve`/`reject` paths are unaffected — they use a separate `.select('*')`.)

The unit tests don't catch this — they only exercise POST; the GET path is unmocked/untested.

**Fix options:**
1. Drop the embed. Fetch profiles separately: collect `artist_id`s from the requests, then `artist_profiles.select('artist_id, verified_at').in('artist_id', ids)`, and map `link_back_completed` in code. Simple and boring — matches the codebase style.
2. Or add an explicit PostgREST relationship hint, but there's no real FK to hint at, so option 1 is the right call.

Please also add a GET-path test so this can't regress silently.

---

### 🟡 `plausible` (minor) — admin email logged to Sentry

```ts
Sentry.captureMessage('Verification request approved', { extra: { ..., adminEmail: admin.email } });
```
CLAUDE.md says don't log personal data. `adminId` is already captured and is sufficient for the audit trail; `adminEmail` is staff PII landing in Sentry. Recommend dropping `adminEmail` (keep `adminId`). Low risk since it's internal staff, but easy to tighten.

---

### 🟢 `needs-manual-verification` (nits)

- **`ownership_verified_by` typed `TEXT`** while it stores `admin.userId` (a UUID). Harmless, but `UUID` would be consistent with the rest of the schema and the existing `user_id` columns.
- **Self-attestation gate is a UI/audit speed-bump, not a technical control** — an authenticated admin can send `ownershipVerified: true` directly. That's by design per the PR description (rubber-stamp deterrent + audit trail), just flagging it so it's not mistaken for enforcement.
- Migration 019 and the new test file are missing trailing newlines (lint nit).

---

### What's good
- Approve gate (400 on missing/false, generic non-leaky error message), audit columns, and Sentry trail are correct and covered by the 4 new tests.
- `?? '(unknown)'` / array-normalization on the embed result is defensive.
- Migration is idempotent (`IF NOT EXISTS`), nullable, no backfill needed, existing RLS covers the new columns.

The GET embed is a hard runtime break, not a style preference — recommend fixing before merge.

---

PR #294, opened 2026-06-27
feat(UNS-31): user settings page (username + password change)

### LINKED ISSUE #31: Allow a listener to share or publicize a saved artist list
Useful for sharing/inviting folks to add artists to check out. Could also be a form of social proof for which artists you've supported

brandonlucasgreen:
## What Problem This Solves

Issue #31 asks for the ability for a listener to share or publicize their saved artist list. This requires a username (a public handle for the user). PR 1 of 2 adds the **/settings page** where users can set a username and change their password. PR 2 (after this merges) will add the public sharing endpoints and `/u/{handle}` page.

## Claude Code Review — Bugs Fixed

This update addresses all 5 confirmed runtime bugs + 3 security/policy flags + 2 minor fixes from Claude Code's senior staff review on PR #294:

1. **`.from('auth.users')` is wrong** — PostgREST can't query `auth.users` via dotted table names. Replaced with `auth.admin.getUserById` for email/has_password and a new `public.usernames` table for username.
2. **Custom `username` column on `auth.users` is unreadable** — GoTrue admin API doesn't surface custom columns. Moved username to `public.usernames` table with RLS, FK to `auth.users(id)` ON DELETE CASCADE, UNIQUE + CHECK constraints, indexed on `username`.
3. **`userMetadata` vs `user_metadata`** — admin API uses snake_case; `userMetadata` was silently ignored. Fixed to `user_metadata` so `has_password` is actually persisted.
4. **Pre-existing password users misclassified** — Users who set a password before the `has_password` flag existed had no flag → locked out of password change form. Migration now backfills `user_metadata.has_password = true` for all users with `encrypted_password IS NOT NULL`.
5. **Migration not re-runnable** — `ADD CONSTRAINT` failed on rerun. Now uses `DROP CONSTRAINT IF EXISTS` before `ADD CONSTRAINT`.

**Security/policy flags fixed:**
- Flag 6: Custom column on `auth.users` with no RLS → replaced by `public.usernames` with full RLS policies (Bug 2)
- Flag 7: `me-password.ts` switched from `'standard'` to `'strict'` rate-limit tier (10/min)
- Flag 8: CORS debt documented with comments matching `saved-artists.ts` pattern

**Structural fixes:**
- Added `api/tsconfig.json` for function typechecking (scoped to me-* files) + root `tsconfig.json` referencing both apps/web and api
- Wired `npx tsc -b` into the build pipeline — this is why `userMetadata` shipped silently (tsc only ran in apps/web)
- Added handler-level tests with mocked Supabase client for all three endpoints

**Minor fixes:**
- Trailing newlines added to all new files

## Why This Change Was Made

- **Username moved from `auth.users.username` to `public.usernames`**: Supabase Auth manages `auth.users` and the GoTrue admin API doesn't surface custom columns. Modifying the managed `auth` schema also violates CLAUDE.md RLS guidance. A `public.usernames` table is PostgREST-accessible, supports RLS, and follows the same pattern as `saved_artists`.
- **Password change via signInWithPassword + admin updateUserById**: Supabase's `auth.updateUser({ password })` doesn't verify the current password. We verify it by attempting `signInWithPassword` with the user's email + current password first, then update via the admin API with `user_metadata` (snake_case).
- **Service role for all writes**: Following the pattern in `saved-artists.ts`, all DB writes use `getClient()` (service role). The anon key is only used for token validation.
- **SPA-only**: /settings is a client-rendered route, no edge function SSR (per CLAUDE.md — one renderer only).

## User Impact

Logged-in users can now:
1. Go to `/settings` from the nav bar ("Settings" link appears when signed in)
2. Set a public username (3-20 chars, lowercase alphanumeric + hyphens)
3. Change their password (requires current password verification)

Users without a password (magic-link-only accounts) see a message directing them to the password reset flow to set one.

## Evidence

- **346 unit tests pass** (22 test files), including new handler-level tests with mocked Supabase client
- **`npx tsc -b` passes** — now includes `api/tsconfig.json` for function typechecking
- **Full `npm run build` passes** — typecheck + unit tests + Vite build + sitemap
- **No password values logged** — verified by handler test + code review

## Migration

Migration 020 (`migration-020-user-username.sql`) is **dropped** — it added a custom column to `auth.users` which is unreadable via PostgREST.

Migration 021 (`migration-021-public-usernames.sql`) **replaces it**:
- Creates `public.usernames` table with FK to `auth.users(id)` ON DELETE CASCADE
- UNIQUE constraint on `username` + CHECK for format (3-20 chars, lowercase alphanumeric + hyphens)
- RLS policies: owner read/write, anyone can read username column (for PR 2 public endpoint)
- Backfill: sets `user_metadata.has_password = true` for users with `encrypted_password IS NOT NULL`
- Re-runnable: uses `DROP CONSTRAINT IF EXISTS` + `DROP POLICY IF EXISTS` before creating

Fixes #31


brandonlucasgreen (commented):
## Senior staff review — PR #294 (UNS-31 PR 1: /settings page)

@brandonlucasgreen Heads up: this PR has **three confirmed runtime bugs that make the whole feature non-functional in production**, and they are invisible to the test suite. The "346 tests pass / build clean" evidence is misleading here, because the new function files are never type-checked (`npm run build` only runs `tsc -b` inside `apps/web`; nothing type-checks `api/functions/`), and the unit tests only re-declare and exercise the regex in isolation — they never call the actual handlers or their Supabase queries. So the green checkmarks don't cover the code that's broken.

Recommend **not merging as-is**. Details below.

---

### 🔴 Regression / correctness (blocking)

**1. `.from('auth.users')` does not query the auth schema — entire read/write path fails. [confirmed]**
`me-settings.ts:61` and `me-username.ts:86/97/108` call `serviceClient.from('auth.users')`. supabase-js `.from(name)` resolves `name` as a table in the *exposed* schema (public); a dotted string is treated as a literal table name, not a schema qualifier (you'd need `.schema('auth').from('users')`). On top of that, the `auth` schema is **not exposed to PostgREST by default**, so even the qualified form would be rejected. Net effect at runtime:
- `GET /api/me/settings` → the username/email select errors → returns 500 "Failed to load settings", so the page never loads.
- `POST /api/me/username` → the existing-value check, collision check, and update all error → 500 "Internal server error". Usernames can never be saved.

No existing endpoint in the repo queries `auth` via PostgREST — `saved-artists.ts` only touches `public` tables, and the correct precedent for reading auth users is the admin API (`auth.admin.getUserById`, which this PR already uses for `hasPassword`).

**2. A custom `username` column on `auth.users` is unreadable through supabase-js anyway. [confirmed]**
Even if (1) were fixed, the GoTrue admin API (`getUserById`) returns a fixed `User` shape and will not surface a custom column you added to `auth.users`. There is no working read path for `username` as designed. Combined with Supabase's strong guidance against modifying the managed `auth` schema (their own migrations can clobber it), this column should live in a **`public.profiles` / `public.usernames` table with RLS** (also satisfies CLAUDE.md's "RLS on every new table/column", which the current migration doesn't address). That table is readable/writable via PostgREST the same way `saved_artists` is.

**3. `userMetadata` is the wrong key for `admin.updateUserById` → `has_password` is never persisted. [confirmed]**
`me-password.ts:91-94` passes `{ password, userMetadata: { has_password: true } }`. The admin API's `AdminUserAttributes` field is `user_metadata` (snake_case); `userMetadata` is silently ignored. So after a successful password change, the flag is never written. Then `GET /settings` reads `user_metadata.has_password` (`me-settings.ts:75`) → `hasPassword: false` → the UI shows "Your account was created with a magic link" and **hides the password-change form for users who just set a password**. (Again, untyped `api/`, so `tsc` doesn't flag the bad property.)

**4. Pre-existing password users are misclassified as magic-link-only. [confirmed — changed assumption]**
`hasPassword` depends on `user_metadata.has_password`, which only gets set by the new/updated-password flow. Any account that set a password *before* this flag existed has no flag → shown the magic-link message and locked out of the change-password form. You need a backfill, or a more reliable signal for "has password" than this metadata flag.

**5. Migration is not re-runnable. [plausible]**
`ADD COLUMN IF NOT EXISTS` is idempotent, but `ADD CONSTRAINT username_format` (`migration-020:8`) is not — re-running (or running when the constraint already exists) fails with "constraint already exists". Wrap in a guard or `DROP CONSTRAINT IF EXISTS` first. (Moot if #2 moves this to a `public` table, but worth fixing in the new migration too.)

---

### 🟠 Security

**6. Custom column on the Supabase-managed `auth.users` table, no RLS. [needs-manual-verification]**
Covered in #2 — the security angle is that CLAUDE.md mandates RLS for new tables/columns, and modifying `auth.users` sidesteps that model entirely. Moving to a `public` table with explicit RLS policies (owner can read/write own row; username readable for the future public-sharing feature) is the right fix.

**7. Password-change endpoint uses the `standard` (30/min) rate-limit tier. [plausible]**
`me-password.ts:39` verifies the current password via `signInWithPassword` — an auth-sensitive operation. A `strict` (10/min) tier already exists for exactly this. The caller must already be authenticated as themselves, so blast radius is limited, but the stricter tier is the conservative choice.

**8. `Access-Control-Allow-Origin: '*'` on authenticated endpoints. [plausible — low]**
All three endpoints hand-roll `*` CORS rather than going through `api/functions/middleware.ts`, which CLAUDE.md asks for (public endpoints restricted to `unstream.stream`). Because auth is a Bearer token rather than a cookie, `*` doesn't directly enable credential theft, and `saved-artists.ts` already does the same thing — so this is consistency-with-existing-debt, not a new hole. Flagging for the record; consider routing through the shared middleware.

**Good, for the record:** passwords are never logged (only `error.message` via `console.error`) [confirmed]; username is regex-validated before use and the query builder parameterizes inputs, so no injection [confirmed]; current-password verification before update is the right pattern [confirmed].

---

### Minor
- New files all lack a trailing newline (`\ No newline at end of file`) — lint may complain.
- Password change doesn't revoke other active sessions (typical hardening; out of scope, noting it).
- The JS and Postgres username regexes match (3–20, no leading/trailing hyphen) — no drift there. 👍

---

### Suggested path forward
1. Move `username` to a new `public.usernames` (or `public.profiles`) table with RLS, FK to `auth.users(id)`, unique + format constraints. Read/write it via the normal PostgREST client like `saved_artists`.
2. Read `username`/`email` via `auth.admin.getUserById` (email) + the new public table (username); drop all `.from('auth.users')` calls.
3. Fix `userMetadata` → `user_metadata` in `me-password.ts`, and decide on a backfill/reliable signal for `has_password`.
4. Add an `api/` tsconfig (or include `api/` in the build's typecheck) so these classes of error are caught — otherwise the function layer ships untyped.
5. Add at least one handler-level test that mocks the Supabase client and asserts the query path, so "tests pass" actually covers this code.

Happy to pair on the `public.profiles` migration + the handler rewrite if useful.

---

brandonlucasgreen (commented):
## Re-review — PR #294 @ `dfd9a46`

@brandonlucasgreen Nice turnaround. I re-checked all the items from the first pass plus the new code. **The five confirmed bugs and all three flags are genuinely fixed** — verified below. But the new migration introduces **one confirmed blocking bug** (it won't apply), and there's one RLS exposure I'd reconsider before merge.

### ✅ Verified fixed
1. **`.from('auth.users')`** — gone. `me-settings.ts` / `me-username.ts` now use `public.usernames` via the service-role client, exactly like `saved_artists`. [confirmed]
2. **Custom column unreadable** — username moved to `public.usernames` (PostgREST-accessible, RLS, FK + unique + check). Email/has_password now read via `auth.admin.getUserById`. [confirmed]
3. **`userMetadata` → `user_metadata`** — fixed in `me-password.ts:101`. The admin API merges metadata keys, so other fields are preserved. [confirmed]
4. **Rate-limit tier** — `me-password.ts` is now `strict` (10/min). [confirmed]
5. **Typecheck wiring** — root `tsconfig.json` references both projects and the build runs root `npx tsc -b` + `vitest run api/functions/`, so `api/` is finally typechecked and the handler tests run in the pipeline. This is the structural fix that matters most — it's what would have caught the original `userMetadata` typo. [confirmed]
6. **Migration 020 deleted**, replaced by 021. [confirmed]

### 🔴 New blocking bug — migration 021 will fail to apply [confirmed]
The backfill targets a column that doesn't exist on `auth.users`:

```sql
UPDATE auth.users
SET user_metadata = jsonb_set(COALESCE(user_metadata, '{}'::jsonb), '{has_password}', 'true'::jsonb)
WHERE encrypted_password IS NOT NULL
  AND COALESCE(user_metadata, '{}'::jsonb)->>'has_password' IS NULL;
```

GoTrue's `auth.users` stores metadata in **`raw_user_meta_data`** (jsonb) — there is no `user_metadata` column on the table (that name only exists in the *API* response shape, which is why it's correct in the TS handlers but wrong in raw SQL). This statement errors with `column "user_metadata" does not exist`. Since the SQL editor runs the file as one implicit transaction, the failure rolls back the whole migration — **the `public.usernames` table never gets created either.**

Fix — both occurrences:
```sql
UPDATE auth.users
SET raw_user_meta_data = jsonb_set(COALESCE(raw_user_meta_data, '{}'::jsonb), '{has_password}', 'true'::jsonb)
WHERE encrypted_password IS NOT NULL
  AND COALESCE(raw_user_meta_data, '{}'::jsonb)->>'has_password' IS NULL;
```
(Worth applying to a throwaway/staging project once before merge — this is exactly the class of error the build can't catch, since migrations run by hand.)

### 🟠 RLS — `Anyone can read usernames` exposes the whole table now [plausible]
```sql
CREATE POLICY "Anyone can read usernames" ON public.usernames FOR SELECT USING (true);
```
This makes `(user_id, username)` for **every user** readable by the anon key (which ships in the frontend bundle) — i.e. anyone can dump the full table and enumerate every handle and its auth `user_id`. It's labeled "for PR 2," but PR 2 doesn't exist yet, so this ships an open read surface ahead of any feature that uses it. Two suggestions:
- **Defer this policy to PR 2**, where the public `/u/{handle}` endpoint actually lands. For PR 1, the owner-read policy is all the app needs (and the service role bypasses RLS anyway).
- When you do add it, resolve handle→profile through a `SECURITY DEFINER` function or a view that returns only what the public page needs, rather than exposing the raw `user_id` and enabling full enumeration.

Also minor: with the public `USING(true)` SELECT policy present, the `Users can read own username` policy is redundant (permissive policies are OR'd). Not a bug — just dead weight you can drop if you keep the public one.

### Minor / no action needed
- If migration 020 was ever applied anywhere (e.g. local testing), 021 doesn't drop the orphaned `auth.users.username` column + unique index it left behind. Likely moot for prod, but worth a manual check on any env where you ran 020. [needs-manual-verification]
- `me-username.ts` keeps the collision pre-check **and** the `23505` fallback — good, that closes the TOCTOU race. 👍

### Verdict
Down to **one true blocker** (the `raw_user_meta_data` column name) and **one policy call** (defer the public-read RLS). Fix the column name and this is mergeable; I'd hold the open `USING(true)` policy for PR 2. Happy to confirm once the migration line is updated.

---

brandonlucasgreen (commented):
## Final review — PR #294 (through `a99625c` + my `d748900`)

Verified the latest state locally (fresh `npm install`, full `npm run build`, and Netlify's own `zip-it-and-ship-it` bundler). Here's where things stand.

### ✅ Everything from the prior reviews is now resolved
- **Migration backfill column** (`da021ea`) — now writes `raw_user_meta_data` (the real `auth.users` column). Correct. The blocking bug is gone.
- **Netlify deploy failure** (`a99625c`) — root cause confirmed: the `me-*.test.ts` files were sitting in `api/functions/`, which is Netlify's `functions` directory, so the deploy tried to ship them as serverless functions (bundling `vitest`/jsdom). Moving them to `functions/__tests__/` fixes it — the deploy preview is now **✅ ready**, and I reproduced the bundler locally: 37 real functions, zero `.test` functions.
- **My earlier CI fix** (`011d4bb`) — the test-mock typecheck error that broke the first build.

### 🔧 One follow-up I pushed (`d748900`)
When the tests moved to `__tests__/`, `api/tsconfig.json` still listed the old `functions/me-*.test.ts` paths. TypeScript treats a non-matching `include` entry as an empty glob (no error), so the handler tests had **silently dropped out of typechecking** — which would have quietly defeated the whole point of wiring `api/` into `tsc -b`. I repointed `include` at the new paths; confirmed all 3 test files are back in the TS program and `tsc -b` + full build are green. (Small and unambiguous, so I pushed it directly per the autofix arrangement — revert if you'd rather it ride in your own commit.)

### 🟠 One open decision (not a blocker for PR 1 to function)
The `Anyone can read usernames` RLS policy (`USING (true)`) is still in migration 021. As flagged last round, this makes every `(user_id, username)` row readable by the anon key now, before the PR 2 feature that needs it exists. Recommend deferring it to PR 2 and resolving handle→profile via a `SECURITY DEFINER` function/view rather than exposing raw `user_id` + full enumeration. Your call — just don't want it shipping silently.

### ⚠️ Supabase Preview check (❌) is a pre-existing infra issue, not this PR
The `Supabase Preview` check fails at **migration 006** (`relation "artists" does not exist`), long before anything this PR touches. Cause: the `artists` table is defined in `supabase/schema.sql`, but Supabase Branching only replays the numbered `migration-*.sql` files — it never applies `schema.sql` — so migration 006's `REFERENCES artists(id)` has nothing to point at. This would fail identically on any branch; PR #294 didn't introduce it. Two implications:
1. It's **not a reason to hold this PR** on its own merits.
2. But it does mean the branching check **never actually validates migration 021** — so 021's correctness rests on code review (now good) until the schema-as-migration gap is fixed. Worth a separate ticket: either seed the base schema as `migration-001` or point branching at a baseline. Happy to scope that separately if you want.

### Verdict
Functionally and for CI (Netlify), this is **mergeable**. Before merge: (a) decide on the public-read RLS policy, and (b) since the Supabase branching check can't exercise it, apply migration 021 once against a throwaway/staging DB to confirm it lands clean. The rest is solid — nice iteration.

---

PR #295, opened 2026-06-27
feat(UNS-31 PR 2): public sharing of saved artists list

brandonlucasgreen:
## Summary

Public sharing of saved artists list (UNS-31 PR 2, depends on PR #294 merged).

**Spec:** `ia-writer://open?path=Brain:%20/projects/Unstream/specs/UNS-share-saved-artists-list-spec.md`
**GH Issue:** #31

## Amendments (2026-06-27 18:07 ET) — all reflected

1. **Rate limit:** `standard` tier (30/min/IP) on `GET /api/public/saved-artists/:handle` — `checkRateLimit(ip, "standard", ...)` in `public-saved-artists.ts`.
2. **Static HTML edge function:** `api/edge/u-handle.ts` renders `/u/:handle`, same pattern as `api/edge/artist-page-static.ts`. React hydration is progressive enhancement (Copy URL button only). One renderer per CLAUDE.md.
3. **No view counter, no notification emails in v1.** Public endpoint is read-only with no side effects.
4. **Reserved-handle list:** server-side validator (`api/lib/reserved-handles.ts`) used in `user-sharing.ts`. 24 handles: `admin, api, settings, login, signup, signin, register, logout, support, about, privacy, terms, dashboard, u, a, artist, www, mail, ftp, root, help, docs, status, blog`.

### 🔴 FIX 1 — Stored XSS in edge function (BLOCKING, fixed)
Artist `slug` was interpolated into an `href` attribute without `escapeHtml()` in `api/edge/u-handle.ts` SSR output. Fixed by:
- `escapeHtml(slug)` in the `href` render — consistent with `name` and `imageUrl` escaping.
- Save-side slug charset validation (`^[a-z0-9](?:[a-z0-9-]{1,18}[a-z0-9])$`) in `saved-artists.ts handleSave` as defense-in-depth. Rejects malformed slugs with 400.
- 15 new tests in `xss-defense.test.ts`.

### 🟠 FIX 2 — user_public_ids dropped (Option A: boring beats clever)
The `user_public_ids` table was never read by either renderer — both `u-handle.ts` and `public-saved-artists.ts` resolve via `public.usernames` filtered on `username` + `saved_artists_public`. Chose **Option A** (drop the table) because:
- Two tables tracking the same concept (handle → user_id) is architectural debt.
- The rename-sync in `me-username.ts` was a no-op write — reads key off `usernames.username`, so a rename takes effect immediately regardless.
- Reserved-handle enforcement is already handled server-side by `isReservedHandle()` in `user-sharing.ts` (the only path that flips `saved_artists_public` to true) + migration 021's CHECK constraints on `usernames.username`.
- Migration 022 kept (sequential numbering preserved) with an architecture note documenting the decision. `me-username.ts` rename-sync code deleted (not commented out — dead code drifts).

### FIX 3 — .single() → .maybeSingle()
`u-handle.ts` now uses `.maybeSingle()` for consistency with `public-saved-artists.ts`. `usernames.username` is unique so functionally identical, but `.single()` logs a noisy error per unknown handle (bot traffic).

### New
- `supabase/migration-022-public-sharing.sql` — `saved_artists_public` column on `usernames` (user_public_ids table dropped in round 2)
- `api/lib/reserved-handles.ts` — shared reserved-handle list
- `api/functions/user-sharing.ts` — `GET/POST /api/me/saved-artists-sharing`
- `api/functions/public-saved-artists.ts` — `GET /api/public/saved-artists/:handle`
- `api/edge/u-handle.ts` — static HTML renderer for `/u/:handle`
- `apps/web/src/components/SharingControls.tsx` — three-state widget
- `apps/web/src/pages/PublicSavedArtistsPage.tsx` — SPA hydration for Copy URL
- `api/functions/__tests__/user-sharing.test.ts` — 10 handler tests
- `api/functions/__tests__/public-saved-artists.test.ts` — 7 handler tests
- `api/functions/__tests__/xss-defense.test.ts` — 15 XSS defense tests (round 2)
- `apps/web/tests/unit/SharingControls.test.tsx` — 5 component tests

### Modified
- `api/functions/me-username.ts` — removed user_public_ids rename-sync (round 2)
- `api/functions/saved-artists.ts` — added slug validation in handleSave (round 2)
- `api/functions/__tests__/me-username.test.ts` — updated mocks for removed sync
- `api/functions/__tests__/user-sharing.test.ts` — updated mocks for removed user_public_ids
- `apps/web/src/main.tsx` — `/u/:handle` route
- `apps/web/src/pages/DashboardPage.tsx` — mount `<SharingControls>` below saved artists
- `netlify.toml` — edge function route + API redirects
- `data/shipped-features.json` — entry on merge

## Test convention

Handler tests are in `api/functions/__tests__/` (NOT in `api/functions/` root — Netlify bundles top-level `.ts` files as functions; co-located `.test.ts` breaks deploys). Component tests in `apps/web/tests/unit/`. This follows the PR 1 convention.

## Verification

- `npx tsc -b` — clean
- `npm run test:unit` — 351 tests pass (23 files)
- `npx vitest run api/functions/` — 53 tests pass (6 files, including 15 new XSS defense tests)
- `npm run build` — all tests + Vite build pass
- `npm run lint` — no new lint errors in changed files

## Key decisions

- **Single source of truth:** `public.usernames` is the only table for handle resolution. `saved_artists_public` flag lives on `usernames`. No separate public-handle table.
- **Public endpoint join:** goes through `public.usernames` (where username + sharing flag live) → `saved_artists` by `user_id`. Never touches `auth.users` directly.
- **Edge function 404:** falls through to `context.next()` (SPA), which renders the `PublicSavedArtistsPage` component — it fetches the API and shows a 404 message. This matches the `artist-page-static.ts` pattern.
- **Slug escaping:** all artist fields (slug, name, imageUrl) are `escapeHtml()`-d in edge function SSR output. Slug is also validated on save as defense-in-depth.

### DIFF
--- a/api/edge/u-handle.ts
+++ b/api/edge/u-handle.ts
@@ -1,0 +1,239 @@
+// Edge function: /u/{handle}
+// Static HTML renderer for public saved-artists pages.
+// Models after api/edge/artist-page-static.ts — same pattern, same RLS-aware data fetching.
+// The edge function is the single renderer per CLAUDE.md "one route, one renderer".
+// React app hydrates client-side for the Copy URL button only (progressive enhancement).
+
+import { Context } from "https://edge.netlify.com";
+import { createClient } from "https://esm.sh/@supabase/supabase-js@2";
+
+function escapeHtml(str: string): string {
+  return str.replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/>/g, '&gt;').replace(/"/g, '&quot;').replace(/'/g, '&#39;');
+}
+
+const LOGO_SVG = '<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 110 110" aria-hidden="true"><defs><filter id="gs"><feColorMatrix type="saturate" values="0"/></filter></defs><g transform="translate(22,22) scale(1.8333)" filter="url(#gs)"><path fill="#50A5E6" d="M30 22c-3 0-6.688 7.094-7 10-.421 3.915 2 4 2 4h11V26s-3.438-4-6-4z"/><ellipse transform="rotate(-60 27.574 28.49)" fill="#1C6399" cx="27.574" cy="28.489" rx="5.848" ry="1.638"/><path fill="#F9CA55" d="M20.086 0c1.181 0 2.138.957 2.138 2.138 0 .789.668 10.824.668 10.824L17.948 18V2.138C17.948.957 18.905 0 20.086 0z"/><path fill="#FFDC5D" d="M18.875 4.323c0-1.099.852-1.989 1.903-1.989 1.051 0 1.903.891 1.903 1.989 0 0 .535 5.942 1.192 9.37.878 1.866 1.369 4.682 1.261 6.248.054.398 5.625 5.006 5.625 5.006-.281 1.813-2.259 6.155-4.759 8.159l-3.521-2.924c-2.885-.404-4.458-3.331-4.458-4.264 0-2.984.854-21.595.854-21.595z"/><path fill="#50A5E6" d="M6 22c3 0 6.688 7.094 7 10 .421 3.915-2 4-2 4H0V26s3.438-4 6-4z"/><ellipse transform="rotate(-30 8.424 28.489)" fill="#1C6399" cx="8.426" cy="28.489" rx="1.638" ry="5.848"/><path fill="#F9CA55" d="M16.061.011c-1.266-.127-2.333.864-2.333 2.103 0 .78-.184 10.319-.184 10.319L17.895 18l.062-15.765c0-1.106-.795-2.114-1.896-2.224z"/><path fill="#FFDC5D" d="M17.125 4.323c0-1.099-.852-1.989-1.903-1.989-1.051 0-1.903.891-1.903 1.989 0 0-.535 5.942-1.192 9.37-.878 1.866-1.369 4.682-1.261 6.248-.054.398-5.625 5.006-5.625 5.006C5.522 26.76 7.5 31.102 10 33.106l3.521-2.924c2.885-.404 4.458-3.331 4.458-4.264 0-2.984-.854-21.595-.854-21.595z"/><path fill="#F9CA55" d="M17.958 25.823c-.414 0-.75-.336-.75-.75V2.792c0-.414.336-.75.75-.75s.75.336.75.75v22.282c.001.413-.335.749-.75.749z"/></g><path d="M14,52 A41,41 0 0,1 96,52" fill="none" stroke="currentColor" stroke-width="8" stroke-linecap="round"/><line x1="14" y1="52" x2="14" y2="64" stroke="currentColor" stroke-width="7" stroke-linecap="round"/><line x1="96" y1="52" x2="96" y2="64" stroke="currentColor" stroke-width="7" stroke-linecap="round"/><rect x="3" y="60" width="22" height="28" rx="9" fill="currentColor"/><rect x="85" y="60" width="22" height="28" rx="9" fill="currentColor"/></svg>';
+
+const CSS = `
+  :root { --bg: #0d0d0d; --bg2: #1a1a1a; --text: #f0f0f0; --muted: #999; --border: #2a2a2a; --accent: #ff6b35; --footer-border: #1a1a1a; }
+  @media (prefers-color-scheme: light) {
+    :root { --bg: #ffffff; --bg2: #f5f5f5; --text: #1a1a1a; --muted: #555; --border: #e0e0e0; --accent: #e55a2b; --footer-border: #e0e0e0; }
+  }
+  * { margin: 0; padding: 0; box-sizing: border-box; }
+  body { font-family: 'Golos Text', system-ui, sans-serif; background: var(--bg); color: var(--text); min-height: 100vh; display: flex; flex-direction: column; }
+  a { color: inherit; }
+  .container { max-width: 640px; margin: 0 auto; padding: 0 24px; width: 100%; }
+  .page-content { position: relative; flex: 1; display: flex; flex-direction: column; }
+  .site-header { padding: 16px; border-bottom: 1px solid var(--border); display: flex; align-items: center; justify-content: space-between; gap: 16px; }
+  .site-header .brand { display: flex; align-items: center; gap: 8px; font-size: 20px; font-weight: 700; color: var(--text); text-decoration: none; flex-shrink: 0; }
+  .site-header .brand:hover { opacity: 0.8; }
+  .site-header .brand svg { flex-shrink: 0; }
+  .artist-card { display: flex; align-items: center; gap: 16px; padding: 16px; border-radius: 12px; border: 1px solid var(--border); background: var(--bg2); text-decoration: none; color: var(--text); transition: background 0.15s; }
+  .artist-card:hover { background: var(--border); }
+  .artist-avatar { width: 56px; height: 56px; border-radius: 50%; object-fit: cover; flex-shrink: 0; background: var(--border); display: flex; align-items: center; justify-content: center; font-size: 24px; font-weight: 600; color: var(--muted); overflow: hidden; }
+  .artist-name { font-weight: 600; font-size: 16px; }
+  .supported-badge { display: inline-flex; align-items: center; gap: 4px; padding: 2px 8px; border-radius: 999px; font-size: 11px; font-weight: 500; background: rgba(255,107,53,0.15); color: var(--accent); }
+  .copy-btn { display: inline-flex; align-items: center; gap: 8px; padding: 8px 16px; border-radius: 8px; border: 1px solid var(--border); background: var(--bg2); color: var(--text); font-size: 14px; cursor: pointer; transition: background 0.15s; }
+  .copy-btn:hover { background: var(--border); }
+  .copy-btn.copied { color: #22c55e; border-color: rgba(34,197,94,0.3); }
+`;
+
+export default async function handler(request: Request, context: Context) {
+  try {
+    const url = new URL(request.url);
+    const handle = url.pathname.replace(/^\/u\//, '').replace(/\/$/, '');
+
+    if (!handle) return context.next();
+
+    const supabaseUrl = Deno.env.get("SUPABASE_URL");
+    const supabaseKey = Deno.env.get("SUPABASE_SERVICE_KEY");
+    if (!supabaseUrl || !supabaseKey) return context.next();
+
+    const supabase = createClient(supabaseUrl, supabaseKey);
+
+    // 5-second timeout for Supabase fetches
+    const controller = new AbortController();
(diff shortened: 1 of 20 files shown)


api/edge/u-handle.ts
```diff
@@ -0,0 +1,239 @@
...
+            ? '<span class="supported-badge"><svg width="12" height="12" fill="currentColor" viewBox="0 0 20 20"><path fill-rule="evenodd" d="M16.707 5.293a1 1 0 010 1.414l-8 8a1 1 0 01-1.414 0l-4-4a1 1 0 011.414-1.414L8 12.586l7.293-7.293a1 1 0 011.414 0z" clip-rule="evenodd"/></svg> Supported</span>'
+            : '';
+
+          return `<a href="${profileUrl}" class="artist-card">${avatarHtml}<div style="flex:1;min-width:0"><div class="artist-name">${escapeHtml(name)}</div>${supportedBadge}</div></a>`;
+        }).join('')
+      : '<p style="color:var(--muted);text-align:center;padding:48px 0">No saved artists yet.</p>';
+
+    // JSON-LD structured data
+    const jsonLd = {
+      '@context': 'https://schema.org',
```
brandonlucasgreen: **Security — stored XSS · confirmed** @brandonlucasgreen

`name` and `imageUrl` are passed through `escapeHtml()`, but `slug` is not — it flows into `profileUrl` and then straight into the `href` attribute:

```js
const slug = artistRow?.slug || row.artist_slug || '';
const profileUrl = slug ? `https://unstream.stream/a/${slug}` : '#';
...
return `<a href="${profileUrl}" class="artist-card">...`;
```

`row.artist_slug` is attacker-controlled and unsanitized. In `api/functions/saved-artists.ts` → `handleSave`, the saved slug is `(body.artistId as string) || (body.slug as string)` and is written to `saved_artists.artist_slug` verbatim — no regex, no charset check, and migration 014 dropped the FK so it's free text. For an unclaimed artist (the common case) `artist_id` is null, the `artists!left` join yields nothing, so `artistRow` is null and the raw `row.artist_slug` is rendered here.

Exploit: an authenticated user saves an artist with
`artistId = ">``&lt;img src=x onerror=alert(document.cookie)&gt;'``, enables sharing, and the rendered HTML becomes
`<a href="https://unstream.stream/a/">``&lt;img src=x onerror=...&gt;``" ...>` — script runs in the browser of **every visitor** to `/u/<handle>`, which is exactly the shareable public link.

The SPA renderer (`PublicSavedArtistsPage.tsx`) is safe because React escapes attributes; only this edge HTML path is vulnerable.

Fix: `escapeHtml(slug)` before building `profileUrl` (and ideally validate the slug charset on save in `saved-artists.ts` as defense in depth). Same applies to `imageUrl` being attacker-controlled — escaping prevents attribute breakout, but note a malicious `artist_image_url` still loads in visitors' browsers as a tracking/IP-logging beacon; consider validating the host on save.

---

brandonlucasgreen (commented):
## Senior staff review — PR #295 (public sharing of saved artists)

@brandonlucasgreen — one **confirmed stored XSS** that should block merge, plus a notable architectural/regression issue and some smaller notes. Findings are labeled `confirmed` / `plausible` / `needs-manual-verification` and limited to what's in the code. (Submitting as a comment since GitHub won't let me formally request changes on my own PR — treat #1 as a blocker.)

### 🔴 Security

**1. Stored XSS in `api/edge/u-handle.ts` — `confirmed`** (see inline comment on the `href` render)
The artist `slug` is interpolated into an `href` attribute **without** `escapeHtml()`, while `name`/`imageUrl` right next to it *are* escaped. The slug is attacker-controlled: `saved-artists.ts → handleSave` stores `body.artistId` verbatim into `saved_artists.artist_slug` (no validation, FK dropped in migration 014). A signed-in user can save an artist with a slug like `">``&lt;img src=x onerror=...&gt;'``, enable sharing, and execute script in every visitor's browser on the public `/u/<handle>` page. **Fix:** `escapeHtml(slug)`; validate slug charset on save as defense-in-depth. The SPA path is safe (React escapes).

**2. Unvalidated `artist_image_url` rendered as `` on a public page — `plausible` (low)**
Escaping blocks attribute breakout, but the URL itself is attacker-set and loads in visitors' browsers — usable as an IP/tracking beacon on the shared page. Consider validating the host on save.

**3. Auth / authz — `confirmed` OK.** Write endpoints (`user-sharing.ts`, `me-username.ts`) require a verified Bearer token via `auth.getUser`; mutations are scoped by `.eq('user_id', user.userId)`. Read endpoints are intentionally public, gated on `saved_artists_public`, rate-limited at `standard` (30/min/IP).

**4. Data exposure — `confirmed` OK.** `public-saved-artists.ts` and the edge function select `user_id` only for the internal join and never emit it; response shape is `{owner_display_name, saved_artists:[{slug,name,image_url,supported}]}`. A test asserts no `user_id`/`email` leak. 404 (not 403) on private/unknown handles avoids existence leaks. Good.

**5. Injection (SQL) — `confirmed` OK.** All queries use parameterized `supabase-js` `.eq()`; no string-built SQL. JSON-LD is `JSON.stringify(...).replace(/</g,'\\u003c')` — `</script>` breakout is handled.

**6. Reserved-handle enforcement — `confirmed` adequate.** Enabling sharing runs `isReservedHandle(handle)` server-side (`user-sharing.ts`), so reserved names can't be turned public even though reads key off `usernames` rather than the constrained `user_public_ids` table.

### 🟠 Regression / correctness

**7. `user_public_ids` is write-only — never read by either renderer — `confirmed` (architectural / dead weight).**
Both read paths (`u-handle.ts`, `public-saved-artists.ts`) resolve via `usernames` filtered on `username` + `saved_artists_public`. The new `user_public_ids` table — its UNIQUE/format/reserved CHECK constraints, RLS, and the rename-sync added to `me-username.ts` — is never consulted when serving `/u/:handle`. Consequences:
  - The `me-username.ts` comment *"the public endpoint will 404 on the stale handle until this syncs"* is **factually wrong** — reads use `usernames.username`, so a rename takes effect immediately and a sync failure has no visible effect. Misleading for the next maintainer.
  - Its CHECK constraints (format/reserved) therefore protect nothing on the read surface; the real guards are migration 021's `usernames` constraints + the runtime `isReservedHandle` check. Per the repo's "boring beats clever / no dead weight" principles, either wire reads through `user_public_ids` or drop the table and the sync and keep `saved_artists_public` on `usernames` as the single source of truth.

**8. Rename does not clobber the sharing flag — `confirmed` OK (verified).** The `me-username` upsert payload is `{user_id, username, updated_at}`; `ON CONFLICT DO UPDATE` only touches those columns, so `saved_artists_public` is preserved across a rename. Good — worth keeping in mind if the payload ever grows.

**9. Null/empty behavior — `confirmed` OK.** `supported: row.supported === true` coerces null→false; empty saved list renders the "No saved artists yet" branch; missing image falls back to the initial-letter avatar. Edge-function `.catch → context.next()` degrades to the SPA.

**10. `.single()` vs `.maybeSingle()` drift — `plausible` (low).** The edge function uses `.single()` (errors on 0 rows) where the API function uses `.maybeSingle()`. `usernames.username` is unique so it's functionally fine, but `.single()` logs a noisy error per unknown handle (and unknown handles are exactly what bots will hit). Prefer `.maybeSingle()` for consistency and quieter logs.

### API contract
`GET /api/public/saved-artists/:handle` and the edge OG/meta are consistent (`owner_display_name` == username). No drift between the JSON API and the SSR HTML beyond the XSS escaping gap in #1.

---
**Bottom line:** #1 (stored XSS) should be fixed before merge; #7 is worth resolving now while the design is fresh (two-sources-of-truth ambiguity that breeds later bugs). Everything else is minor.

---

brandonlucasgreen (commented):
## Second-pass review — PR #295

@brandonlucasgreen — re-reviewed focusing on the RLS layer (since the read endpoints use the service-role key, I wanted to check what the *anon* key shipped in the web bundle can reach directly). Two new findings; the XSS in my first review (`u-handle.ts`, unescaped `slug` in `href`) remains the merge blocker.

### 🔴 New — Security

**11. Anon-readable `user_id` + sharing status — `confirmed` (RLS comment is wrong).**
`migration-022` adds `CREATE POLICY "Anyone can read public handles" ON user_public_ids FOR SELECT USING (true)` with the comment *"Only the public_handle column is publicly readable; user_id is never exposed to clients."* That's false — **RLS is row-level, not column-level**, so `USING (true)` exposes the whole row. With the public anon key, anyone can run `supabase.from('user_public_ids').select('user_id, public_handle')` and harvest every `(auth UUID, handle)` pair. The same holds for `usernames` (migration 021's `"Anyone can read usernames" USING (true)`), which **this PR extends with `saved_artists_public`** — so every user's auth UUID *and* sharing status become anon-enumerable there too.

The clean part: **nothing legitimately needs these anon-read policies.** Both `public-saved-artists.ts` and `user-sharing.ts`, plus the edge function, use the service-role client (`getClient()` / `SUPABASE_SERVICE_KEY`) which bypasses RLS; the SPA only calls `/api/...` and never queries these tables directly. So the `USING (true)` policies have no consumer and only create exposure.
- **Fix:** drop the anon `SELECT USING (true)` policy on `user_public_ids` (and reconsider it on `usernames`), keeping only owner-scoped policies. Service-role reads are unaffected.
- Ties into first-review finding #7 (`user_public_ids` is write-only / never read): if the table stays it shouldn't be anon-readable; if it's removed, this resolves itself.
- Scope note: `saved_artists` *is* correctly owner-scoped (`auth.uid() = user_id`), so list **contents** are not directly anon-readable — good. This finding is UUID + sharing-flag enumeration, not list-content leakage.

**12. No CDN cache purge on "Make private" — `plausible` (privacy, low/moderate).**
The edge function sets `Netlify-CDN-Cache-Control: s-maxage=300, stale-while-revalidate=300` and a `Cache-Tag: user-share-${handle}`, but `user-sharing.ts` never purges that tag when a user disables sharing (or removes an artist). So after a user clicks "Make private," their list can keep serving publicly from the CDN for up to ~5 min (s-maxage) and be revalidated-while-stale beyond that — the opposite of what the safety control implies. Since "Make private" is the user's panic button, this window matters.
- **Fix:** call Netlify's purge-by-tag API from `user-sharing.ts` on disable (the `Cache-Tag` is already there for exactly this), or drop `s-maxage` to something small for these pages.

### Unchanged from first review
- **#1 stored XSS (`u-handle.ts`, unescaped `slug` → `href`)** — still the blocker. Fix `escapeHtml(slug)` + validate slug on save.
- #7 `user_public_ids` write-only/never-read (now compounded by #11), #2 image-URL beacon, #10 `.single()` vs `.maybeSingle()` — all still stand.
- Confirmed-OK: token auth + user-scoped mutations, no email/user_id in API responses, parameterized queries, JSON-LD escaping, reserved-handle runtime check, rename preserving the sharing flag.

Net: still two things worth fixing before merge (#1 XSS, #11 anon RLS exposure); #12 is a quick follow-up.

---

PR #296, opened 2026-06-29
feat(UNS-144): user profile improvements

brandonlucasgreen:
Fixes UNS-144

### A. Settings page (`/settings`)
- **Location field**: new free-text input after the username field, capped at 100 chars, trims whitespace. Optional — clears to `null` when empty.
- **Sharing toggle moved**: the public/private toggle (`SharingControls`) has been moved from the Dashboard to a new "Sharing" section on the Settings page. Data path unchanged (`profiles.saved_artists_public` on `public.usernames`).

### B. Public profile (`/u/<handle>`)
- **Subheading removed**: the "a listener's saved artists on Unstream" subheading has been removed from both the edge function (SSR) and the React SPA.
- **Location rendered**: if the user has set a location, it displays below their name on the public profile. Empty/null locations show nothing (no placeholder).

### Migration
- `20260628200000_user-location.sql`: adds `location TEXT` column to `public.usernames` (nullable, no default).

### API
- New endpoint `/api/me/location` (GET + POST) for reading/writing the location field.
- `/api/me/settings` GET now includes `location` in the response.
- `/api/public/saved-artists/:handle` now includes `owner_location` in the response.

### Other
- Fixed pre-existing JSON syntax error in `package.json` (missing closing brace — was already broken on `main`).

## Verification
- `npx tsc -b` passes
- `npm run build` passes (full build + sitemap)
- `npx vitest run` — all 23 tests pass across me-settings, me-location, and public-saved-artists suites

## To verify after merge
- Supabase auto-migrate picks up the new column
- Settings page shows location field + sharing toggle
- Public profile shows location (when set) and no subheading
- Dashboard no longer shows sharing controls

### DIFF
--- a/api/functions/me-location.ts
+++ b/api/functions/me-location.ts
@@ -90,13 +90,20 @@ export async function handler(event: {
     // Allow null to clear the field
     if (raw === null) {
       try {
-        const { error: updateError } = await client
+        const { error: upsertError } = await client
           .from('usernames')
-          .update({ location: null })
-          .eq('user_id', user.userId);
+          .upsert(
+          { user_id: user.userId, location: null },
+          { onConflict: 'user_id' }
+        );
 
-        if (updateError) {
-          console.error('[me-location] Error clearing location:', updateError.message);
+        if (upsertError) {
+          // No usernames row exists — username column is NOT NULL, so insert fails.
+          // Location is already effectively null, so this is the desired state.
+          if (upsertError.code === '23502') {
+            return { statusCode: 200, headers: CORS_HEADERS, body: JSON.stringify({ location: null }) };
+          }
+          console.error('[me-location] Error clearing location:', upsertError.message);
           return { statusCode: 500, headers: CORS_HEADERS, body: JSON.stringify({ error: 'Failed to update location' }) };
         }
 
@@ -114,13 +121,20 @@ export async function handler(event: {
     }
 
     try {
-      const { error: updateError } = await client
+      const { error: upsertError } = await client
         .from('usernames')
-        .update({ location: trimmed || null })
-        .eq('user_id', user.userId);
+        .upsert(
+          { user_id: user.userId, location: trimmed || null },
+          { onConflict: 'user_id' }
+        );
 
-      if (updateError) {
-        console.error('[me-location] Error updating location:', updateError.message);
+      if (upsertError) {
+        // No usernames row — username column is NOT NULL, so insert fails.
+        // User must set a username before they can set a location.
+        if (upsertError.code === '23502') {
+          return { statusCode: 400, headers: CORS_HEADERS, body: JSON.stringify({ error: 'Set a username before setting your location.' }) };
+        }
+        console.error('[me-location] Error updating location:', upsertError.message);
         return { statusCode: 500, headers: CORS_HEADERS, body: JSON.stringify({ error: 'Failed to update location' }) };
       }
 
--- a/api/functions/me-location.ts
+++ b/api/functions/me-location.ts
@@ -1,0 +1,139 @@
+// API endpoint: /api/me/location
+// GET  — returns the current user's location string (or null).
+// POST — validates and sets the user's location.
+// Body: { location: string | null }
+// Returns the new location on success, or a friendly error on failure.
+
+import { createClient } from '@supabase/supabase-js';
+import { getClient } from './db';
+import { checkRateLimit, getClientIp } from './ratelimit';
+
+const CORS_HEADERS = {
+  'Content-Type': 'application/json',
+  'Access-Control-Allow-Origin': '*',
+  'Access-Control-Allow-Headers': 'Content-Type, Authorization',
+  'Access-Control-Allow-Methods': 'GET, POST, OPTIONS',
+};
+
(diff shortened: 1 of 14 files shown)


api/functions/me-location.ts
```diff
@@ -0,0 +1,139 @@
...
+        .update({ location: trimmed || null })
+        .eq('user_id', user.userId);
+
+      if (updateError) {
+        console.error('[me-location] Error updating location:', updateError.message);
+        return { statusCode: 500, headers: CORS_HEADERS, body: JSON.stringify({ error: 'Failed to update location' }) };
+      }
+
+      return {
+        statusCode: 200,
```
brandonlucasgreen, line 142: **Regression / edge case — `confirmed`** @brandonlucasgreen

This `POST` uses `.update().eq('user_id', ...)`, but `location` lives on the `usernames` table and a row only exists once the user has set a username (`me-username.ts` creates it via `upsert`). For a signed-in user who has **not** set a username yet, this `UPDATE` matches **zero rows, succeeds silently**, and the handler still returns `200` with `{ location: trimmed }` — so the UI shows "Location saved." but nothing was persisted (and it's gone on reload).

This path is reachable: on `SettingsPage.tsx` the `LocationField` sits in the same section as `UsernameField`, so a brand-new account can fill in a location before ever picking a username.

Fix options:
- `upsert({ user_id, location, updated_at }, { onConflict: 'user_id' })` like `me-username.ts` does — but note the `usernames` row may have a `NOT NULL`/unique `username`, so an upsert that omits `username` could fail; verify the table constraints first.
- Or check the update's affected-row count and return a clear error (e.g. "Set a username before adding a location") when no row was updated.

Whichever way, the silent success is the part to remove.

---

brandonlucasgreen (commented):
## Review — PR #296 (UNS-144: user profile improvements)

@brandonlucasgreen — reviewed as a security + regression pass. Overall this is clean and consistent with the existing `me-*` endpoint patterns. **One confirmed regression** (silent no-op on save), inline on `me-location.ts`. Nothing security-blocking.

### Security
- **XSS / injection — `confirmed` safe.** Location is escaped server-side in the edge SSR (`u-handle.ts` → `escapeHtml`) and auto-escaped by React in the SPA (`PublicSavedArtistsPage.tsx`). No `dangerouslySetInnerHTML`. ✓
- **Authorization — `confirmed` correct.** `me-location` GET/POST authenticate via Bearer token and scope all DB access to `user_id`. No way to read/write another user's location. ✓
- **Data exposure — `confirmed` safe.** `owner_location` is only returned/rendered **after** the `saved_artists_public` check in both the API (`public-saved-artists.ts:66`) and the edge function (`u-handle.ts:75`). A private profile leaks nothing — same as the existing 404-not-403 pattern. ✓
- **Input validation — `confirmed` adequate.** Length capped at 100, trimmed, `null` clears, non-string coerced to empty → null. No crash on bad input. Minor (non-blocking): newlines/control chars aren't stripped, but since every render path escapes, this is cosmetic only.
- **CORS — `confirmed` consistent (not new debt).** `Access-Control-Allow-Origin: '*'` matches the documented pattern in `me-settings.ts`/`me-username.ts`/`saved-artists.ts`. Acceptable because auth is via Bearer token, not cookies. Not introduced by this PR.
- **Secrets / unsafe defaults — none found.** No secrets logged; error logs use `.message` only. RLS isn't newly required — the column is added to an existing table and accessed via the service client.

### Regression
- **`confirmed`: silent no-op when no `usernames` row exists** — see inline comment on `me-location.ts`. `UPDATE ... WHERE user_id` affects zero rows for a user who hasn't set a username, but the handler returns `200` + the submitted value. This is the one item I'd fix before merge.
- **API contract drift — `confirmed`, handled.** `/api/me/settings` (+`location`) and `/api/public/saved-artists/:handle` (+`owner_location`) both grew fields; all consumers (`SettingsPage`, `PublicSavedArtistsPage`, edge SSR) and tests were updated in-PR. No stale shape. ✓
- **Null/empty behavior — `confirmed` covered.** `?? null` everywhere; empty/whitespace renders nothing (no placeholder), matching the stated intent. Tests exercise set/null/empty/over-limit. ✓
- **State move (SharingControls Dashboard → Settings) — `confirmed` clean.** Import removed from `DashboardPage`, added to `SettingsPage`; data path (`saved_artists_public`) unchanged.

### Nits (non-blocking)
- Several new files (`me-location.ts`, `LocationField.tsx`, migration, test) are missing a trailing newline.
- `package.json` brace fix is unrelated to UNS-144 but harmless — worth a one-line mention in the PR body (already noted).

---

api/functions/me-location.ts
brandonlucasgreen: Fixed — switched to `.upsert({user_id, location}, {onConflict: 'user_id'})` for both POST branches. The `usernames` table has `user_id` as PRIMARY KEY so `onConflict: 'user_id'` is correct. Since `username` is NOT NULL, upserting without it fails with Postgres 23502 — the handler now detects this and returns 400 'Set a username before setting your location.' (SET branch) or 200 with `location: null` (CLEAR branch, desired state already achieved). Added regression tests for the no-row case in both branches. Pushed commit af3dfcc.

PR #297, opened 2026-06-29
fix(UNS-149): migration 002 self-contained

brandonlucasgreen:
## fix(UNS-149): migration 002 self-contained

Fixes UNS-149

### Problem

Every Supabase branch preview check on every PR was failing with:

```
ERROR: relation "artists" does not exist (SQLSTATE 42P01)
```

Migration 002 (`20260401000000_artist-profiles.sql`) opens with `ALTER TABLE artists` but the `artists` table was never created by any migration file — it only exists in `supabase/schema.sql` (a one-time dashboard snapshot from before migrations existed). On fresh branch preview DBs, `artists` doesn't exist, so the ALTER fails immediately.

### Fix

Added `CREATE TABLE IF NOT EXISTS public.artists (...)` at the top of migration 002 (both the timestamped and legacy copies), sourced verbatim from `supabase/schema.sql:12-24`. Also added the two indexes (`idx_artists_slug`, `idx_artists_updated`) with `IF NOT EXISTS` guards.

### Column list in the CREATE

```sql
id uuid primary key default gen_random_uuid(),
slug text unique not null,
name text not null,
image_url text,
match_confidence text check (match_confidence in ('verified', 'unverified')),
source text not null default 'auto',
created_at timestamptz not null default now(),
updated_at timestamptz not null default now(),
last_enriched_at timestamptz
```

This matches `schema.sql` exactly. The original CHECK constraint (`'verified' | 'unverified'`) is included because migration 002 immediately drops it and replaces it with the expanded version (`'verified' | 'unverified' | 'claimed'`).

### Idempotency

- **Fresh DB (branch preview):** CREATE runs → table exists → ALTER succeeds → rest of migration applies.
- **Existing DB (prod, post-ledger-repair):** `CREATE IF NOT EXISTS` is a no-op → ALTER succeeds → no behavior change.

### Files changed

- `supabase/migrations/20260401000000_artist-profiles.sql` — added CREATE TABLE + indexes at top
- `supabase/migration-002-artist-profiles.sql` — identical change (legacy copy kept in sync)

### DIFF
--- a/supabase/migration-002-artist-profiles.sql
+++ b/supabase/migration-002-artist-profiles.sql
@@ -1,6 +1,22 @@
 -- Migration 002: Artist Profiles & Claim Flow
 -- Run this in your Supabase SQL editor after enabling Supabase Auth.
 
+-- Foundation: ensure the artists table exists (originally created via dashboard before migrations)
+CREATE TABLE IF NOT EXISTS public.artists (
+  id uuid primary key default gen_random_uuid(),
+  slug text unique not null,
+  name text not null,
+  image_url text,
+  match_confidence text check (match_confidence in ('verified', 'unverified')),
+  source text not null default 'auto',
+  created_at timestamptz not null default now(),
+  updated_at timestamptz not null default now(),
+  last_enriched_at timestamptz
+);
+
+create index if not exists idx_artists_slug on public.artists(slug);
+create index if not exists idx_artists_updated on public.artists(updated_at);
+
 -- Update match_confidence to support 'claimed'
 alter table artists drop constraint if exists artists_match_confidence_check;
 alter table artists add constraint artists_match_confidence_check
--- a/supabase/migrations/20260401000000_artist-profiles.sql
+++ b/supabase/migrations/20260401000000_artist-profiles.sql
@@ -1,6 +1,22 @@
 -- Migration 002: Artist Profiles & Claim Flow
 -- Run this in your Supabase SQL editor after enabling Supabase Auth.
 
+-- Foundation: ensure the artists table exists (originally created via dashboard before migrations)
+CREATE TABLE IF NOT EXISTS public.artists (
+  id uuid primary key default gen_random_uuid(),
+  slug text unique not null,
+  name text not null,
+  image_url text,
+  match_confidence text check (match_confidence in ('verified', 'unverified')),
+  source text not null default 'auto',
+  created_at timestamptz not null default now(),
+  updated_at timestamptz not null default now(),
+  last_enriched_at timestamptz
+);
+
+create index if not exists idx_artists_slug on public.artists(slug);
+create index if not exists idx_artists_updated on public.artists(updated_at);
+
 -- Update match_confidence to support 'claimed'
 alter table artists drop constraint if exists artists_match_confidence_check;
 alter table artists add constraint artists_match_confidence_check


brandonlucasgreen (commented):
## Senior staff review — PR #297 (UNS-149)

@brandonlucasgreen The diagnosis is correct and the change is safe for prod, but the fix is **incomplete**: it relocates the branch-preview failure rather than eliminating it. Details below.

### Security

- **No prod security regression — confirmed.** `CREATE TABLE IF NOT EXISTS public.artists` and `create index if not exists ...` are no-ops on the existing prod DB (table + RLS already present), so prod behavior is unchanged. Idempotency claim in the PR body checks out.
- **Missing RLS on fresh (migration-built) DBs — plausible (unsafe default / data exposure, preview-scoped).** In `schema.sql` the `artists` table ships with `alter table artists enable row level security` plus the `Public read access` / `Service insert` / `Service update` policies (`schema.sql:108-117`). The new `CREATE TABLE` in this migration adds **none** of that. On a fresh DB built from migrations only (i.e. branch previews), `artists` will exist with **RLS disabled**. Branch previews are ephemeral so the real-world impact is low, but it's a genuine divergence between a schema.sql-built DB and a migration-built DB. Prod is unaffected only by luck of `IF NOT EXISTS`.
- No injection / secret-handling / auth-authorization concerns in this diff.

### Regression

- **The migration will still fail on a fresh branch-preview DB — confirmed (static), CI outcome needs-manual-verification.** Migration 002 ends with:
  ```sql
  create trigger artist_profiles_updated_at
    before update on artist_profiles
    for each row execute function update_updated_at();
  ```
  `update_updated_at()` is defined **only in `schema.sql:123-129`** — it is not created by any file in `supabase/migrations/`. (Verified: `grep -rln "function update_updated_at" supabase/migrations` returns only the *usage* in this file, never a definition.) Since the branch preview starts from an empty DB and runs migrations only (that's why `artists` was missing in the first place), this migration will now get past the `ALTER TABLE artists` line and fail a few lines later with **`function update_updated_at() does not exist`**. The fix moves the error from line 21 to line 61. The "Supabase Preview" check is still `in_progress` as of this review — worth confirming it actually goes green rather than failing on the next missing object.

- **Same root cause hits a later migration — confirmed.** `20260525110000_featured-embed.sql` runs `ALTER TABLE artist_links ADD COLUMN IF NOT EXISTS display_name text;`, but `artist_links` is **never created by any migration** (only `schema.sql:25-37`). That migration will also fail on a fresh DB. So even patching `artists` + `update_updated_at()` into 002 won't get branch previews fully green.

- **`artists.updated_at` auto-update missing on preview DBs — plausible (changed assumption).** `schema.sql` also creates the `artists_updated_at` trigger (`schema.sql:131-133`); this migration doesn't. Minor, preview-only, but another schema.sql/migration divergence.

### Recommendation

The underlying issue is that the pre-migration `schema.sql` baseline (the `artists` + `artist_links` tables, their RLS/policies/triggers, and the `update_updated_at()` function) was never captured as a migration. Patching just the `artists` table into 002 fixes one symptom of that.

The boring, complete fix is a single **baseline migration** that runs before 002 — e.g. `supabase/migrations/20260331000000_baseline.sql` containing the pre-migration schema verbatim from `schema.sql` (tables, indexes, `update_updated_at()`, RLS + policies + triggers), every statement guarded with `IF NOT EXISTS` / `create or replace` / `drop ... if exists` so it's a no-op on prod. Then revert the `artists` CREATE out of 002 to keep one source of truth. That makes a migrations-only DB match a schema.sql DB and gets every dependent migration (002, featured-embed, etc.) to apply cleanly.

If a full baseline is too much scope right now, the minimum to unblock *this* migration is to also add `update_updated_at()` (and ideally `artist_links` + the `artists` RLS/trigger) here — but that's the partial path that tends to produce another "next missing object" round-trip.

Happy to push the baseline-migration version to this branch if you want it.

---

brandonlucasgreen (commented):
## Senior staff review — @brandonlucasgreen

`confirmed`: this PR is **superseded by #298**, which reverts the migration-002 `CREATE TABLE artists` added here and instead introduces a proper `20260331000000_baseline.sql` that also closes the same-root-cause gaps this PR leaves open (`artist_links`, the `update_updated_at()` function, and the `featured-embed` migration's dependency on `artist_links`).

The change here is correct and idempotent on its own, but partial. **Recommendation: close this PR in favor of #298** rather than merging both — they overlap, and #298 is the complete fix. No security or regression issues in the diff itself beyond the incompleteness.

---

brandonlucasgreen (commented):

### Assessment — superseded (confirmed)

The fix here is correct as far as it goes: adding `CREATE TABLE IF NOT EXISTS public.artists` + guarded indexes to the top of migration 002 resolves the immediate `relation "artists" does not exist (42P01)` on fresh preview DBs, and the idempotency reasoning (no-op on prod) is sound.

But it's **incomplete and overlapping with #298**. The same root cause (pre-migration `schema.sql` objects never captured as a migration) also leaves `artist_links`, the `update_updated_at()` function, and the RLS/policies/triggers missing on fresh DBs — `20260525110000_featured-embed.sql` and migration 002's trigger would still fail. #298 captures the full baseline in a single `20260331000000_baseline.sql` and reverts the need for these in-002 edits.

**Recommendation: close this PR unmerged in favor of #298.** Merging both would leave two divergent edits to migration 002 (the `CREATE TABLE artists` block here vs. #298 owning it in the baseline). No security or regression issues with the diff itself beyond that overlap.

---

PR #298, opened 2026-06-29
fix(UNS-149): baseline migration for pre-migration schema.sql state

brandonlucasgreen:
## Summary

Round 1 (PR #297) patched one symptom — the missing `artists` table — but Claude's review identified three more gaps from the same root cause: the pre-migration `schema.sql` baseline was never captured as a migration.

### Root cause
Branch previews build from `supabase/migrations/` only. Several objects defined only in `schema.sql` were never in any migration:
- `artists` table (round 1 patched this)
- `artist_links` table (needed by `20260525110000_featured-embed.sql`)
- `update_updated_at()` function (needed by migration 002's `artist_profiles_updated_at` trigger)
- RLS + policies + triggers on `artists` and `artist_links`

### Fix
1. **New baseline migration** `20260331000000_baseline.sql` (runs before 002) capturing the full pre-migration `schema.sql` state:
   - `CREATE TABLE IF NOT EXISTS` for `artists`, `artist_links`, `verification_requests`
   - All indexes with `IF NOT EXISTS`
   - `CREATE OR REPLACE FUNCTION update_updated_at()`
   - All `ALTER TABLE ... ENABLE ROW LEVEL SECURITY`
   - All policies using `DROP POLICY IF EXISTS ... ; CREATE POLICY ...` pattern
   - All triggers using `DROP TRIGGER IF EXISTS ... ; CREATE TRIGGER ...` pattern
   - Every statement idempotent → no-op on prod

2. **Reverted round-1 changes** — `CREATE TABLE artists` + indexes removed from both copies of migration 002. The baseline owns those now.

3. **Patched migration 006** (`verification-requests.sql`) — added `IF NOT EXISTS` to indexes and `DROP POLICY IF EXISTS` before `CREATE POLICY`, since the baseline now creates that table and 006's non-idempotent statements would fail on a fresh DB.

### Prod safety
- Baseline is fully idempotent (every statement guarded) → no-op on prod
- Migration 006 patches are also idempotent → no-op on prod (indexes/policies already exist, DROP IF EXISTS + CREATE is safe)
- No schema changes on prod; only affects fresh/preview DBs

### Checklist coverage
Every `CREATE TABLE` / `CREATE INDEX` / `CREATE FUNCTION` / `CREATE TRIGGER` / `CREATE POLICY` / `ENABLE ROW LEVEL SECURITY` from `schema.sql` that predates migration 002 is captured in the baseline. Items created by 002 itself (artist_profiles table, its indexes, RLS, policies, trigger) remain in 002.

### What to verify
- [ ] Supabase branch preview check passes (no 42P01, no `function does not exist`)
- [ ] `20260525110000_featured-embed.sql` applies cleanly (artist_links now exists)
- [ ] No regression on prod (baseline + 006 patches are no-ops)

Closes #297 (supersedes).
Refs UNS-149.

### DIFF
--- a/supabase/migrations/20260331000000_baseline.sql
+++ b/supabase/migrations/20260331000000_baseline.sql
@@ -1,0 +1,131 @@
+-- Baseline migration: captures pre-migration schema.sql state
+-- so that fresh DBs (branch previews) built from migrations only
+-- match the prod DB that was originally set up via schema.sql.
+--
+-- Every statement is idempotent (IF NOT EXISTS / CREATE OR REPLACE /
+-- DROP ... IF EXISTS) so this is a no-op on prod where schema.sql
+-- was applied long ago.
+--
+-- Source: supabase/schema.sql (canonical baseline)
+
+-- ── Tables ──────────────────────────────────────────────────────────
+
+CREATE TABLE IF NOT EXISTS public.artists (
+  id uuid primary key default gen_random_uuid(),
+  slug text unique not null,
+  name text not null,
+  image_url text,
+  match_confidence text check (match_confidence in ('verified', 'unverified')),
+  source text not null default 'auto',
+  created_at timestamptz not null default now(),
+  updated_at timestamptz not null default now(),
+  last_enriched_at timestamptz
+);
+
+CREATE TABLE IF NOT EXISTS public.artist_links (
+  id uuid primary key default gen_random_uuid(),
+  artist_id uuid not null references artists(id) on delete cascade,
+  platform text not null,
+  url text not null,
+  source text not null default 'search',
+  is_direct boolean not null default true,
+  latest_release jsonb,
+  display_order integer,
+  created_at timestamptz not null default now(),
+  updated_at timestamptz not null default now(),
+  unique(artist_id, platform)
+);
+
+CREATE TABLE IF NOT EXISTS public.verification_requests (
+  id uuid primary key default gen_random_uuid(),
+  artist_id uuid not null references artists(id) on delete cascade,
+  user_id uuid not null,
+  email text not null,
+  message text not null,
+  status text not null default 'pending' check (status in ('pending', 'approved', 'rejected')),
+  created_at timestamptz not null default now(),
+  reviewed_at timestamptz,
+  reviewer_notes text,
+  ownership_verified_by uuid,
+  ownership_verified_at timestamptz
+);
+
+-- ── Indexes ─────────────────────────────────────────────────────────
+
+CREATE INDEX IF NOT EXISTS idx_artists_slug ON public.artists(slug);
+CREATE INDEX IF NOT EXISTS idx_artists_updated ON public.artists(updated_at);
+CREATE INDEX IF NOT EXISTS idx_artist_links_artist_id ON public.artist_links(artist_id);
+CREATE INDEX IF NOT EXISTS idx_verification_requests_artist_id ON public.verification_requests(artist_id);
+CREATE INDEX IF NOT EXISTS idx_verification_requests_status ON public.verification_requests(status);
+
+-- ── Function ────────────────────────────────────────────────────────
+
+CREATE OR REPLACE FUNCTION public.update_updated_at()
+RETURNS trigger AS $$
+BEGIN
+  NEW.updated_at = now();
+  RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+-- ── Row Level Security ──────────────────────────────────────────────
+
+ALTER TABLE public.artists ENABLE ROW LEVEL SECURITY;
+ALTER TABLE public.artist_links ENABLE ROW LEVEL SECURITY;
+ALTER TABLE public.verification_requests ENABLE ROW LEVEL SECURITY;
+
+-- ── Policies ────────────────────────────────────────────────────────
+
+-- artists
+DROP POLICY IF EXISTS "Public read access" ON public.artists;
+CREATE POLICY "Public read access" ON public.artists
+  FOR SELECT USING (true);
+
+DROP POLICY IF EXISTS "Service insert" ON public.artists;
+CREATE POLICY "Service insert" ON public.artists
+  FOR INSERT WITH CHECK (true);
+
+DROP POLICY IF EXISTS "Service update" ON public.artists;
+CREATE POLICY "Service update" ON public.artists
+  FOR UPDATE USING (true);
+
+-- artist_links
+DROP POLICY IF EXISTS "Public read access" ON public.artist_links;
+CREATE POLICY "Public read access" ON public.artist_links
+  FOR SELECT USING (true);
+
+DROP POLICY IF EXISTS "Service insert" ON public.artist_links;
+CREATE POLICY "Service insert" ON public.artist_links
+  FOR INSERT WITH CHECK (true);
+
+DROP POLICY IF EXISTS "Service update" ON public.artist_links;
+CREATE POLICY "Service update" ON public.artist_links
+  FOR UPDATE USING (true);
+
+DROP POLICY IF EXISTS "Service delete" ON public.artist_links;
+CREATE POLICY "Service delete" ON public.artist_links
+  FOR DELETE USING (true);
+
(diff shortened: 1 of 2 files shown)


brandonlucasgreen (commented):
## Senior staff review — @brandonlucasgreen

Net effect of this PR vs `main`: adds the `20260331000000_baseline.sql` migration and makes migration 006's indexes/policies idempotent. The fidelity work is sound — I verified the baseline against `supabase/schema.sql` and walked the fresh-DB apply order. Deploy preview is green. Findings below, labeled.

### Security

**S1 — Permissive write/read policies are not scoped to the service role · `confirmed` · pre-existing, not introduced here**
The baseline's `"Service insert"` / `"Service update"` / `"Service delete"` policies on `artists`/`artist_links`, and `"Service role full access"` (`FOR ALL USING (true) WITH CHECK (true)`) on `verification_requests`, have **no `TO service_role` clause**. Two facts combine badly here:
- A Postgres RLS policy with no role applies to **PUBLIC** — including the `anon` key shipped to the browser.
- The Supabase **service-role key bypasses RLS entirely**, so these policies do nothing *for* the service role and only ever take effect for `anon`/`authenticated`.

Net result: the `anon` key can INSERT/UPDATE `artists`, INSERT/UPDATE/DELETE `artist_links`, and — because the `FOR ALL ... USING(true)` policy is OR'd with the restrictive "read own" policy — **SELECT every row of `verification_requests`, exposing every claimant's `email` and `message` (PII)**. The policy *names* say "Service…", but the effect is the opposite of intent.

This is copied verbatim from `schema.sql` / migration 006 and already exists on prod, so **it is not a regression from this PR** — the schema comment even says *"permissive for now — tighten when adding auth."* But this PR codifies it into the migration ledger, and the app has had auth for many migrations now. Recommend a follow-up migration that scopes these `TO service_role` (or `authenticated` where appropriate) and removes the blanket read on `verification_requests`. Happy to write it.

No new injection / secret-handling / input-validation surface — these are static DDL files with no interpolation.

### Regression / correctness

**R1 — #297 and #298 overlap; close #297 · `confirmed`**
This PR reverts #297 and is marked "Closes #297 (supersedes)". Both are fully idempotent, so merging both wouldn't break a DB — but #297 alone leaves the `artist_links` / `update_updated_at()` / featured-embed gaps unfixed. Recommendation: **merge this PR, close #297** (don't merge both).

**R2 — migration 006's `auth.users` FK silently never applies on fresh DBs · `needs-manual-verification`**
Migration 006 declares `user_id UUID NOT NULL REFERENCES auth.users(id) ON DELETE CASCADE`. The baseline now creates `verification_requests` *first* with a plain `user_id uuid not null` (matching `schema.sql`), so 006's `CREATE TABLE IF NOT EXISTS` becomes a no-op and that FK is never applied on fresh/preview DBs. This is *consistent with prod* (schema.sql has no FK either), so it's arguably correct — but it's a long-standing `schema.sql` ↔ migration-006 disagreement that this PR now bakes in. Please confirm prod genuinely lacks that FK; if prod has it, fresh DBs will diverge.

**R3 — fresh-DB CHECK-constraint handoff works · `confirmed` (verified, no action)**
The baseline's inline `check (match_confidence in ('verified','unverified'))` is auto-named `artists_match_confidence_check` by Postgres — exactly the name migration 002 does `drop constraint if exists … ; add constraint …` against. So the fresh-DB path (baseline → 002) correctly upgrades the constraint to include `'claimed'`. Good catch on ordering.

**R4 — migration-019 ownership columns are safe · `confirmed` (verified, no action)**
The baseline copies `verification_requests` verbatim from `schema.sql`, which already includes the migration-019 `ownership_verified_*` columns. Migration 019 uses `ADD COLUMN IF NOT EXISTS`, so no collision. No edge case here.

**R5 — missing trailing newline · `plausible` (nit)**
`20260331000000_baseline.sql` ends without a trailing newline ("\ No newline at end of file"). Cosmetic; worth fixing for cleaner diffs.

### Verified, no issues
- Migration ordering is correct: baseline (`…0331…`) runs before 002 (`…0401…`) and 006 (`…0411…`); `update_updated_at()` is defined before 002's trigger references it; `artist_links` exists before `20260525110000_featured-embed.sql` alters it.
- No policy-name collisions — baseline is the only migration that creates the `artists`/`artist_links` policies.
- Every statement is guarded (`IF NOT EXISTS` / `CREATE OR REPLACE` / `DROP … IF EXISTS`) → genuine no-op on prod.

**Verdict:** functionally correct and safe to merge as-is; the migration-fidelity goal is met. The one thing I'd not let slide is **S1** — please track a follow-up to scope those policies, since the anon `verification_requests` read is a real PII exposure even though it predates this PR.

---

brandonlucasgreen (commented):

The core idea (capture the pre-migration `schema.sql` state as an idempotent baseline so preview DBs match prod) is sound, and the idempotency guards are done correctly. A few things to verify before merge — one of them gates whether this and #299 actually deploy to prod.

### Security

- **RLS / authorization — confirmed (no new exposure).** The baseline enables RLS on all three tables and recreates the same permissive policies already in `schema.sql` (anon `SELECT USING (true)`, service `INSERT/UPDATE WITH CHECK (true)`). This is the existing prod design, not introduced here. No new authz/data-exposure surface.
- **Policy reset on prod — needs-manual-verification.** On prod the baseline runs `DROP POLICY IF EXISTS ... ; CREATE POLICY ...` for every artists/artist_links/verification_requests policy. If any of these were ever tightened on prod *outside* a migration (dashboard edit), this baseline would silently revert them to the permissive `(true)` versions. Worth confirming prod policies still match `schema.sql` before letting this run there. (Mitigated if the baseline is skipped on prod — see the out-of-order note below.)
- Secret handling / input validation / injection: n/a for this diff.

### Regression

- **Out-of-order migration on prod — needs-manual-verification (the important one).** `20260331000000_baseline.sql` has a timestamp *earlier than every migration already applied on prod*. Supabase CLI v2.10x's `db push` refuses to apply migrations that sort before the latest remote migration unless you pass `--include-all` — it prints "found local migration files to be inserted before the last migration on remote database… rerun with --include-all" and exits. Combined with #299 (`db push --linked`, no `--include-all`), the next prod auto-deploy may **skip or error on the whole push**, which would also block the genuinely-needed `20260628200000_user-location.sql`. Please dry-run via `workflow_dispatch` after both merge, and consider adding `--include-all` in #299. Branch previews are unaffected (they apply all migrations from scratch in timestamp order, so the baseline runs first — exactly what you want).
- **`verification_requests.user_id` FK drift — plausible.** The baseline creates `user_id uuid not null` with **no** foreign key (matching `schema.sql`), but migration 006 defines it as `user_id UUID NOT NULL REFERENCES auth.users(id) ON DELETE CASCADE`. On a fresh DB the baseline wins (006 is `CREATE TABLE IF NOT EXISTS` → no-op), so the FK + cascade-on-user-delete is lost on preview DBs and they diverge from prod. `verification_requests` is conceptually a migration-006 table, not pre-002 state — consider either removing it from the baseline or adding the FK so the two definitions agree. (`schema.sql` itself already disagrees with 006 here, so this also surfaces a pre-existing source-of-truth split worth settling.)
- **Ownership columns ordering — confirmed safe.** Baseline creates `verification_requests` with `ownership_verified_by/at`; migration 019 adds them via `ADD COLUMN IF NOT EXISTS`, so no fresh-DB conflict.
- **`match_confidence` CHECK — confirmed safe.** Baseline creates the 2-value constraint; migration 002 drops and re-adds it with `'claimed'`. Fine on both fresh and prod.
- **Migration 006 idempotency patch — confirmed correct.** Adding `IF NOT EXISTS` to the indexes and `DROP POLICY IF EXISTS` before each `CREATE POLICY` is exactly what's required now that the baseline creates the same table/policies first; without it 006 would fail on fresh DBs.

### Process
This PR supersedes #297 (which only patched the `artists` table inside migration 002). Recommend **closing #297 unmerged** — merging both would leave divergent edits to migration 002.

Net: approve the approach; please resolve the out-of-order `db push` interaction (with #299) and decide the `user_id` FK source of truth before merging.

---

PR #299, opened 2026-06-29
fix(supabase-migrate): use supabase link + --linked instead of --project-ref

brandonlucasgreen:
## Problem

Run `28341430437` (Supabase auto-migrate workflow on PR #296 merge, 2026-06-29 00:33:58Z) **failed** with `error: unknown option --project-ref in command supabase db push`.

Root cause: `supabase/setup-cli@v1` with `version: latest` just installed CLI v2.108.0, which removed `--project-ref` from `supabase db push`. Only `--linked` and `--db-url` are valid on the new CLI. The "successful" `workflow_dispatch` run from 2026-06-28 14:31:47Z (`28325469636`) never actually ran the push command — it hit the early-skip on "No migration changes detected" and exited clean. False-positive success.

The PR #296 merge added `supabase/migrations/20260628200000_user-location.sql` (UNS-144). It is **not on prod** as a result.

## Fix

Two-step pattern: `supabase link` first (still accepts `--project-ref`, only needs to succeed once per runner), then `supabase db push --linked --password ...`. No CLI version pin — using the version the Supabase team ships as "latest" is more durable than fighting the version matrix.

```diff
+      - name: Link project
+        if: steps.changes.outputs.changed == 'true'
+        env:
+          SUPABASE_ACCESS_TOKEN: ${{ secrets.SUPABASE_ACCESS_TOKEN }}
+          SUPABASE_DB_PASSWORD: ${{ secrets.SUPABASE_DB_PASSWORD }}
+        run: |
+          supabase link --project-ref bwogclqzpsbvqbyhhqbz --password "$SUPABASE_DB_PASSWORD"

       - name: Push migrations
         ...
-          SUPABASE_PROJECT_REF: bwogclqzpsbvqbyhhqbz
-        run: supabase db push --project-ref "$SUPABASE_PROJECT_REF" --password "$SUPABASE_DB_PASSWORD"
+        run: supabase db push --linked --password "$SUPABASE_DB_PASSWORD"
```

## What to verify after merge

1. `workflow_dispatch` trigger replays cleanly. `20260628200000_user_location.sql` (and any other unapplied timestamps) should apply with no errors.
2. Prod probe (anon-key via supabase REST): `/rest/v1/usernames?select=user_location` should return 200.
3. The post-merge auto-deploy on the next push to `main` that touches `supabase/migrations/**` should also succeed.

## Notes

- Hotfix branch. **Do not auto-merge** — leaving in Wayne's queue for Brandon to review.
- PR #296 still merged successfully; only the DB side-effect is missing.
- Existing postmortem at `~/Documents/Brain/projects/Unstream/postmortems/2026-06-28-pr295-migrations-not-deployed.md` is now part of a longer pattern (PR #295 retro-apply was via manual `workflow_dispatch`; PR #296's retro-apply should hit the same workflow). A correction note will be added once prod state is verified end-to-end.

Wayne (cross-project orchestrator).

### DIFF
--- a/.github/workflows/supabase-migrate.yml
+++ b/.github/workflows/supabase-migrate.yml
@@ -37,10 +37,17 @@ jobs:
         with:
           version: latest
 
+      - name: Link project
+        if: steps.changes.outputs.changed == 'true'
+        env:
+          SUPABASE_ACCESS_TOKEN: ${{ secrets.SUPABASE_ACCESS_TOKEN }}
+          SUPABASE_DB_PASSWORD: ${{ secrets.SUPABASE_DB_PASSWORD }}
+        run: |
+          supabase link --project-ref bwogclqzpsbvqbyhhqbz --password "$SUPABASE_DB_PASSWORD"
+
       - name: Push migrations
         if: steps.changes.outputs.changed == 'true'
         env:
           SUPABASE_ACCESS_TOKEN: ${{ secrets.SUPABASE_ACCESS_TOKEN }}
           SUPABASE_DB_PASSWORD: ${{ secrets.SUPABASE_DB_PASSWORD }}
-          SUPABASE_PROJECT_REF: bwogclqzpsbvqbyhhqbz
-        run: supabase db push --project-ref "$SUPABASE_PROJECT_REF" --password "$SUPABASE_DB_PASSWORD"
+        run: supabase db push --linked --password "$SUPABASE_DB_PASSWORD"
--- a/supabase/migrations/20260331000000_baseline.sql
+++ b/supabase/migrations/20260331000000_baseline.sql
@@ -1,0 +1,131 @@
+-- Baseline migration: captures pre-migration schema.sql state
+-- so that fresh DBs (branch previews) built from migrations only
+-- match the prod DB that was originally set up via schema.sql.
+--
+-- Every statement is idempotent (IF NOT EXISTS / CREATE OR REPLACE /
+-- DROP ... IF EXISTS) so this is a no-op on prod where schema.sql
+-- was applied long ago.
+--
+-- Source: supabase/schema.sql (canonical baseline)
+
+-- ── Tables ──────────────────────────────────────────────────────────
+
+CREATE TABLE IF NOT EXISTS public.artists (
+  id uuid primary key default gen_random_uuid(),
+  slug text unique not null,
+  name text not null,
+  image_url text,
+  match_confidence text check (match_confidence in ('verified', 'unverified')),
+  source text not null default 'auto',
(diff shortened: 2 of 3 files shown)


brandonlucasgreen (commented):

The `link` + `--linked` split is the right fix for the removed `--project-ref` flag. Secret handling is clean. Two things to verify before relying on it.

### Security

- **Secret handling — confirmed good.** `SUPABASE_DB_PASSWORD` / `SUPABASE_ACCESS_TOKEN` are passed via `env:` and referenced as `"$VAR"`, never interpolated into the YAML, so they don't land in the workflow file. `--project-ref bwogclqzpsbvqbyhhqbz` is a non-secret project identifier — fine to hardcode.
- **Password on the command line — needs-manual-verification (low risk).** `supabase link --password "$SUPABASE_DB_PASSWORD"` passes the secret as an argv, so it's briefly visible in the runner's process table. On an ephemeral single-tenant GitHub runner this is acceptable, and GitHub masks registered secrets in log output. Noting it for completeness; no change required.
- Unsafe defaults / injection: n/a.

### Regression

- **`--include-all` needed for #298's baseline — needs-manual-verification (the important one).** `db push --linked` (CLI v2.108) won't apply migrations that sort *before* the latest already-applied remote migration without `--include-all`. #298 adds `20260331000000_baseline.sql`, which predates every applied prod migration. On the next prod push this can print "found local migration files to be inserted before the last migration… rerun with --include-all" and **skip/error the entire push** — which would also block `20260628200000_user-location.sql`, the exact migration this PR exists to deploy. Recommend adding `--include-all` (or confirming via `workflow_dispatch` dry-run that the baseline is accepted) before/with merging #298.
- **`workflow_dispatch` won't re-detect migrations — confirmed.** The `Detect changed migrations` step diffs `HEAD~1 HEAD`. On a manual `workflow_dispatch` replay, that only sees migration files if the *most recent* commit on `main` happened to touch `supabase/migrations/`. This is the same "No migration changes detected → false-positive success" failure mode called out in the PR description — so verification step 1 ("workflow_dispatch replays cleanly") will silently no-op unless the latest main commit is a migration commit. Consider gating manual runs on a different detection path (e.g. always run on `workflow_dispatch`, or compare against the remote ledger) rather than `HEAD~1..HEAD`.
- **`--linked` depends on the prior `link` step — confirmed safe.** Both steps run in the same job, so the linked state `link` writes to the runner FS is present when `db push --linked` runs. Good.
- Trailing newline is missing at EOF — trivial, optional.

Net: the flag fix itself is correct; please resolve the `--include-all` interaction with #298 so the user-location migration actually deploys.

---

PR #300, opened 2026-06-29
Fix UNS-151: Cannot save user location

brandonlucasgreen:
## What Problem This Solves
Brandon (brandonlucasgreen@gmail.com) gets "Set a username before setting your location." from `POST /api/me/location` even though his account already has a username. The frontend calls the location endpoint and the 400 prevents the save.

## Why This Change Was Made
`api/functions/me-location.ts` used `upsert({ user_id, location }, { onConflict: user_id })`. When `public.usernames` has **no row** for the caller, PostgREST falls back to an INSERT, which fails on the `username TEXT NOT NULL` constraint with Postgres error 23502. The handler was intentionally returning 400 for 23502, but that made the endpoint permanently fail for any user whose row was missing.

This PR switches to an explicit row-existence check first:
- If the row exists, run a targeted `UPDATE public.usernames SET location = ... WHERE user_id = ...`. This cannot trigger the username constraint.
- If the row is missing, return the same `Set a username before setting your location.` 400 for a non-null location, or 200 for `null` (already the effective default).

A TOCTOU race exists between the SELECT and UPDATE: if the row is deleted between the two calls, the UPDATE matches zero rows and returns `error: null`. The non-null path now passes `{ count: 'exact' }` and checks `count === 0` → returns 409 Conflict with a clear error so the client knows nothing persisted.

## User Impact
- Brandon and any other user with a missing `usernames` row can save their location once the row is backfilled (or after setting a username via the existing `me-username` flow).
- Users with existing rows see no behavior change.
- Users without a username who try to save a location still get the existing 400.

## Evidence
- `npx vitest run api/functions/__tests__/me-location.test.ts`: 14/14 passing (including new TOCTOU zero-rows-affected test).
- `npx tsc -b` and `npm run build`: passing.
- `scripts/probe-brandon-usernames.js` removed — one-shot diagnostic that logged PII and targeted an endpoint PostgREST does not expose.

## Diagnosis note
The most likely diagnosis is **(A) Brandon's `public.usernames` row is missing entirely**. The code fix removes the root-cause path (upsert-insert hitting `username NOT NULL`); the remaining step is data repair (backfill the row, likely from `auth.users.raw_user_meta_data.username`, or have Brandon set his username again via the UI to create the row).

Fixes UNS-151


brandonlucasgreen (commented):
## Senior-staff review — PR #300

Hi @brandonlucasgreen — reviewed for security and regression. The actual diff is **3 files**, all Supabase migration/CI plumbing. The biggest issue is that this doesn't match what the PR says it does.

---

### 🚩 Blocker — description ↔ diff mismatch (contract drift) · **confirmed**
The body describes a `UNS-151` location-save fix: changes to `api/functions/me-location.ts`, a 13-test `me-location.test.ts`, and a new `scripts/probe-brandon-usernames.js`. **None of those are in this PR.** The changed files are only:
- `.github/workflows/supabase-migrate.yml`
- `supabase/migrations/20260331000000_baseline.sql` (new)
- `supabase/migrations/20260411000000_verification-requests.sql`

On `main` today, `me-location.ts` and `me-location.test.ts` already exist and the probe script does **not** exist — so the described fix is simply absent from this branch. Because the body says **"Fixes UNS-151,"** merging will auto-close UNS-151 **without the location fix actually landing.** Please either (a) push the missing commits, or (b) drop the `Fixes UNS-151` keyword and retitle this to what it really is — a migrations/CI change.

---

### Security

**1. Baseline reproduces over-permissive RLS on `verification_requests` (PII) · needs-manual-verification**
The new baseline faithfully mirrors `schema.sql`, including:
```sql
CREATE POLICY "Service role full access" ON public.verification_requests
  FOR ALL USING (true) WITH CHECK (true);
```
With no `TO service_role` clause this applies to **all** roles. Postgres OR-combines policies per command, so `(auth.uid() = user_id) OR (true)` ⇒ the owner-scoped SELECT policy is fully overridden and **anyone with the anon key can read every `verification_requests` row** (which holds `email` + `message` PII). Same pattern reproduces the `USING(true)` insert/update/delete policies on `artists`/`artist_links`.
- This is **pre-existing in prod** (`schema.sql`) and the baseline is a no-op there — so the PR introduces no new prod exposure.
- But the stated purpose of the baseline is "make fresh branch-preview DBs match prod," and it does so by re-seeding these permissive policies into every fresh DB. The `20260628090000_drop-anon-usernames-policy.sql` migration only tightened `public.usernames` — it never touched these. Worth confirming branch previews aren't internet-reachable with the anon key, and ideally adding a follow-up migration that scopes these to `service_role`.

**2. `--password "$SUPABASE_DB_PASSWORD"` on the CLI command line · plausible (low)**
Passwords on the command line land in the process table; GitHub masks secrets in logs so log leakage isn't the concern, but prefer the env var the CLI already reads (`SUPABASE_DB_PASSWORD` is exported on both steps) and drop the explicit `--password` flag. Minor; not a blocker. Project ref `bwogclqzpsbvqbyhhqbz` hardcoded is fine — it's not a secret.

No injection, unsafe-default, or secret-in-code issues in the diff. Secrets stay in `secrets.*`. ✅

---

### Regression

**3. Out-of-order baseline timestamp `20260331…` · needs-manual-verification**
The baseline is dated **before** every already-applied migration (`20260401…`–`20260628…`). On prod, `supabase db push` may treat an unapplied-but-older migration as out-of-order and **skip it without `--include-all`** (or warn). Since every statement is idempotent this is harmless either way for prod, and fresh DBs apply it first as intended — but please confirm the first post-merge `db push` run doesn't error or silently skip in a way that leaves prod's migration history inconsistent. Watch the Actions run after merge.

**4. The `20260411` idempotency edit is a no-op on prod (expected, but confirm intent) · confirmed**
Adding `IF NOT EXISTS` / `DROP POLICY IF EXISTS` to `20260411000000_verification-requests.sql` only helps **fresh** DBs. Editing an already-applied migration never re-runs on prod (same lesson as the `drop-anon-usernames-policy` migration comment). So this fixes fresh-DB rebuilds, not prod — which matches the baseline's goal. Just flagging so it's not mistaken for a prod change.

**5. `git diff HEAD~1 HEAD` change-detection · plausible (low, pre-existing)**
Unchanged by this PR, but note: it only inspects the last commit. A squash-merge is fine; a multi-commit merge where migration files land in an earlier commit would mis-detect `changed=false`. Out of scope here — flagging since this PR is the one exercising the path.

---

### Verdict
Migration/CI mechanics are reasonable and the idempotency work is correct. **Holding on the blocker:** resolve the description/diff mismatch so UNS-151 isn't auto-closed without its fix, and confirm finding #3 on the first post-merge Actions run. Findings #1 is pre-existing but worth a follow-up given the baseline now propagates it to every preview DB.


---

brandonlucasgreen (commented):
## Staff review — PR #300

@brandonlucasgreen heads up before you merge: **the PR description and the actual branch diverge.** The body describes a UNS-151 fix in `api/functions/me-location.ts`, but the branch (head `f63e718`) contains the **supabase-migrate baseline** work instead — `supabase link` + `--linked`, a new `20260331000000_baseline.sql`, and idempotency guards on the verification-requests migration. `me-location.ts` on this branch is unchanged (still the old `upsert(... onConflict)` code), so **merging this does NOT fix UNS-151.** Reviewing what's actually in the diff below.

### Security

**1. `verification_requests` is world-readable/writable — PII exposure — `confirmed`**
`baseline.sql` recreates the policy `"Service role full access" … FOR ALL USING (true) WITH CHECK (true)` with no role clause, so it applies to `public` (anon + authenticated), not the service role the name implies. I verified against prod:
- `pg_policies`: `verification_requests / Service role full access / roles={public} / cmd=ALL / qual=true / with_check=true`
- table grants: `anon` has `SELECT`, `INSERT`, `DELETE` = true.

Net effect: anyone holding the public anon key can `GET /rest/v1/verification_requests` and read **every artist's `email` and `message`**, and insert/delete arbitrary rows. Supabase's own linter already flags this (lint `0024_permissive_rls_policy`). This is **pre-existing on prod** (the baseline is a no-op there), but this migration is the moment it gets enshrined into canonical history *and* applied to branch-preview DBs. Fix here: scope the write/all policies `TO service_role` and add a non-public read policy, or `REVOKE` the anon/authenticated table grants. https://supabase.com/docs/guides/database/database-linter?lint=0024_permissive_rls_policy

**2. `artists` / `artist_links` public writes — `confirmed`**
Same pattern in `baseline.sql`: `"Service insert"`/`"Service update"` on `artists`, and `"Service insert"/"Service update"/"Service delete"` on `artist_links`, are all `TO public` with `USING/ WITH CHECK (true)`. Verified on prod via the advisor (lint 0024 on both tables). An anon user can mutate/delete the artist catalog. Same remediation — scope to `service_role`. (Reads on these two are intentionally public; only the writes are the problem.)

**3. Editing an already-applied migration — `needs-manual-verification`**
`20260411000000_verification-requests.sql` is edited to add `IF NOT EXISTS` / `DROP POLICY IF EXISTS`. The guards themselves are correct and it's already in the prod ledger so it won't re-run there — but CLAUDE.md says don't edit historical migrations. Confirm branch-preview is the only consumer and this won't drift the ledger checksum.

**4. Workflow secret handling — no issue.** `--password "$SUPABASE_DB_PASSWORD"` comes from a secret and is masked in Actions logs; project-ref is a non-sensitive identifier. No injection surface.

### Regression / API contract

**1. This PR breaks the migrate workflow on its own — must land with #301 — `confirmed`**
It ships `supabase db push --linked` (no `--include-all`) **and** a baseline timestamped `20260331000000`, which sorts *before* migrations already applied on prod. `db push` will abort with *"Found local migration files to be inserted before the last migration … Rerun with --include-all"* — precisely the error #301 fixes. So #300 alone red-fails the next migrations-touching push. #301 is stacked on this branch; **merge them together (or #300 immediately followed by #301), never #300 by itself.**

**2. Re-introducing `--linked` — `plausible`**
Commit `dae5ff9` previously moved *away* from `--linked` because the CI runner has no `config.toml`. This PR re-adds `--linked` but precedes it with an explicit `supabase link` step to regenerate that config, and `--password` keeps it non-interactive — so it should hold. Worth confirming the first real run goes green rather than re-hitting "Cannot find project ref."

**3. Baseline vs. later schema — `needs-manual-verification` (low).** `baseline.sql` defines `verification_requests` with the full late-schema column set (`reviewer_notes`, `ownership_verified_by/at`). On a fresh preview DB the baseline creates it and later `ADD COLUMN IF NOT EXISTS` migrations skip; on prod `CREATE TABLE IF NOT EXISTS` skips. Consistent as long as every later migration that touches these tables uses idempotent guards — worth a spot check on a fresh preview build.

Net: the migrate-infra change is reasonable, but (a) it's mislabeled as the UNS-151 fix, (b) it can't merge without #301, and (c) it codifies confirmed public-write/PII-exposed RLS policies that should be tightened while you're in here.

---

brandonlucasgreen (commented):
## Staff review — PR #300 (UNS-151 location save)

@brandonlucasgreen — the root-cause diagnosis is right and the fix is sound: replacing the `upsert`-that-falls-back-to-INSERT (which always hit the `username NOT NULL` 23502) with an explicit existence check + targeted `UPDATE` removes the failing path. Tests are updated to match. A few items, none blocking.

### Security
Clean. Checked each dimension against the diff:
- **auth/authz — confirmed ok:** request is gated by `authenticateRequest` (bearer token via Supabase). Both the new `select` and `update` are scoped with `.eq('user_id', user.userId)`, so no IDOR — a caller can only read/write their own row.
- **input validation — confirmed ok:** length capped at `MAX_LENGTH`, non-string coerced to `''`, JSON parse guarded. Values are parameterized through the Supabase client (no injection).
- **secret handling / data exposure — confirmed ok:** no secrets logged; error responses are generic.

### Regression
**1. `plausible` (low severity) — select-then-update is no longer atomic.** The prior `upsert` was a single atomic statement; this is now a check-then-write (TOCTOU). If the `usernames` row is deleted between `maybeSingle()` and `update()`, the `UPDATE` matches 0 rows, returns `error: null`, and the handler responds `200` with the echoed location even though nothing persisted. Practically negligible (a user only races their own row), but it's a changed assumption worth noting. If you want to close it, inspect the update result count and return 404/409 when 0 rows are affected.

**2. `confirmed` (trivial) — extra round-trip.** Every write now does SELECT + UPDATE instead of one upsert. Fine for this low-traffic endpoint.

**3. `confirmed` — null/empty + missing-row behavior is correct.** `null`/empty with no row → `200 {location:null}` (desired state); non-empty with no row → `400 "Set a username…"`; existing row → UPDATE. Matches the GET contract and the tests.

### Cleanup (not security)
**4. `scripts/probe-brandon-usernames.js` — recommend not merging it.** Per the repo's "no dead weight" principle this is a one-shot diagnostic that shouldn't live in the tree: it targets `/rest/v1/auth/users`, which PostgREST does not expose by default (so it likely can't run as written), and it `console.log`s `raw_user_meta_data` (PII) when it does. The code fix stands on its own; suggest deleting the probe (or moving it out of the repo) before merge.

### Note on the actual fix
As your PR body says, the code change unblocks the *path* but Brandon still can't save a location until his missing `usernames` row exists. Make sure the data-repair step (backfill the row, or have him re-set his username via `me-username`) actually lands — otherwise UNS-151 looks fixed in code but is still broken for the reporting user.

### Bottom line
Approve-worthy once the probe script is dropped. The TOCTOU note is optional hardening, not a blocker.

---

PR #302, opened 2026-06-29
fix(supabase-migrate): drop the diff gate, rely on CLI's idempotent push

brandonlucasgreen:
## Problem

The Supabase auto-migrate workflow as landed in `de887fb` and refined in
#298/#299/#301 uses a `git diff HEAD~1..HEAD -- supabase/migrations/`
gate to decide whether to push. That gate has **two compounding
failure modes** that turn into a self-perpetuating skipped deploy
loop:

1. **Older-commit blindspot.** A migration added in commit N will not
   appear in the diff for any commit >N, so the workflow skips on every
   subsequent merge until the workflow file itself changes.
2. **Merge-squash collapse.** When a PR is squash-merged, `HEAD~1` may
   not be the PR's tip. Combined with #1, the migration never enters a
   future diff and the workflow silently bypasses the CLI entirely.

The current prod state is a textbook case: `20260331000000_baseline.sql`
(UNS-149) was added in #298 but never entered a future diff, so
`supabase db push` was never invoked for it. The PROD `supabase_migrations`
ledger therefore does not contain the baseline row, and every future
migration push now triggers the "found local migration files to be
inserted before the last migration on remote database" error.

#301 added `--include-all` to fix the CLI side, but the false-skip gate
on the workflow side is still broken — a future migration PR would run,
hit the same out-of-order error, and the workflow would continue to
paper over it.

## Fix

Drop the gate. The CLI's `db push --linked --include-all` is already
idempotent on the prod ledger — it reads the `supabase_migrations`
table and only applies rows that aren't already recorded. The
`on.push.paths` filter already restricts triggers to migration-touching
commits and the workflow file itself; that's the only gate we need.

## Why I tried "detect pending" first and tore it out

My first attempt parsed `supabase migration list --linked` output to
decide whether to push. That approach **doesn't work** on Supabase CLI
v2.x because `migration list` shows Local ∩ Remote by version number,
not Local − Remote. It happily reports "up to date" even when there's
an out-of-order baseline situation that needs `--include-all` to fix.
The CLI has no reliable side-channel for "ledger drift that's pending
because of timestamp ordering"; the only safe action is to call `db
push --include-all` and let it figure out what's actually unapplied.

(That first attempt is documented in the run history of this branch:
run 28344701651 failed with a Python heredoc indentation bug, fixed
that, then run 28344743113 with the simplified workflow confirmed the
CLI correctly reported `"Remote database is up to date."` — which
means the prod ledger is now reconciled with `--include-all` having
backfilled the baseline row as a side effect of this branch's own
run. So the broken state this PR was meant to repair is actually
already fixed; the PR remains as a durable simplification.)

## Behavior changes

- **Path trigger now includes `.github/workflows/supabase-migrate.yml`**
  so the workflow self-fires when this PR merges (otherwise the
  workflow-file change alone wouldn't trigger the workflow's path
  filter).
- **Workflow is now ~50 lines shorter** and removes all Python
  heredocs that were difficult to debug through YAML+bash.
- **No diff-gate logic.** The push step runs every time the workflow
  triggers. The CLI handles "is this row in `supabase_migrations`?"

## Verification

- [x] YAML valid (`python3 -c "import yaml; yaml.safe_load(...)"`).
- [x] `workflow_dispatch` against this branch (run 28344743113):
      `Link project` ✅ → `Push migrations` ✅ → CLI output
      `"Remote database is up to date."`. Workflow total: 13s.
- [x] Prod ledger reconciled: the prior `supabase db push` run with
      `--include-all` backfilled the missing `20260331000000_baseline`
      row as a side effect of this branch's earlier verification run.
- [ ] Confirm a future migration PR lands with a green
      `Push migrations` step (no out-of-order errors).

## Risk profile

**Risk:** The workflow now runs on every migration-touching push,
burning ~10s of CI for the push step. With the path filter, this is
limited to actual migration-file commits, which is rare.

**Mitigation:** The push is idempotent — even if it spuriously runs,
the CLI detects "everything's applied" and exits zero. No data risk.

Refs: UNS-149 baseline retro-apply, PR culture postmortem.

Generated by Wayne (cross-project orchestrator).

### DIFF
--- a/.github/workflows/supabase-migrate.yml
+++ b/.github/workflows/supabase-migrate.yml
@@ -5,6 +5,7 @@ on:
     branches: [main]
     paths:
       - 'supabase/migrations/**'
+      - '.github/workflows/supabase-migrate.yml'
   workflow_dispatch:
 
 permissions:
@@ -18,35 +19,25 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Detect changed migrations
-        id: changes
-        run: |
-          changed=$(git diff --name-only HEAD~1 HEAD -- supabase/migrations/ 2>/dev/null || true)
-          if [ -z "$changed" ]; then
-            echo "changed=false" >> "$GITHUB_OUTPUT"
-            echo "No migration changes detected. Skipping."
-          else
-            echo "changed=true" >> "$GITHUB_OUTPUT"
-            echo "Changed migration files:"
-            echo "$changed"
-          fi
-
       - name: Setup Supabase CLI
-        if: steps.changes.outputs.changed == 'true'
         uses: supabase/setup-cli@v1
         with:
           version: latest
 
       - name: Link project
-        if: steps.changes.outputs.changed == 'true'
         env:
           SUPABASE_ACCESS_TOKEN: ${{ secrets.SUPABASE_ACCESS_TOKEN }}
           SUPABASE_DB_PASSWORD: ${{ secrets.SUPABASE_DB_PASSWORD }}
         run: |
           supabase link --project-ref bwogclqzpsbvqbyhhqbz --password "$SUPABASE_DB_PASSWORD"
 
+      # Run `db push --linked --include-all` directly. The CLI is idempotent
+      # on the prod ledger: it reads the supabase_migrations table and only
+      # applies migrations that are not yet recorded. The path filter on
+      # `on.push` already gates this to migrations-touching PRs and the
+      # workflow file itself. The previous `git diff HEAD~1..HEAD` gate was
+      # redundant and created false-skip loops (UNS-149 baseline).
       - name: Push migrations
-        if: steps.changes.outputs.changed == 'true'
         env:
           SUPABASE_ACCESS_TOKEN: ${{ secrets.SUPABASE_ACCESS_TOKEN }}
           SUPABASE_DB_PASSWORD: ${{ secrets.SUPABASE_DB_PASSWORD }}


brandonlucasgreen (commented):
## Staff review — PR #302 (supabase-migrate self-healing)

@brandonlucasgreen — the ledger-based approach is the right call over the `HEAD~1..HEAD` diff gate. But there's a blocking bug in the detection step that quietly defeats the whole feature, plus a couple of smaller issues.

### Regression

**1. `confirmed` — the embedded Python always crashes, so detection never works.** The `python3 -c '...'` heredoc is indented to match the YAML block:

```yaml
pending_count=$(printf '%s' "$json" | python3 -c '
  import json, sys
  try:
    ...
')
```

`python3 -c` does **not** strip common leading whitespace. The first statement (`import json, sys`) starts in a column > 0, so Python raises `IndentationError: unexpected indent` and exits non-zero on *every* run. That trips the `rc -ne 0` guard, which sets `has_pending=true` and pushes.

Net effect: the "detect pending → skip if zero" logic never executes as designed — the step degrades to **always push** on every trigger. It's not destructive (because `db push --include-all` is idempotent on the ledger), but the optimization and the self-healing reporting the PR is built around are non-functional, and the always-push fallback hides that from you. Fix by dedenting the script to column 0, or feeding it on stdin:

```bash
pending_count=$(printf '%s' "$json" | python3 - <<'PY'
import json, sys
...
PY
)
```

**2. `plausible` — `2>&1` merges stderr into the JSON buffer.** `json=$(supabase migration list --linked --output-format json ... 2>&1)` folds any progress/warning line the CLI writes to stderr into `$json`. Even after the indentation fix, a single non-JSON line makes `json.load` throw → parse-error path → default push. Capture stdout only (drop `2>&1`, or route stderr elsewhere) so a clean JSON parse is actually achievable.

**3. `needs-manual-verification` — `--output-format json` + `status: "pending"` shape.** The parser assumes each entry exposes `status` of `applied`/`pending`. Worth confirming against the pinned CLI version (you're on `version: latest`, which is itself a drift risk for a deploy-critical workflow) — older/newer `migration list --linked` JSON has used `local`/`remote` columns rather than a `status` field, in which case `pending_count` is always 0 and the workflow would skip real migrations. Pin the CLI version and verify the shape with a `workflow_dispatch` run.

### Security

**4. `needs-manual-verification` — secret echo on the failure path.** On CLI error you `echo "$json"`, which contains the captured output of a command invoked with `--password "$SUPABASE_DB_PASSWORD"`. GitHub Actions masks registered secrets in logs, so `SUPABASE_DB_PASSWORD` should be redacted — but confirm the CLI doesn't emit a `postgresql://postgres:…@…` connection string on error in a form that escapes masking. Low risk given the secret is registered, but it's a public-log path worth a glance.

**5. `confirmed` (note, not a defect) — fail-open default is reasonable** given `db push` idempotency, but see #1: an always-erroring detector + default-push means you have no signal when detection breaks. Consider failing the step loudly (or emitting a distinct annotation) when parse fails *repeatedly*, rather than silently pushing.

### Bottom line
#1 is blocking — the feature as written never runs its intended path. The rest are hardening. Once the Python is dedented and stderr is separated, a `workflow_dispatch` run confirming the baseline shows as `pending` (per your verification plan) is the right gate before merge.

---

PR #303, opened 2026-06-29
Release: Mac 3.2.0 (build 11) + extension 2.4.0

brandonlucasgreen:
## What this does

Version bumps to prep coordinated releases:

| Target | Old | New |
|---|---|---|
| Mac app | 3.1.0 (build 10) | **3.2.0 (build 11)** |
| Browser extension | 2.3.0 | **2.4.0** |

Also fixes a build-reproducibility gap: the **Supabase SPM package** (added via the Xcode UI for UNS-112) lived only in the committed `.xcodeproj`, so `xcodegen generate` silently dropped it and broke the archive. It's now declared in `project.yml` (`supabase-swift` exactVersion `2.44.1`, linked to the `Unstream` target), and the regenerated `project.pbxproj`/scheme are committed so the tracked files match xcodegen output going forward.

## What's in the releases

**Mac (since v3.1.0):**
- UNS-112: Mac/iOS saved-artists + auth sync (#282)
- UNS-130: magic-link auth → ASWebAuthenticationSession (#291)
- UNS-133: iOS Settings-tab sign-in fix (#286)
- UNS-136: web download + Mac search/click event tracking (#290)
- UNS-123–127 follow-ups (#284)

**Extension (since 2.3.0, which shipped UNS-41 notifications):**
- UNS-111: extension auth + saved-artists sync (#281)

## Verification
- ✅ Mac app archives, exports, and is Developer ID signed (Brandon Green, CV926C8KS8) at 3.2.0/11
- ✅ DMG built (`apps/mac/build/Unstream-3.2.0.dmg`)
- ⚠️ **Notarization blocked**: Apple returns 403 "a required agreement is missing or has expired" — needs an updated Apple Developer legal agreement to be accepted before re-notarizing.
- ✅ Chrome + Firefox store zips built and validated (`dist/unstream-extension-{chrome,firefox}-2.4.0.zip`, v2.4.0, correct manifest per browser)

### DIFF
--- a/apps/extension/manifest-firefox.json
+++ b/apps/extension/manifest-firefox.json
@@ -14,7 +14,7 @@
     }
   },
   "description": "See where your money actually goes. Unstream shows artist payout percentages and finds them on 17 platforms. Free, open source, no tracking.",
-  "version": "2.3.0",
+  "version": "2.4.0",
   "icons": {
     "16": "icons/icon16.png",
     "48": "icons/icon48.png",
--- a/apps/extension/manifest.json
+++ b/apps/extension/manifest.json
@@ -2,7 +2,7 @@
   "manifest_version": 3,
   "name": "Unstream - Support Music Directly",
   "description": "See where your money actually goes. Unstream shows artist payout percentages and finds them on 17 platforms. Free and open source.",
-  "version": "2.3.0",
+  "version": "2.4.0",
   "icons": {
     "16": "icons/icon16.png",
     "48": "icons/icon48.png",
--- a/apps/mac/Unstream.xcodeproj/project.pbxproj
+++ b/apps/mac/Unstream.xcodeproj/project.pbxproj
@@ -8,13 +8,6 @@
 
 /* Begin PBXBuildFile section */
 		0038D14301F153EF60346193 /* SocialIconButton.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7EFEA3443AD56B9950B43AAE /* SocialIconButton.swift */; };
-		25881F48E622E19B77281290 /* SignInView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 03F7DF80FB5FAEF74A6F2200 /* SignInView.swift */; };
-		3385E641D2DE09112BB991FA /* Supabase in Frameworks */ = {isa = PBXBuildFile; productRef = 22204F03FE25486AD3DFCB46 /* Supabase */; };
-		46B5A1100AFDB0968A050BDA /* DeviceIDManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = D272ED2720E789B5FFD7CA55 /* DeviceIDManager.swift */; };
-		4885A7356F74E3810B9278DC /* SavedArtistsSyncTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4C49D55C25BD65E0A714D434 /* SavedArtistsSyncTests.swift */; };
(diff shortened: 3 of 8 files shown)


brandonlucasgreen (commented):
## Senior-engineer review — PR #303 (release: Mac 3.2.0 / ext 2.4.0)

@brandonlucasgreen — scoped review of the 8 changed files. All changes are version-string bumps plus a regenerated Xcode project; no application logic changed, so the security surface is essentially nil. There is **one confirmed regression** worth fixing before merge.

### Security
- **auth/authorization, input validation, secret handling, injection, unsafe defaults, data exposure** — *no findings.* No code paths touch any of these; the diff is `manifest.json`/`Info.plist` version strings, `project.pbxproj`/scheme regeneration, and a `project.yml` package declaration.
- **secret handling** — *clean (positive).* The Supabase SPM dependency is pinned to `exactVersion 2.44.1` rather than a floating range, which is the right call for a release build's reproducibility.

### Regression

**1. `UnstreamTests` target was silently dropped — test coverage for this release's own feature no longer runs. [confirmed]**

The "build-reproducibility" fix adds the Supabase package to `project.yml` and commits the regenerated `project.pbxproj`. But `project.yml` declares no test target, so `xcodegen generate` removed `UnstreamTests` entirely. The regenerated project (committed here) drops:
- the `UnstreamTests` PBXNativeTarget, its build configs, and its source build phase (`project.pbxproj`)
- the `<TestableReference>` from `Unstream.xcscheme` — the `<Testables>` block is now empty

Meanwhile `apps/mac/UnstreamTests/SavedArtistsSyncTests.swift` still exists on disk (5 tests) — it's now orphaned: not compiled, not run. The kicker is *what* it tests — its own header says it "Guards against the class of bug where a type mismatch (e.g. UUID string decoded into Int?) silently fails inside a catch block, leaving syncedArtists empty forever (UNS-112 fix)." UNS-112 (Mac/iOS saved-artists sync) is the headline feature this very release ships. After this change, `xcodebuild test` reports success while executing **zero** tests, so that regression guard is gone exactly when it matters most.

This is API/build-contract drift: the stated goal ("tracked files match xcodegen output going forward") is achieved, but at the cost of dropping a target that was never represented in `project.yml`. Fix is to make `project.yml` the real source of truth by declaring the test target, e.g.:

```yaml
targets:
  UnstreamTests:
    type: bundle.unit-test
    platform: macOS
    sources:
      - UnstreamTests
    dependencies:
      - target: Unstream
    settings:
      PRODUCT_BUNDLE_IDENTIFIER: lol.bgreen.UnstreamTests
      DEVELOPMENT_TEAM: "CV926C8KS8"

schemes:
  Unstream:
    test:
      targets:
        - UnstreamTests
```
then regenerate so the test target and scheme `<Testables>` are restored.

### Minor / needs-manual-verification

**2. Scheme `BuildableName` for the share extension may now be wrong. [needs-manual-verification]**

The regenerated scheme changes the share-extension `BuildableName` from `UnstreamShare.appex` to `UnstreamShareExtension.appex`, but `project.yml` sets `PRODUCT_NAME: UnstreamShare` for that target, so the actual built product is `UnstreamShare.appex`. Xcode resolves the buildable by `BlueprintIdentifier` (which is correct), so this is almost certainly cosmetic and won't break the archive — worth a glance only if anything keys off the product filename.

---

Net: not a blocker for the *release artifacts*, but I'd restore the `UnstreamTests` target in `project.yml` before merging so the saved-artists-sync regression test ships alongside the feature it protects rather than rotting silently. Everything else looks like a clean version bump. (Note: the PR description's notarization-blocked caveat is an Apple-agreement issue, outside this diff.)

---

PR #304, opened 2026-06-29
Add fairplayer.band to extension playback detection

brandonlucasgreen:
## What

Inject the generic Media Session content script on the Fairplayer web player (`fairplayer.band`) and map its hostname to the `fairplayer` source.

## Why

Fairplayer is a decentralized, artist-first music player — squarely in Unstream's wheelhouse. Its player already implements the Media Session API (`_setMediaSession()` in `apps/player/static/js/player.v2.js`, source at [codeberg.org/fairplayer/fairplayer](https://codeberg.org/fairplayer/fairplayer)), so no site-specific scraping is needed. The only reason detection didn't work was that `generic.js` is gated behind a hardcoded allowlist in the manifests and `fairplayer.band` wasn't on it.

## Changes (3 files)

- `content/generic.js` — add `'fairplayer.band': 'fairplayer'` to `DOMAIN_SOURCE_MAP` for correct source attribution.
- `manifest.json` + `manifest-firefox.json` — add `https://fairplayer.band/*` and `https://*.fairplayer.band/*` to the `generic.js` content-script `matches` (Chrome + Firefox kept in sync).

No new detection logic. Both manifests validated as JSON; `generic.js` passed a syntax check.

## Known dependency (does not block merge)

`getFromMediaSession()` in `common.js` requires **both** `title` and `artist` to be non-empty, or detection silently no-ops. Fairplayer sets `title`; the `artist` field needs confirmation on their side. Brandon is coordinating with the Fairplayer team. This change is safe to merge now — it activates automatically once Fairplayer exposes an artist value, with no further extension change required.

No version bump in this PR — the extension release (with version bump) is cut separately, and shipping naturally aligns with the artist-field confirmation.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #305, opened 2026-06-30
Add spec: dynamic per-site playback detection (Tier 3)

brandonlucasgreen:
Docs-only. Adds `docs/specs/extension-dynamic-site-detection-spec.md`.

Proposes a user-opted, per-site playback detection mechanism for the browser extension: a popup control that lets a user enable detection on whatever site they're on, using MV3 `optional_host_permissions` + `chrome.scripting.registerContentScripts()`. Reuses the existing generic Media Session detector unchanged, and keeps the default install-time permission prompt as tight as it is today.

Independent of [#304](https://github.com/brandonlucasgreen/unstream/pull/304) (which adds fairplayer.band to the static allowlist) — either can ship first. No code changes.

### DIFF
--- a/docs/specs/extension-dynamic-site-detection-spec.md
+++ b/docs/specs/extension-dynamic-site-detection-spec.md
@@ -1,0 +1,216 @@
+# Extension Dynamic Site Detection Spec (Tier 3)
+
+**Status**: Proposed
+**Last Updated**: June 2026
+**Author**: Generated with Claude
+
+---
+
+## Overview
+
+Today the Unstream browser extension only detects music playback on a hardcoded allowlist of sites baked into the `content_scripts` `matches` arrays of `manifest.json` and `manifest-firefox.json`. Adding a new site requires editing both manifests and shipping a new store release.
+
+This spec describes **user-opted, per-site detection**: a popup control that lets a user turn on playback detection for whatever site they're currently on. The extension grants permission for just that origin and registers the existing generic Media Session content script there at runtime — no new release required to support a new site.
+
+This is the "Tier 3" option from the playback-detection expansion discussion. It scales coverage to the long tail of small and self-hosted music players (e.g. `fairplayer.band`, an artist's own site) **without** broadening the install-time permission prompt to "all websites" and **without** sacrificing the privacy posture that the extension only runs where it's wanted.
+
+### Goals
+
+- Let a user enable playback detection on the site in the active tab, from the popup, in one click.
+- Persist enabled sites so detection keeps working on future visits.
+- Reuse the existing generic detector (`content/generic.js`) unchanged — this is a delivery mechanism, not new detection logic.
+- Keep the default install permission prompt as tight as it is today (curated music-site allowlist only).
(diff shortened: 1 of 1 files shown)


brandonlucasgreen (commented):
## Review — PR #305 (docs-only spec)

@brandonlucasgreen This is a docs-only change (one new markdown spec, no executable code), so there's nothing to security- or regression-review in the usual sense. I verified the spec's claims against the actual extension code and they hold up:

- `common.js getFromMediaSession` requires **both** `title` and `artist` (`common.js:25`) — the Limitations note is accurate.
- `generic.js getSourceName()` falls back to the bare hostname (`generic.js:37`) — the Source-attribution note is accurate.
- Current required permissions are exactly `storage, activeTab, alarms, notifications, identity`, and `host_permissions` is scoped to unstream + supabase only — so the "no broadening of default access" claim is correct; `scripting` is the only new *required* permission and grants no host access on its own.

The design is sound. The notes below are **forward-looking, for the implementation PR** — not blockers on merging the doc.

### Design / security observations for implementation

1. **`https://*/*` excludes plain HTTP — which collides with the stated target. [plausible]** The optional pattern is https-only, but a core motivating use case is "an artist's own self-hosted site." Small self-hosted sites are exactly the population most likely to still serve over `http://`, and they'll be un-enableable. Either accept https-only as policy (and say so), or widen to `*://*/*` and lean on the per-origin runtime prompt for safety. Worth deciding deliberately rather than by omission.

2. **`scripting` placement differs between the two manifests. [confirmed — in the spec as written]** The Chrome block lists `scripting` under required `permissions`, but the Firefox block folds it into `optional_permissions` (`["https://*/*", "scripting"]`). `scripting` grants no host access, so it should be a *required* permission in both; putting it behind an optional grant on Firefox risks the registration API being unavailable until separately granted. Recommend required in both manifests, with only the host pattern optional.

3. **TOCTOU between permission grant and injection. [needs-manual-verification]** In the `enableSite(origin)` sketch, `origin` is resolved from the active tab in the popup, then `chrome.tabs.query({active, currentWindow})` is re-queried *inside* `enableSite` and `executeScript` injects into whatever is active at that later moment. If the active tab changes between the grant and the inject, you could `executeScript` into a tab whose origin wasn't the one just granted. Suggest capturing `tab.id` once up front, and verifying the tab's current origin still matches the granted origin before `executeScript`.

4. **User-influenced `source` now reaches the backend. [plausible — data exposure / input-trust]** Today `source` comes from a curated allowlist; with dynamic sites it becomes an arbitrary user-enabled hostname flowing into search + analytics. No injection risk in the extension itself (only bundled scripts run, mediaSession read only — confirmed), but the search/analytics backend should treat that `source` string as untrusted/unbounded input (length cap, no assumption it maps to a known platform).

Nits already handled well in the spec: duplicate-`id` registration race, `persistAcrossSessions` + manual reconciliation drift, and privileged-page (`chrome://`/`about:`) guarding are all called out under Edge cases. Good.

No regression risk — no code changed, no API contract touched.

---

PR #306, opened 2026-06-30
Fix Xcode Cloud build: commit Package.resolved

brandonlucasgreen:
## Problem

Xcode Cloud Mac builds are failing:

> a resolved file is required when automatic dependency resolution is disabled and should be placed at `…/project.xcworkspace/xcshareddata/swiftpm/Package.resolved`. Running resolver because the following dependencies were added: 'supabase-swift'

## Cause

The Mac app gained its **first SPM dependency** (`supabase-swift`) in the 3.2.0 release ([#303](https://github.com/brandonlucasgreen/unstream/pull/303)), which makes a `Package.resolved` lockfile necessary. Xcode Cloud builds with automatic dependency resolution disabled and require that lockfile to be committed.

But `apps/mac/Unstream.xcodeproj/project.xcworkspace/` was **never tracked in git** — so `Package.resolved` only ever existed locally. Every Xcode Cloud build since #303 has been missing it.

[#304](https://github.com/brandonlucasgreen/unstream/pull/304) didn't cause this — it was extension-only. It just happened to be the next push to `main`, which triggered the build that surfaced the pre-existing gap from #303.

## Fix

Commit the lockfile (pinning supabase-swift 2.44.1 + its transitive deps) and the standard `contents.xcworkspacedata`. `xcuserdata` remains gitignored, so no user-specific state is committed.

After merge, the Xcode Cloud build should resolve the exact pinned versions instead of failing.

### DIFF
--- a/apps/mac/Unstream.xcodeproj/project.xcworkspace/contents.xcworkspacedata
+++ b/apps/mac/Unstream.xcodeproj/project.xcworkspace/contents.xcworkspacedata
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Workspace
+   version = "1.0">
+   <FileRef
+      location = "self:">
+   </FileRef>
+</Workspace>
--- a/apps/mac/Unstream.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/apps/mac/Unstream.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -1,0 +1,69 @@
+{
+  "originHash" : "ccc727310637079b7355ccf59e7d153c26677020c1a43e177b78cc9df7963c0f",
+  "pins" : [
+    {
+      "identity" : "supabase-swift",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/supabase/supabase-swift.git",
+      "state" : {
+        "revision" : "06ae7b34ec21406cbd3e643bee7a8a54206fa8f5",
+        "version" : "2.44.1"
+      }
+    },
+    {
+      "identity" : "swift-asn1",
+      "kind" : "remoteSourceControl",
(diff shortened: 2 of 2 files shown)


brandonlucasgreen (commented):
Senior staff engineer review — @brandonlucasgreen

**Scope:** Config-only. Two added files: `contents.xcworkspacedata` (standard single self-reference workspace) and a SwiftPM `Package.resolved` lockfile. No application code, so the security surface (auth/authorization, input validation, secret handling, injection, unsafe defaults, data exposure) is effectively N/A here.

### Security
- **Secret handling — confirmed clean.** The lockfile contains only public repo URLs, version tags, and commit revisions. No tokens, keys, or credentials. `xcuserdata` stays gitignored, so no user-specific state leaks.
- **Dependency provenance — confirmed.** All pins resolve to well-known upstreams: `supabase/supabase-swift`, `apple/swift-asn1`, `apple/swift-crypto`, `apple/swift-http-types`, and the `pointfreeco` test/concurrency helpers. Locations use canonical `github.com` HTTPS sources pinned to explicit revisions — the desired hardening (Xcode Cloud resolves exact commits instead of floating).
- No injection / unsafe-default / data-exposure vectors introduced.

### Regression
- **API-contract / spec drift — confirmed no drift.** `apps/mac/project.yml` declares `Supabase … exactVersion: 2.44.1`, and the lockfile pins `supabase-swift` at exactly `2.44.1` (rev `06ae7b3…`). XcodeGen-generated project and lockfile agree.
- **State transition — confirmed correct fix.** Addresses the root cause from #303 (first SPM dependency added) rather than #304; committing the lockfile is the right resolution for "resolved file required when automatic dependency resolution is disabled."
- **Edge cases / maintenance note (needs-manual-verification, low):** standard committed-lockfile caveat — it must be regenerated whenever `project.yml`'s version pins change, or Xcode Cloud will fail again the same way. No action needed for this PR.

**Verdict:** No security or regression findings. Looks good to merge.

---

PR #310, opened 2026-07-11
Fix Linear webhook registration script: require allPublicTeams when no team given

brandonlucasgreen:
## Summary
- Linear's `webhookCreate` mutation rejects requests with neither `teamId` nor `allPublicTeams` set — there's no implicit "whole workspace" default, contrary to what the script's original comment assumed
- Fix: when `LINEAR_TEAM_ID` isn't provided, explicitly pass `allPublicTeams: true`

## Test plan
- [x] Script typechecks
- [ ] Re-run `npm run register:linear-webhook` after merge and confirm it succeeds

### DIFF
--- a/scripts/linear-register-webhook.ts
+++ b/scripts/linear-register-webhook.ts
@@ -3,7 +3,9 @@
 //   LINEAR_API_KEY=... npx tsx scripts/linear-register-webhook.ts
 //
 // Optional: LINEAR_TEAM_ID=... to scope the webhook to a single team
-// (omit to receive Issue events from every team in the workspace).
+// (omit to receive Issue events from every public team in the workspace —
+// Linear's webhookCreate mutation requires either a teamId or an explicit
+// allPublicTeams: true; there's no implicit "whole workspace" default).
 //
 // Requires a workspace-admin personal API key, or an OAuth app token with
 // the `admin` scope — per Linear's docs, only admins can create webhooks.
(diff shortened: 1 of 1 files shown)


brandonlucasgreen (commented):

**Security**
- Auth/authorization: unchanged — script still requires `LINEAR_API_KEY` (workspace-admin key) via env var; no auth logic touched. **confirmed** (no change)
- Secret handling: `LINEAR_API_KEY` and the returned webhook signing secret are handled the same way as before (env var in, printed to stdout for manual copy into Netlify) — not touched by this diff. **confirmed** (no change, pre-existing pattern)
- Injection risk: the new `allPublicTeams` field is passed through GraphQL `variables`, not string-interpolated into the query text — no injection surface introduced. **confirmed**
- Unsafe defaults: the new `else` branch sets `allPublicTeams: true` whenever `LINEAR_TEAM_ID` is unset. This isn't a *new* default — it restores the behavior the original comment already documented ("omit to receive Issue events from every team"), which was silently broken because Linear's API rejects a request with neither `teamId` nor `allPublicTeams`. **confirmed**, intentional and correctly scoped (this is a manually-run, admin-key-gated ops script, not reachable from user input).
- Data exposure: none introduced by this diff.

**Regression**
- Changed assumption: the prior code assumed omitting `teamId` alone was enough to get workspace-wide events. It isn't — Linear's `webhookCreate` requires an explicit `teamId` or `allPublicTeams: true`. That's the bug being fixed here, not a new regression. **confirmed**
- Edge case: `LINEAR_TEAM_ID=""` (empty string) still falls through to the `else` branch via the existing truthy check — same behavior as before the diff, not newly introduced. **confirmed** (no change)
- API contract drift: `allPublicTeams` is a genuinely new field added to the local `input` type for the `webhookCreate` mutation. I can't confirm from the diff alone that this field name/shape matches Linear's live GraphQL schema (`WebhookCreateInput`) — the PR's own test plan leaves this unchecked ("Re-run `npm run register:linear-webhook` after merge and confirm it succeeds"). **needs-manual-verification**

No blocking issues. Small, well-contained fix to a one-time manual ops script (not part of the deployed request path), consistent with the documented intent. Recommend merging once the post-merge verification step in the test plan is actually run.

---

PR #311, opened 2026-07-11
Add temporary debug logging to linear-webhook

brandonlucasgreen:
## Summary
Temporary diagnostic PR — UNS-157 test issue produced one 6.38ms invocation of `linear-webhook` with nothing downstream and nothing in Linear's own delivery-failure log, so we can't tell which check (signature, timestamp, or event-type filter) is rejecting the payload. Adds Sentry info-level breadcrumbs at each decision point (signature fail, stale-timestamp fail, event-ignored, dispatching) to pinpoint it on the next test issue.

**Remove this instrumentation once diagnosed** — not meant to be permanent.

## Test plan
- [x] `npm run typecheck:api` / `npm run test:api` pass
- [ ] Merge, redeploy, file another test issue, check Sentry for which breadcrumb fires

### DIFF
--- a/api/functions/linear-webhook.ts
+++ b/api/functions/linear-webhook.ts
@@ -54,6 +54,11 @@ export async function handler(event: {
 
   const signature = event.headers['linear-signature'];
   if (!signature || !isValidSignature(event.body, signature)) {
+    // TEMPORARY debug instrumentation — remove once the live-test issue is resolved.
+    Sentry.captureMessage('linear-webhook DEBUG: signature check failed', {
+      level: 'info',
+      extra: { hasSignatureHeader: !!signature, headerKeys: Object.keys(event.headers) },
+    });
     return { statusCode: 401, body: 'Invalid signature' };
   }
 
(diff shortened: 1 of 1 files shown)


api/functions/linear-webhook.ts
```diff
@@ -54,6 +54,11 @@ export async function handler(event: {
 
   const signature = event.headers['linear-signature'];
   if (!signature || !isValidSignature(event.body, signature)) {
+    // TEMPORARY debug instrumentation — remove once the live-test issue is resolved.
+    Sentry.captureMessage('linear-webhook DEBUG: signature check failed', {
+      level: 'info',
+      extra: { hasSignatureHeader: !!signature, headerKeys: Object.keys(event.headers) },
+    });
```
brandonlucasgreen, line 61: @brandonlucasgreen **Security — confirmed: unauthenticated Sentry event on every failed-signature request**

This `Sentry.captureMessage` fires *before* the HMAC signature is validated — i.e. on the `return 401` path for any request that fails `isValidSignature`. Since `linear-webhook` is a public, unauthenticated endpoint with no rate limiting (unlike `search-sources`, `resolve-url`, etc., which route through `api/functions/ratelimit.ts`), anyone can POST to it with a garbage/missing `linear-signature` header and trigger a Sentry ingest event on every single request, at whatever volume they choose.

On `main` today, an invalid-signature request is a pure, side-effect-free `401` (confirmed via diff against `origin/main`). This PR turns that into an outbound Sentry call per rejected request — a straightforward vector to burn through Sentry event quota / cost, or bury real alerts in noise, from an endpoint that takes no attacker-supplied secret to reach.

Worth noting this also breaks with the codebase's existing convention: the analogous signature check in `discord-interaction.ts` (`event.headers['x-signature-ed25519']` / Ed25519 verify) logs nothing on failure — it just returns 401. This PR's pre-auth breadcrumb is the first instance of capturing telemetry before signature verification succeeds.

Suggested fix: move this specific breadcrumb (or at least gate it) so it can't be driven by unauthenticated traffic — e.g. only fire when `signature` is present but invalid (still somewhat attacker-triggerable, but at least requires guessing an HMAC), or better, count failures in-memory/via existing rate limiter and only sample/report at a capped rate. The other three debug breadcrumbs added in this PR (stale timestamp, event-ignored, dispatching) are fine — they all execute after the signature check has already passed, so they're authenticated Linear traffic only.

Also flagging as a **regression/process risk** (not a code bug): all four breadcrumbs are marked "TEMPORARY... remove once diagnosed" but nothing enforces removal — if this lands and the diagnosis is deferred, the endpoint permanently ships with per-request Sentry calls until someone remembers to revert. Given finding #1, I'd suggest fixing the pre-auth placement before merge rather than shipping it "temporarily."

No other auth/input-validation/secret-handling/injection issues found — signature verification still uses `timingSafeEqual`, secret is read from env and never logged, and the new `extra` payloads only include header *names* (not values) and Linear's own issue title/identifier (not sensitive, no cross-tenant exposure).

---

brandonlucasgreen (commented):
Security review of this diagnostic PR — one confirmed finding, rest looks fine. See inline comment on `linear-webhook.ts` for details. @brandonlucasgreen

---

PR #314, opened 2026-07-11
Add user-opted dynamic per-site playback detection (UNS-152)

brandonlucasgreen:
Implements [UNS-152](https://linear.app/unstream/issue/UNS-152) — the "Tier 3" per-site playback detection from the expansion discussion.

## What it does

A user on a music site that isn't in the curated allowlist (e.g. `fairplayer.band`, an artist's own site) can open the popup and click **Enable on {host}**. The browser shows its native per-origin permission prompt; on accept, the existing generic Media Session detector is registered for that origin and injected into the current tab immediately (no reload). The site is remembered and auto-registers on future visits, and can be removed from a managed list in Settings (which revokes the permission).

This scales coverage to the long tail of small/self-hosted players **without** a new store release per site and **without** broadening the install-time permission prompt.

## Changes

- **New `apps/extension/lib/custom-sites.js`** — single home for the logic, imported by both popup and service worker: `enableSite`/`disableSite`/`reconcileCustomSites`, a standard match-pattern matcher (`isBuiltInSite`, unit-checked against the real allowlist incl. `amazon.com/music/*`), and the `customSites` storage source of truth.
- **Popup** — Enable card on Discover (shown only on scriptable, non-allowlist, not-already-enabled sites), a gentle "watching but no track info" idle note, and a Custom Sites remove-list in Settings. A 100-site sanity cap is surfaced in the UI.
- **Service worker** — reconcile stored origins vs granted permissions vs live registrations on `onInstalled`/`onStartup`, pruning any site whose permission was revoked out-of-band via browser settings.
- **Manifests** — add `scripting` + optional broad-host pattern; bump to **2.5.0**.

## Note on one spec deviation

The spec put `scripting` in Firefox's `optional_permissions`. This PR puts `scripting` in the **required** `permissions` for *both* manifests, and keeps only the broad host pattern (`https://*/*`) optional. Rationale: if `scripting` were optional in Firefox, `chrome.scripting` would be undefined until granted, breaking startup reconciliation. `scripting` alone grants no host access, so it doesn't widen the install prompt. Per-origin host access remains fully opt-in and user-initiated.

Also: the Enable button captures the target origin up front so `permissions.request` runs directly in the click gesture with no preceding `await` (Firefox rejects otherwise).

## Testing

Static checks pass (JS syntax, JSON validity, match-pattern unit test). The extension isn't covered by `npm run lint` or the build pipeline, and the enable flow depends on native permission dialogs, so end-to-end testing is manual (load unpacked):

- **Chrome:** enable `fairplayer.band`, confirm the prompt names only that origin, confirm detection without reload, restart browser and confirm persistence, remove and confirm script + permission are gone.
- **Firefox:** repeat with the Firefox manifest (manifest swap per usual flow).
- **Reconciliation:** revoke the permission in browser settings, restart, confirm the orphaned registration is cleaned up and the site drops from the list.
- **Negative:** enable a site with no Media Session support → idle "no track info" note, no crash.
- **Allowlist/privileged:** confirm the Enable control is hidden on built-in sites and `chrome://`/`about:` pages.

Ships in a normal extension release (version bump + manual zip workflow for Chrome + Firefox). Independent of the fairplayer.band static-allowlist PR (#304).

### DIFF
--- a/apps/extension/popup/popup.js
+++ b/apps/extension/popup/popup.js
@@ -3,6 +3,15 @@
 import { isBandcampFriday } from '../lib/bandcamp-friday.js';
 import { ALLOWED_RELEASE_DOMAINS, SOURCE_CONFIG, PAYOUT_PERCENTAGES } from '../lib/constants.js';
 import { signInWithPassword, signInWithOtp, signOut, getStoredSession, getAccessToken, getDeviceId } from '../lib/supabase.js';
+import {
+  getCustomSites,
+  enableSite,
+  disableSite,
+  originPattern,
+  isScriptableUrl,
+  isBuiltInSite,
+  MAX_CUSTOM_SITES,
+} from '../lib/custom-sites.js';
 
 const API_BASE = 'https://unstream.stream/api';
 
@@ -39,6 +48,10 @@ const elements = {
   // Discover tab
   nowPlaying: document.getElementById('now-playing'),
   idleState: document.getElementById('idle-state'),
+  detectSiteSection: document.getElementById('detect-site-section'),
+  detectTitle: document.getElementById('detect-title'),
+  detectDesc: document.getElementById('detect-desc'),
+  detectEnableBtn: document.getElementById('detect-enable-btn'),
   resultsSection: document.getElementById('results-section'),
   resultsGrid: document.getElementById('results-grid'),
   socialSection: document.getElementById('social-section'),
@@ -76,6 +89,8 @@ const elements = {
   releaseAlertsSection: document.getElementById('release-alerts-section'),
   checkNowBtn: document.getElementById('check-now-btn'),
   lastCheckTime: document.getElementById('last-check-time'),
+  customSitesList: document.getElementById('custom-sites-list'),
+  noCustomSites: document.getElementById('no-custom-sites'),
 };
 
 // State
@@ -86,6 +101,11 @@ let currentSocialLinks = null;
 let currentLocation = null;
 let newReleases = [];
 let authSession = null; // null = unknown/loading, false = signed out, object = signed in
+
+// Active-tab origin the "Enable" button will request, captured up front so the
+// permission request runs directly inside the user's click (no await before it,
+// which would break the required user gesture). { origin, tabId } or null.
+let detectTarget = null;
 
 // Slug lookup cache: { artistName → slug }
 const SLUG_CACHE_KEY = 'artistSlugCache';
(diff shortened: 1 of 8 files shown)


apps/extension/popup/popup.js
```diff
@@ -457,6 +477,141 @@ function formatLocation(location) {
...
+  const sites = await getCustomSites();
+
+  if (sites.includes(origin)) {
+    // Already enabled. If nothing is playing/detected, explain the idle state
+    // rather than silently showing nothing.
+    const { currentTrack } = await chrome.storage.local.get('currentTrack');
+    const hasFreshTrack = currentTrack && Date.now() - currentTrack.timestamp < 5 * 60 * 1000;
+    if (!hasFreshTrack) showDetectIdleNote();
+    return;
+  }
```
brandonlucasgreen, line 525: @brandonlucasgreen **Regression / state-transition — plausible.**

`setupDetectionControl()` treats membership in the stored `customSites` list as proof the site is actively watched, but it never calls `chrome.permissions.contains({ origins: [origin] })` here. If the user revokes the per-origin host permission out-of-band (via the browser's own site/extension permission UI) while the service worker is still alive, this branch still fires and shows "Watching this site" — even though the content script can no longer be injected. The user has no obvious way to recover from the popup itself; they'd need to notice, go to Settings → Custom Sites, remove the entry, and re-enable (re-triggering the permission prompt), or wait for the next service-worker restart to trigger `reconcileCustomSites()`.

Root cause is in `apps/extension/lib/custom-sites.js`: reconciliation is only wired to `chrome.runtime.onInstalled`/`onStartup` (see `service-worker.js`), not to `chrome.permissions.onRemoved`, which exists specifically for this case and would let you prune/reflect the revoked permission immediately instead of relying on the next SW eviction.

---

apps/extension/popup/popup.js
```diff
@@ -457,6 +477,141 @@ function formatLocation(location) {
...
+  const origin = originPattern(tab.url);
+  const host = new URL(tab.url).hostname;
+  const sites = await getCustomSites();
+
+  if (sites.includes(origin)) {
+    // Already enabled. If nothing is playing/detected, explain the idle state
+    // rather than silently showing nothing.
+    const { currentTrack } = await chrome.storage.local.get('currentTrack');
+    const hasFreshTrack = currentTrack && Date.now() - currentTrack.timestamp < 5 * 60 * 1000;
+    if (!hasFreshTrack) showDetectIdleNote();
```
brandonlucasgreen, line 523: @brandonlucasgreen **Regression / edge case — plausible.**

`hasFreshTrack` is computed from the single global `currentTrack` value in storage without checking that it actually came from the active tab/site. If music was detected on a different tab (e.g. a Spotify tab) within the last 5 minutes, `hasFreshTrack` is `true`, so `showDetectIdleNote()` is skipped for the *currently active* newly-enabled custom site — and since this branch also doesn't call `showDetectEnablePrompt`, the whole `detect-site-section` stays hidden (it was reset to hidden at the top of the function). Net effect: a user on a freshly-enabled site with nothing playing gets no feedback at all instead of the intended "watching, no track info yet" note, purely because of stale state from an unrelated tab.

Worth scoping the freshness check to the current site (e.g. compare `currentTrack.source`/origin) rather than just recency.

---

brandonlucasgreen (commented):
@brandonlucasgreen Security + regression pass on #314 (UNS-152 custom-site detection). Overall the permission model is sound: host-permission requests are narrowly scoped to a single origin, requested synchronously inside the click handler (correct for both Chrome and Firefox gesture requirements), `isScriptableUrl` correctly excludes privileged pages, and the new UI renders all user-derived strings via `textContent` (no injection risk there). `optional_host_permissions`/`optional_permissions` are both scoped to `https://*/*` only (no bare `http`), and `scripting` alone (without host access) is inert, so making it a required permission is low risk on its own.

Two plausible regressions left inline, both in the new `setupDetectionControl()` flow in `apps/extension/popup/popup.js`:
1. **Permission-revoked staleness** — no `chrome.permissions.onRemoved` listener, so a site revoked out-of-band shows as "Watching" until the next service-worker restart.
2. **Cross-tab stale-track false negative** — the idle check reads the global `currentTrack` without scoping to the current site, so a recent track from an unrelated tab can suppress the "watching, no track info" note entirely.

One item flagged for manual confirmation but not inline-commentable (touches only pre-existing, unmodified lines in `service-worker.js`/`content/generic.js`, outside this diff's hunks): this feature lets `trackAppEvent`'s `streaming_service` context carry arbitrary user-chosen hostnames instead of the previous fixed platform list — worth confirming that's consistent with the "no tracking" claim in the manifest description.

No confirmed auth/injection/secret-handling issues found. Labels: findings above are `plausible`; the analytics-scope note is `needs-manual-verification`.

---

PR #316, opened 2026-07-24
Replace blank loading screens with branded skeleton UI

brandonlucasgreen:
Every loading state in the web app was either a bare "Loading..." string on
an otherwise empty page or a generic spinner. Signed-in pages were the worst:
the dashboard, settings and profile editor dropped the header and footer
entirely while they fetched, so the whole app disappeared for a beat.

Adds a small set of loading primitives and page-shaped skeletons built on
them, then wires them through every loading branch:

- Skeleton.tsx: shimmer block, circle, text lines, equalizer bars and an
  accessible SkeletonScreen wrapper.
- LoadingSkeletons.tsx: skeletons shaped like the content they stand in for
  (search results, artist profile, dashboard, artist rows, article list and
  body, forms, stat tiles).
- PageSkeleton.tsx: full-page shell that keeps the real header and footer, so
  a loading page still looks like Unstream.
- AppFallback.tsx: the router's Suspense fallback plus a real error fallback.
  The error boundary previously rendered the loading skeleton, which left
  users staring at a placeholder forever after a crash.

The shimmer sweeps a warm accent tint across each block and the equalizer
bars use the three accent colours, so the waiting state reads as ours rather
than as a generic grey pulse. Both are theme-aware and both stop under
prefers-reduced-motion. Skeleton screens fade in after 120ms so fast loads
never flash a placeholder.

Claude-Session: https://claude.ai/code/session_01BdfxTVsXrxVukhcMAgox5Y

### DIFF
--- a/apps/web/src/components/PageSkeleton.tsx
+++ b/apps/web/src/components/PageSkeleton.tsx
@@ -1,0 +1,30 @@
+import { Header } from './Header';
+import { Footer } from './Footer';
+import { SkeletonScreen } from './Skeleton';
+
+/**
+ * Full-page loading shell. Renders the real header and footer — they need no
+ * data — and puts a content skeleton where the page body will land, so a
+ * loading page still looks like the app instead of a blank screen.
+ */
+export function PageSkeleton({
+  label,
+  maxWidth = 'max-w-2xl',
+  children,
+}: {
+  label: string;
+  maxWidth?: string;
+  children: React.ReactNode;
+}) {
+  return (
+    <div className="min-h-screen bg-bg-primary text-text-primary flex flex-col">
+      <Header />
+      <main className="flex-1 p-6">
+        <div className={`${maxWidth} mx-auto`}>
+          <SkeletonScreen label={label}>{children}</SkeletonScreen>
+        </div>
+      </main>
+      <Footer />
+    </div>
+  );
+}
--- a/apps/web/src/App.tsx
+++ b/apps/web/src/App.tsx
@@ -5,6 +5,8 @@ import { SearchBar } from './components/SearchBar';
 import { ResultCard } from './components/ResultCard';
 
 import { Header } from './components/Header';
+import { LoadingLabel, SkeletonScreen } from './components/Skeleton';
+import { SearchResultsSkeleton } from './components/LoadingSkeletons';
 import type { SearchResult } from './types';
 import { sources, sourceCategories, searchPlatforms, resolveArtistUrl, fetchMusicBrainzData, mergeWithMusicBrainzData } from './services/sources';
 import { analytics } from './services/analytics';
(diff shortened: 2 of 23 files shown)


apps/web/src/components/PageSkeleton.tsx
brandonlucasgreen, line 1: **Regression — plausible:** `PageSkeleton` mounts the real `<Header />` (not a static placeholder). `Header` has a `useEffect` that fires an authenticated `fetch('/api/admin/verify', ...)` whenever `isAdmin && session?.access_token` are set:

```tsx
// Header.tsx
useEffect(() => {
  if (!isAdmin || !session?.access_token) return;
  fetch('/api/admin/verify', { headers: { Authorization: `Bearer ${session.access_token}` } })
    ...
}, [isAdmin, session?.access_token]);
```

Since `isAdmin`/`session` typically resolve in `AuthContext` before a given page's own `loading` flag clears, admin users will mount `Header` once inside the skeleton (firing the fetch) and again when the real page renders (firing it a second time) — on every page that uses `PageSkeleton` (`DashboardPage`, `ArtistDashboardPage`, `AdminAnalyticsPage`, `ArtistEditPage`, `LoginPage`, `ResetPasswordPage`, `PublicSavedArtistsPage`) plus `AppLoadingFallback` for every lazily-loaded route. Not a security bug (same bearer token, same endpoint an admin can already call), just a duplicated request per navigation for admin users — worth a quick manual check to confirm it's an acceptable trade-off vs. e.g. rendering the skeleton without a live `Header`.

Label: **regression, plausible** (verified by reading `Header.tsx`'s effect and dependency timing; not confirmed against a live admin session).

---

brandonlucasgreen (commented):
@brandonlucasgreen Automated staff-engineer review of this PR.

**Scope:** this is a pure frontend/CSS change (loading skeletons, primitives, one new test file). No touches to `api/`, auth logic, Supabase, middleware, or any user-input handling.

### Security

| Area | Finding |
|---|---|
| Auth/authorization | No changes to auth/authz logic. **confirmed** clean. |
| Input validation | No new external/user input is handled anywhere in this diff — all skeleton props are static strings/counts authored by page components. **confirmed** clean. |
| Secret handling | No secrets touched. **confirmed** clean. |
| Injection risks | No `dangerouslySetInnerHTML`, no string concatenation into HTML/URLs with untrusted data — all `className` template strings are built from hardcoded Tailwind tokens, not user data. **confirmed** clean. |
| Unsafe defaults | Component default props (`count`, `sections`, `fields`, `bars`) are cosmetic only, no security implication. **confirmed** clean. |
| Data exposure | No new data is fetched, logged, or rendered by this PR. **confirmed** clean. |

No security findings.

### Regression

- One plausible finding left as an inline comment on `PageSkeleton.tsx`: mounting the real `<Header />` inside the loading skeleton causes a duplicate `/api/admin/verify` fetch for admin users on every page that uses `PageSkeleton`/`AppLoadingFallback`, since `Header` re-mounts when the real page takes over. Not a security issue, just a possible extra network round-trip — flagged **needs-manual-verification** since I haven't reproduced it against a live admin session, only traced it through the code.
- `main.tsx`'s `Sentry.ErrorBoundary` fallback and the `Suspense` fallback were previously the same component; they're now split into `AppErrorFallback` (plain `<a href="/">`, correctly avoiding `<Link>` since it can render outside Router context) and `AppLoadingFallback`. This looks correct and is actually a fix — the old shared fallback made a crash look like an infinite loading state.
- Skeleton components carry no real business logic (just `Array.from({length})` renders), so no off-by-one or null/empty-state risk beyond cosmetic layout.
- No API contracts changed.

CI's `check` run is green; Netlify deploy-preview checks were still in progress at review time.

---

PR #317, opened 2026-07-26
Report Bandcamp bot-challenge instead of failing silently (UNS-152)

brandonlucasgreen:
## Problem

`bandcamp.com/search` is behind a Fastly bot challenge that responds with **HTTP 200** and a ~3KB interstitial. `searchBandcampForArtistUrl` therefore parsed zero results and returned `null` — indistinguishable from "this artist isn't on Bandcamp." A wrong confident answer rather than a visible failure, with nothing reaching Sentry.

## Change

- Adds `isBandcampChallenge(html)` to `api/functions/search-parsers.ts` — pure, no I/O, sits with the other testable parsers. Detects Fastly's `/_fs-ch-` asset path, with a >20KB size pre-filter so a stray mention inside a real 100KB+ page can't disable a working scrape.
- Wires it into `searchBandcampForArtistUrl`, reporting via `Sentry.captureMessage` deduped through the existing `checkSentryDedup` at a 6h TTL.

The function still returns `null`. `/search` is also `Disallow`ed in Bandcamp's `robots.txt`, so it is superseded by the slug-probe lookup described in `docs/specs/bandcamp-coverage-research.md`. This PR makes the current failure observable in the meantime.

Also adds the research doc: verified live code map, the robots-allowed `discover_web` API contract, measured slug-probe recall, and the squatter-filter analysis.

## Verification

Checked against real captured responses from live probing:

| Input | Bytes | Detected |
|---|---|---|
| live challenge page | 3,036 | `true` |
| live `/search?item_type=b` | 3,036 | `true` |
| live artist page | 168,103 | `false` |
| live album page | 383,715 | `false` |

`tsc -b` clean · `typecheck:api` clean · api tests 69/69 · unit tests 359/359

Note: `npm run lint` reports 80 pre-existing problems across ~40 files in `apps/web/` — identical before and after this change, none in files touched here, and `lint` is not in the build pipeline. Left alone rather than expanding scope.

### DIFF
--- a/api/functions/search-parsers.ts
+++ b/api/functions/search-parsers.ts
@@ -13,6 +13,24 @@ import {
 // Bandcamp
 // ---------------------------------------------------------------------------
 
+/**
+ * True if this HTML is a Fastly bot-challenge interstitial rather than real content.
+ *
+ * Bandcamp serves the challenge with HTTP 200, so an `!response.ok` check never
+ * catches it. Without this, a challenge parses to zero results and is
+ * indistinguishable from "the artist genuinely isn't on Bandcamp" — a wrong
+ * confident answer rather than a visible failure.
+ *
+ * Fastly's challenge page loads its assets from a `/_fs-ch-<token>/` path and
+ * carries a restrictive inline CSP; the asset path is the stable marker.
+ */
+export function isBandcampChallenge(html: string): boolean {
+  if (!html) return false;
+  // Challenge pages are small; real Bandcamp pages are 100KB+. Cheap pre-filter.
+  if (html.length > 20_000) return false;
+  return html.includes('/_fs-ch-');
+}
+
 /** Parse Bandcamp search results HTML into PlatformResult[] */
 export function parseBandcampSearchResults(html: string, query: string): PlatformResult[] {
   const results: PlatformResult[] = [];
--- a/api/search/enrichment.ts
+++ b/api/search/enrichment.ts
@@ -1,9 +1,12 @@
 // Shared enrichment functions for MusicBrainz data extraction
 // These functions are used both by the MusicBrainz Netlify function and search-sources.ts
 
+import { Sentry } from '../lib/sentry';
 import { cacheGetOrFetch, artistCacheKey } from '../functions/cache';
 import { persistEnrichment } from '../functions/db';
 import { isUrlHostnameAllowed } from '../functions/middleware';
+import { checkSentryDedup } from '../functions/ratelimit';
+import { isBandcampChallenge } from '../functions/search-parsers';
 
 // Social platform types
 export type SocialPlatform =
@@ -466,6 +469,11 @@ export function mergeLocations(...sources: (ArtistLocation | null | undefined)[]
 // Search Bandcamp for an artist by name and return the first matching artist/label URL.
 // Mirrors the Phase 1 Bandcamp scrape but runs server-side within Phase 2 enrichment,
 // so no client-supplied URLs are involved.
+//
+// NOTE: bandcamp.com/search is behind a Fastly bot challenge and is `Disallow`ed in
+// Bandcamp's robots.txt, so this currently always returns null. It is superseded by the
+// slug-probe lookup — see docs/specs/bandcamp-coverage-research.md. Until that lands, the
+// challenge is detected and reported rather than being silently read as "not on Bandcamp".
 export async function searchBandcampForArtistUrl(artistName: string, timeoutMs = 4000): Promise<string | null> {
   try {
     const searchUrl = `https://bandcamp.com/search?q=${encodeURIComponent(artistName)}&item_type=b`;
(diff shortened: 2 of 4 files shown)


api/functions/search-parsers.ts
brandonlucasgreen, line 1: **Regression — brittle detection, same failure mode it fixes (plausible)**

`isBandcampChallenge` gates entirely on `html.length > 20_000` + a literal `/_fs-ch-` substring match against Fastly's current challenge markup. If Fastly changes the challenge page's asset-path prefix or pushes its size over 20KB (extra inline JS, a longer CSP nonce list, etc.), this silently returns `false` again — and the code falls straight back to the exact bug this PR fixes: a 200 response with zero parsed results, read as "artist not on Bandcamp," nothing reaching Sentry. Worth a `needs-manual-verification` flag on your side rather than a blocking issue, since the PR is explicit that this is a stopgap superseded by the slug-probe work — but since there's no independent verification path here (e.g. logging "0 searchresult blocks parsed AND not flagged as challenge" as its own weaker signal), a future Fastly change reintroduces silent failure with no alarm. Consider a looser secondary heuristic (e.g. "response < 20KB AND zero `.searchresult` matches" also fires a lower-confidence capture) so a marker change doesn't fully re-silence this.

@brandonlucasgreen

---

api/search/enrichment.ts
brandonlucasgreen, line 1: **Regression — Redis load added to a hot, high-frequency path (needs-manual-verification)**

`searchBandcampForArtistUrl` is reached from `enrichLocationFallback`, which fires whenever MusicBrainz has no location for a query (`search-sources.ts:521/528/540`, `search-musicbrainz.ts:91/98/111`), plus a direct call at `search-musicbrainz.ts:300`. Per the PR's own research doc, `/search` is now blocked with a **100% hit rate**, so every one of those calls now takes the new branch and does a Redis `GET` via `checkSentryDedup` — for what is likely a very common case (indie/alt artists sparsely covered in MusicBrainz are exactly Unstream's core audience). The 6h dedup keeps actual Sentry writes rare, but the `GET` itself isn't gated — it's on every fallback call. Given Upstash Redis is billed and shared with real rate-limiting, worth a quick gut-check on request volume before merge; if it's negligible relative to existing rate-limit traffic, no action needed. Not blocking either way.

Separately, confirmed as correct: `checkSentryDedup` fails open on Redis error/unavailability (`ratelimit.ts:369-380`), so if Redis has an outage, this branch would emit an unbounded Sentry `captureMessage` per call instead of degrading gracefully — same fail-open pattern already used elsewhere in this file (`uc5`/`uc6` dedup keys in `search-sources.ts`), so this isn't a new risk class, just inherited exposure from a pre-existing utility.

**Data exposure — none found.** `extra.searchUrl` embeds the encoded artist-name query string, consistent with existing captures in `search-sources.ts:1720` (`extra: { platform, query, ... }`) and `:1950` (`extra: { query, ... }`) — not a new pattern, and it's a public search term, not PII.

**Input validation / injection — confirmed clean.** `artistName` is passed through `encodeURIComponent` before landing in `searchUrl`; `isBandcampChallenge` does a plain substring `.includes()`, no regex, no ReDoS surface. No client-supplied URL is fetched here (matches the existing comment above the function).

**Behavioral regression — none.** Before this PR, a challenge response already fell through the existing `.searchresult` parsing loop and returned `null` (zero matches). This PR returns `null` at the same point, just earlier and now observable — the actual return value for every caller is unchanged.

@brandonlucasgreen

---

brandonlucasgreen (commented):
@brandonlucasgreen Reviewed. Small, well-scoped, self-contained change — no auth/authz surface, no secrets, no injection risk (artist name is `encodeURIComponent`-encoded before reaching the URL; the new detector is a plain substring check, no regex/ReDoS). Behavior for every existing caller is unchanged (a challenge response already resolved to `null` before this PR; now it does so earlier and observably).

**Security**
- Auth/authz: N/A — no change.
- Input validation: confirmed clean — `artistName` properly encoded, no client-supplied URLs fetched.
- Secrets: N/A.
- Injection: confirmed clean — no regex/ReDoS in `isBandcampChallenge`.
- Data exposure: confirmed clean — `extra.searchUrl` in the Sentry payload matches existing capture patterns elsewhere in `search-sources.ts` (query terms are already logged there); not a new exposure class.
- Unsafe defaults: none found.

**Regression**
- Detection is brittle by construction (20KB size cap + literal `/_fs-ch-` marker) — see inline comment on `search-parsers.ts`. If Fastly changes the challenge page's shape, this silently regresses to the exact bug it fixes, with no fallback signal. `needs-manual-verification` / worth a follow-up, not blocking.
- `checkSentryDedup`'s Redis `GET` is now on a hot path (every `enrichLocationFallback` fallback call, likely frequent given Unstream's long-tail-artist audience) — see inline comment on `enrichment.ts`. Fail-open behavior on Redis outage is pre-existing and consistent with other dedup keys in the codebase, not new risk. `needs-manual-verification` on volume, not blocking.
- No edge-case/off-by-one/state-transition issues found; test coverage (empty string, whitespace, large page with a stray marker, real vs. challenge HTML) is solid for the function's scope.

Nothing here blocks merge — both notes are follow-up-worthy, not defects.

---

PR #318, opened 2026-07-26
Fix ReferenceError that has disabled MusicBrainz enrichment since March

brandonlucasgreen:
## The bug

`api/functions/search-musicbrainz.ts` calls `artistCacheKey()` at line 372, but line 4 only imports `cacheGetOrFetch`:

```ts
import { cacheGetOrFetch } from './cache';   // artistCacheKey missing
...
const cacheKey = artistCacheKey('musicbrainz', normalizedQuery);   // ReferenceError
```

`artistCacheKey` is exported from `./cache` and has no local declaration, so this is a hard `ReferenceError` on every invocation, thrown before the handler does any work.

## Why nobody saw it

The failure is silent at every layer:

1. The handler's own `try/catch` swallows the ReferenceError and returns **HTTP 500 with a well-formed all-nulls payload** (`artistName: null`, `socialLinks: []`, …).
2. The web client does `if (!response.ok) return null` — with **no Sentry capture** on that branch (the capture only covers thrown network errors).

So Phase 2 enrichment just returns nothing, and nothing anywhere reports a problem.

## How long

Introduced **2026-03-14** in `7d83630` (the v2.0 repo reorg) — roughly **four months**. Everything Phase 2 provides has been dead that whole time: official site, Discogs, social links, Wikipedia, MusicBrainz location, and the MusicBrainz-derived Bandcamp URL.

## Verification

Invoked the real handler with `query=radiohead`:

**Before**
```
MusicBrainz endpoint error: ReferenceError: artistCacheKey is not defined
statusCode: 500
artistName: null | officialUrl: null | socialLinks: []
```

**After**
```
statusCode: 200
artistName: Radiohead | officialUrl: http://www.radiohead.com/
socialLinks: facebook, instagram, tiktok, youtube, bluesky
9 platform URLs found, including https://radiohead.bandcamp.com/
```

Worth noting that last line: the MusicBrainz route to Bandcamp links was working the whole time — it just never got to run.

`tsc -b` clean · api tests 69/69 · unit tests 359/359

## Follow-up worth doing

This is invisible to the build because `api/tsconfig.json` has `include` limited to six named files. Nothing else under `api/` — ~40 functions and all the search modules — is typechecked. Put this file in scope and `tsc` flags it immediately as `TS2304`. Widening that coverage would have caught this in March; I'd suggest it as its own PR since there are ~30 pre-existing errors to work through.

### DIFF
--- a/api/functions/search-musicbrainz.ts
+++ b/api/functions/search-musicbrainz.ts
@@ -1,7 +1,7 @@
 // MusicBrainz enrichment server function
 // Imports shared functions from enrichment.ts, keeps Netlify handler here
 
-import { cacheGetOrFetch } from './cache';
+import { cacheGetOrFetch, artistCacheKey } from './cache';
 import { persistEnrichment } from './db';
 import { checkRateLimit, getClientIp } from './ratelimit';
 import { validateQuery, isUrlHostnameAllowed } from './middleware';
--- a/api/search/enrichment.ts
+++ b/api/search/enrichment.ts
@@ -585,7 +585,11 @@ export async function fetchBandcampLocation(bandcampUrl: string, timeoutMs = 400
 export async function fetchMirloLocation(artistSlug: string, timeoutMs = 4000): Promise<ArtistLocation | null> {
   if (!artistSlug) return null;
   try {
-    const apiUrl = `https://api.mirlo.space/v1/artists?urlSlug=${encodeURIComponent(artistSlug)}`;
+    // Must be the by-slug endpoint. The list form (`/v1/artists?urlSlug=…`) silently
+    // IGNORES the filter and returns the first page of all artists, so reading
+    // results[0] handed every artist on the site the same location — "Holyoke, MA",
+    // whoever happens to be first. This endpoint 404s when the artist isn't on Mirlo.
+    const apiUrl = `https://api.mirlo.space/v1/artists/${encodeURIComponent(artistSlug)}`;
     const controller = new AbortController();
     const timeout = setTimeout(() => controller.abort(), timeoutMs);
     const response = await globalThis.fetch(apiUrl, {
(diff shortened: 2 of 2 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Reviewed. This is a minimal, well-scoped fix — one-line diff restoring the `artistCacheKey` import that was dropped in the March reorg (`7d83630`).

**Security**

- **Auth/authorization** — no changes touched; handler still gated by `checkRateLimit`/`getClientIp` before any work runs. N/A.
- **Input validation** — unaffected; `validateQuery(event.queryStringParameters?.query)` still runs before the query reaches `normalizeSearchQuery`/`artistCacheKey`. **confirmed** no regression.
- **Secret handling** — no secrets touched.
- **Injection risks** — `artistCacheKey('musicbrainz', normalizedQuery)` builds a Redis key via `query.toLowerCase().trim().replace(/\s+/g, '_')` (cache.ts:134-136), not a raw string concatenation into SQL/shell/URL. Matches the exact call signature and usage pattern already in production at `search-sources.ts:972,1046,1087,1175,1247` and `search/enrichment.ts`. **confirmed** no injection risk introduced.
- **Unsafe defaults / data exposure** — none. The fix doesn't change response shape, CORS headers, or what's persisted via `persistEnrichment`.

No security findings.

**Regression**

- **Changed assumptions / API contract** — response shape, cache TTL (30 min), and the `emptyResult` fallback are all unchanged. Restoring the working code path (real enrichment data instead of the all-nulls 500 payload) is a pure improvement, not a behavior change fans would notice as a regression. **confirmed** safe.
- **Edge cases / null-empty** — no-match and low-confidence-score (`score < 95`) fallback paths (`enrichLocationFallback`) are untouched by this diff.
- **Test coverage gap (needs-manual-verification)** — this PR ships no test asserting the import binds / no smoke test guarding this bug class from recurring. The author's own PR description flags that `api/tsconfig.json`'s narrow `include` list is why `tsc -b` didn't catch this for 4 months, and defers widening it to a follow-up PR. Agreed with deferring the tsconfig widening (real scope, ~30 pre-existing errors) — but flagging that until that lands, a plain `import { artistCacheKey } from './cache'` typo/removal here would silently reintroduce this exact failure mode with no CI signal. Worth a cheap unit test importing the handler module (which would fail fast on any top-level `ReferenceError`) as a stopgap even before the tsconfig follow-up ships.

Nothing blocking. Nice catch on a fix that's been silently degrading Phase 2 enrichment since March — the "swallowed 500 → all-nulls payload → client returns null with no Sentry capture" failure chain you traced is exactly the kind of thing worth a follow-up: consider adding Sentry capture on the client's `!response.ok` branch too, not just thrown network errors, so a future regression here doesn't take another 4 months to surface.

---

PR #319, opened 2026-07-26
Find Bandcamp artists by probing subdomains instead of the blocked search (UNS-152)

brandonlucasgreen:
Layer A of the Bandcamp coverage work. Rebased onto main after #317 and #318 (both squash-merged, so this replays only the Layer A commit).

## Why

`bandcamp.com/search` is behind a Fastly bot challenge **and** `Disallow`ed in Bandcamp's `robots.txt`. It can't be used to look artists up. This replaces it with candidate-subdomain probing against `<slug>.bandcamp.com/music` — a path `robots.txt` permits.

## How

One request per candidate answers everything:

| Signal | Meaning |
|---|---|
| `HTTP 404` | no such account |
| `data-band="{id,name}"` | authoritative identity |
| `.music-grid-item` `data-item-id="album-…"/"track-…"` | album / track counts |

**Both verification steps are load-bearing:**

- A slug existing ≠ the right artist. `thebeths.bandcamp.com` resolves but belongs to an account named **"no content"**.
- A matching name ≠ a real presence. The accounts at `beyonce`, `sufjan` and `jackwhite` all match by name and hold **zero releases**. The real Jack White is at `officialjackwhite`.

Three candidates, in measured recall order over 3,018 known artists: base **66.3%**, +strip-"the" **69.3%**, +hyphenated **71.1%**. The curve then flattens — eleven candidates reach only 74.8% — so three is a deliberate stop. True rate is nearer **80%**, because many artists hold more than one valid account and Bandcamp redirects some alias slugs.

## Caching

New table `bandcamp_slug_probes`, **negatives included**. Without negative caching, every search for an artist who isn't on Bandcamp would re-probe forever. Positives never expire; negatives refresh after 30 days since an artist may join later.

Server-only: RLS enabled with **no policies by design** (service role bypasses, anon and authenticated get nothing).

## The `undecided` verdict

`probeBandcampArtist` returns a fifth verdict for a timeout, network error, or bot challenge. **The migration's CHECK constraint deliberately omits it**, so a transient failure can never be written as a cached negative — enforced in both `putBandcampProbe` and the schema.

This is the failure mode that caused the original coverage loss (a 200 challenge read as "not on Bandcamp") and the Mirlo bug in #318 ("no data" indistinguishable from "someone else's data"). Belt and braces this time.

## Verified against live Bandcamp

| Query | Verdict | Alb/Trk | Result |
|---|---|---|---|
| Boy Harsher | `accepted` | 15/1 | `boyharsher` |
| Radiohead | `accepted` | 15/0 | `radiohead` |
| Sufjan Stevens | `accepted` | 15/1 | `sufjanstevens` |
| The Mountain Goats | `accepted` | 14/2 | `themountaingoats` |
| Robyn Hitchcock | `accepted` | 16/0 | `robynhitchcock` |
| Beyonce | `rejected_empty` | 0/0 | squatter caught |
| Jack White | `rejected_empty` | 0/0 | squatter caught |
| Panda Bear / Butcher Brown / MESH | `rejected_empty` | 0/0 | real ones are `…music` / `meshphilly` |
| The Beths | `rejected_name` | 0/0 | `thebeths` is "no content" |
| King Gizzard & The Lizard Wizard | `absent` | 0/0 | real one is truncated `kinggizzard` |
| nonsense query | `absent` | 0/0 | correct |

**No case returned a wrong URL.** Every failure mode declines cleanly.

## Checks

- `tsc -b` clean
- Explicit strict `tsc` over all touched files + import graph: **zero new errors** (31 baseline on main, 31 with this branch). Necessary because `api/tsconfig.json` only covers six files — see #318's note.
- api tests **69/69**
- unit tests **378/378** (up from 359 — 19 new)

New unit tests cover `parseBandcampBandIdentity` (escaped JSON, the "no content" squatter case, malformed input, wrong types), `parseBandcampReleaseCounts` (album/track split, dedupe, grid-only scoping), and `bandcampSlugCandidates` (order, dedupe, accents, 3-char floor, mid-name "the").

## Migration

`20260726120000_bandcamp-slug-probes.sql` — auto-deploys on merge via the Supabase GitHub Action. `IF NOT EXISTS` guards throughout, RLS enabled, two indexes (one partial, for the future `/admin/verify` queue).

## Note on scope

Cross-source release matching as an ambiguity resolver (`pending_review` → `/admin/verify`) is **not** in this PR. The verdict value exists in the CHECK constraint so it doesn't need a second migration later, but nothing produces it yet. That's a follow-on after Layer B, per the agreed order.

### DIFF
--- a/api/search/bandcamp-probe.ts
+++ b/api/search/bandcamp-probe.ts
@@ -215,15 +215,23 @@ export async function probeBandcampArtist(
  * cached too, so repeated searches for an artist who isn't on Bandcamp are free
  * after the first. Undecided outcomes are never cached, so a transient failure
  * doesn't become a permanent miss.
+ *
+ * `budgetMs` caps the whole probe round, not each request. Callers running inside
+ * a Netlify function's 10s ceiling should pass their own budget rather than take
+ * the default — measured latency is 270-980ms, but the cap is what bounds the bad
+ * case if Bandcamp is slow.
  */
-export async function findBandcampArtistUrl(query: string): Promise<string | null> {
+export async function findBandcampArtistUrl(
+  query: string,
+  budgetMs = DEFAULT_BUDGET_MS,
+): Promise<string | null> {
   const queryNorm = normalizeForComparison(query);
   if (!queryNorm) return null;
 
   const cached = await getBandcampProbe(queryNorm);
   if (cached) return cached.artist_url;
 
-  const result = await probeBandcampArtist(query);
+  const result = await probeBandcampArtist(query, budgetMs);
 
   // Don't know != not there. Leave the cache empty so we retry next time.
   if (result.verdict === 'undecided') return null;
--- a/api/search/bandcamp-probe.ts
+++ b/api/search/bandcamp-probe.ts
@@ -1,0 +1,243 @@
+// Find a Bandcamp artist URL by probing candidate subdomains.
+//
+// bandcamp.com/search is behind a Fastly bot challenge and is Disallow'ed in
+// Bandcamp's robots.txt, so it cannot be used to look artists up. Instead we
+// derive candidate slugs from the artist name and ask <slug>.bandcamp.com/music
+// directly — a path robots.txt permits.
+//
+// One request per candidate answers everything we need:
+//   - HTTP 404                      -> no such account
+//   - data-band="{id,name}"         -> authoritative identity
+//   - .music-grid-item data-item-id -> album / track counts
+//
+// Both verification steps are load-bearing. A slug existing does not mean it is
+// the right artist (thebeths.bandcamp.com is an unrelated account named "no
+// content"), and a matching name does not mean it is a real presence — the
+// accounts at beyonce, sufjan and jackwhite all match by name and hold no
+// releases at all. The real Jack White is at officialjackwhite.
+//
+// See docs/specs/bandcamp-coverage-research.md for the measurements behind this.
+
+import { Sentry } from '../lib/sentry';
+import { getBandcampProbe, putBandcampProbe } from '../functions/db';
+import { isUrlHostnameAllowed } from '../functions/middleware';
+import { checkSentryDedup } from '../functions/ratelimit';
+import {
+  bandcampSlugCandidates,
+  namesMatch,
+  normalizeForComparison,
+} from '../functions/search-utils';
+import {
+  isBandcampChallenge,
+  parseBandcampBandIdentity,
+  parseBandcampReleaseCounts,
+} from '../functions/search-parsers';
+
+/**
+ * Outcome of probing for an artist.
+ *
+ * `undecided` means we could not find out — a network error, a timeout, or a bot
+ * challenge. It is deliberately NOT a value the cache accepts: caching it would
+ * turn a transient outage into a permanent "this artist isn't on Bandcamp".
+ */
+export type BandcampProbeVerdict =
+  | 'accepted'
+  | 'absent'
+  | 'rejected_empty'
+  | 'rejected_name'
+  | 'undecided';
+
+export interface BandcampProbeResult {
+  verdict: BandcampProbeVerdict;
+  /** Resolved artist URL. Non-null only when verdict is 'accepted'. */
+  artistUrl: string | null;
+  bandName?: string;
+  bandId?: number;
+  albumCount: number;
+  trackCount: number;
+  matchedSlug?: string;
+}
+
+const DEFAULT_BUDGET_MS = 5000;
+const MIN_REQUEST_MS = 1200;
+
+interface CandidateOutcome {
+  verdict: Exclude<BandcampProbeVerdict, 'absent'>;
+  identity?: { id: number; name: string };
+  counts: { albums: number; tracks: number };
+}
+
+/** Fetch and classify a single candidate slug. Returns null if the slug has no account. */
+async function probeCandidate(slug: string, timeoutMs: number): Promise<CandidateOutcome | null> {
+  const url = `https://${slug}.bandcamp.com/music`;
+
+  // Slugs come from normalizeForComparison / accent-stripping, so they contain
+  // only [a-z0-9-] and cannot alter the host. Validated anyway as defence in depth.
+  if (!isUrlHostnameAllowed(url)) return null;
+
+  const controller = new AbortController();
+  const timeout = setTimeout(() => controller.abort(), timeoutMs);
+  let response: Response;
+  try {
+    response = await globalThis.fetch(url, {
+      signal: controller.signal,
+      headers: { 'User-Agent': 'Unstream/1.0 (+https://unstream.stream)' },
+    });
+  } finally {
+    clearTimeout(timeout);
+  }
+
+  // No such account. This is a real answer, not a failure.
+  if (response.status === 404) return null;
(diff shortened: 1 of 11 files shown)


api/search/bandcamp-probe.ts
```diff
@@ -0,0 +1,243 @@
...
+      artistUrl: `https://${slug}.bandcamp.com`,
+      bandName: identity.name,
+      bandId: identity.id,
+      albumCount: counts.albums,
+      trackCount: counts.tracks,
+      matchedSlug: slug,
+    };
+  }
+
+  if (sawUndecided && fallback.verdict === 'absent') {
```
brandonlucasgreen, line 205: **[regression — confirmed]** This final check only promotes the result to `undecided` when `fallback.verdict === 'absent'`. It does not implement the invariant stated three lines above ("`undecided` outranks both [rejections]") once a rejection has already been recorded.

Concrete failure: candidate 1 is an unrelated squatter (`rejected_name`, so `fallback` becomes `rejected_name`), candidate 2 — the artist's real slug — hits a bot challenge (`undecided`, `sawUndecided = true`), candidate 3 404s. The loop ends with `fallback.verdict === 'rejected_name'` (not `'absent'`), so this function returns `rejected_name` instead of `undecided`.

`findBandcampArtistUrl` then caches that as a real negative (`artist_url: null`) for up to `NEGATIVE_PROBE_TTL_DAYS` (30 days) — exactly the "transient failure written as a cached negative" bug this PR's description says was fixed "belt and braces." A real artist temporarily blocked by Bandcamp's bot challenge on their actual slug can be miscached as a name-mismatch for a month.

Suggested fix: track `sawUndecided` independently of `fallback` and check it first —
```ts
if (sawUndecided) {
  return { verdict: 'undecided', artistUrl: null, ...empty };
}
return fallback;
```
(or, if a confirmed rejection should still win over a later undecided, that needs to be stated explicitly rather than falling out of `fallback.verdict === 'absent'` incidentally.)

@brandonlucasgreen

---

api/search/enrichment.ts
```diff
@@ -615,7 +538,7 @@ export async function enrichLocationFallback(query: string): Promise<ArtistLocat
   const mirloSlug = query.toLowerCase().replace(/\s+/g, '');
   const FALLBACK_TIMEOUT = 3000;
 
-  const bandcampUrl = await searchBandcampForArtistUrl(query, FALLBACK_TIMEOUT);
+  const bandcampUrl = await findBandcampArtistUrl(query);
```
brandonlucasgreen: **[regression — confirmed]** `FALLBACK_TIMEOUT = 3000` (line 539) is still passed to `fetchBandcampLocation` and `fetchMirloLocation` below, but `findBandcampArtistUrl(query)` here has no way to receive it — it uses `probeBandcampArtist`'s own internal `DEFAULT_BUDGET_MS = 5000`.

Net effect: this fallback path's worst-case latency silently grew from ~3s (old `searchBandcampForArtistUrl(query, FALLBACK_TIMEOUT)`) to ~5s for the probe alone, plus up to another 3s for the subsequent `fetchBandcampLocation` call if a URL is found — up to ~8s total where the constant's name implies a 3s cap. Since this runs in the Phase 2 MusicBrainz enrichment path (background, but still function-time-bounded), worth confirming this doesn't push any caller closer to a Netlify function timeout, and either thread `FALLBACK_TIMEOUT` through to `probeBandcampArtist`'s `budgetMs` or rename/adjust the constant so it reflects reality.

@brandonlucasgreen

---

api/search/bandcamp-probe.ts
brandonlucasgreen, line 1: **[test-coverage — confirmed]** The new unit tests cover the pure parsing helpers (`parseBandcampBandIdentity`, `parseBandcampReleaseCounts`, `bandcampSlugCandidates`) well, but nothing exercises `probeBandcampArtist`'s orchestration state machine (budget exhaustion, fallback-priority ordering, the undecided-vs-rejection interaction flagged in the other comment) or `getBandcampProbe`/`putBandcampProbe`'s cache read/TTL logic in `api/functions/db.ts`. That state machine is where the actual risk in this PR lives — the docs/spec table of live verdicts is good evidence the happy paths work, but doesn't cover the interleaved-outcome cases (e.g. one candidate rejected, another undecided) that a mocked-fetch unit test could exercise deterministically and that live testing is unlikely to hit reliably.

@brandonlucasgreen

---

api/search/bandcamp-probe.ts
```diff
@@ -0,0 +1,243 @@
...
+      signal: controller.signal,
+      headers: { 'User-Agent': 'Unstream/1.0 (+https://unstream.stream)' },
+    });
+  } finally {
+    clearTimeout(timeout);
+  }
+
+  // No such account. This is a real answer, not a failure.
+  if (response.status === 404) return null;
+  if (!response.ok) return { verdict: 'undecided', counts: { albums: 0, tracks: 0 } };
```
brandonlucasgreen, line 92: **[SSRF — needs-manual-verification]** `isUrlHostnameAllowed` validates the request URL, but the subsequent `globalThis.fetch(url, ...)` doesn't set `redirect: 'manual'`, so a redirect response from `<slug>.bandcamp.com` would be followed without re-validating the final hostname against the allowlist. Slugs themselves can't be steered off `*.bandcamp.com` (charset is `[a-z0-9-]` only, matching the comment above), so this isn't attacker-controlled via the query — the residual question is whether Bandcamp lets an artist/label configure a redirect response on their own subdomain, which would let them redirect this server-side fetch elsewhere (e.g. an internal address) using an account whose slug happens to match a searched name. This mirrors the existing `fetchBandcampLocation` in this same file (unchanged, same gap), so it's not a new pattern, but flagging since it's now invoked automatically as part of every search rather than only the old (currently broken) `searchBandcampForArtistUrl` path. Worth a quick check of whether Bandcamp exposes any user-controllable redirect/webhook config for subdomains before writing this off as unreachable.

@brandonlucasgreen

---

brandonlucasgreen (commented):
Staff-eng security/regression pass on the Layer A Bandcamp slug-probe work (UNS-152). Nice, well-documented change overall — the squatter/name-verification design and negative-caching TTL are sound, and RLS-with-no-policies on the new table is exactly right for a server-only cache.

**Security** (auth/authz, input validation, secrets, injection, unsafe defaults, data exposure): no issues found. Slugs are derived through `normalizeForComparison`/accent-stripping so they can't carry anything outside `[a-z0-9-]`, Supabase access goes through parameterized calls, no secrets touched, and the new table's RLS (service-role only, no anon/authenticated policies) is correct. One defense-in-depth note on redirect handling below, but it isn't new to this PR and may not be reachable at all — flagged as needs-manual-verification, not a blocker by itself.

**Regression**: found one real logic bug and one real timeout regression, both inline. The first one (in `bandcamp-probe.ts`) directly undermines the "never cache an undecided outcome as a negative" guarantee this PR's own description calls out as the whole point of the `undecided` verdict — recommend treating it as blocking before merge even though GitHub won't let this review carry a formal "request changes" on your own PR.

Also flagging a test-coverage gap in the orchestration logic (`probeBandcampArtist`) where that bug lives — the new tests are solid for the parsing helpers but don't reach the state machine.

@brandonlucasgreen

---

PR #320, opened 2026-07-26
Surface Bandcamp probe results in search, and cut enrichment latency

brandonlucasgreen:
Two related things: the probe from #319 was finding Bandcamp URLs and throwing them away, and the enrichment path had a worst case well over its function ceiling.

## 1. The probe's results never reached users

#319 wired the probe only into the location path. Both callers did this:

```ts
const searchedUrl = await findBandcampArtistUrl(artist.name, 2500);
return searchedUrl ? fetchBandcampLocation(searchedUrl) : null;   // URL discarded
```

Meanwhile Phase 1's `searchBandcamp` still returned `[]` — dead code the compiler had been flagging as unused (`TS6133`). So the only Bandcamp link a user saw came from a MusicBrainz relation, or a generic `bandcamp.com/search?q=` fallback.

Measured over 100 artists:

| Source | Finds a Bandcamp URL |
|---|---|
| MusicBrainz relations | 9% |
| **The probe** | **19%** |
| Either | 21% |

So ~12 points of coverage were being computed and discarded.

`searchBandcamp` now calls the probe and joins the Phase 1 fan-out. Verified end to end:

```
"Boy Harsher"  -> https://boyharsher.bandcamp.com/   [DIRECT]
"Radiohead"    -> https://radiohead.bandcamp.com/    [DIRECT]
"Beyonce"      -> none                               (empty-catalogue gate rejects the squatter)
```

Also diffed `Beyonce`'s platform list against main to confirm no search-fallback link was lost: identical before and after.

## 2. Location now rides along, saving a request

The probe fetches `<slug>.bandcamp.com/music`, and that response already carries the location in a `class="location"` element — **89%** of the time (16/18 long-tail artists; both misses have no location in Bandcamp's own discover API either). Returning it removes a second fetch of a page already in hand.

Migration `026` caches it, so a cache hit yields URL *and* location together.

`findBandcampArtist` is the new cached entry point; `findBandcampArtistUrl` remains as a thin wrapper. Phase 1 uses the cached form deliberately — it's the hot path, and a repeat query should cost one DB read rather than three HTTP requests.

## 3. Latency

`/api/search/musicbrainz` ran **three sequential network phases**, worst case ~19.6s in a 10s function. Measured real latency was 3.9s, so this was tail risk — but a timeout there produces the same silent all-nulls 500 that hid #318's crash for four months.

- **Merged two MusicBrainz calls.** The `url-rels` lookup and the release-group browse were separate requests with a mandatory 1100ms gap. `inc=url-rels+release-groups` returns both. Cold-cache A/B on fresh artists: **2.11s → 0.51s**. Verified the combined response carries `first-release-date` on every release group (25/25), which is all the pre-2005 Hoopla/Freegal check reads.
- **Dropped the third sequential fetch** in the location lambda (see §2).
- **Back off on HTTP 429.** Bandcamp *does* rate-limit — I triggered it during measurement. A 429 was already classified as `undecided` (never cached as a negative), but the loop then tried the remaining candidates anyway. It now abandons the round and reports to Sentry. This matters more for Layer B's sweep than for search.
- **`fetchBandcampLocation` default 4000 → 2500ms** (measured ~670ms).

## Checks

- `tsc -b` clean
- Explicit strict `tsc` over every touched file + import graph: **zero new errors**, and it clears **3 pre-existing** ones (31 → 28), including the dead `searchBandcamp`. Needed because `api/tsconfig.json` still only covers six files.
- api tests **69/69** · unit tests **384/384** (6 new, covering `parseBandcampPageLocation`)

## Note

Migration `20260726180000_bandcamp-probe-location.sql` adds one nullable column and auto-deploys on merge. The `Supabase Preview` check will exercise it on the preview branch first.

### DIFF
--- a/api/functions/db.ts
+++ b/api/functions/db.ts
@@ -208,6 +208,8 @@ export interface BandcampProbeRow {
   track_count: number;
   matched_slug: string | null;
   verdict: 'accepted' | 'absent' | 'rejected_empty' | 'rejected_name' | 'pending_review';
+  /** Raw location string from the probed /music page, e.g. "Oxford, UK". */
+  location: string | null;
   checked_at: string;
 }
 
@@ -230,7 +232,7 @@ export async function getBandcampProbe(queryNorm: string): Promise<BandcampProbe
   try {
     const { data, error } = await client
       .from('bandcamp_slug_probes')
-      .select('query_norm, artist_url, band_name, band_id, album_count, track_count, matched_slug, verdict, checked_at')
+      .select('query_norm, artist_url, band_name, band_id, album_count, track_count, matched_slug, verdict, location, checked_at')
       .eq('query_norm', queryNorm)
       .maybeSingle();
 
--- a/api/functions/search-musicbrainz.ts
+++ b/api/functions/search-musicbrainz.ts
@@ -32,7 +32,7 @@ import {
   fetchMirloLocation,
   enrichLocationFallback,
 } from '../search/enrichment';
-import { findBandcampArtistUrl } from '../search/bandcamp-probe';
+import { findBandcampArtist } from '../search/bandcamp-probe';
 
 // Cache TTL for MusicBrainz lookups (30 minutes)
 const MUSICBRAINZ_CACHE_TTL = 30 * 60;
@@ -114,8 +114,11 @@ async function searchMusicBrainz(query: string): Promise<MusicBrainzSearchRespon
     // Wait 1.1 seconds to respect MusicBrainz rate limit (1 req/sec)
     await new Promise(resolve => setTimeout(resolve, 1100));
 
-    // Fetch artist details with URL relations
-    const artistUrl = `https://musicbrainz.org/ws/2/artist/${artist.id}?inc=url-rels&fmt=json`;
+    // Fetch URL relations AND release groups in a single lookup. These used to be two
+    // requests with a mandatory 1.1s gap between them; `inc=url-rels+release-groups`
+    // returns both, saving a round trip plus the gap (~1.6s measured). The release
+    // groups carry `first-release-date`, which is all the pre-2005 check needs.
+    const artistUrl = `https://musicbrainz.org/ws/2/artist/${artist.id}?inc=url-rels+release-groups&fmt=json`;
 
     const artistResponse = await globalThis.fetch(artistUrl, {
       headers: {
@@ -132,6 +135,7 @@ async function searchMusicBrainz(query: string): Promise<MusicBrainzSearchRespon
     const platformUrls: string[] = [];
 
     let mbLocation: ArtistLocation | undefined;
+    let hasPre2005Release = false;
 
     if (artistResponse.ok) {
       const artistData = await artistResponse.json() as {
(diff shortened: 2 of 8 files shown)


brandonlucasgreen (commented):

## Security

No blocking issues found.

- **SSRF** — `probeCandidate` fetches `https://${slug}.bandcamp.com/music` through `isUrlHostnameAllowed`, and `slug` is constrained to `[a-z0-9-]` by `bandcampSlugCandidates`/`normalizeForComparison` (`api/functions/search-utils.ts`), so the host can't be redirected. **confirmed safe.**
- **Injection** — the new `location` column is written/read through the Supabase query builder (`.upsert(...)`, `.select(...)`) in `db.ts`, no raw SQL concatenation. **confirmed safe.**
- **Data exposure / XSS** — `location` is scraped via a bounded regex (`parseBandcampPageLocation`, 120-char cap, no nested-quantifier ReDoS risk) and only ever reaches the client through `parseLocationString` → `ArtistLocation.city/country`, rendered as a plain template string in `RichArtistProfile.tsx` (React-escaped, no `dangerouslySetInnerHTML`). **confirmed safe.**
- **Rate-limit handling** — the new 429/503 → `rateLimited` → abandon-round path plus Sentry dedup (`uns152:bandcamp-rate-limited`) is a reasonable addition, no issues.
- Auth/authorization and secret handling are untouched by this diff — N/A.

## Regression

**Confirmed — the latency fix wasn't applied to the code path that actually runs on every search.**

The PR's "Latency" section (merging `url-rels`+`release-groups` into one MusicBrainz call, dropping the sequential 1.1s-gapped release-group fetch) is applied to `api/functions/search-musicbrainz.ts`. But there's a second, separately-maintained `searchMusicBrainz` in `api/functions/search-sources.ts:498-729` — same name, same shape, different file — that is invoked from `searchAllPlatforms()` (`search-sources.ts:1709`), i.e. **Phase 1**, on every single search request. That copy still has the exact pattern the PR describes as the risk: `await delay(1100)` (line 688) followed by a separate `release-group/?artist=...&limit=20` fetch (lines 690-715), on top of the fetches already before it. Per `CLAUDE.md`, Phase 1 is supposed to be the fast ~1-2s path with MusicBrainz enrichment deferred to Phase 2 — this embedded call contradicts that and still carries the worst-case-latency risk the PR is fixing elsewhere. Not introduced by this PR, but it means the PR's stated latency win doesn't cover the hot path, worth confirming that's understood/intentional (or tracked as a follow-up) rather than assumed fixed.

**Plausible — silently tightened timeout on an unverified call site.**

`fetchBandcampLocation`'s default timeout dropped from 4000ms → 2500ms (`api/search/enrichment.ts:469`), justified by "measured ~670ms" for the paths this PR touched. That default is also inherited by the untouched call at `search-sources.ts:724` (`bandcampUrl ? fetchBandcampLocation(bandcampUrl) : ...`), inside the same slow, unpatched `searchMusicBrainz` above. No measurement was presented for that call site specifically, and it's now running with less budget on a function that already has more upstream latency than the ones that were profiled. Low blast radius (location is best-effort/non-fatal on failure), but flagging since it's an untested side effect of a shared default change.

**needs-manual-verification — release-group completeness via `inc=`.**

Swapping the dedicated `release-group/?artist=X&limit=20` browse call for `inc=release-groups` on the artist lookup changes how MusicBrainz paginates the result (lookup `inc=` typically returns the full set rather than a `limit=20` page). The PR's 25/25 sample check is reassuring but doesn't rule out an artist with >20 release groups where set membership near the boundary differs — worth a spot-check against a high-release-count artist if one exists in the fixture set.

Everything else (probe/DB schema change, migration, new location parsing + tests) looks solid — bounded input, idempotent migration, good test coverage on the new parser including the "substring of location" and oversized-match edge cases.

---

PR #321, opened 2026-07-26
Stop caching MusicBrainz failures as "this artist has no links"

brandonlucasgreen:
Found while verifying #320 in production.

## Symptom

`Radiohead` returned `artistName: null` on **four consecutive** requests, while `Aphex Twin`, `Big Thief` and `Explosions in the Sky` all returned full data — and the *identical* code returned Radiohead correctly when run locally.

The only meaningful difference between those two environments is Redis.

## Cause

`cacheGetOrFetch` caches whatever the fetch returns. `searchMusicBrainz` returned its populated all-nulls `emptyResult` when MusicBrainz answered non-2xx:

```ts
if (!response.ok) {
  console.log('MusicBrainz artist search failed:', response.status);
  return emptyResult;      // <- indistinguishable from "no such artist", and cached
}
```

`artistCacheKey` lowercases and trims, so every casing of a query shares one key. **One MusicBrainz hiccup made that artist look link-less for the full 30-minute TTL.**

This is the third instance of the same root pattern today:

| PR | Collapsed "source unavailable" into… |
|---|---|
| #317 | a bot-challenge page → "not on Bandcamp" |
| #318 | Mirlo's unfiltered list → "this artist's location" |
| **this** | a MusicBrainz 503 → "this artist has no links" — *and cached it* |

## Fix

- `searchMusicBrainz` returns `null` **only** for a non-answering upstream (non-2xx, or a thrown network error/timeout). The legitimate no-match paths — no results, low score, name mismatch — keep returning `emptyResult` with fallback location, and stay cacheable.
- `cacheGetOrFetch` gains an optional `shouldCache` predicate. Existing callers are unaffected (omitted predicate means "always cacheable").
- The unavailable response sends `Cache-Control: no-store`, so the CDN doesn't hold it for 5 minutes either, and reports to Sentry deduped at 30 minutes.

## Verification

Simulated an outage and recovery against the real handler:

```
1. MusicBrainz down      -> HTTP 200  Cache-Control: no-store              artist null
2. MusicBrainz recovered -> HTTP 200  Cache-Control: s-maxage=300, swr     artist "Radiohead"
   3 MusicBrainz fetch attempts — the failure was retried, not remembered.
```

`tsc -b` clean · explicit strict `tsc` **zero new errors** (28 baseline, 28 after) · api tests **74/74** (5 new, covering the predicate contract including the re-run-after-failure case) · unit tests **384/384**

### DIFF
--- a/api/functions/search-musicbrainz.ts
+++ b/api/functions/search-musicbrainz.ts
@@ -1,9 +1,10 @@
 // MusicBrainz enrichment server function
 // Imports shared functions from enrichment.ts, keeps Netlify handler here
 
+import { Sentry } from '../lib/sentry';
 import { cacheGetOrFetch, artistCacheKey } from './cache';
 import { persistEnrichment } from './db';
-import { checkRateLimit, getClientIp } from './ratelimit';
+import { checkRateLimit, checkSentryDedup, getClientIp } from './ratelimit';
 import { validateQuery, isUrlHostnameAllowed } from './middleware';
 import { normalizeAccents, normalizeSearchQuery } from './search-utils';
 
@@ -52,8 +53,14 @@ interface MusicBrainzSearchResponse {
   location?: ArtistLocation;
 }
 
-// Search MusicBrainz for artist info including official website, Discogs, social links, and release history
-async function searchMusicBrainz(query: string): Promise<MusicBrainzSearchResponse> {
+// Search MusicBrainz for artist info including official website, Discogs, social links, and release history.
+//
+// Returns null when MusicBrainz itself did not answer (non-2xx). That is deliberately
+// distinct from "MusicBrainz has no such artist", which returns a populated empty
+// result: the caller must not cache the former. Caching an upstream failure turned a
+// single MusicBrainz hiccup into 30 minutes of "this artist has no links" for that
+// query — the same mistake as reading a bot-challenge page as "not on Bandcamp".
+async function searchMusicBrainz(query: string): Promise<MusicBrainzSearchResponse | null> {
   const emptyResult: MusicBrainzSearchResponse = {
     query,
     artistName: null,
@@ -79,8 +86,9 @@ async function searchMusicBrainz(query: string): Promise<MusicBrainzSearchRespon
     });
 
     if (!response.ok) {
+      // Upstream did not answer — signal "unknown" so this is not cached.
       console.log('MusicBrainz artist search failed:', response.status);
-      return emptyResult;
+      return null;
     }
 
     const data = await response.json() as { artists?: { id: string; name: string; score: number }[] };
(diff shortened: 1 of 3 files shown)


api/functions/search-musicbrainz.ts
```diff
@@ -359,16 +386,39 @@ export async function handler(event: { queryStringParameters?: Record<string, st
...
+      }
+      return {
+        statusCode: 200,
+        headers: {
+          'Content-Type': 'application/json',
+          'Access-Control-Allow-Origin': '*',
+          // Do not let the CDN cache an unavailable upstream either.
+          'Cache-Control': 'no-store',
+        },
+        body: JSON.stringify(unavailableResult(query)),
```
brandonlucasgreen, line 418: **Regression — confirmed: `query` field drifts on the unavailable path**

Every other response shape reflects the *normalized* query: `searchMusicBrainz(normalizedQuery)` sets `.query = query` internally (used for the success case and all three `emptyResult` no-match branches), so callers always see the accent-normalized form back.

This new branch instead calls `unavailableResult(query)` with the outer handler's pre-normalization `query` (from `queryResult.query`, line 381), not `normalizedQuery`. So a request for an accented name (e.g. `"Tanerélle"`) gets `query: "Tanerélle"` back on a transient MusicBrainz outage, but `query: "Tanerelle"` on every other path (success, no-match, low-confidence, name-mismatch). Low impact (cosmetic API-contract drift, not a correctness bug in the enrichment data itself), but it's a real inconsistency a consumer diffing responses could trip on.

Suggest: `unavailableResult(normalizedQuery)` to match every other branch.

@brandonlucasgreen

---

brandonlucasgreen (commented):
Reviewed for security and regressions. @brandonlucasgreen

**Security** — no findings.
- Auth/authz: unaffected, no changes.
- Input validation: `query` still goes through the existing `validateQuery` (length-capped, control chars stripped) before reaching this code; the raw value used in the new "unavailable" response is already-validated, not new untrusted surface.
- Secret handling: no secrets touched.
- Injection: no new fetch/query construction; the MusicBrainz URL still uses `encodeURIComponent`.
- Unsafe defaults: `shouldCache` is optional and defaults to "always cache," preserving existing behavior for the other `cacheGetOrFetch` call sites in `search-sources.ts` — confirmed by grep, none of them pass the new param and none regress.
- Data exposure: the new Sentry capture logs the search query, same as multiple pre-existing `console.log` calls in this file — not a new exposure. `checkSentryDedup` is a pre-existing helper (already on `main`), reused correctly with a fixed dedup key so an outage doesn't flood Sentry.

**Regression** — one confirmed finding, left as an inline comment on `search-musicbrainz.ts`:
- The new "MusicBrainz unavailable" response builds its body from the pre-normalization `query` instead of `normalizedQuery`, unlike every other response branch in this function. Minor API-contract drift (cosmetic), not a data-correctness issue.

Everything else checks out: the `shouldCache` predicate is exercised by new unit tests (including the recovery-after-failure case that mirrors the reported bug), the `!response.ok` early-return still happens before the 1.1s rate-limit wait / second fetch (no behavior change to MusicBrainz call volume), and the CDN `no-store` header correctly stops the outage response from being cached for 5 minutes the way it used to be.

---

PR #322, opened 2026-07-26
Fix Qobuz link and artist photo vanishing from search results

brandonlucasgreen:
Regression from #320, reported in production for **Anna von Hausswolff**: her Bandcamp link resolves correctly now, but the Qobuz link and the artist photo both disappeared.

## One cause, two symptoms

`fetchReleasesForDisambiguation` runs *every* release fetch inside a single fixed race:

```ts
await Promise.race([
  Promise.allSettled(promises),
  new Promise(resolve => setTimeout(resolve, 4000)),
]);
```

Before #320, Phase 1 returned no Bandcamp results, so that 4s budget was spent on Qobuz. #320 started returning them — adding **two Bandcamp requests per result** — which crowded the Qobuz fetches out. Qobuz then arrived with no release data, `removeDeadQobuzLinks` dropped it as a dead link, and because **the artist image comes from the Qobuz match**, the photo went with it.

Confirmed against production before fixing:

```
result: name='Anna von Hausswolff'  imageUrl=None
platforms: bandcamp, ampwall, kofi, buymeacoffee, officialsite, discogs, instagram, youtube, facebook
           ^ no qobuz
```

And ruled out the obvious alternative — Qobuz *does* list her, and the names normalize identically on both sides, so matching was never the problem:

```
normalizeForComparison('Anna von Hausswolff') = annavonhausswolff
probe data-band name                          = "Anna von Hausswolff" -> annavonhausswolff
parseQobuzSearchResults                       = [["annavonhausswolff",
                                                  ".../interpreter/anna-von-hausswolff/1078840"]]
```

## Fix

The probe already fetches `<slug>.bandcamp.com/music` — **the same page `getBandcampReleaseTitles` was refetching**. So the titles now come back with the probe and are cached (migration 027), and disambiguation skips the Bandcamp title fetch when they're already there. That halves Bandcamp's share of the budget and leaves room for Qobuz.

`PlatformResult` gains `allReleaseTitles`, and `aggregateResults` carries it through — it previously copied only `sourceId` and `url`. I deliberately did **not** also start copying `latestRelease` there, since that would change existing merge behaviour.

## Two side benefits

Bandcamp now arrives *with* release titles, so:
- `crossPlatformReleaseComparison` can actually compare catalogues instead of `continue`-ing on empty titles
- `splitSuspiciousPlatforms` no longer treats Bandcamp as a release-less "suspicious" platform

Both were silently inert for Bandcamp since #320.

## Verification

Probe returns titles from the page it already read, no extra requests:

```
Anna von Hausswolff  accepted  10 albums  "Gothenburg, Sweden"  13 titles
Boy Harsher          accepted  15 albums  "Northampton, MA"     16 titles
```

`tsc -b` clean · explicit strict `tsc` **zero new errors** (28 baseline, 28 after — it caught the `PlatformResult` field error on the first attempt) · api tests **74/74** · unit tests **384/384**

Will confirm on the deploy preview that Qobuz and the photo return before merging.

### DIFF
--- a/api/functions/db.ts
+++ b/api/functions/db.ts
@@ -210,6 +210,8 @@ export interface BandcampProbeRow {
   verdict: 'accepted' | 'absent' | 'rejected_empty' | 'rejected_name' | 'pending_review';
   /** Raw location string from the probed /music page, e.g. "Oxford, UK". */
   location: string | null;
+  /** Normalized release titles from the probed /music page. */
+  release_titles: string[] | null;
   checked_at: string;
 }
 
@@ -232,7 +234,7 @@ export async function getBandcampProbe(queryNorm: string): Promise<BandcampProbe
   try {
     const { data, error } = await client
       .from('bandcamp_slug_probes')
-      .select('query_norm, artist_url, band_name, band_id, album_count, track_count, matched_slug, verdict, location, checked_at')
+      .select('query_norm, artist_url, band_name, band_id, album_count, track_count, matched_slug, verdict, location, release_titles, checked_at')
       .eq('query_norm', queryNorm)
       .maybeSingle();
 
--- a/api/functions/search-sources.ts
+++ b/api/functions/search-sources.ts
@@ -88,6 +88,12 @@ async function searchBandcamp(query: string): Promise<PlatformResult[]> {
     name: match.bandName ?? query,
     type: 'artist',
     url: match.url,
+    // Carried from the /music page the probe already read. This must be populated:
+    // fetchReleasesForDisambiguation shares one 4s budget across all its release
+    // fetches, so a Bandcamp platform with no titles forces two more requests into
+    // that budget and starves the Qobuz ones — which drops the Qobuz link, and with
+    // it the artist image.
+    allReleaseTitles: match.releaseTitles.length > 0 ? match.releaseTitles : undefined,
   }];
 }
 
@@ -1333,7 +1339,11 @@ async function fetchReleasesForDisambiguation(aggregated: AggregatedResult[]): P
     const bc = result.platforms.find(p => p.sourceId === 'bandcamp');
     if (bc) {
       promises.push(getBandcampLatestRelease(bc.url).then(r => { if (r) bc.latestRelease = r; }));
-      promises.push(getBandcampReleaseTitles(bc.url).then(t => { if (t.length > 0) bc.allReleaseTitles = t; }));
+      // Skip when the probe already supplied titles. All of these fetches share one
+      // 4s race below, so redundant Bandcamp requests directly cost Qobuz its data.
+      if (!bc.allReleaseTitles || bc.allReleaseTitles.length === 0) {
+        promises.push(getBandcampReleaseTitles(bc.url).then(t => { if (t.length > 0) bc.allReleaseTitles = t; }));
+      }
     }
 
     const qz = result.platforms.find(p => p.sourceId === 'qobuz');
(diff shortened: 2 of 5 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Staff-eng pass on this one. Scope is tight (probe cache gains a column, disambiguation skips a redundant fetch when it's populated) and the root-cause writeup checks out against the code. Summary below; nothing here blocks merge.

## Security

- **Auth/authorization** — no change to any authenticated/admin path. `needs-manual-verification`: n/a, not touched.
- **Input validation** — `parseBandcampReleaseTitles` (search-parsers.ts:220) parses Bandcamp's own HTML via the existing `node-html-parser`/`textContent` pattern already used for titles/location elsewhere in this file, then runs the result through `normalizeForComparison`. No new raw-HTML or string-concat parsing path introduced. **confirmed** safe, consistent with existing parsers.
- **Secret handling** — none touched.
- **Injection risk** — the new `release_titles` column is read via Supabase's parameterized `.select()`/`.eq()` client (db.ts), no interpolated SQL. **confirmed** no injection surface.
- **Unsafe defaults** — `bandcamp_slug_probes` already has RLS enabled with zero policies (service-role only, anon/authenticated get nothing per the migration-025 comment); migration 027 only adds a nullable `TEXT[]` column, doesn't touch RLS. **confirmed** no new exposure.
- **Data exposure** — release titles are public Bandcamp catalog metadata (album/track names), same sensitivity class as the `location`/`band_name` fields already cached in this table. No PII. **confirmed** non-issue.

## Regression

- **Changed assumptions / edge cases** — `match.releaseTitles` is typed as required `string[]` on `BandcampArtistMatch` (bandcamp-probe.ts:262) and both construction sites (`cached.release_titles ?? []`, `result.releaseTitles ?? []`) default to `[]`, so `search-sources.ts:96`'s `match.releaseTitles.length > 0` never sees `undefined`. **confirmed** safe null handling.
- **State transitions (cache staleness)** — the probe cache treats `verdict: 'accepted'` rows as stable and only re-probes negatives on a schedule (per migration 025's own comment: "positives are stable"). Once `release_titles` is populated for an artist, it's frozen at first-probe time — no path re-fetches it when the artist releases new albums. That's fine for the dead-link check (`removeDeadQobuzLinks`, any non-empty list works), but `crossPlatformReleaseComparison` and `splitSuspiciousPlatforms` use these titles for match-scoring against Qobuz's live catalog, so a long-lived cached artist could see its catalog-overlap score drift down over time as both sides diverge, since Bandcamp is what dead links to fresh Qobuz release. `needs-manual-verification`: whether this is an acceptable tradeoff (same staleness model as the pre-existing `band_name`/`location` fields) or worth a periodic re-probe for accepted rows too — worth a one-line call-out either way since it's now feeding disambiguation logic, not just display data.
- **API contract drift** — `PlatformResult.allReleaseTitles` and the `platforms[].allReleaseTitles` field are additive/optional; `aggregateResults`' existing-platform-merge branch (search-utils.ts:420) doesn't backfill `allReleaseTitles` onto an already-existing platform entry, but that matches the pre-existing behavior for other fields in that branch (e.g. a second `url` for the same platform is likewise dropped, not merged) — not a new gap. **plausible**, low severity, pre-existing pattern.
- **Semantic-revert bot flag** — the automated check flagged line deletions in `db.ts`/`bandcamp-probe.ts` matching #320's additions. Reviewed the diff: these are reformats (single-line object literals expanded to multi-line to add `releaseTitles`/`release_titles`), and the `location`/`band_name` data from #320 is preserved verbatim in the new object shape. **confirmed false positive** — not an actual revert of #320's work.

Nice, minimal fix for the actual bug (shared 4s budget getting starved by redundant Bandcamp requests) — the probe-already-has-the-page insight is the right lever. Only outstanding question is the cache-staleness tradeoff above, which is a design call rather than a blocker.

---

PR #323, opened 2026-07-26
Stop caching failed platform searches as "artist not on this platform"

brandonlucasgreen:
#321 fixed this for MusicBrainz. It's the same bug in **four more places** — and one of them is what actually broke Anna von Hausswolff's card.

## The bug

`searchQobuz`, `searchPatreon`, `searchBeatport` and `searchEven` each do this:

```ts
const { data } = await cacheGetOrFetch(cacheKey, async () => {
  const results = [];
  try {
    const response = await fetchWithTimeout(url, {...}, 5000);
    if (!response.ok) return results;      // <- empty, and about to be cached
    ...
  } catch (error) { /* logs, falls through */ }
  return results;                          // <- same
}, PLATFORM_CACHE_TTL);
```

An empty array from a *failure* is cached exactly like a genuine "no matches". One hiccup drops that platform from results for the whole TTL.

## How this was diagnosed

| Environment | Result |
|---|---|
| Local (no Redis) | Qobuz present, **19 release titles**, artist image resolved |
| Deploy preview (shares prod Redis) | no Qobuz, no image — reproduced exactly |
| Qobuz search URL directly | **HTTP 200 in 0.58s**, and it does list her |

So the cause was a **cached empty Qobuz search**, not the code path. And because the artist photo comes *from* the Qobuz match (`qobuzMatches` is `Map<string, {url, imageUrl}>`), one cached failure removed both the link and the photo — the two symptoms reported.

Worth noting: **the deploy preview was never a valid control here**, since it shares production's Redis. That's what made this hard to see.

## Fix

Each of the four sets a local `fetchFailed` flag at both bail-out points and passes `() => !fetchFailed` as `shouldCache`.

Deliberately kept to a local flag rather than changing four return types and their consumers — a shared-shape change is exactly what caused the #322 regression.

`searchAmpwall` is left alone on purpose: it's still a stub with no outbound request, so it cannot fail and its empty result is a real answer. Noted in a comment so the omission doesn't read as an oversight.

## One thing worth flagging about process

My first pass at this used a regex sweep and inserted the flag into **six unrelated catch blocks**. The explicit strict `tsc` baseline caught all six as `TS2304`.

`npm run build` would **not** have caught them — `api/tsconfig.json` still covers only six files. That gap has now hidden a four-month outage (#318) and nearly shipped this. It's worth its own PR.

## Checks

`tsc -b` clean · explicit strict `tsc` **zero new errors** (28 baseline, 28 after) · api tests **74/74** · unit tests **384/384**

Verified degradation: with Qobuz forced to 503 the search still returns HTTP 200 with the remaining nine platforms and does not crash.

### DIFF
--- a/api/functions/search-sources.ts
+++ b/api/functions/search-sources.ts
@@ -989,6 +989,9 @@ async function searchJamcoop(query: string): Promise<Map<string, string>> {
 
 async function searchPatreon(query: string): Promise<Map<string, string>> {
   const cacheKey = artistCacheKey('patreon', query);
+  // Set when the upstream did not answer. A failure must not be cached as
+  // "this artist isn't on patreon" -- see the shouldCache predicate below.
+  let fetchFailed = false;
 
   const { data } = await cacheGetOrFetch<[string, string][]>(
     cacheKey,
@@ -1005,7 +1008,7 @@ async function searchPatreon(query: string): Promise<Map<string, string>> {
           },
         }, 5000);
 
-        if (!response.ok) return results;
+        if (!response.ok) { fetchFailed = true; return results; }
 
         const data = await response.json() as {
           data?: {
@@ -1044,6 +1047,7 @@ async function searchPatreon(query: string): Promise<Map<string, string>> {
         }
       } catch (error: unknown) {
         const err = error as { name?: string; message?: string };
+        fetchFailed = true;
         if (err.name !== 'AbortError') {
           console.error('Patreon search error:', err.message);
         }
@@ -1051,7 +1055,9 @@ async function searchPatreon(query: string): Promise<Map<string, string>> {
 
       return results;
     },
-    PLATFORM_CACHE_TTL
+    PLATFORM_CACHE_TTL,
+    // Never cache the empty result of a failed fetch.
+    () => !fetchFailed,
   );
 
   return new Map(data);
@@ -1063,7 +1069,8 @@ const AMPWALL_CACHE_TTL = 30 * 60;
 
 async function searchAmpwall(query: string): Promise<Map<string, string>> {
   const cacheKey = artistCacheKey('ampwall', query);
-
+  // No shouldCache guard here: this is still a stub with no outbound request, so it
+  // cannot fail and its empty result is a genuine answer rather than a hidden error.
   const { data, cached } = await cacheGetOrFetch<[string, string][]>(
     cacheKey,
     async () => {
@@ -1104,6 +1111,9 @@ async function searchAmpwall(query: string): Promise<Map<string, string>> {
 
 async function searchQobuz(query: string): Promise<Map<string, { url: string; imageUrl?: string }>> {
   const cacheKey = artistCacheKey('qobuz', query);
+  // Set when the upstream did not answer. A failure must not be cached as
+  // "this artist isn't on qobuz" -- see the shouldCache predicate below.
+  let fetchFailed = false;
 
   const { data } = await cacheGetOrFetch<[string, { url: string; imageUrl?: string }][]>(
     cacheKey,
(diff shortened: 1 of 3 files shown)


api/functions/search-sources.ts
brandonlucasgreen, line 1: **Regression — plausible: `searchBeatport` still has two un-guarded empty-return paths that reproduce the exact bug this PR fixes**

The `fetchFailed` flag only gets set at the `!response.ok` check and in the `catch` block. But `searchBeatport` (lines ~1229-1243, unchanged by this diff) has two more early `return results;` points inside the *same try block*, both reachable after a `200 OK` response, that don't set it:

```ts
if (!response.ok) { fetchFailed = true; return results; }
const html = await response.text();
const root = parse(html);
const scriptEl = root.querySelector('script#__NEXT_DATA__');
if (!scriptEl) return results;          // <- not flagged
...
if (!artists) return results;           // <- not flagged
```

`__NEXT_DATA__` is the Next.js app-shell hydration payload — a normal Beatport search page (even a zero-match one) should always include it. If Beatport serves a bot-check/interstitial page, a redirect, or changes markup, the response is `200 OK` with no `__NEXT_DATA__`, and this silently caches "no such artist on Beatport" for the full 30-minute `PLATFORM_CACHE_TTL` — precisely the failure mode described in this PR's own writeup, just one branch lower in the same function. Label: **plausible** (strong structural evidence from the code; would need a live repro against beatport.com to confirm).

The `if (!artists) return results;` branch a few lines below is the same shape but weaker evidence either way — it's plausible this is a legitimate "no matches" case rather than a schema-drift failure. Label: **needs-manual-verification**.

Suggest applying the same `fetchFailed = true` treatment to both bail-outs for consistency with the rest of this PR's fix.

---

**Regression — plausible, low severity: `searchEven`'s missing-`ALGOLIA_API_KEY` branch is also un-guarded**

```ts
if (!algoliaApiKey) {
  console.warn('[EVEN] Missing ALGOLIA_API_KEY env var, skipping Even search');
  return results;             // <- not flagged, still cached as "not found"
}
```

Same class of bug (env issue treated as "no results"), pre-existing before this PR rather than introduced by it, and lower-impact since a missing key is deterministic across the TTL window rather than transient — but worth the same one-line fix for consistency if this PR is meant to close out this bug class.

---

**Security:** nothing to flag. This diff only introduces a local `fetchFailed` boolean and a `shouldCache` predicate passed into the existing `cacheGetOrFetch` helper — no new auth/authz, input-validation, secret-handling, injection, or data-exposure surface. `query` continues to flow into URLs via `encodeURIComponent` exactly as before, and the Algolia key handling is unchanged.

**Other regression angles checked, no issues found:** `cacheGetOrFetch`'s `shouldCache` param is optional and additive — no signature break for existing call sites (e.g. `searchAmpwall` still calls it with 3 args). No race condition on the `fetchFailed` closure: `cacheGetOrFetch` fully `await`s `fetchFn()` (including its catch block) before evaluating `shouldCache`, so the flag is always settled by the time it's read.

@brandonlucasgreen

---

brandonlucasgreen (commented):
@brandonlucasgreen Security review: no confirmed or plausible issues — this diff only adds a local failure flag and a `shouldCache` predicate to an existing cache helper, no new auth/input-validation/secret/injection/data-exposure surface.

Regression review: the fix is incomplete. `searchBeatport` has two more `200 OK` bail-out paths (missing `__NEXT_DATA__`, missing `artists` array) that still get cached as "artist not found" for the full TTL — the same bug class this PR is fixing, just further down the same function. `searchEven`'s missing-`ALGOLIA_API_KEY` bail-out has the same gap (pre-existing, lower severity). Details and suggested fix inline.

---

PR #324, opened 2026-07-26
Stop deleting Qobuz links when the release lookup merely timed out

brandonlucasgreen:
This is the actual cause of the Anna von Hausswolff card — the fifth and last instance of one pattern found today.

## The bug

`removeDeadQobuzLinks` deletes any Qobuz link with no release data behind it. The reasoning is sound: a name-only match with no catalogue is probably a different artist.

Except the release data is fetched inside **one fixed 4s race shared by every platform**:

```ts
await Promise.race([
  Promise.allSettled(promises),
  new Promise(resolve => setTimeout(resolve, 4000)),
]);
```

So "no release data" also covers **"Qobuz didn't answer in time"**. Absence of data was being read as evidence of absence.

That explains the observed pattern precisely: Radiohead kept its Qobuz link and image (popular query, warm at Qobuz's edge, wins the race), while Boy Harsher, Aphex Twin, Big Thief and Anna von Hausswolff lost theirs. And because **the artist image comes from the Qobuz match**, the photo disappeared with the link.

## Why my earlier attempts missed it

- **#322** reduced contention for that budget (real improvement, wrong target) — the deletion still fired.
- **#323** stopped caching failed *searches* (real bug, also not this) — the Qobuz search was succeeding all along; it was the *release lookup* losing the race.

## Fix

`fetchReleasesForDisambiguation` now returns the platform entries whose release lookups actually settled. `removeDeadQobuzLinks` deletes only when the lookup **finished and genuinely found nothing**.

A Qobuz platform counts as checked only once **both** its lookups settle — one empty answer isn't enough to conclude anything.

The set is an optional parameter, so previous semantics are preserved for any caller that omits it, and the existing tests still assert that.

## Verification

Made Qobuz's interpreter pages take 6s so they lose the race:

```
[Cleanup] Keeping Qobuz link for "Anna von Hausswolff" — release lookup did not finish
platforms: ["qobuz","bandcamp","ampwall","kofi",...]   qobuz kept: YES   imageUrl: present
```

Under the old code that same run deleted the link and the image.

`tsc -b` clean · explicit strict `tsc` **zero new errors** (28 baseline, 28 after) · api tests **77/77** · unit tests **388/388** (4 new, covering lookup-didn't-finish, lookup-did-finish, has-releases, and the no-set backward-compatible path)

## Deliberate trade-off

When a lookup times out we now keep a Qobuz link that release verification might have rejected — i.e. potentially a name-only match on a different artist.

I think that's the lesser harm: every other platform is already kept on a name match alone, and the old behaviour was demonstrably hiding *correct* links for every artist whose Qobuz pages are slow. But it is a real trade, so worth your call.

### DIFF
--- a/api/functions/search-sources.ts
+++ b/api/functions/search-sources.ts
@@ -21,6 +21,7 @@ import {
   createQobuzOnlyResults,
   preferBandcampFeaturedRelease,
   removeDeadQobuzLinks,
+  type AggregatedPlatform,
   crossPlatformReleaseComparison,
   deduplicateQobuzUrls,
   createOrphanedQobuzStandalones,
@@ -1368,8 +1369,18 @@ async function searchEven(query: string): Promise<Map<string, string>> {
 // Phase 3: Fetch releases & disambiguate
 // ---------------------------------------------------------------------------
 
-async function fetchReleasesForDisambiguation(aggregated: AggregatedResult[]): Promise<void> {
+/**
+ * Fetch release data used for disambiguation, bounded by one shared 4s race.
+ *
+ * Returns the platform entries whose release lookups actually finished. That set is what
+ * lets removeDeadQobuzLinks tell "Qobuz has no releases" apart from "Qobuz didn't answer
+ * in time" — losing this race must not look like an empty catalogue.
+ */
+async function fetchReleasesForDisambiguation(
+  aggregated: AggregatedResult[],
+): Promise<Set<AggregatedPlatform>> {
   const promises: Promise<void>[] = [];
+  const completed = new Set<AggregatedPlatform>();
 
   for (const result of aggregated) {
     if (result.type !== 'artist') continue;
@@ -1386,8 +1397,14 @@ async function fetchReleasesForDisambiguation(aggregated: AggregatedResult[]): P
 
     const qz = result.platforms.find(p => p.sourceId === 'qobuz');
     if (qz) {
-      promises.push(getQobuzLatestRelease(qz.url).then(r => { if (r) qz.latestRelease = r; }));
-      promises.push(getQobuzReleaseTitles(qz.url).then(t => { if (t.length > 0) qz.allReleaseTitles = t; }));
+      // Marked complete only once BOTH lookups settle. If just one came back empty we
+      // still don't know whether Qobuz has releases.
+      promises.push(
+        Promise.allSettled([
+          getQobuzLatestRelease(qz.url).then(r => { if (r) qz.latestRelease = r; }),
+          getQobuzReleaseTitles(qz.url).then(t => { if (t.length > 0) qz.allReleaseTitles = t; }),
+        ]).then(() => { completed.add(qz); }),
+      );
     }
   }
 
@@ -1395,6 +1412,8 @@ async function fetchReleasesForDisambiguation(aggregated: AggregatedResult[]): P
     Promise.allSettled(promises),
     new Promise(resolve => setTimeout(resolve, 4000)),
   ]);
+
+  return completed;
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Automated staff-engineer review — 1 open PR this cycle.

## Security

- **Auth/authorization** — N/A. No change to any auth-checked path.
- **Input validation** — N/A. No new external input is parsed; the change only reorders internal bookkeeping (a `Set` of platform objects) around an existing, unmodified fetch path (`getQobuzLatestRelease` / `getQobuzReleaseTitles`). *(confirmed)*
- **Secret handling** — No secrets touched. *(confirmed)*
- **Injection risks** — None. The new `console.log` lines interpolate `result.name` / `p.url`, same pattern as the pre-existing log line right above them — no new sink. *(confirmed)*
- **Unsafe defaults** — `removeDeadQobuzLinks`'s new second parameter is optional and defaults to `undefined`, which preserves the old (more aggressive) deletion behavior for any caller that doesn't pass it. That's the safe direction for a default (fails toward the previous, already-shipped behavior rather than a new one). *(confirmed)*
- **Data exposure** — The functional effect of the fix is to *keep* a Qobuz link (and its artist image) in ambiguous cases where the release lookup didn't finish in time, rather than deleting it. Worth naming explicitly since it's a data-accuracy/trust trade-off rather than a vulnerability: a kept-but-unverified Qobuz link could occasionally be a name-only match on a different artist. The PR description already calls this out and argues it's the lesser harm (every other platform is already kept on name-match alone) — I agree with that framing. *(confirmed, flagged as intentional trade-off, not a bug)*

No SSRF, CORS, RLS, or middleware surface is touched by this diff.

## Regression

- **Changed assumptions / object identity** — The new `completed`/`releaseLookupsCompleted` `Set<AggregatedPlatform>` relies on reference equality: the same platform object pushed into the set in `fetchReleasesForDisambiguation` must still be the same object `removeDeadQobuzLinks` filters over. I traced the one function that runs in between, `preferBandcampFeaturedRelease`, and it mutates the Qobuz platform object in place (`delete qz.latestRelease`) rather than replacing it — so identity holds. *(confirmed, verified by reading `search-utils.ts:580-593`)*
- **Multiple-Qobuz-per-result edge case** — `fetchReleasesForDisambiguation` and `removeDeadQobuzLinks` both use `.find(p => p.sourceId === 'qobuz')`, i.e. only the first Qobuz platform per result is tracked. I checked `attachQobuzAndSearchLinks`, the only place a `qobuz` platform is pushed onto a result, and it's guarded by a single `usedPlatformUrls` push per artist — so a result can't end up with two Qobuz platforms today. If that ever changes, a second Qobuz entry would silently always be "kept" (never in the completed set). Low risk given current call sites, but worth a comment or an assertion if another Qobuz-attaching path is ever added. *(needs-manual-verification — contingent on future code, not a bug today)*
- **Off-by-one / null/empty behavior** — `hasReleases` still short-circuits correctly when `allReleaseTitles` is `undefined` vs empty array; unchanged from prior logic. *(confirmed, no regression)*
- **API contract drift** — `removeDeadQobuzLinks` is exported from `search-utils.ts`; the new parameter is additive and optional, so any other import site (if one exists outside this file) keeps compiling and keeps old behavior. I didn't find any other caller in this diff. *(confirmed)*
- **Test coverage** — The four new tests directly cover the two branches this PR adds (lookup-didn't-finish → kept, lookup-did-finish → removed) plus a backward-compat case for no-set callers and a has-releases case regardless of set. That maps cleanly to the actual branches in the new code. CI (`check`) is green. *(confirmed)*

## Summary

Small, well-scoped, well-tested fix with a clearly labeled trade-off. No security concerns found. One low-priority structural note (single-Qobuz-per-result assumption) for future-proofing only — not blocking.

---

PR #325, opened 2026-07-26
Source the artist photo from Bandcamp instead of Qobuz

brandonlucasgreen:
Prerequisite for retiring the Qobuz search, and it fixes the missing-photo symptom on its own.

## Why

The artist image comes from the Qobuz match — `attachQobuzAndSearchLinks` does `result.imageUrl = qobuzData.imageUrl`. Since Qobuz currently returns nothing in production, **no artist has a photo right now**, and removing the Qobuz search would make that permanent.

## Fix

Bandcamp is the natural replacement, and it's free: the `/music` page the probe already fetches carries the band photo in its `og:image`.

Verified band-level rather than album art — `radiohead/music` yields `f4.bcbits.com/img/0040867508_23.jpg`, which is **exactly** the image production already displays for Radiohead.

```
Anna von Hausswolff    https://f4.bcbits.com/img/0006059542_23.jpg
Boy Harsher            https://f4.bcbits.com/img/0007772900_23.jpg
Low Hum                https://f4.bcbits.com/img/0034700558_23.jpg
```

`parseBandcampImage` rejects two things deliberately:
- Bandcamp's `blank.gif` placeholder — otherwise artists with no photo get a 1×1 transparent gif as their avatar
- any non-`https` value, to avoid emitting a mixed-content URL

Cached in migration `028` so a cache hit carries the image too.

## Verification

End to end with Qobuz forced to **403**:

```
imageUrl:  https://f4.bcbits.com/img/0006059542_23.jpg
platforms: ["bandcamp","ampwall","kofi","buymeacoffee"]
```

The photo no longer depends on Qobuz at all.

`tsc -b` clean · explicit strict `tsc` **zero new errors** (28 baseline, 28 after) · api tests **77/77** · unit tests **393/393** (5 new)

## Next

This unblocks retiring the Qobuz search, which is `Disallow`ed in Qobuz's `robots.txt` (`Disallow: /*/search/artists/`). That's a separate PR plus a research doc, per the plan agreed in-session.

### DIFF
--- a/api/functions/db.ts
+++ b/api/functions/db.ts
@@ -212,6 +212,8 @@ export interface BandcampProbeRow {
   location: string | null;
   /** Normalized release titles from the probed /music page. */
   release_titles: string[] | null;
+  /** Artist photo from the probed page's og:image. */
+  image_url: string | null;
   checked_at: string;
 }
 
@@ -234,7 +236,7 @@ export async function getBandcampProbe(queryNorm: string): Promise<BandcampProbe
   try {
     const { data, error } = await client
       .from('bandcamp_slug_probes')
-      .select('query_norm, artist_url, band_name, band_id, album_count, track_count, matched_slug, verdict, location, release_titles, checked_at')
+      .select('query_norm, artist_url, band_name, band_id, album_count, track_count, matched_slug, verdict, location, release_titles, image_url, checked_at')
       .eq('query_norm', queryNorm)
       .maybeSingle();
 
--- a/api/functions/search-parsers.ts
+++ b/api/functions/search-parsers.ts
@@ -216,6 +216,25 @@ export function parseBandcampPageLocation(html: string): string | null {
   return raw.length > 0 && raw.length <= 120 ? raw : null;
 }
 
+/**
+ * Read the artist photo from a Bandcamp page's og:image.
+ *
+ * Verified band-level rather than album art: radiohead/music yields
+ * f4.bcbits.com/img/0040867508_23.jpg, which is exactly the image production already
+ * shows for Radiohead. Free — the probe has this HTML in hand.
+ *
+ * Matters because the artist image has been coming from the Qobuz match, and Qobuz's
+ * search path is robots-disallowed and being retired. Bandcamp is the replacement.
+ */
+export function parseBandcampImage(html: string): string | null {
+  const match = html.match(/<meta property="og:image" content="([^"]+)"/);
+  if (!match) return null;
+  const url = match[1].trim();
+  // Bandcamp uses blank.gif as a placeholder; treat it as no image.
+  if (!url.startsWith('https://') || url.includes('/blank.gif')) return null;
+  return url;
+}
+
 /** Parse Bandcamp /music page HTML to extract release titles */
 export function parseBandcampReleaseTitles(html: string): string[] {
   const root = parse(html);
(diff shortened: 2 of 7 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Automated senior-staff-engineer review.

## Security

**No confirmed or plausible findings.**

- **Auth/authorization** — n/a, no auth-surface code touched.
- **Input validation** — `parseBandcampImage` (search-parsers.ts) extracts `og:image` from HTML fetched server-side from a Bandcamp subdomain the probe already resolved; not user-supplied. It rejects non-`https://` values and the `blank.gif` placeholder before the URL is ever stored or returned. *(confirmed: adequate for this input source)*
- **Injection risks** — the extracted URL is only ever used as an `` `src` (never concatenated into HTML/SQL), and the `https://` prefix check rules out `javascript:`/`data:` URIs. Supabase writes go through the existing parameterized client. *(confirmed: no injection path)*
- **Secret handling** — none introduced.
- **Unsafe defaults** — `imageUrl: match.imageUrl ?? undefined` and the nullable `image_url` column both default to "no image" rather than a placeholder/blank asset, matching the intent described in the PR. *(confirmed)*
- **Data exposure** — no PII added; `image_url` is a public band photo. Migration only adds a nullable column to `bandcamp_slug_probes`, which already has RLS enabled with **no public policies** (service-role only) from migration `20260726120000`— this change doesn't need its own RLS grant. *(confirmed)*

One **needs-manual-verification** note (not a blocker): `parseBandcampImage`'s regex assumes `property="og:image"` appears before `content="..."` in the tag. If Bandcamp ever emits the attributes in the other order, extraction silently returns `null` (graceful degradation to "no image," not a crash) — worth keeping an eye on if photos start disappearing again, but not worth a fix now given it matches Bandcamp's actual current markup.

## Regression

- **Precedence/changed assumptions**: Verified `attachQobuzAndSearchLinks` (search-utils.ts:522-525) — Qobuz's image now only fills in `if (qobuzData.imageUrl && !result.imageUrl)`, so Bandcamp's image (attached earlier in `aggregateResults`) correctly wins over Qobuz, matching the PR's stated goal of not depending on Qobuz. No conflicting-precedence bug found.
- **Null/empty edge cases**: `findBandcampArtist`'s cache-hit path maps `cached.image_url` directly (can be `null`) into `BandcampArtistMatch.imageUrl`; consumers (`searchBandcamp`) coalesce with `?? undefined`. Handled correctly for both cache-hit and fresh-probe paths.
- **State/cache transition**: rows cached before this migration (or before a future re-probe) will have `image_url = NULL` and are intentionally *not* retroactively refetched — this is called out in the migration comment as a deliberate choice, not a hidden bug, but means some artists will still show no photo until their probe cache naturally expires/refreshes.
- **API contract drift**: `BandcampProbeRow`/`BandcampProbeResult`/`BandcampArtistMatch` all updated consistently across db.ts, bandcamp-probe.ts; the Supabase `select` column list was updated alongside the new column, so no stale-select bug.
- **Test coverage gap**: the 5 new tests cover `parseBandcampImage` in isolation well (og:image present/absent, blank.gif, non-https, whitespace-trimming), but there's no test exercising the full `searchBandcamp` → `aggregateResults` → `attachQobuzAndSearchLinks` chain to lock in the Bandcamp-wins-over-Qobuz precedence end-to-end. Low risk given the code is straightforward, but that precedence is the crux of this PR and a future refactor could invert it silently.

Overall: clean, well-scoped change consistent with the stated goal (Bandcamp photo replacing Qobuz-sourced image). No blocking issues found.

---

PR #326, opened 2026-07-26
Document the Qobuz search removal plan

brandonlucasgreen:
Documentation only — no code changes. Turns §7 of the Qobuz research doc from a to-do list into an executable plan, so the edits can be made in a fresh session.

## What's in it

- **Exact line ranges** for every deletion (verified against `0845da2`), including `createOrphanedQobuzStandalones` at 1914, which sits far below the others and is easy to miss
- **The known-good MusicBrainz attachment code** — this part was written and verified before the revert
- **The ordering constraint**: attaching MB's Qobuz URL must not ship without the removal, or vice versa. Step 2 without step 1 makes Qobuz *disappear* rather than narrow to major artists.
- **Verification commands**, including main's current baseline of **28** strict-tsc errors so a new session can tell its own errors from the pre-existing ones

## Two things recorded so they aren't repeated

**These files must be hand-edited.** My automated attempt used a brace-walking script to find function boundaries. It mis-parsed on braces inside strings and regexes, reported `searchQobuz: 1 line` for a 95-line function, and produced 31 syntax errors. Reverted; main was left clean.

That's the *second* scripted-edit failure in these two files today — the first injected a variable into six unrelated `catch` blocks. At ~2,050 and ~900 lines they're too large and string-dense for structural scripting.

**`npm run build` does not typecheck any of this.** `api/tsconfig.json` covers only six files, none of them these. The plan includes the explicit `tsc` invocation instead.

## One upside worth noting

#324 turned out to be load-bearing after all, contrary to my earlier recommendation to close it. With the Qobuz release fetchers removed, nothing enters its `completed` set — so `removeDeadQobuzLinks` will now *keep* MB-sourced Qobuz links instead of deleting them for having no release data. Merging it was the right call.

### DIFF
--- a/docs/specs/qobuz-coverage-research.md
+++ b/docs/specs/qobuz-coverage-research.md
@@ -166,13 +166,119 @@ worth recording so the next person does not re-litigate them:
 | #324 | Qobuz link deleted when its release lookup merely timed out | **close — mooted by this decision** |
 | #325 | Artist photo sourced from Bandcamp instead of Qobuz | open |
 
-## 7. Remaining work
+## 7. Implementation plan (not yet done)
 
-1. **Remove `searchQobuz`** and the fan-out entry, plus the now-dead validators and Qobuz release
-   fetchers. Depends on #325 landing first so photos survive.
-2. **Attach MusicBrainz's Qobuz URL as a platform.** It is currently used only for disambiguation
-   (`search-sources.ts` ~1617), never attached — so removing the search without this would drop
-   Qobuz entirely.
-3. **Measure Wikidata `P7071` coverage** against real search traffic before deciding whether to add
-   it as a second source.
-4. **Consider asking Qobuz** about sanctioned access, as with Bandcamp.
+PR #325 has landed, so artist photos now come from Bandcamp and the removal is unblocked.
+
+> **Read this first.** An automated attempt at step 2 corrupted both files. A brace-walking script
+> mis-identified function boundaries — braces inside strings and regexes — reported
+> `searchQobuz: 1 line` for a 95-line function, and produced 31 syntax errors. It was reverted; main
+> is clean.
+>
+> **`search-sources.ts` (~2,050 lines) and `search-utils.ts` (~900 lines) must be hand-edited.** Do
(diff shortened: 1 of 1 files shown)


brandonlucasgreen (commented):

**Scope:** the only change is a rewrite of `docs/specs/qobuz-coverage-research.md` (§7 "Remaining work" → "Implementation plan"). No application code, config, migrations, or dependencies are touched. `git diff --stat` for this PR shows one markdown file.

**Security**
- auth/authorization: not applicable — confirmed (no code changed).
- input validation: not applicable — confirmed.
- secret handling: not applicable — confirmed. Skimmed the added markdown for anything resembling a token/key/credential; none present.
- injection risks: not applicable — confirmed. The `tsc` command block in the doc is a static example for a human to copy-paste, not code that executes anywhere in the repo.
- unsafe defaults: not applicable — confirmed.
- data exposure: not applicable — confirmed.

**Regression**
- changed assumptions / state transitions / off-by-one / edge cases: not applicable to this diff — confirmed, since no runtime logic changed. Note the plan *itself* documents a real prior regression risk (attaching MusicBrainz's Qobuz URL must ship in the same change as removing `searchQobuz`, or Qobuz links disappear for artists with no MB relation) — that's correctly called out as an ordering constraint in Step 1, not something this PR introduces.
- API contract drift: not applicable — confirmed.
- The repo's own semantic-revert-check bot flagged this PR for deleting 9 lines added in #325. I read the diff: the deleted lines are the old §7 bullet list, superseded by the new plan, which explicitly opens with "PR #325 has landed, so artist photos now come from Bandcamp and the removal is unblocked." This is a doc reorganization, not a revert of #325's behavior — **plausible false positive**, flagging per instructions rather than silently dismissing since I can't fully rule out intent without the author confirming.

**needs-manual-verification:** none of the code snippets or line numbers in the plan (e.g. `search-sources.ts:1719`, `:1401`, `search-utils.ts:467`) are re-verified here since no code lands in this PR — worth a quick line-number sanity check at implementation time given the doc itself already flags that a prior automated attempt on these same files produced 31 syntax errors.

No blocking findings. Nothing to change before merge.

---

PR #327, opened 2026-07-26
Retire the dedicated Qobuz search, keep Qobuz as a platform

brandonlucasgreen:
Retires the dedicated Qobuz search. Qobuz stays a supported platform — what goes away is the *scraping*; what replaces it is a MusicBrainz relation.

Full research: `docs/specs/qobuz-coverage-research.md`, whose §7 this PR rewrites from a plan into a record of what shipped.

## Why

Qobuz search has been returning nothing in production for **every** artist. The path `searchQobuz` fetched — `/us-en/search/artists/{query}` — is `Disallow`ed in Qobuz's robots.txt, along with `/api.json/` and the album/track/label search paths. It was never permitted; enforcement just caught up.

The Bandcamp playbook doesn't transfer. Bandcamp URLs are guessable (`<slug>.bandcamp.com`); Qobuz needs a numeric ID you cannot derive from a name:

```
/us-en/interpreter/radiohead          -> 404
/us-en/interpreter/radiohead/43840    -> 200
```

No probe fallback is possible, and `open.qobuz.com` is `User-Agent: * / Disallow: /` too.

## What replaces it

`pickQobuzUrl` in `api/functions/search-utils.ts` reads the Qobuz link out of MusicBrainz `url-rels`, preferring the `www.qobuz.com/.../interpreter/...` form over `open.qobuz.com/artist/...` since MB stores both for the same artist. Attached in `applyEnrichmentToResults`, in the Phase-2.2 MB-fallback result, and mirrored client-side in `mergeWithMusicBrainzData`.

Verified against the live MB API: Qobuz arrives as relation type **`purchase for download`** and **`streaming`** — not `streaming music` — and `purchase for download` was already in `platformRelTypes`. **These URLs were being collected into `platformUrls` all along and simply never attached to a result.**

Qobuz therefore becomes a mainstream-only platform. That's the intended outcome, not a regression — most indie artists aren't on Qobuz, and it exists here as a fallback for major artists whose music isn't on Bandcamp.

## What was removed

- `searchQobuz`, its Phase-1 fan-out slot and Sentry monitoring entry — in both the Netlify function and the Vite dev-server copy
- The three release scrapers, and the Qobuz branch of `fetchReleasesForDisambiguation` (now returns `void`)
- **Six validators** that existed only to police name-matched guesses: `removeDeadQobuzLinks`, `crossPlatformReleaseComparison`, `deduplicateQobuzUrls`, `createOrphanedQobuzStandalones`, `createQobuzOnlyResults`, `preferBandcampFeaturedRelease`. An MB relation is authoritative by construction, so all six could only ever have deleted good links.
- `isQobuzVariation`, `qobuzDisplayName`, `parseQobuzSearchResults`, the `AggregatedPlatform` type, the Qobuz branch of `extractPlatformIdentifier`
- `checkQobuz` (it searched the robots-disallowed `/us-en/search/albums/`), Qobuz from the Mac app's and extension's release-check lists, and the unused `QobuzReleaseChecker.swift`
- `www.qobuz.com` from `ALLOWED_OUTBOUND_HOSTNAMES` — nothing fetches Qobuz now, and least privilege means re-adding a scraper should be deliberate

**Kept:** registry entry and ~70% payout, badges, claim-flow URL pattern, extension detection, Mac `PlatformCatalog`, and the user-facing "search Qobuz" link. A human clicking a search URL is not a crawler.

## Also fixes: MusicBrainz dropped all enrichment for accented artist names

This one is independent of Qobuz but would have kept the replacement from working for a whole class of artists, so it ships here.

`searchMusicBrainz` hand-rolled its name check as `.toLowerCase().replace(/[^a-z0-9]/g, '')`, which **deletes** an accented letter rather than folding it:

```
query  "Tanerélle" -> normalizeSearchQuery -> "Tanerelle" -> naive -> "tanerelle"
MB     "Tanerélle" ---------------------------------------> naive -> "tanerlle"   ✗
```

The function then bailed to an all-nulls result, logging `[MusicBrainz] Skipping "Tanerélle" - doesn't match query "Tanerelle"` — silently discarding official site, Discogs, socials, Wikipedia, location **and the new Qobuz link**, for every artist with an accent in their name. Both copies now call `normalizeForComparison`, which folds via NFD. Guarded by a unit test that contrasts the two forms so the inline version doesn't come back.

## Verification

- `npm run build` green — 375 unit tests, 77 api tests, typecheck, vite build
- `api/` isn't covered by the build's typecheck (`api/tsconfig.json` includes six files), so it was checked explicitly against a stashed baseline: **31 pre-existing errors before, 30 after, zero new** — the one that vanished was inside the deleted `searchQobuz`
- Integration suite back to its pre-change **4 failures**, all pre-existing. Radiohead's `["bandcamp","qobuz"]` fixture now passes *because* of the MB path, making it the regression test for this change — though it flakes in full-suite runs when MusicBrainz 503s under ~20 rapid queries (passes 2-of-3 in isolation)
- Mac project regenerated with `xcodegen`; the `.pbxproj` diff is exactly the 4 lines for the deleted Swift file

## Notes for review

- **This supersedes #326**, which is an open PR documenting an implementation plan for this now-completed work. It edits the same `§7` section and will conflict. Recommend closing it.
- Two integration fixtures were wrong and are corrected here rather than deleted: `Morice` → **`Mo-Rice`** (the artist has a hyphen; `mo-rice.bandcamp.com`, 16 releases), and `Tanerélle` now asserts `discogs` since her Bandcamp page resolves but holds zero releases, so the probe correctly rejects it.
- One consequence worth naming: Qobuz coverage now rests entirely on MusicBrainz being reachable. It degrades safely — `searchMusicBrainz` is deliberately not wrapped in `cacheGetOrFetch`, so a 503 costs one request its link and is never written down as "this artist has no Qobuz link", the failure mode #321 and #323 were about.


brandonlucasgreen (commented):

## Security

**SSRF / outbound fetch** — `plausible`, favorable: removing `www.qobuz.com` from `ALLOWED_OUTBOUND_HOSTNAMES` (`api/functions/middleware.ts`) is correct and I confirmed nothing else in `api/` still fetches a `qobuz.com` URL (`claim-artist.ts`'s reference is a regex match against a user-submitted claim URL, not a fetch; `platforms-list.ts`/`discord-search-background.ts`/`og-metadata.ts` only use it for display strings). No functional regression from tightening the allowlist.

**Hostname check in `pickQobuzUrl`** (`api/functions/search-utils.ts:874`, mirrored in `apps/web/src/services/sources.ts`) — `no finding, verified safe`: it uses exact `hostname === 'www.qobuz.com' | 'qobuz.com' | 'open.qobuz.com'`, not a suffix/`endsWith` check (the plan doc at #326 describes an `endsWith('qobuz.com')` version, but the code that actually shipped is the safer exact-match form). Confirmed by the `pickQobuzUrl(['https://qobuz.com.evil.example/artist/1'])` test asserting `null`. Good — a suffix check here would have been a real hostname-spoofing bug.

**Unvalidated URL rendered as `href`** — `needs-manual-verification`: `qobuzUrl` comes straight from MusicBrainz's `url-rels` (external, community-editable data) and is pushed into `platforms` with no scheme check, then rendered via `href={url}` in `SourceBadge.tsx` (confirmed by reading that file). A non-`http(s)` scheme (e.g. `javascript:`) in an MB relation would render as a clickable link. This is not new — `officialUrl`/`discogsUrl`/`bandcampUrl` already flow through the identical unvalidated path — but this PR adds Qobuz as a new member of that same MB-sourced, unvalidated set, so it's worth a scheme allowlist on all of them at some point rather than treating it as this PR's problem to fix.

**Injection/ReDoS** — `no finding, improvement`: deletion of the hand-rolled Qobuz HTML/regex scrapers (`searchQobuz`, `getQobuzLatestRelease`, `getQobuzReleaseTitles`, the Swift `QobuzReleaseChecker`) removes several regex-over-untrusted-HTML surfaces entirely. Net reduction in attack surface, nothing to flag.

**Auth/secrets** — not touched by this diff.

## Regression

**Accent-normalization fix changes MusicBrainz match behavior for *all* platforms, not just Qobuz** (`search-musicbrainz.ts:175`, `search-sources.ts:546`) — `confirmed`, intentional per the PR body, but flagging the blast radius: switching the name-match check from a naive strip to `normalizeForComparison` (NFD-fold) will change which artists pass the MB match gate for every consumer of `searchMusicBrainz`/`EnrichedMusicBrainzResult` — official site, Discogs, socials, location, Bandcamp URL, hasPre2005Release — not just the new `qobuzUrl` field. That's the intended fix, but it's a behavior change wider than the PR title suggests, worth calling out explicitly in case anything downstream assumed the old (broken) matching as a de facto filter.

**Duplicate-Qobuz-across-results guard relies on a control-flow invariant that isn't locally enforced** — `plausible`: `mbPlatforms.push(qobuzUrl...)` in the Phase-2.2 fallback block (`search-sources.ts:~1550`) only runs when `existingMatch` is false, which is what prevents the same MB-sourced Qobuz URL from being attached both to an existing aggregated result (via `applyEnrichmentToResults`) and to a newly-created fallback result. That's correct today, but the invariant lives in the surrounding `if` rather than in `pickQobuzUrl`/the push site itself — there's no `dedupe-by-url-across-results` pass anymore now that `deduplicateQobuzUrls` is deleted. Low risk since one MB call produces one candidate URL per query, but worth a comment noting why no cross-result dedup is needed now, since a future edit to that `if` could silently reintroduce duplicate Qobuz badges on split/merged results.

**State transition: `fetchReleasesForDisambiguation` return type `Set<AggregatedPlatform> → void`** — `confirmed, correctly threaded`: verified the only caller (`search-sources.ts:~1509`) no longer consumes a return value, and `AggregatedPlatform`/`removeDeadQobuzLinks` are fully removed together rather than left half-referenced. No dangling import found in `search-sources.ts`'s updated import list.

**API/contract drift, client-visible** — `confirmed, backward compatible`: `check-releases.ts`'s `PlatformUrls`/`ReleaseResult.platform` union dropped `'qobuz'`, and the Mac/extension release-checkers no longer send a `qobuz` key. Since these are plain TS interfaces (not runtime-validated against a schema), an older extension/Mac client still sending `platforms.qobuz` would just have that key ignored server-side — no break, but confirm the Mac App Store / extension store rollout doesn't need version-gating if an old client somehow expects a Qobuz release check response.

**Edge case, not this PR's bug but adjacent** — the PR body documents Bandcamp probe cache poisoning on punctuated names (`Mo-Rice` permanently hidden by a prior `Morice` miss) as a newly-discovered, separately-tracked issue (§8 of the linked doc). Confirmed it's out of scope here (no code changes to `bandcamp_slug_probes` in this diff) — just flagging so it doesn't get lost since it's a real production data-correctness bug, not merely a test fixture typo.

No confirmed security vulnerabilities introduced by this PR. It's net-positive on attack surface (fewer outbound fetches, tighter allowlist, deleted regex scrapers) and the one new pure function (`pickQobuzUrl`) is correctly hardened against hostname spoofing and covered by tests for that exact case.

---

PR #330, opened 2026-07-27
Stop rejecting single-release Bandcamp artists as empty squatters

brandonlucasgreen:
Fixes the first of the two bugs found by the real-traffic coverage measurement. The second turned out to need no code — details below.

## Bug 1: single-release artists rejected as squatters

`parseBandcampReleaseCounts` only counted `.music-grid-item` elements. Bandcamp serves **two** layouts at `/music`:

- **Discography grid** — what the parser expects.
- **Single release** — `/music` 303s to that release, and the album page it lands on has *no grid at all*. The discography sits in a `#discography` sidebar instead.

Artists in the second layout scored 0 albums / 0 tracks, which is the parked-squatter signature, so the probe rejected them. `rejected_empty` is a cacheable verdict, so each false negative was stored and served indefinitely.

Measured against artists' own self-declared Bandcamp links: **5 of 13 sampled `rejected_empty` verdicts were real artists** — Massive Attack and Yoko Kanno among them, so this is not only a long-tail problem.

The sidebar is read **only when the grid is empty**, so pages that already worked are untouched, and genuinely parked accounts (no grid *and* no sidebar) still score 0/0 and stay rejected.

`parseBandcampReleaseTitles` gets the same fallback. Without it an accepted single-release artist arrives with no titles, and disambiguation spends its shared 4s release budget re-fetching a page the probe already read.

### Verified against live Bandcamp

| Slug | Before | After |
|---|---|---|
| `prairiez` | 0a/0t → rejected | 1a/0t, `subtitlesforblushing` |
| `gizzvanbuskirk` | 0a/0t → rejected | 1a/0t, `collisionsinthedark` |
| `bloodlessgirls` | 0a/0t → rejected | 1a/0t, `pinkpillsforpalepeople` |
| `massive-attack` | 0a/0t → rejected | 1a/0t |
| `yoko-kanno` | 0a/0t → rejected | 1a/0t |
| `boyharsher` (control, grid layout) | 15a/1t | 15a/1t — unchanged |
| `beyonce`, `chatgpt` (real squatters) | 0a/0t | 0a/0t — still rejected |

### Migration `20260727120000`

Clears cached `rejected_empty` rows. The parser fix cannot reach them on its own — they are already-stored false negatives. Safe: the table is a cache, and cleared rows are re-probed at most one round each on the next search. Only `rejected_empty` is touched, so #328's `probed_slugs` work is unaffected.

## Bug 2: no code change needed

The second bug was the probe returning a *different artist with the same name* about 4% of the time. I reported that users were seeing this. **That was asserted without testing, and it was wrong.** Production returns the correct link for all three measured cases, because two independent mitigations were already in place:

1. A claimed DB record is prepended and same-name live results are filtered out.
2. Phase 2 **replaces** any Bandcamp platform with MusicBrainz's relation URL where one exists.

The residual is *unclaimed artist + no MusicBrainz relation + same-name collision* — an intersection with no corroborating source in it, so nothing cheap remains. That needs the §5b cross-source resolver or Layer B.

The one worthwhile change: mitigation 2 was explained only by a comment about the **old disabled scraper** ("MB relations are our primary source for direct Bandcamp links" — untrue since the probe shipped and began finding 80% vs MusicBrainz's 35%). Read literally, it invites a refactor to downgrade the overwrite to fill-in-if-missing and silently reintroduce the bug. Comment-only change; no behaviour difference.

## Checks

- `tsc -b` clean
- 380/380 unit tests pass, including 6 new parser tests covering both layouts
- `npm run lint` reports 80 problems — **identical to the pre-existing baseline**, unchanged by this PR

## Notes

- Full method and numbers: `docs/specs/bandcamp-coverage-research.md` §10.
- Branch was merged up to current `main` after #327/#328/#329 landed; the migration was moved off the `20260727090000` slot #328 had taken.
- Not verified end-to-end in the preview: the affected artists' `rejected_empty` rows are still cached in production, so the preview would serve the stored negative. Migration clears them on merge. The parser itself is verified directly against the five real pages above.


brandonlucasgreen (commented):
@brandonlucasgreen Automated staff-engineer review of #330.

**Scope reviewed:** `api/functions/search-parsers.ts` (new `parseBandcampSidebarDiscography` fallback), the comment-only change in `apps/web/src/services/sources.ts`, new unit tests, and migration `20260727090000_bandcamp-probe-clear-false-empty.sql`. Verified locally: `tsc -b` clean, `npx vitest run apps/web/tests/unit/search-parsers.test.ts` → 77/77 pass.

## Security

- **Auth/authorization** — not touched by this PR. No findings.
- **Input validation** — `parseBandcampSidebarDiscography` parses untrusted third-party HTML (Bandcamp's response) via `node-html-parser`. Extracted `href`/`title` values are only used for substring membership checks (`/album/`, `/track/`), `Set` dedup, and `normalizeForComparison` — never concatenated into a URL, SQL, or rendered as raw HTML. No injection surface introduced. **confirmed: no issue.**
- **Secret handling** — no secrets touched. **confirmed: no issue.**
- **Injection risks** — the new migration is a static, parameterless `DELETE ... WHERE verdict = 'rejected_empty'`; no string interpolation. **confirmed: no issue.**
- **Unsafe defaults** — none identified.
- **Data exposure** — the scraped sidebar `href`/`title` values don't appear to be surfaced to clients beyond aggregate counts and normalized titles already produced by the pre-existing grid path. I did not trace every downstream consumer of `releaseTitles` in `bandcamp-probe.ts`/`search-sources.ts` outside this diff. **needs-manual-verification:** confirm nothing downstream renders the raw scraped `href` (as opposed to the constructed `https://<slug>.bandcamp.com` URL) directly to users.

## Regression

- **Changed assumptions** — the `sources.ts` hunk is comment-only; the executable line (`newPlatforms[existingBandcamp] = { sourceId: 'bandcamp', url: bandcampUrl }`) is byte-for-byte unchanged. **confirmed: no behavior change**, despite the surrounding text reading like a policy change.
- **Edge cases / null-empty** — the sidebar fallback in both `parseBandcampReleaseCounts` and `parseBandcampReleaseTitles` is gated behind "the grid produced nothing," so existing grid-layout pages take the old path unconditionally (covered by the new "prefers the grid over the sidebar when both are present" test — but that test only exists for `parseBandcampReleaseCounts`, not `parseBandcampReleaseTitles`). Traced the actual gating: `parseBandcampReleaseTitles`'s fallback triggers on `titles.length === 0`, not on grid presence, so if a real Bandcamp grid page ever had `.music-grid-item` elements without a `.title` child (markup drift), the titles fallback would consult the sidebar while the counts function wouldn't — a latent inconsistency, not something this diff introduces. **plausible, low severity:** add a "grid present but no titles → still prefers grid's empty result over sidebar" test for symmetry with the counts test, or accept the asymmetry as intentional (titles are best-effort, counts are the gating signal).
- **Off-by-one** — the `titles.length >= 20` cap is applied identically in both the grid loop and the new sidebar-fallback loop. **confirmed: no off-by-one.**
- **State transitions** — verified against `api/search/bandcamp-probe.ts:227-239`: `rejected_empty` is written only when `counts.albums === 0 && counts.tracks === 0`, which is exactly the state this PR's parser fix changes for single-release artists, and exactly the verdict the migration clears. No other verdict (`accepted`, `absent`, `rejected_name`) is affected. **confirmed: correct, minimal-blast-radius cache invalidation.**
- **API contract drift** — no exported function signatures changed; the `HTMLElement` import is type-only. **confirmed: no drift.**

## Process note (not a code defect)

The `semantic-revert-check` bot comment on this PR flags large deletions across ~15 files (Qobuz retirement, punctuation-cache fix, CLAUDE.md, etc. from #327-#329). This looks like a **false positive in that workflow**, not a real problem in this PR: the PR's base sha (`6c35ba52`) matches the current tip of `main`, and the actual PR diff (`get_diff`) only touches the 5 files described above. Worth a quick look at the semantic-revert-check workflow itself if it's misfiring, since a real revert warning on a future PR could get ignored if this one is noise.

No blocking findings. The one actionable item is the low-severity test-coverage gap noted above.

---

PR #331, opened 2026-07-27
Parallelize network I/O across functions for latency reduction

brandonlucasgreen:
## Summary

This PR refactors multiple API functions to run independent network I/O operations concurrently using `Promise.all()` and `Promise.allSettled()`, reducing cumulative latency on the critical path. Rate-limit checks, authentication, database queries, and cache operations that have no data dependencies now execute in parallel rather than sequentially.

### `api/functions/db.ts`
- **Artist slug lookup**: Consolidated four sequential fallback queries (exact, case-insensitive, hyphen variants, camelCase) into a single PostgREST `or()` query with client-side filtering. Eliminates up to 3 sequential round-trips on cache misses.
- **Profile data fetching**: Parallelized artist links and profile data fetches for claimed artists using `Promise.all()`.
- **Search result persistence**: Converted sequential artist/link upserts to concurrent operations via `Promise.all()` and bulk link upserts instead of one-per-link loops.

### `api/functions/ratelimit.ts`
- **`checkRateLimit()`**: Daily and per-minute rate-limit checks now run in parallel; daily check no longer blocks per-minute token consumption.
- **`checkApiRateLimit()`**: Same parallelization for API key rate limits (daily + per-minute).

### `api/functions/saved-artists.ts`
- Rate-limit and authentication checks run concurrently at handler entry.
- Artist lookup (exact + case-insensitive) parallelized via `Promise.all()`.
- Soft-delete operation parallelized with CDN cache purge lookup; error handling improved with `.catch()` fallback.

### `api/functions/analytics-stats.ts`
- Rate-limit and authentication checks run concurrently.
- Artist resolution and ownership verification queries parallelized; 404 (no artist) still takes precedence over 403 (unauthorized).

### `api/functions/artist-auth.ts`
- Rate-limit and authentication checks run concurrently.
- Artist details now embedded via PostgREST FK join (`artists (...)`) instead of separate query; eliminates sequential round-trip.
- Email backfill for profiles converted to concurrent `Promise.all()` map.

### `api/functions/me-settings.ts`
- Rate-limit and authentication checks run concurrently.
- Refactored `authenticateRequest()` to return `hasPassword` flag from token validation (`auth.getUser()`) instead of requiring a separate `admin.getUserById()` call.
- Removed redundant admin API call; email and password status now come from the same token validation round-trip.

### `api/functions/saved-artists-sync.ts`
- Rate-limit and authentication checks run concurrently.

### `api/functions/search-sources.ts`
- Merge overrides fetch started concurrently with platform fan-out instead of sequentially after Phase 1 completes.
- Claimed artist lookup runs concurrently with `searchAllPlatforms()` instead of blocking on it.

### `api/functions/__tests__/me-settings.test.ts`
- Updated test mocks to reflect that `hasPassword` now comes from `auth.getUser()` rather than `admin.getUserById()`.

## Implementation Details

- **Input validation**: Added slug regex check in `getArtistBySlug()` before interpolating into PostgREST filters (security + clarity).
- **Error handling**: Concurrent operations use `.catch()` fallbacks where appropriate (e.g., CDN purge lookup) to prevent one failure from blocking the response.
- **Deduplication**: Bulk link upserts use a `Map` to deduplicate by platform before sending to Postgres (prevents conflict key violations).
- **Precedence**: Where multiple checks exist (e.g., 404 vs 403), sequential evaluation is preserved via `await` on the first check before returning early.

These changes reduce API latency by eliminating sequential round-trips on the hot path (search, auth, rate-limit checks) while maintaining correctness and error handling.

https://claude.ai/code/session_01LxN7t8Ro5HSiFvDrcBmxq9

### DIFF
--- a/api/functions/db.ts
+++ b/api/functions/db.ts
@@ -334,59 +334,45 @@ export async function getArtistBySlug(slug: string): Promise<ArtistResult | null
   const client = getClient();
   if (!client) return null;
 
+  // Stored slugs only ever contain [a-z0-9-] (see artistSlug). Anything else
+  // can't match a real artist, and rejecting it here keeps the slug safe to
+  // interpolate into the PostgREST or() filter below (no ilike wildcards or
+  // filter-syntax characters).
+  if (!/^[A-Za-z0-9-]+$/.test(slug)) return null;
+
   try {
-    // Try exact match first
-    let { data: artist, error: artistError } = await client
+    // The lookup tolerates three slug variants, checked in priority order:
+    //   1. exact match, 2. case-insensitive (e.g. "KingTriumph"),
+    //   3. hyphens stripped ("king-triumph" → "kingtriumph"),
+    //   4. hyphens added at camelCase boundaries ("kingTriumph" → "king-triumph").
+    // They're fetched in ONE query — the old one-query-per-variant fallback
+    // chain cost up to 4 sequential round-trips on every miss, and this
+    // function runs at the front of every search.
+    const noHyphens = slug.includes('-') ? slug.replace(/-/g, '') : null;
+    const camelHyphenated = !slug.includes('-')
+      ? slug.replace(/([a-z])([A-Z])/g, '$1-$2').toLowerCase()
+      : null;
+
+    const orFilters = [`slug.ilike.${slug}`]; // ilike without wildcards = case-insensitive match, covers exact too
+    if (noHyphens) orFilters.push(`slug.eq.${noHyphens}`);
+    if (camelHyphenated && camelHyphenated !== slug) orFilters.push(`slug.eq.${camelHyphenated}`);
+
+    const { data: candidates, error: artistError } = await client
       .from('artists')
       .select('*')
-      .eq('slug', slug)
-      .single();
+      .or(orFilters.join(','));
 
-    // If no exact match, try case-insensitive and hyphen-variant matches
-    if (artistError || !artist) {
-      // Try case-insensitive match (e.g. "KingTriumph" → "kingtriumph")
-      const { data: ciData } = await client
-        .from('artists')
-        .select('*')
-        .ilike('slug', slug)
-        .single();
-      if (ciData) {
-        artist = ciData;
-        artistError = null;
-      }
-    }
+    if (artistError || !candidates || candidates.length === 0) return null;
 
-    // Try hyphenated variant (e.g. "king-triumph" → "kingtriumph")
-    if ((artistError || !artist) && slug.includes('-')) {
-      const noHyphens = slug.replace(/-/g, '');
-      const { data: nhData } = await client
-        .from('artists')
-        .select('*')
-        .eq('slug', noHyphens)
-        .single();
-      if (nhData) {
-        artist = nhData;
-        artistError = null;
-      }
-    }
+    const rows = candidates as ArtistRow[];
+    const lowerSlug = slug.toLowerCase();
+    const artist =
+      rows.find(r => r.slug === slug) ||
+      rows.find(r => r.slug.toLowerCase() === lowerSlug) ||
+      (noHyphens ? rows.find(r => r.slug === noHyphens) : undefined) ||
+      (camelHyphenated ? rows.find(r => r.slug === camelHyphenated) : undefined);
 
-    // Try adding hyphens at camelCase boundaries (e.g. "kingtriumph" → "king-triumph")
-    if ((artistError || !artist) && !slug.includes('-')) {
-      const hyphenated = slug.replace(/([a-z])([A-Z])/g, '$1-$2').toLowerCase();
-      if (hyphenated !== slug) {
-        const { data: hyData } = await client
-          .from('artists')
-          .select('*')
-          .eq('slug', hyphenated)
-          .single();
-        if (hyData) {
-          artist = hyData;
-          artistError = null;
-        }
-      }
-    }
-
-    if (artistError || !artist) return null;
+    if (!artist) return null;
 
     const row = artist as ArtistRow;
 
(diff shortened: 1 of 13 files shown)


api/functions/me-settings.ts
```diff
@@ -40,8 +47,12 @@ export async function handler(event: {
     return { statusCode: 204, headers: CORS_HEADERS, body: '' };
   }
 
+  // Rate-limit check (Redis) and token validation (Supabase Auth) are both
+  // network round-trips with no data dependency — run them concurrently.
   const ip = getClientIp(event.headers);
-  const rl = await checkRateLimit(ip, 'standard', CORS_HEADERS);
+  const rlPromise = checkRateLimit(ip, 'standard', CORS_HEADERS);
+  const userPromise = authenticateRequest(event.headers.authorization).catch(() => null);
```
brandonlucasgreen, line 54: **Security — auth/authorization (plausible), same pattern repeats in `artist-auth.ts:49`, `saved-artists.ts:87`, `saved-artists-sync.ts:48`, `analytics-stats.ts:48`**

Before this PR, `checkRateLimit` was awaited first, and `authenticateRequest` (a real network round-trip to Supabase Auth's `auth.getUser`) only ran if the request wasn't rate-limited. Now `userPromise`/`authPromise` is kicked off *before* `rl` is even awaited, so every request — including ones that end up 429'd, and ones carrying garbage/invalid bearer tokens — still costs a call to Supabase's auth server.

This weakens the rate limiter's role as a cost gate in front of the auth service: an attacker can still cheaply generate load against `auth.getUser()` even though the local rate limiter rejects the request. Confirmed as a code-level behavior change; labeling the actual security impact "plausible" since severity depends on Supabase Auth's own quota/pricing for `getUser()` calls under load.

Suggest either only starting the auth check after confirming `rl.limited === false`, or explicitly noting the tradeoff is accepted (the latency win is real, just worth being deliberate about it given 5 endpoints now do this).

@brandonlucasgreen

---

api/functions/db.ts
```diff
@@ -468,16 +457,19 @@ export async function persistSearchResults(results: ArtistResult[]): Promise<voi
   const client = getClient();
   if (!client) return;
 
-  for (const result of results) {
-    if (result.type !== 'artist') continue;
+  // This runs before the search response is sent, so wall-clock time matters:
+  // artists persist concurrently, and each artist's links go up in one bulk
+  // upsert instead of one round-trip per link.
+  await Promise.all(results.map(async result => {
```
brandonlucasgreen, line 463: **Regression — state transitions (needs-manual-verification)**

`persistSearchResults` moved from a sequential `for...of` loop to `Promise.all(results.map(...))`, so every artist in a search response now upserts concurrently. Each iteration does a read (`existing?.match_confidence === 'claimed'`) then a conditional upsert against the same `artists` row keyed by slug, and the per-result link-dedup `Map` only dedupes platforms *within* one result, not across results.

If `aggregateResults`/`splitSuspiciousPlatforms` can ever emit two entries that resolve to the same `artistSlug(name)` (e.g. an imperfectly-deduped near-duplicate before this stage), the two iterations now race: both can read `match_confidence` before either writes, and both upsert independently. The old sequential loop made this impossible by construction — this PR doesn't add a guard against it.

Impact is low (this is cached/derived search-result data, not user data), but worth confirming aggregation guarantees one entry per artist slug reaches this function before treating the concurrency as safe. Category: regression / off-by-one-adjacent state handling.

@brandonlucasgreen

---

brandonlucasgreen (commented):
@brandonlucasgreen Reviewed as senior-staff pass (security + regression). Verified `npm run test:api` passes on this branch (87/87) and manually typechecked the files this PR touches that fall outside `api/tsconfig.json`'s narrow `include` (`db.ts`, `analytics-stats.ts`, `artist-auth.ts`, `ratelimit.ts`, `saved-artists.ts`, `saved-artists-sync.ts`, `search-sources.ts`) — no new type errors vs. main; the pre-existing errors there (`saved-artists.ts` null-client checks, `search-sources.ts` `PlatformLink`/nullable-string mismatches) are untouched by this diff.

**Security**
- `db.ts` `getArtistBySlug`: switching the slug-variant fallback to a single `.or()` PostgREST filter is a good change, but building `.or()` from string-interpolated filter clauses (`slug.ilike.${slug}`, `slug.eq.${noHyphens}`) would normally be a filter-injection risk (comma/dot/paren are PostgREST filter syntax). The PR closes this correctly with the `^[A-Za-z0-9-]+$` guard before any interpolation — confirmed safe, no action needed, just flagging that this is load-bearing and shouldn't be loosened later without re-checking the injection angle.
- Auth/rate-limit ordering: see inline comment on `me-settings.ts:54` (same pattern in 4 other files) — token validation now fires unconditionally in parallel with the rate-limit check, so rate-limited/garbage-token requests still cost a Supabase Auth round-trip. Labeled plausible.
- No new secret handling, injection, or unsafe-default issues found in the rest of the diff (analytics-stats ownership check, artist-auth FK embed, me-settings auth source change all look behaviorally equivalent to what they replaced).

**Regression**
- `db.ts` `persistSearchResults`: see inline comment at line 463 — sequential loop → `Promise.all` introduces a race if two aggregated results ever resolve to the same artist slug. Labeled needs-manual-verification.
- Everything else (rate-limit daily/per-minute parallelization, `analytics-stats.ts` ownership-query parallelization, `artist-auth.ts` FK embed + backfill parallelization, `search-sources.ts` overrides/claimed-artist prefetch) traced through carefully and is behaviorally equivalent to the sequential version — no state-transition or off-by-one issues found there.

Nice latency win overall; the two flagged items are worth a quick look before merging but neither looks like a blocker to me.

---

PR #332, opened 2026-07-27
Stop a stale MusicBrainz link and a homonym from merging into one artist

brandonlucasgreen:
Reported by the artist Honeycrush: searching `honeycrush` returned a single result named "Honey Crush" whose Bandcamp button opened a **Bandcamp signup form**, mixed together with the real Honeycrush's location, Instagram and Patreon.

Not a regression from the probe-cache work in #328. Three independent faults stacked up.

## Root causes

**1. MusicBrainz has a stale Bandcamp relation.** MB artist `d4268f62…` ("Honeycrush", Brooklyn) lists `https://honeyyycrush.bandcamp.com/`. That subdomain was retired and now answers:

```
303 → https://bandcamp.com/signup?new_domain=honeyyycrush
```

Nothing ever fetched it, and MB's URL unconditionally overwrote the probe's, so the dead link shipped.

**2. A homonym passes the probe.** `bandcampSlugCandidates("honeycrush")` yields slug `honeycrush`, which is a real and unrelated Orlando band literally named "Honey Crush". Name match and release count both pass — neither check can see it's someone else. The true artist is at `honeycrushing`, a slug not derivable from the name.

| slug | name | location | releases |
|---|---|---|---|
| `honeycrush` | Honey Crush | Orlando, Florida | 1984, Trying |
| `honeycrushing` | Honeycrush | New York, New York | Man-Size, Dogwoods, Marigold, … |

**3. The two merged**, producing one result assembled from two artists plus a link belonging to neither live account.

## Fixes

**1. `checkBandcampSubdomain()`** — one `HEAD` with `redirect: 'manual'` before we trust an MB Bandcamp relation. Tri-state `'live' | 'dead' | 'unknown'`: only a redirect to `bandcamp.com/signup` counts as dead, while a 429, 5xx, timeout or network error returns `unknown` and leaves the link alone, so a Bandcamp outage can't strip good links off every result. Verdicts cache per-URL for a week; `unknown` is never cached.

Applied in **both** search functions — Phase 2 is the client's fallback path, so fixing only Phase 1 would let the dead link arrive a second later instead of not at all.

**2. An MB relation no longer overwrites a probe-verified URL.** The probe fetched the page and checked identity against the query; an MB relation is a stored string nobody re-checked. Only a `bandcamp.com/search` placeholder is still replaced.

**3. Subdomain-disagreement check.** MB is authoritative about *which* Bandcamp account belongs to an artist, and that claim survives the account being retired. MB says `honeyyycrush`, the probe matched `honeycrush` → different accounts → MB's enrichment no longer attaches. Phase 2.2's coverage check excludes conflicting homonyms too; otherwise rejecting the impostor would delete the real artist from the response instead of listing them separately.

### Two heuristics deliberately rejected

The obvious disambiguators both reject the **correct** artist, so they're not here:

- **Location.** MB says `Brooklyn`; the correct Bandcamp account says `New York, New York`. Same place, different strings. Country level is useless — the wrong artist is also `US`.
- **Release-title overlap.** MB knows one release group ("Unfold"); the correct account lists eight different titles. Zero overlap.

MB's coverage of small independent artists is thin, so "the reference source doesn't corroborate this" almost always means it doesn't know — absence treated as a negative, the same error class as `never cache uncertainty`. Hence an authoritative identity claim rather than attribute comparison.

## Performance

Roughly 0ms added on cache hits. On a cold miss: one bodyless `HEAD`, measured at ~200ms for a dead subdomain and ~275–450ms for a live one, running *inside* the existing parallel enrichment block that already budgets 4s — so `max(phase, 0.3s)`, not `phase + 0.3s`. Only fires when MB actually has a Bandcamp relation. Fixes 2 and 3 are pure logic, zero network.

## Testing

- 15 new tests in `api/functions/__tests__/bandcamp-identity-conflict.test.ts`. One caught a real bug during development — the first cut classified a 429 as `live`.
- `npm run build` green end to end. API 110/110, unit 380/380.
- `api/` isn't in the build's typecheck include, so `tsc` was run explicitly against each changed file and diffed against a stashed baseline: four pre-existing errors in `search-sources.ts`, unchanged, zero new. Lint: 78 pre-existing problems before and after, none from this branch.
- The pieces are verified; the assembled pipeline is not verified against production — worth a look at the deploy preview.

## Scope limit

Honeycrush is still not discoverable **by name** — `honeycrushing` can't be derived from "honeycrush". That needs the upstream MusicBrainz record corrected (the artist has been contacted) or a merge override, which has been applied. What this PR fixes is that the signup link is gone and the two artists no longer merge into one.

### DIFF
--- a/api/functions/search-musicbrainz.ts
+++ b/api/functions/search-musicbrainz.ts
@@ -30,6 +30,7 @@ import {
   parseLocationString,
   mergeLocations,
   fetchBandcampLocation,
+  checkBandcampSubdomain,
   fetchMirloLocation,
   enrichLocationFallback,
 } from '../search/enrichment';
@@ -146,7 +147,7 @@ async function searchMusicBrainz(query: string): Promise<MusicBrainzSearchRespon
     let wikipediaUrl: string | null = null;
     const socialLinks: SocialLink[] = [];
     const seenPlatforms = new Set<SocialPlatform>();
-    const platformUrls: string[] = [];
+    let platformUrls: string[] = [];
 
     let mbLocation: ArtistLocation | undefined;
     let hasPre2005Release = false;
@@ -286,7 +287,7 @@ async function searchMusicBrainz(query: string): Promise<MusicBrainzSearchRespon
     // Fetch additional social links from Discogs, official site, PeerTube, Wikipedia,
     // and platform locations (Bandcamp, Mirlo) in parallel.
     // Bandcamp location: prefers MB relation URL; falls back to live Bandcamp search by name.
-    const [discogsSocialLinks, officialSiteResult, peertubeLink, wikipediaResult, bandcampLocation, mirloLocation] = await Promise.all([
+    const [discogsSocialLinks, officialSiteResult, peertubeLink, wikipediaResult, bandcampLocation, mirloLocation, bandcampStatus] = await Promise.all([
       discogsUrl ? fetchDiscogsSocialLinks(discogsUrl) : Promise.resolve([]),
       officialUrl ? fetchOfficialSiteSocialLinks(officialUrl) : Promise.resolve({ socialLinks: [], linktreeUrl: null, discoveredPlatforms: [] }),
       searchPeerTubeChannels(artist.name),
@@ -304,7 +305,17 @@ async function searchMusicBrainz(query: string): Promise<MusicBrainzSearchRespon
         return match?.location ? parseLocationString(match.location) : null;
       })(),
       fetchMirloLocation(mirloSlug),
+      mbBandcampUrl ? checkBandcampSubdomain(mbBandcampUrl) : Promise.resolve('unknown' as const),
     ]);
+
+    // Phase 1 strips retired subdomains at the same point; this is the Phase 2 path the
+    // client falls back to when server-side enrichment didn't land in time. Both have to
+    // do it, or the dead link simply arrives a second later. Only a confirmed 'dead' is
+    // dropped — 'unknown' keeps trusting MB, so a Bandcamp outage changes nothing.
+    if (mbBandcampUrl && bandcampStatus === 'dead') {
+      console.log(`[MusicBrainz] Dropping retired Bandcamp subdomain for "${artist.name}": ${mbBandcampUrl}`);
+      platformUrls = platformUrls.filter(u => u !== mbBandcampUrl);
+    }
 
     // Merge locations: MusicBrainz is authoritative; Bandcamp and Mirlo fill in missing fields
     const location = mergeLocations(mbLocation, bandcampLocation, mirloLocation);
--- a/api/functions/__tests__/bandcamp-identity-conflict.test.ts
+++ b/api/functions/__tests__/bandcamp-identity-conflict.test.ts
@@ -1,0 +1,152 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+
+// The bug this guards, reported by the artist Honeycrush:
+//
+//   /?q=honeycrush returned one result named "Honey Crush" whose Bandcamp button
+//   opened a Bandcamp *signup form*, mixed with the real Honeycrush's location,
+//   Instagram and Patreon.
+//
+// Three independent faults stacked up:
+//
+//   1. MusicBrainz lists https://honeyyycrush.bandcamp.com/ for the real artist. That
+//      subdomain was retired and now answers 303 -> bandcamp.com/signup?new_domain=…,
+//      but nothing ever fetched it, so the dead link shipped.
+//   2. The subdomain probe derived slug "honeycrush", which is a real and unrelated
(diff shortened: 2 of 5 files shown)


api/functions/search-musicbrainz.ts
```diff
@@ -304,8 +305,18 @@ async function searchMusicBrainz(query: string): Promise<MusicBrainzSearchRespon
...
+      mbBandcampUrl ? checkBandcampSubdomain(mbBandcampUrl) : Promise.resolve('unknown' as const),
     ]);
 
+    // Phase 1 strips retired subdomains at the same point; this is the Phase 2 path the
+    // client falls back to when server-side enrichment didn't land in time. Both have to
+    // do it, or the dead link simply arrives a second later. Only a confirmed 'dead' is
+    // dropped — 'unknown' keeps trusting MB, so a Bandcamp outage changes nothing.
+    if (mbBandcampUrl && bandcampStatus === 'dead') {
+      console.log(`[MusicBrainz] Dropping retired Bandcamp subdomain for "${artist.name}": ${mbBandcampUrl}`);
+      platformUrls = platformUrls.filter(u => u !== mbBandcampUrl);
```
brandonlucasgreen, line 317: Minor/regression — **confirmed**, low severity: the Phase 1 twin of this (`search-sources.ts`) reports a dropped dead Bandcamp relation to Sentry via `checkSentryDedup` + `Sentry.captureMessage` (`"MusicBrainz Bandcamp relation points at a retired subdomain"`), so the stale-MB-data problem gets tracked for an upstream fix. This Phase 2 path only `console.log`s the same event — it won't surface in Sentry. Not a bug, just an observability asymmetry between the two functions worth closing to match.

@brandonlucasgreen

---

brandonlucasgreen (commented):
@brandonlucasgreen Security + regression pass on this PR (the only one open right now).

## Security

- **Auth/authorization** — no findings. No auth-gated code paths touched.
- **Input validation** — no findings. `checkBandcampSubdomain()` correctly routes its fetch through `isUrlHostnameAllowed()` before hitting the network, consistent with the SSRF-allowlist requirement in `CLAUDE.md`.
- **Secret handling** — no findings. Nothing new logged or cached carries credentials; cache keys are derived from the (already-public) Bandcamp URL, not user PII.
- **Injection risks** — no findings. No new HTML/SQL/shell interpolation; `console.log`/`Sentry.captureMessage` calls only carry artist name, URL, and query text, matching existing logging patterns elsewhere in the file.
- **Unsafe defaults** — no findings. The tri-state `'live' | 'dead' | 'unknown'` model defaults failures to `'unknown'` and only `'dead'` (one specific redirect target) causes a link to be dropped, which is the correct default per the project's "never cache uncertainty" rule. `cacheGetOrFetch`'s `shouldCache` predicate (`status => status !== 'unknown'`) correctly refuses to cache the uncertain case.
- **Data exposure** — no findings. The new `bandcampSubdomain` field stays internal to `EnrichedMusicBrainzResult` in `search-sources.ts` and is never serialized into the public API response.

## Regression

- **Confirmed — the fix is incomplete for the Phase 2 client fallback path.** Left as an inline comment on `api/functions/search-musicbrainz.ts:317`, but the more important instance couldn't be anchored because the file isn't in this diff: `apps/web/src/services/sources.ts`'s `mergeWithMusicBrainzData` (lines ~733–757) still overwrites a result's Bandcamp URL by name-match alone, with no equivalent of the new `bandcampSubdomainConflicts()` check. That function only runs when `hasPendingEnrichment` is true (`App.tsx:156`) — i.e. precisely when Phase 1's server-side enrichment (the code this PR fixes) didn't land in time. `MusicBrainzSearchResponse` never gained a `bandcampSubdomain` field, and neither did the client's `MusicBrainzData` type, so the client has no way to detect the MB-vs-probe account disagreement at all. Net effect: a homonym with a *live* (not retired) but different Bandcamp account than MB's relation can still get merged into one result via the fallback path — same bug class as the reported issue, just not gated on the link being dead. The PR body's "Applied in both search functions" claim is accurate only for the retired-subdomain check, not the subdomain-disagreement check.
- **Edge cases / null-empty behavior** — no issues found. `bandcampSubdomainConflicts` correctly treats a missing side as absence rather than conflict (verified against the new test file), and custom domains are deliberately excluded rather than guessed at.
- **State transitions** — the `'live'/'dead'/'unknown'` tri-state is handled consistently in both places it's read (drop-link condition explicitly checks `=== 'dead'`, never treats `unknown` as either verdict).
- **Off-by-one logic** — none observed; no loop/index changes of that shape in this diff.
- **API contract drift** — none found in the public/external API surface. The internal `bandcampSubdomain` field is a same-request-lifetime implementation detail, not persisted or re-read across requests, so no drift risk there.

Nothing here should block merge — the core Phase 1 fix (the reported bug) looks solid and well-tested — but the Phase 2 gap is worth a fast follow so the fallback path doesn't quietly reproduce the same class of bug it just fixed.

---

PR #333, opened 2026-07-28
Move mobile header nav into a hamburger menu, follow system theme

brandonlucasgreen:
The header nav ran out of room on phones once signed in, and more items are
coming. Below `sm` the nav now collapses into a hamburger button in the
top-right that opens a dropdown: Login when signed out; the account email,
Dashboard, Settings and Sign out when signed in (plus the admin Verify alert
when there is one, which was otherwise unreachable on mobile).

Desktop keeps the inline nav. Both renderers read the same `navItems` array,
so future items only need adding once. The email moves to `md` and up so the
inline nav still fits at small-desktop widths.

Also drops the light/dark/system toggle: the app now always follows the OS
setting. Light theme is plain `@media (prefers-color-scheme: light)` in
index.css instead of a `data-theme` attribute, which removes the blocking
inline script in index.html along with its flash-of-wrong-theme workaround,
and matches how the edge functions already theme their HTML. The logo takes
its colour from `currentColor` rather than a resolved-theme prop.

Claude-Session: https://claude.ai/code/session_01DTx8Mxp13p1UjB8ALUYbL3


brandonlucasgreen (commented):

## Security

This PR is UI-only (mobile hamburger nav + removal of the theme toggle in favor of `prefers-color-scheme`). No new attack surface.

- **Auth/authorization** — confirmed, no issues. The admin-only `/admin/verify` link visibility (`isAdmin && pendingVerifyCount > 0`) is carried over unchanged into the shared `navItems` array; still a client-side display gate only (same as before this PR — actual route protection, if any, lives elsewhere and wasn't touched).
- **Input validation** — confirmed, no issues. No user-supplied input is parsed, stored, or sent anywhere new.
- **Secret handling** — confirmed, no issues. The only storage-related change is *removing* the `unstream-theme` `localStorage` key (no secrets involved).
- **Injection risks** — confirmed, no issues. `user?.email` and nav labels render as plain JSX text (React auto-escapes); no `dangerouslySetInnerHTML`, no raw HTML interpolation.
- **Unsafe defaults** — confirmed, no issues. `MobileNav`'s outside-click/Escape listeners are only attached while `open` and are cleaned up via the `useEffect` return, so there's no dangling global listener.
- **Data exposure** — confirmed, no issues. The email is shown only inside the dropdown, gated by `session`, same exposure surface as the inline `<span>` it replaces.

## Regression

- **CSS specificity change — needs-manual-verification.** Light theme moved from `html[data-theme="light"] { ... }` (attribute selector, specificity `0,1,1`, order-independent) to `@media (prefers-color-scheme: light) { :root { ... } }` (specificity `0,1,0`, equal to the base `@theme`/`:root` defaults). Correctness now depends on this block staying *after* the dark defaults in `index.css`'s source order — true today, but a future edit that adds another `:root` var block later in the file, or an unexpected build-tool reorder, would silently break light mode with no error. Worth a mental note, not a blocker.
- **Intentional product change, not a bug.** The manual light/dark/system toggle is deliberately removed per the PR description — confirming that's the intent and not accidental scope creep, since it's a user-facing feature removal.
- **State transitions / edge cases** — no issues found. `onSignOut` closes the menu (`setOpen(false)`) before invoking sign-out, synchronously; `email={session ? user?.email : undefined}` preserves the prior optional-chaining behavior for the loading-session edge case.
- **API contract drift** — none; no backend/API surface touched.

No blocking findings.

---

PR #336, opened 2026-07-29
feat: artist_releases + release_links data foundation

brandonlucasgreen:
## Summary

Milestone 1 (Data Foundation) for Unstream Releases feature.

### Changes

1. **Supabase migration 029** (20260728120000_artist-releases.sql) — creates artist_releases and release_links tables with RLS, indexes, and updated_at triggers. Public read, service write (same pattern as artists/artist_links).

2. **schema.sql updated** — canonical baseline now reflects the new tables.

3. **scripts/migrate-releases.ts** — one-time idempotent migration script that lifts existing latest_release jsonb data from artist_links into the new tables. Deduplicates by normalized title across platforms (same release on Bandcamp + Mirlo becomes one artist_releases row, two release_links rows).

4. **api/functions/db.ts** — persistSearchResults now also writes to artist_releases + release_links when platform links have latestRelease data. Additive — the jsonb column on artist_links is still written for backward compatibility (search disambiguation still uses it).

5. **api/functions/check-releases.ts** — now persists detected releases to the new tables (fire-and-forget). The function still returns the same response shape.

## What to verify

- The migration SQL runs cleanly against the production DB (uses IF NOT EXISTS guards)
- npx tsx scripts/migrate-releases.ts works with SUPABASE_URL and SUPABASE_SERVICE_KEY env vars set
- npm run test:api — 87 tests pass (2 pre-existing failures from missing jose package, unrelated)
- npx tsc -b — only pre-existing jose error, no new type errors

## Risks

- The migration script runs once and is idempotent, but should be tested against a staging DB first to verify deduplication quality
- persistSearchResults now does more DB writes per search (up to N additional upserts for N releases). This is fire-and-forget and should be negligible, but worth monitoring function execution time
- The latest_release jsonb on artist_links is NOT removed or stopped — it is still written for backward compatibility

## Not in this PR

- Release pages (/a/{slug}/{release-slug}) — Phase 2
- Artist page Releases section — Phase 3
- RSS/ICS feeds — Phase 4
- iTunes Search API integration — Phase 5

### DIFF
--- a/api/functions/check-releases.ts
+++ b/api/functions/check-releases.ts
@@ -1,4 +1,7 @@
 import { parse } from 'node-html-parser';
+import { getClient, artistSlug } from './db.js';
+import { mapReleaseType, releaseSlug, isStreamingPlatform, bandcampReleaseType } from './release-utils.js';
+import { checkRateLimit, getClientIp } from './ratelimit.js';
 
 interface PlatformUrls {
   bandcamp?: string;
@@ -11,6 +14,7 @@ interface ReleaseResult {
   releaseDate: string; // ISO format
   releaseUrl: string;
   platform: 'bandcamp' | 'faircamp' | 'mirlo';
+  releaseType: 'album' | 'track';
 }
 
 interface CheckReleasesRequest {
@@ -22,6 +26,119 @@ interface CheckReleasesResponse {
   artistName: string;
   release: ReleaseResult | null;
   error?: string;
+}
+
+// ── Release persistence (artist_releases + release_links) ─────────────────
+//
+// When check-releases finds a release, it writes to the new release tables
+// in addition to returning the data. The jsonb latest_release on artist_links
+// is still written by persistSearchResults in db.ts for backward compatibility.
+// These tables are additive.
+//
+// Security: before persisting, we validate that the release URL's domain
+// matches a platform link already stored on the artist in artist_links.
+// This prevents arbitrary URL injection via the unauthenticated endpoint.
+
+/**
+ * Persist a detected release to artist_releases + release_links.
+ * Idempotent: upserts on (artist_id, slug) and (release_id, platform).
+ * Fire-and-forget — errors are logged but don't fail the response.
+ */
+async function persistRelease(
+  artistName: string,
+  release: ReleaseResult,
+): Promise<void> {
+  const client = getClient();
+  if (!client) return;
+
+  try {
+    const slug = artistSlug(artistName);
+
+    // Look up the artist by slug
+    const { data: artist, error: artistError } = await client
+      .from('artists')
+      .select('id')
+      .eq('slug', slug)
+      .single();
+
+    if (artistError || !artist) {
+      // Artist not in DB — nothing to persist
+      return;
+    }
+
+    const artistId = artist.id;
+
+    // Validate that the release URL's domain matches a platform link already
+    // stored on this artist. This prevents arbitrary URL injection via the
+    // unauthenticated endpoint — only releases from known artist platform URLs
+    // are persisted.
+    const { data: artistLinks } = await client
+      .from('artist_links')
+      .select('platform, url')
+      .eq('artist_id', artistId);
+
+    if (!artistLinks || artistLinks.length === 0) return;
+
+    const releaseDomain = new URL(release.releaseUrl).hostname.replace(/^www\./, '');
+    const hasMatchingLink = artistLinks.some((link: { platform: string; url: string }) => {
+      try {
+        const linkDomain = new URL(link.url).hostname.replace(/^www\./, '');
+        return linkDomain === releaseDomain;
+      } catch {
+        return false;
+      }
+    });
+
+    if (!hasMatchingLink) {
+      console.log(`[check-releases] Skipping persist: release URL domain ${releaseDomain} does not match any artist_link`);
+      return;
+    }
+
+    const relSlug = releaseSlug(release.releaseName);
+    const releaseType = mapReleaseType(release.releaseType);
+
+    // Upsert the release
+    const { data: releaseRow, error: releaseError } = await client
+      .from('artist_releases')
+      .upsert(
+        {
+          artist_id: artistId,
+          title: release.releaseName,
+          slug: relSlug,
+          release_type: releaseType,
+          release_date: release.releaseDate,
+          source: 'auto',
+        },
+        { onConflict: 'artist_id,slug' }
+      )
+      .select('id')
+      .single();
+
+    if (releaseError || !releaseRow) {
+      console.error('[check-releases] Failed to upsert artist_releases:', releaseError?.message);
+      return;
+    }
+
+    // Upsert the release link for this platform
+    const { error: linkError } = await client
+      .from('release_links')
+      .upsert(
+        {
+          release_id: releaseRow.id,
+          platform: release.platform,
+          url: release.releaseUrl,
+          is_streaming: isStreamingPlatform(release.platform),
+          source: release.platform, // 'bandcamp' | 'faircamp' | 'mirlo'
+        },
+        { onConflict: 'release_id,platform' }
+      );
+
+    if (linkError) {
+      console.error('[check-releases] Failed to upsert release_links:', linkError.message);
+    }
+  } catch (err) {
+    console.error('[check-releases] persistRelease error:', err);
+  }
 }
 
 // Helper to fetch with timeout
@@ -139,6 +256,7 @@ async function checkBandcamp(artistUrl: string): Promise<ReleaseResult | null> {
       releaseDate,
       releaseUrl: fullUrl,
       platform: 'bandcamp',
+      releaseType: bandcampReleaseType(fullUrl),
     };
   } catch (error) {
     console.error('Bandcamp check error:', error);
@@ -190,6 +308,7 @@ async function checkFaircamp(faircampUrl: string): Promise<ReleaseResult | null>
       releaseDate,
       releaseUrl: linkMatch[1].trim(),
       platform: 'faircamp',
+      releaseType: 'album',
     };
   } catch (error) {
     console.error('Faircamp check error:', error);
@@ -266,6 +385,7 @@ async function checkMirlo(mirloUrl: string): Promise<ReleaseResult | null> {
       releaseDate: latestRelease.pubDate,
       releaseUrl: latestRelease.link,
       platform: 'mirlo',
+      releaseType: 'album',
     };
   } catch (error) {
     console.error('Mirlo check error:', error);
@@ -312,6 +432,7 @@ async function checkAllPlatforms(platforms: PlatformUrls): Promise<ReleaseResult
 export async function handler(event: {
   httpMethod?: string;
   body?: string;
+  headers?: Record<string, string | undefined>;
   queryStringParameters?: Record<string, string>;
 }) {
   // Handle CORS preflight
@@ -338,6 +459,17 @@ export async function handler(event: {
       body: JSON.stringify({ error: 'Method not allowed. Use POST.' }),
     };
   }
+
+  const corsHeaders = {
+    'Content-Type': 'application/json',
+    'Access-Control-Allow-Origin': '*',
+  };
+
+  // Rate limit: strict tier (10 req/min, 500/day) — this endpoint fetches
+  // external pages and writes to the DB, so it's expensive.
+  const ip = getClientIp(event.headers || {});
+  const rl = await checkRateLimit(ip, 'strict', corsHeaders);
+  if (rl.limited) return rl.response;
 
   // Parse request body
   let request: CheckReleasesRequest;
@@ -384,6 +516,13 @@ export async function handler(event: {
 
   try {
     const release = await checkAllPlatforms(request.platforms);
+
+    // Persist to artist_releases + release_links (fire-and-forget, additive to jsonb)
+    if (release) {
+      persistRelease(request.artistName, release).catch(err =>
+        console.error('[check-releases] Background persist failed:', err)
+      );
+    }
 
     const response: CheckReleasesResponse = {
       artistName: request.artistName,
--- a/api/functions/check-releases.ts
+++ b/api/functions/check-releases.ts
@@ -1,7 +1,13 @@
 import { parse } from 'node-html-parser';
-import { getClient, artistSlug } from './db.js';
-import { mapReleaseType, releaseSlug, isStreamingPlatform, bandcampReleaseType } from './release-utils.js';
+import { isUrlHostnameAllowed } from './middleware.js';
+import { bandcampReleaseType } from './release-utils.js';
 import { checkRateLimit, getClientIp } from './ratelimit.js';
+
+// Note on releaseType mapping: the API returns 'album' | 'track' from the
+// platform parsers, while the DB stores the enum 'album' | 'single' | 'ep' |
+// 'compilation'. mapReleaseType() in release-utils.ts handles the mapping
+// ('track' → 'single', 'album' → 'album'). This endpoint returns the raw
+// platform values; persistence (via db.ts) applies the mapping.
 
 interface PlatformUrls {
   bandcamp?: string;
@@ -26,119 +32,6 @@ interface CheckReleasesResponse {
   artistName: string;
   release: ReleaseResult | null;
   error?: string;
-}
-
-// ── Release persistence (artist_releases + release_links) ─────────────────
-//
-// When check-releases finds a release, it writes to the new release tables
-// in addition to returning the data. The jsonb latest_release on artist_links
-// is still written by persistSearchResults in db.ts for backward compatibility.
-// These tables are additive.
-//
-// Security: before persisting, we validate that the release URL's domain
-// matches a platform link already stored on the artist in artist_links.
-// This prevents arbitrary URL injection via the unauthenticated endpoint.
-
-/**
- * Persist a detected release to artist_releases + release_links.
- * Idempotent: upserts on (artist_id, slug) and (release_id, platform).
- * Fire-and-forget — errors are logged but don't fail the response.
- */
-async function persistRelease(
-  artistName: string,
-  release: ReleaseResult,
-): Promise<void> {
-  const client = getClient();
-  if (!client) return;
-
-  try {
-    const slug = artistSlug(artistName);
-
-    // Look up the artist by slug
-    const { data: artist, error: artistError } = await client
-      .from('artists')
-      .select('id')
-      .eq('slug', slug)
-      .single();
-
-    if (artistError || !artist) {
-      // Artist not in DB — nothing to persist
-      return;
-    }
-
-    const artistId = artist.id;
-
-    // Validate that the release URL's domain matches a platform link already
-    // stored on this artist. This prevents arbitrary URL injection via the
-    // unauthenticated endpoint — only releases from known artist platform URLs
-    // are persisted.
-    const { data: artistLinks } = await client
-      .from('artist_links')
-      .select('platform, url')
-      .eq('artist_id', artistId);
-
-    if (!artistLinks || artistLinks.length === 0) return;
-
-    const releaseDomain = new URL(release.releaseUrl).hostname.replace(/^www\./, '');
-    const hasMatchingLink = artistLinks.some((link: { platform: string; url: string }) => {
-      try {
-        const linkDomain = new URL(link.url).hostname.replace(/^www\./, '');
-        return linkDomain === releaseDomain;
-      } catch {
-        return false;
-      }
-    });
-
-    if (!hasMatchingLink) {
-      console.log(`[check-releases] Skipping persist: release URL domain ${releaseDomain} does not match any artist_link`);
-      return;
-    }
-
-    const relSlug = releaseSlug(release.releaseName);
-    const releaseType = mapReleaseType(release.releaseType);
-
-    // Upsert the release
-    const { data: releaseRow, error: releaseError } = await client
-      .from('artist_releases')
-      .upsert(
-        {
-          artist_id: artistId,
-          title: release.releaseName,
-          slug: relSlug,
-          release_type: releaseType,
-          release_date: release.releaseDate,
-          source: 'auto',
-        },
-        { onConflict: 'artist_id,slug' }
-      )
-      .select('id')
-      .single();
-
-    if (releaseError || !releaseRow) {
-      console.error('[check-releases] Failed to upsert artist_releases:', releaseError?.message);
-      return;
-    }
-
-    // Upsert the release link for this platform
-    const { error: linkError } = await client
-      .from('release_links')
-      .upsert(
-        {
-          release_id: releaseRow.id,
-          platform: release.platform,
-          url: release.releaseUrl,
-          is_streaming: isStreamingPlatform(release.platform),
-          source: release.platform, // 'bandcamp' | 'faircamp' | 'mirlo'
-        },
-        { onConflict: 'release_id,platform' }
-      );
-
-    if (linkError) {
-      console.error('[check-releases] Failed to upsert release_links:', linkError.message);
-    }
-  } catch (err) {
-    console.error('[check-releases] persistRelease error:', err);
-  }
 }
 
 // Helper to fetch with timeout
@@ -466,7 +359,7 @@ export async function handler(event: {
   };
 
   // Rate limit: strict tier (10 req/min, 500/day) — this endpoint fetches
-  // external pages and writes to the DB, so it's expensive.
+  // external pages, so it's expensive.
   const ip = getClientIp(event.headers || {});
   const rl = await checkRateLimit(ip, 'strict', corsHeaders);
   if (rl.limited) return rl.response;
@@ -490,10 +383,7 @@ export async function handler(event: {
   if (!request.artistName || !request.platforms) {
     return {
       statusCode: 400,
-      headers: {
-        'Content-Type': 'application/json',
-        'Access-Control-Allow-Origin': '*',
-      },
+      headers: corsHeaders,
       body: JSON.stringify({ error: 'artistName and platforms are required' }),
     };
   }
@@ -506,23 +396,36 @@ export async function handler(event: {
   if (!hasPlatform) {
     return {
       statusCode: 400,
-      headers: {
-        'Content-Type': 'application/json',
-        'Access-Control-Allow-Origin': '*',
-      },
+      headers: corsHeaders,
       body: JSON.stringify({ error: 'At least one platform URL is required' }),
     };
+  }
+
+  // SSRF protection: validate every inbound platform URL against the hostname
+  // allowlist before any fetch. This endpoint is called by the Mac app and
+  // browser extension with no auth — without this check, it would be an open proxy.
+  const platformUrls = [
+    request.platforms.bandcamp,
+    request.platforms.faircamp,
+    request.platforms.mirlo,
+  ].filter(Boolean) as string[];
+
+  for (const url of platformUrls) {
+    if (!isUrlHostnameAllowed(url)) {
+      return {
+        statusCode: 400,
+        headers: corsHeaders,
+        body: JSON.stringify({ error: `URL not allowed: ${url}` }),
+      };
+    }
   }
 
   try {
     const release = await checkAllPlatforms(request.platforms);
 
-    // Persist to artist_releases + release_links (fire-and-forget, additive to jsonb)
-    if (release) {
-      persistRelease(request.artistName, release).catch(err =>
-        console.error('[check-releases] Background persist failed:', err)
-      );
-    }
+    // Note: this endpoint is read-only. It does NOT persist to the database.
+    // All writes to artist_releases + release_links go through persistSearchResults
+    // in db.ts, which runs from the authenticated search pipeline.
 
     const response: CheckReleasesResponse = {
       artistName: request.artistName,
--- a/api/functions/db.ts
+++ b/api/functions/db.ts
@@ -2,7 +2,7 @@
 // All operations are optional — if Supabase is not configured, they no-op gracefully.
 
 import { createClient, SupabaseClient } from '@supabase/supabase-js';
-import { mapReleaseType, releaseSlug, isStreamingPlatform } from './release-utils';
+import { mapReleaseType, releaseSlugWithCollision, isStreamingPlatform, normalizeReleaseTitle } from './release-utils';
 
 let supabase: SupabaseClient | null = null;
 
@@ -38,24 +38,38 @@ async function persistReleasesForArtist(
   artistId: string,
   platforms: PlatformLink[],
 ): Promise<void> {
-  // Group by normalized title for deduplication across platforms
+  // Group by normalized title for deduplication across platforms.
+  // Uses the shared normalizeReleaseTitle (includes .trim()) to avoid
+  // normalization drift between the migration script and the live pipeline.
   const byNormTitle = new Map<string, { title: string; type: 'album' | 'track'; platform: PlatformLink }[]>();
-  const normalize = (s: string) => s.toLowerCase().replace(/[^a-z0-9]+/g, '-').replace(/^-|-$/g, '');
 
   for (const p of platforms) {
     if (!p.latestRelease?.title) continue;
-    const norm = normalize(p.latestRelease.title);
+    const norm = normalizeReleaseTitle(p.latestRelease.title);
     if (!norm) continue;
     const existing = byNormTitle.get(norm) || [];
     existing.push({ title: p.latestRelease.title, type: p.latestRelease.type, platform: p });
     byNormTitle.set(norm, existing);
   }
 
+  // Fetch existing slugs for this artist to detect collisions (issue #3).
+  // Two different titles can produce the same slug (e.g. "Album Name" and
+  // "Album Name." both become "album-name"). The upsert on (artist_id, slug)
+  // would silently clobber the first with the second. We detect this and
+  // append a 6-char hash to disambiguate.
+  const { data: existingReleases } = await client
+    .from('artist_releases')
+    .select('slug,title')
+    .eq('artist_id', artistId);
+  const existingTitlesBySlug = new Map<string, string>(
+    (existingReleases || []).map((r: { slug: string; title: string }) => [r.slug, r.title])
+  );
+
   for (const [, entries] of byNormTitle) {
     // Prefer 'album' type if any platform reports it
     const bestType = entries.some(e => e.type === 'album') ? 'album' : entries[0].type;
     const title = entries[0].title;
-    const slug = releaseSlug(title);
+    const slug = releaseSlugWithCollision(title, existingTitlesBySlug);
 
     // Best artwork and date across platforms
     const artwork = entries.find(e => e.platform.latestRelease?.imageUrl)?.platform.latestRelease?.imageUrl || null;
@@ -105,17 +119,24 @@ async function persistReleasesForArtist(
         continue;
       }
 
-      // Upsert release links for each platform that had this release
+      // Upsert release links for each platform that had this release.
+      // Skip links where lr.url is missing — a release link pointing to the
+      // artist's profile page (the old fallback) is worse than no link.
       for (const entry of entries) {
-        const lr = entry.platform.latestRelease!;
+        if (!entry.platform.latestRelease) continue; // explicit guard (issue #8)
+        const lr = entry.platform.latestRelease;
+        if (!lr.url) continue; // skip if no release-specific URL (issue #5)
+
         const { error: linkError } = await client
           .from('release_links')
           .upsert(
             {
               release_id: releaseRow.id,
               platform: entry.platform.sourceId,
-              url: lr.url || entry.platform.url,
+              url: lr.url,
               is_streaming: isStreamingPlatform(entry.platform.sourceId),
+              // 'source' tracks provenance: 'auto' vs 'claimed' per spec.
+              // The platform name goes in the 'platform' column, not 'source' (issue #4).
               source: 'auto',
             },
             { onConflict: 'release_id,platform' }
--- a/api/functions/db.ts
+++ b/api/functions/db.ts
@@ -2,6 +2,7 @@
 // All operations are optional — if Supabase is not configured, they no-op gracefully.
 
 import { createClient, SupabaseClient } from '@supabase/supabase-js';
+import { mapReleaseType, releaseSlug, isStreamingPlatform } from './release-utils';
 
 let supabase: SupabaseClient | null = null;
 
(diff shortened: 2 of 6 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Security/regression review of the release-persistence changes.

## Security

**1. [confirmed] Unauthenticated endpoint now writes attacker-influenced data into a public, permanent table** — `api/functions/check-releases.ts`

`handler()` has no auth check and no rate limit (unlike `search-sources.ts`, which calls `checkRateLimit`). It accepts `artistName` and `platforms.{bandcamp,faircamp,mirlo}` as raw, unvalidated strings from the POST body, and `checkBandcamp`/`checkFaircamp`/`checkMirlo` fetch whatever URL is supplied with no `isUrlHostnameAllowed()` check (pre-existing, but previously the output was just returned to the caller — never persisted).

This PR adds `persistRelease(request.artistName, release)`, which resolves `artist_id` purely by slugifying the client-supplied `artistName` and upserts `release.releaseName` / `release.releaseUrl` from whatever URL the client pointed at into `artist_releases`/`release_links`. There's no check that the fetched release actually belongs to the named artist — a request like `{"artistName": "Real Existing Artist", "platforms": {"bandcamp": "https://some-unrelated-artist.bandcamp.com"}}` will happily attach that unrelated artist's release title/URL to the real artist's public record. Since this endpoint is called directly by the Mac app and browser extension (`https://unstream.stream/.netlify/functions/check-releases`) with no key, this is exploitable by anyone, at scale, to poison the release data that Phase 2/3 will surface on artist pages.
Recommend: require the caller to prove ownership of the artist (match against the platform URLs already stored on that artist in `artist_links`, not just a name match), and/or route through the SSRF allowlist, and/or gate the endpoint like other functions (`checkRateLimit` at minimum).

**2. [confirmed] New RLS policies grant write access to `anon`/`authenticated`, not just the service role** — `supabase/migrations/20260728120000_artist-releases.sql`, `supabase/schema.sql`

```sql
create policy "Service insert" on public.artist_releases for insert with check (true);
create policy "Service update" on public.artist_releases for update using (true);
create policy "Service delete" on public.artist_releases for delete using (true);
```
None of these specify `to service_role`. Per Postgres RLS semantics, a policy with no `TO` clause applies to `PUBLIC` — i.e. every role, including `anon`. The server functions here use the service-role key (`db.ts` `getClient()` reads `SUPABASE_SERVICE_KEY`), which bypasses RLS entirely regardless of policies, so these permissive policies do nothing for legitimate writes — their only real effect is that anyone holding the public anon key (shipped in the web/extension bundles) can insert/update/delete arbitrary rows in `artist_releases`/`release_links` directly via the Supabase REST API (fake releases, arbitrary outbound URLs, etc.).
This exact pattern already exists on `artists`/`artist_links` in the current schema, so it's not net-new to the codebase, but this PR extends the same exposure to two more tables rather than scoping the new policies `to service_role`. Worth fixing here rather than propagating further.

**3. [needs-manual-verification] Secret handling in `scripts/migrate-releases.ts`** — reads `SUPABASE_SERVICE_KEY` from env and never logs it; looks fine, but it's a one-off script intended to run against production — worth confirming it's only ever run locally/CI with the key injected via a secret store, never committed to shell history.

## Regression

**4. [confirmed] `mapReleaseType` is fed a constant, not the actual release type** — `api/functions/check-releases.ts`

```ts
const releaseType = mapReleaseType(release.platform === 'bandcamp' ? 'album' : 'album');
```
Both branches of the ternary are `'album'`, so `mapReleaseType`'s `'track' → 'single'` branch is unreachable from this call site — every release detected via `check-releases.ts` is persisted with `release_type = 'album'`, even singles/tracks. (`ReleaseResult` in this file doesn't even carry an album/track distinction, unlike the `latest_release.type` field used in `db.ts`'s parallel `persistReleasesForArtist`.) This looks like a copy/paste leftover from the `db.ts` version of the same helper.

**5. [needs-manual-verification] Triplicated logic drifting across three files** — `mapReleaseType`, `releaseSlug`/`slugify`, `STREAMING_PLATFORMS`/`isStreamingPlatform` are independently redefined in `api/functions/check-releases.ts`, `api/functions/db.ts`, and `scripts/migrate-releases.ts`. Not a bug today, but the next platform added to `STREAMING_PLATFORMS` (or fix to date parsing) has to be applied in three places — easy to miss one and get silently inconsistent `is_streaming`/`release_type` values across the auto-search path vs. the check-releases path vs. the one-time migration.


---

api/functions/check-releases.ts
```diff
@@ -139,6 +149,7 @@ async function checkBandcamp(artistUrl: string): Promise<ReleaseResult | null> {
       releaseDate,
       releaseUrl: fullUrl,
       platform: 'bandcamp',
+      releaseType: bandcampReleaseType(fullUrl),
```
brandonlucasgreen, line 152: **[plausible] SSRF: the allowlist check added in this PR doesn't cover `checkBandcamp`'s second fetch (lines 125-128)**

The SSRF guard added lower in this file (~line 404) validates `request.platforms.bandcamp/faircamp/mirlo` with `isUrlHostnameAllowed()` before `checkAllPlatforms()` runs. But `checkBandcamp` does a *second* fetch to the album page, and that URL is never re-validated:

```ts
// lines 125, 128
const fullUrl = href.startsWith('http') ? href : new URL(href, baseUrl).toString();
const albumResponse = await fetchWithTimeout(fullUrl, { ... });   // no isUrlHostnameAllowed() call
```

`href` comes from `link.getAttribute('href')` scraped out of the first page's HTML (`.music-grid-item a`). When it's a relative path, `new URL(href, baseUrl)` stays same-origin as the already-validated `baseUrl`, which is safe. But when `href.startsWith('http')` (an absolute URL), `fullUrl = href` is used verbatim with **no allowlist check at all** — not even a same-hostname check against `baseUrl`.

`*.bandcamp.com` is a wildcard entry in `ALLOWED_OUTBOUND_HOSTNAMES` (`middleware.ts`), so anyone can pass the *first* check with their own free `*.bandcamp.com` page. If that page's release-grid markup (or a Bandcamp Pro custom domain) can be made to emit an absolute `href` pointing off-allowlist, this unauthenticated endpoint (called by the Mac app/extension) will fetch whatever that points to. This is a confirmed code-level gap — I can't verify from this repo how much control an artist has over that specific markup on Bandcamp's platform, so I'd call exploitability **plausible / needs-manual-verification** rather than fully confirmed. Given this PR explicitly adds SSRF protection "before any fetch" for this endpoint, this second hop should go through `isUrlHostnameAllowed()` too (or at minimum be pinned to the same hostname as `baseUrl`).

@brandonlucasgreen

---

api/functions/db.ts
```diff
@@ -24,6 +25,133 @@ export function artistSlug(name: string): string {
...
+  // Fetch existing slugs for this artist to detect collisions (issue #3).
+  // Two different titles can produce the same slug (e.g. "Album Name" and
+  // "Album Name." both become "album-name"). The upsert on (artist_id, slug)
+  // would silently clobber the first with the second. We detect this and
+  // append a 6-char hash to disambiguate.
+  const { data: existingReleases } = await client
+    .from('artist_releases')
+    .select('slug,title')
+    .eq('artist_id', artistId);
+  const existingTitlesBySlug = new Map<string, string>(
```
brandonlucasgreen, line 64: **[plausible] Regression / state-transition edge case: TOCTOU race in slug-collision detection can silently clobber a different release's data**

`existingTitlesBySlug` is fetched once (lines 60-66) and then used to compute collision-free slugs for every release found in this call (`releaseSlugWithCollision` at line 72), but the snapshot is never refreshed as releases are upserted in the loop below, and there's no locking between concurrent invocations.

`persistSearchResults` runs from the search pipeline, so `persistReleasesForArtist` can plausibly be invoked twice concurrently for the same `artist_id` — e.g. two users searching the same popular artist at close to the same time, each request hitting a cold cache. Both calls read the same `existingTitlesBySlug` snapshot (missing each other's in-flight inserts), both compute a slug that looks collision-free from their own snapshot, and both `upsert(..., { onConflict: 'artist_id,slug' })`. If the two concurrent runs happen to produce releases that land on the same slug (the exact scenario this collision detection exists to prevent), the second upsert silently overwrites the first release's `title`/`release_type`/`release_date`/`artwork_url` — and any `release_links` already written against the first release's `id` now point at a row with the wrong title, with no error surfaced anywhere.

This is the same "never cache uncertainty" class of bug called out in CLAUDE.md, just applied to a check-then-act read instead of a cache — the collision check is correct against *already-committed* data but not race-safe against concurrent writers. Since I can't confirm from the code alone how likely two searches for the same artist are to overlap in production, I'd flag this as **plausible / needs-manual-verification** rather than confirmed. `scripts/migrate-releases.ts` has the identical pattern, but is lower risk since it's a single-threaded one-time run.

@brandonlucasgreen

---

brandonlucasgreen (commented):
@brandonlucasgreen Security + regression pass on the release-persistence PR. Two inline findings (SSRF gap in `check-releases.ts`, TOCTOU race in `db.ts`); rest of the diff (migrations, RLS policies, rate limiting, `release-utils.ts`) looked solid.

## Security

1. **[plausible] SSRF** — `checkBandcamp`'s second fetch (album page, `check-releases.ts:125-128`) isn't covered by the allowlist check this PR adds. See inline comment. Category: injection/SSRF.
2. **[needs-manual-verification] Input validation — field/host mismatch** — `isUrlHostnameAllowed()` validates `platforms.bandcamp/faircamp/mirlo` against the *general* outbound allowlist (`middleware.ts`), not that the URL actually belongs to that specific platform. e.g. `platforms.bandcamp` set to a `patreon.com` or `beatport.com` URL still passes, and `checkBandcamp` will happily fetch `<that-host>/music`. Not an open proxy (still bounded by the allowlist) and low real-world impact, but worth a second look since the field name implies a platform-specific trust boundary that isn't actually enforced.
3. Rate limiting (`strict` tier, IP-keyed) and the SSRF allowlist check added to `check-releases.ts` are both correctly placed *before* body parsing / before any fetch for the paths they do cover — good.
4. Migration 029 RLS: public read + service-role-only write, consistent with the existing `artists`/`artist_links` pattern; no anon/authenticated write policies. No secrets or PII touched by these tables. No SQL injection surface — all writes go through the Supabase client with parameterized upserts.

## Regression

1. **[plausible] TOCTOU race → silent data clobber** — `persistReleasesForArtist`'s collision-detection snapshot (`db.ts:60-72`) isn't race-safe against concurrent search requests for the same artist. See inline comment. Same pattern is duplicated in `scripts/migrate-releases.ts` but lower risk there (single-threaded one-time run).
2. Edge cases around empty titles, missing `latestRelease.url`, and missing `latestRelease` (issues #4/#5/#8 per the inline code comments) are already explicitly guarded — good defensive coverage.
3. API contract: `releaseType` is an additive field on `ReleaseResult`/`CheckReleasesResponse` — backward compatible for existing Mac app / extension consumers.

Everything else — index coverage, `unique(artist_id, slug)` / `unique(release_id, platform)` conflict targets, idempotent upserts, the `IF NOT EXISTS` guards in the migration — looked correct.

---

brandonlucasgreen (commented):

## Security

**1. [needs-manual-verification] `checkBandcamp`'s second fetch (the album page) isn't re-validated against the SSRF allowlist the way the PR's headline fix implies.**

The new handler-level check validates the three *inbound* platform URLs (`request.platforms.{bandcamp,faircamp,mirlo}`) via `isUrlHostnameAllowed()` before calling `checkAllPlatforms` — good, and it closes what the PR's own comment correctly calls out as "this endpoint... would be an open proxy" without it. `checkFaircamp` and `checkMirlo` only ever fetch fixed, same-host, or hardcoded URLs after that point, so they're fully covered.

`checkBandcamp` is different: after validating the artist URL, it scrapes an `href` out of the artist's `/music` page and builds `fullUrl = href.startsWith('http') ? href : new URL(href, baseUrl).toString()`, then fetches `fullUrl` directly — without running it back through `isUrlHostnameAllowed()`. In the normal case `href` is a same-domain relative link from Bandcamp's own template, so this isn't attacker-reachable through the request body (the requester doesn't control page content). But it means the allowlist guarantee this PR is adding isn't actually end-to-end for this handler — worth either re-validating `fullUrl` before the second fetch, or leaving a comment explaining why it's considered safe to skip (e.g. "href always same-origin, verified because...").

`api/functions/check-releases.ts`, `checkBandcamp` (the `fullUrl` fetch, existing code — not new in this PR, but this PR is what added the allowlist check elsewhere in the same function and is the natural place to close the gap).

The rate limiting (`checkRateLimit(ip, 'strict', ...)`) is correctly placed ahead of body parsing so it can't be bypassed by a malformed request, and the RLS policies on the two new tables (public read / service-role-only write, no anon policies) match the existing `artists`/`artist_links` pattern.

## Regression

Nothing confirmed. The slug-collision handling (`releaseSlugWithCollision`) is correctly defensive, though worth noting `existingTitlesBySlug` is fetched once per `persistReleasesForArtist`/migration-artist call and not updated as releases are inserted within that same call — currently safe only because `releaseSlug()` and `normalizeReleaseTitle()` apply materially the same transform, so two distinct dedup groups can't collide on the same base slug within one call. That coupling isn't enforced anywhere, so a future edit to either function that lets them diverge would silently reintroduce the collision (the `onConflict: 'artist_id,slug'` upsert would then clobber the earlier release). Not an active bug today, just fragile — a comment noting the two functions must stay in lockstep, or a shared implementation, would remove the ambiguity.

Test/verification coverage per the PR description (idempotent migration script, `npm run test:api`, `tsc -b`) looks appropriately scoped for a data-foundation PR with no user-facing surface yet.

---

brandonlucasgreen (commented):

**Security**
- Adding `isUrlHostnameAllowed()` SSRF validation plus `strict`-tier rate limiting to `check-releases.ts` is a solid fix — this was previously an unauthenticated endpoint that fetched caller-supplied URLs with no allowlist check. Confirmed correct, not a new issue.
- **needs-manual-verification**: the new hostname check validates the *input* URL against the allowlist before calling `checkBandcamp`/`checkFaircamp`/`checkMirlo`, but those fetchers weren't touched by this diff. If their underlying `fetch()` calls follow redirects (Node's default), an allowlisted URL could 302 to a non-allowlisted host and bypass the new check. Worth confirming those fetchers disable redirect-following or that the final resolved URL is re-checked.
- RLS/migration hygiene looks right: `IF NOT EXISTS` guards, public-read/service-write policies matching the existing `artists`/`artist_links` pattern, no new columns needing extra RLS thought.

**Regression**
- `releaseSlugWithCollision` (`api/functions/release-utils.ts`) — **confirmed**: the collision-disambiguation "hash" is `Buffer.from(title).toString('hex').slice(0, 6)`, i.e. the literal hex encoding of the title's first 3 bytes — not an actual hash function, so it has no collision resistance. Two different titles that already collide on `releaseSlug()` almost always share a common prefix (that's typically *why* they collided — e.g. differing only by trailing punctuation like "Album Name" vs. "Album Name."), so they'll produce the *same* disambiguation suffix too, and the upsert on `(artist_id, slug)` will silently overwrite one release with the other instead of disambiguating them. The function doesn't provide the collision resistance its own comment claims.
- `persistReleasesForArtist`'s "prefer album" heuristic (`entries.some(e => e.type === 'album') ? 'album' : entries[0].type` in `api/functions/db.ts`) combined with `checkFaircamp`/`checkMirlo` hardcoding `releaseType: 'album'` regardless of the actual release (`api/functions/check-releases.ts`) means any single/EP that's cross-listed on Faircamp or Mirlo gets persisted as `release_type: 'album'` even when Bandcamp correctly identifies it as a track. **Confirmed** — this is systematic (any artist with non-album releases cross-listed on those two platforms), not a rare edge case.

---

brandonlucasgreen (commented):

**Security**

1. **[confirmed]** SSRF allowlist bypass via a second-hop fetch in `checkBandcamp` (`api/functions/check-releases.ts`). The handler's `isUrlHostnameAllowed()` loop only validates the *input* `request.platforms.bandcamp/faircamp/mirlo` URLs. `checkBandcamp` then scrapes the artist's own music page and does a second, unvalidated fetch:
   ```ts
   const fullUrl = href.startsWith('http') ? href : new URL(href, baseUrl).toString();
   const albumResponse = await fetchWithTimeout(fullUrl, {...});
   ```
   `href` comes from HTML on that page — content anyone can control by registering a `*.bandcamp.com` subdomain and editing it. An absolute or protocol-relative `href` resolves to an arbitrary host via `new URL()` and gets fetched server-side with no allowlist check. This predates this PR functionally, but this PR's new comment ("without this check, it would be an open proxy") implies the endpoint is now protected when this path isn't — worth closing in this PR or a fast follow-up.

2. **[confirmed]** Collision map is stale within a single persist batch. `persistReleasesForArtist` (`api/functions/db.ts`) and `scripts/migrate-releases.ts` both fetch `existingTitlesBySlug` once before looping over releases and never update it as releases are upserted in the same run. Two releases processed in the same call that produce the same `releaseSlug()` output (e.g. punctuation variants) won't be caught by the collision guard — the second `upsert(... onConflict: 'artist_id,slug')` silently overwrites the first row's `title`/`release_date`/`artwork_url`, while the first release's `release_links` stay attached to the now-mismatched row. This defeats the collision-guard fix the code comments describe.

3. **[confirmed, informational]** The new migration's RLS comment says the write policy matches the existing `artists`/`artist_links` pattern, but those tables' insert policies have no role restriction (defaulting to PUBLIC), while the new tables correctly scope writes `to service_role`. Not a regression — the new tables are actually stricter — but the comment undersells a real pre-existing gap on the older tables, worth a separate look.

4. Secrets/injection: clean. `scripts/migrate-releases.ts` reads `SUPABASE_SERVICE_KEY` from env only, never logs/hardcodes it; all writes go through the parameterized Supabase client, no raw SQL interpolation.

**Regression risk**

5. **[plausible]** Non-ASCII title normalization in `release-utils.ts` strips accented characters as separators rather than transliterating ("Café" → `caf`, "Cafe" → `cafe`), so the same release across platforms can fail to merge, while punctuation-heavy distinct titles can collapse together.

6. **[plausible]** When titles merge, the "best" date/artwork is picked via `entries.find(...)` — first match by array order, not most recent — so a reissue merging with an original release can silently take the wrong date/artwork with no discrepancy surfaced.

7. **[needs-manual-verification]** New rate limit on `check-releases.ts` (10/min, 500/day per IP) is a behavior change for what looks like a background-polling endpoint; if a client (e.g. the Mac app) checks many followed artists from one IP, it could get silently rate-limited. The polling caller isn't in this diff to confirm real-world impact.

---

brandonlucasgreen (commented):
@brandonlucasgreen Automated staff-eng review.

## Security

- **SSRF — confirmed correct**: `check-releases.ts` now validates every inbound `request.platforms.*` URL against `isUrlHostnameAllowed()` before `checkAllPlatforms()` fetches it. Given this endpoint is called unauthenticated by the Mac app and browser extension, this closes a real open-proxy gap. Correct fix.
- **Auth/rate limiting — confirmed correct**: added `checkRateLimit(ip, 'strict', corsHeaders)` ahead of the existing external fetches. Appropriate given the endpoint has no auth and does outbound HTTP per call.
- **Secret handling — confirmed correct**: `migrate-releases.ts` takes `SUPABASE_SERVICE_KEY` from env only, with a comment warning against passing it via shell args/history.
- **RLS — confirmed correct**: both new tables (`artist_releases`, `release_links`) enable RLS with public `select`, and `insert`/`update`/`delete` scoped to `to service_role` only — consistent with the existing `artists`/`artist_links` pattern. No policy gap for anon/authenticated write.
- **Injection**: all new DB access goes through the Supabase client's parameterized `upsert`/`select`, no raw SQL interpolation. Confirmed clean.

## Regression — CONFIRMED bug

`releaseSlugWithCollision()` in `api/functions/release-utils.ts` doesn't actually prevent the collision it's named for (the thing its own comment calls "issue #3"):

```js
const hash = Buffer.from(title).toString('hex').slice(0, 6);
```

`Buffer.from(title).toString('hex')` hex-encodes the *whole* title; `.slice(0, 6)` then keeps only the first 6 hex chars, i.e. the encoding of just the **first 3 characters** of the title. It is not a hash of the full string. But two titles that collide on the base slug almost always share the *same leading characters* (the whole reason they collide is a punctuation/case/whitespace difference elsewhere in the string) — so the "disambiguation" suffix collides too. Concretely: `"Live at Nowhere"` and `"Live At Nowhere!"` both slugify to `live-at-nowhere`, and both produce the same 6-hex-char suffix (both start `Liv`), so the second `upsert` (`onConflict: 'artist_id,slug'`) silently overwrites the first release row's title/artwork/date with the second one's — the exact clobbering bug this code exists to prevent.

Compounding it: `existingTitlesBySlug` (in both `persistReleasesForArtist` in `db.ts` and the per-artist loop in `scripts/migrate-releases.ts`) is fetched **once** from the DB before the loop and never updated as releases are inserted within that same call/run. Two *new* releases in the same batch that collide only with each other (not with anything already in the DB) won't be flagged as colliding at all, since neither is in the map yet.

Recommend: derive the suffix from a real hash of the full normalized title (or just use the generated row id / a random suffix), and update `existingTitlesBySlug` after each insert inside the loop.

## Other

- **Test coverage**: `release-utils.ts` is new pure-logic code (explicitly modeled in its own header comment on the tested `search-utils.ts` pattern) but ships with zero unit tests. A single test on `releaseSlugWithCollision` with two colliding, same-prefix titles would have caught the bug above immediately.
- **Doc drift — needs-manual-verification**: `supabase/schema.sql`'s comment on `release_links.source` lists platform names as possible values (`'auto' | 'claimed' | 'bandcamp' | 'mirlo' | 'musicbrainz' | 'itunes'`), while migration `20260728120000_artist-releases.sql`'s comment on the same column says "platform provenance is in the platform column" (i.e. only `'auto' | 'claimed'`). Since `schema.sql` is supposed to mirror the migrations for fresh installs, these two should say the same thing — as written they contradict each other about the column's contract.
- **Off-by-one risk — needs-manual-verification**: the duplicated date-parsing logic (`persistReleasesForArtist` in `db.ts`, and the equivalent block in `migrate-releases.ts`) does `new Date(monthDayYearString)` then `.toISOString().split('T')[0]`. `new Date(string)` parses in the process's local timezone; if that timezone isn't UTC, a date near local midnight can shift by a day once converted to UTC. Likely a non-issue if the Netlify Functions runtime is UTC (typical), but there's no explicit UTC parsing here, so worth a quick confirmation.
- **Typecheck coverage gap**: none of `release-utils.ts`, the new `persistReleasesForArtist` in `db.ts`, or the `check-releases.ts` changes were added to `api/tsconfig.json`'s narrow `include` list (unlike PR #343, which added its touched files). Per this repo's own documented gotcha, that means this new code — including the bug above — isn't typechecked by CI and only surfaces at runtime. Worth adding, especially given the confirmed bug found here.

---

brandonlucasgreen (commented):
@brandonlucasgreen Staff-engineer review pass (security + regression), based only on the diff and the current state of the branch.

## Security

- **`api/functions/check-releases.ts` — SSRF allowlist covers only the first hop** — `needs-manual-verification`
  This PR adds `isUrlHostnameAllowed()` validation on the inbound `platforms.bandcamp/faircamp/mirlo` URLs before any fetch, closing what the PR comment calls an "open proxy" risk. That's correct for the *first* fetch. But `checkBandcamp()` (unchanged by this PR) does a second fetch to `fullUrl`, which is built from an `href` scraped out of the first page's HTML — and when that `href` is absolute (`href.startsWith('http')`), it's used verbatim with no re-check against `isUrlHostnameAllowed`. In practice a legitimate `*.bandcamp.com` page won't emit a cross-host absolute link here, so exploitability is low, but the guarantee the PR describes ("without this check, it would be an open proxy") isn't actually complete for this handler. Worth a conscious call on whether the second-hop URL should be validated too, now that the allowlist pattern is established for this endpoint.

- No new auth/authz, input-validation, or secret-handling issues found in `db.ts`, `release-utils.ts`, or the migration. RLS on both new tables (public read / service-role write, no anon write) matches the existing `artists`/`artist_links` pattern, and `scripts/migrate-releases.ts` documents the service-key handling correctly.

## Regression

- **`releaseSlugWithCollision()` doesn't actually hash the title — confirmed**, `api/functions/release-utils.ts` (also used from `scripts/migrate-releases.ts`):
  ```ts
  const hash = Buffer.from(title).toString('hex').slice(0, 6);
  ```
  `Buffer.from(title).toString('hex')` hex-encodes the *entire* string; `.slice(0, 6)` then just takes the first 6 hex characters — i.e. the first 3 bytes of the title, not a hash of it. Two titles that share a common prefix produce the *identical* disambiguation suffix. That includes the exact motivating example in the function's own comment — `"Album Name"` vs `"Album Name."` — both start with `"Alb"` (`0x41 0x6c 0x62`), so both get suffix `416c62` and the slugs still collide on `unique(artist_id, slug)`. The upsert (`onConflict: 'artist_id,slug'`) then silently overwrites one release's title/artwork/date with the other's instead of disambiguating — the exact data-loss scenario "issue #3" was meant to prevent.

- **Collision detection doesn't see same-batch collisions — plausible**, `persistReleasesForArtist()` in `api/functions/db.ts`: `existingTitlesBySlug` is fetched once from the DB *before* looping over the current call's `byNormTitle` groups, so it only catches collisions against releases persisted in earlier calls. If two colliding titles (e.g. the same "Album Name" / "Album Name." pair) arrive for the same artist within a single `persistSearchResults` batch, the map won't contain the first one yet when the second is upserted, and detection is skipped for that pair too. Compounds the finding above.

Nothing else stood out for off-by-one logic, state transitions, or API contract drift — the new fields/tables are additive and the jsonb fallback path is untouched.

---

api/functions/check-releases.ts
brandonlucasgreen, line 152: **Security — SSRF, confirmed (code-level gap)**

This PR adds an SSRF allowlist check on `request.platforms.*` before `checkAllPlatforms()` runs (below, "SSRF protection: validate every inbound platform URL... before any fetch"). But `checkBandcamp()` (a few lines up) does a *second*, unchecked fetch: `fullUrl` is built from `href`, a value scraped out of the artist's own `/music` page HTML, and when `href.startsWith('http')` it's used as-is with no `isUrlHostnameAllowed(fullUrl)` call before the follow-up `fetchWithTimeout(fullUrl, …)` that fetches the album page.

Since the entry URL only has to be *some* `*.bandcamp.com` subdomain (free to create), if that page's `.music-grid-item a` markup can ever contain an attacker-supplied absolute URL, this endpoint fetches it server-side with no allowlist check on the second hop — the exact open-proxy pattern the new check is meant to close. The endpoint is unauthenticated by design (Mac app/extension), so it's reachable by anyone. Actual exploitability depends on how much of Bandcamp's grid markup a seller can control, which I can't verify from this repo, but the code doesn't do what the comment two paragraphs down claims ("before any fetch"). Suggest re-validating `fullUrl` with `isUrlHostnameAllowed()` before this second fetch — it's a no-op for legitimate same-host links, so it's free defense-in-depth.

@brandonlucasgreen

---

api/functions/check-releases.ts
```diff
@@ -1,4 +1,13 @@
 import { parse } from 'node-html-parser';
+import { isUrlHostnameAllowed } from './middleware.js';
```
brandonlucasgreen, line 2: **Regression risk — needs-manual-verification**

These three new imports (`middleware.js`, `release-utils.js`, `ratelimit.js`) use an explicit `.js` extension. Every other relative import in `api/functions/` (including this PR's own `db.ts` change, which imports `from './release-utils'` with no extension) omits it — `git grep "from '\./"` across `api/functions/*.ts` on `main` shows 100% no-extension. If the Netlify build's module resolution doesn't map `.js` specifiers back to the sibling `.ts` file the way it does elsewhere, this fails only at runtime, not at build time — `api/tsconfig.json` has a narrow `include` (per CLAUDE.md, only `me-*` functions), so `tsc -b` won't catch a broken import here either. Worth confirming this file actually loads in a Netlify deploy preview before merging, since a break wouldn't show up in `npm run build`.

@brandonlucasgreen

---

api/functions/db.ts
```diff
@@ -24,6 +25,133 @@ export function artistSlug(name: string): string {
...
+    existing.push({ title: p.latestRelease.title, type: p.latestRelease.type, platform: p });
+    byNormTitle.set(norm, existing);
+  }
+
+  // Fetch existing slugs for this artist to detect collisions (issue #3).
+  // Two different titles can produce the same slug (e.g. "Album Name" and
+  // "Album Name." both become "album-name"). The upsert on (artist_id, slug)
+  // would silently clobber the first with the second. We detect this and
+  // append a 6-char hash to disambiguate.
+  const { data: existingReleases } = await client
```
brandonlucasgreen, line 60: **Regression — state transition / race, plausible**

The collision-avoidance scheme (`releaseSlugWithCollision`, "issue #3") reads `existingReleases` once here, then loops over `byNormTitle` upserting on `(artist_id, slug)`. This is a check-then-act, not atomic: two concurrent searches that both persist results for the *same artist* (a popular artist searched by two users at once, which is exactly the traffic pattern that makes an artist popular enough to have ambiguous release titles) will both read the same pre-upsert snapshot, both compute the same slug for two genuinely different titles that collide (e.g. "Reissue" vs "Reissue!"), and the second upsert will silently overwrite the first's row via `onConflict: 'artist_id,slug'` — losing one of the two distinct releases. That's the same silent-clobber failure mode this code is explicitly trying to prevent, just moved from "different titles, same slug" to "different titles, same slug, *and* concurrent."

Narrow trigger condition (needs two colliding titles arriving in the same race window), so labeling this plausible rather than confirmed-in-practice — but the code path itself is real and traceable. A unique constraint violation + retry-with-suffix, or a serializable transaction, would close the window; the current select-then-upsert can't.

@brandonlucasgreen

---

api/functions/check-releases.ts
```diff
@@ -374,17 +396,37 @@ export async function handler(event: {
...
+  // SSRF protection: validate every inbound platform URL against the hostname
+  // allowlist before any fetch. This endpoint is called by the Mac app and
+  // browser extension with no auth — without this check, it would be an open proxy.
+  const platformUrls = [
+    request.platforms.bandcamp,
+    request.platforms.faircamp,
+    request.platforms.mirlo,
+  ].filter(Boolean) as string[];
+
+  for (const url of platformUrls) {
```
brandonlucasgreen, line 413: **Regression — edge case, needs-manual-verification**

Before this PR, `check-releases.ts` accepted any `platforms.faircamp` URL. `isUrlHostnameAllowed()`'s Faircamp rule (in `middleware.ts`) only allows a hostname where `"faircamp"` appears as a non-leftmost label (e.g. `artist.faircamp.net`), which won't match a self-hosted Faircamp instance on an unrelated custom domain — and Faircamp is designed to be self-hostable anywhere. If any artists already in the DB have a Faircamp URL like that, release checks for them now silently 400 instead of returning `release: null`. That's a pre-existing constraint of the shared allowlist, not something this PR changes, but this PR is what newly exposes this endpoint to it — worth checking whether real `artist_links.faircamp` URLs in production all match the allowlist's pattern before this ships.

@brandonlucasgreen

---

brandonlucasgreen (commented):
@brandonlucasgreen Security + regression pass on the releases data-foundation PR. Four inline findings, most-severe first:

**Security**
- `confirmed` — `check-releases.ts`: the new SSRF allowlist check covers the entry `platforms.bandcamp/faircamp/mirlo` URL, but `checkBandcamp()`'s second fetch (the album page, built from a scraped `href`) is never re-validated against `isUrlHostnameAllowed()`. The endpoint is unauthenticated by design, so this is the exact open-proxy shape the new check is meant to close, just one hop later. Practical exploitability turns on how much of Bandcamp's own grid markup a seller can control, which I can't verify from the repo — but the code doesn't match its own "before any fetch" comment.
- Everything else checked clean: no auth/authz changes, no injection (all DB access is via the Supabase client, no string-built SQL), no secrets in the diff (the migration script's own comment correctly warns against passing `SUPABASE_SERVICE_KEY` via shell args), RLS on both new tables matches the existing public-read/service-write pattern with no gaps.

**Regression**
- `plausible` — `db.ts` `persistReleasesForArtist`: the collision-avoidance slug check is a read-then-upsert with no transaction/lock, so two concurrent searches for the same artist can race past each other and silently clobber one of two distinct, colliding-slug releases — the same failure class the check exists to prevent, just under concurrency.
- `needs-manual-verification` — `check-releases.ts`'s new imports use an explicit `.js` extension, unlike 100% of existing relative imports elsewhere in `api/functions/` (including this same PR's `db.ts` change). This file isn't in `api/tsconfig.json`'s narrow include, so a resolution mismatch would only surface at runtime in production, not in CI.
- `needs-manual-verification` — the new allowlist check on this endpoint will reject any Faircamp URL that doesn't have "faircamp" as a hostname label, which won't match every self-hostable Faircamp domain. Pre-existing allowlist behavior, but newly enforced here — worth confirming against real `artist_links` data before shipping.

Nothing here should block if the SSRF gap gets a defense-in-depth fix on the second fetch; the rest are worth a look but not launch-blocking.

---

api/functions/check-releases.ts
```diff
@@ -374,17 +396,37 @@ export async function handler(event: {
...
+    if (!isUrlHostnameAllowed(url)) {
+      return {
+        statusCode: 400,
+        headers: corsHeaders,
+        body: JSON.stringify({ error: `URL not allowed: ${url}` }),
+      };
+    }
+  }
```
brandonlucasgreen, line 421: @brandonlucasgreen **Security — SSRF, confirmed**

This allowlist check only validates the *initial* URL string. It doesn't prevent SSRF via HTTP redirect, and combined with an existing gap in `isUrlHostnameAllowed` (`api/functions/middleware.ts:403-407`), that's exploitable end-to-end:

1. `middleware.ts`'s faircamp allowance is `labels.indexOf('faircamp') > 0` — a *non-leftmost label* check, not "the domain is a real Faircamp instance." An attacker who owns any domain can pass this by naming a subdomain `faircamp`, e.g. `release.faircamp.attacker.com` — fully attacker-controlled infrastructure, and it passes `isUrlHostnameAllowed`.
2. `checkFaircamp` (line 163) then does `fetchWithTimeout(`${baseUrl}/feed.rss`, ...)`. Since `fetch()` defaults to `redirect: 'follow'`, the attacker's own server can respond to that request with a 302 to `http://169.254.169.254/latest/meta-data/...` or an internal service — and Node will follow it. Nothing here validates the resolved destination, only the URL the caller supplied.
3. Same gap applies to `checkBandcamp`'s second fetch (line 128: `fetchWithTimeout(fullUrl, ...)`), where `fullUrl` is built from an `href` scraped out of the *first* fetch's HTML response and is never re-checked against the allowlist before being fetched.

The comment above ("without this check, it would be an open proxy") is right that this endpoint needed protection — it just isn't sufficient as written, because the check happens once at the boundary and the actual fetches can be redirected past it. A fix needs either `redirect: 'manual'` + re-validating the `Location` target in a loop, or tightening the faircamp label check to require an anchored/known-good domain shape.

Verdict: **confirmed** (demonstrable from the code: `fetch`'s default redirect behavior + the label-based bypass in `middleware.ts` are both there today, independent of any live test).

---

api/functions/release-utils.ts
```diff
@@ -0,0 +1,74 @@
...
+  existingTitlesBySlug: Map<string, string>,
+): string {
+  const base = releaseSlug(title);
+  const existingTitle = existingTitlesBySlug.get(base);
+  if (existingTitle && existingTitle !== title) {
+    // Collision: a different title produced the same slug. Append a 6-char hash.
+    const hash = Buffer.from(title)
+      .toString('hex')
+      .slice(0, 6);
+    return `${base}-${hash}`;
```
brandonlucasgreen, line 46: @brandonlucasgreen **Regression / data integrity — confirmed**

This isn't a hash — `Buffer.from(title).toString('hex').slice(0, 6)` is the hex encoding of the **first 3 bytes** of `title`, i.e. just its prefix. For the exact scenario in the docstring above and in `db.ts:56-57` ("Album Name" vs "Album Name."), both titles share the same first 3 bytes, so they produce the *same* disambiguating suffix:

```
releaseSlug("Album Name")   -> "album-name"
releaseSlug("Album Name.")  -> "album-name"
hash("Album Name")  -> "416c62"
hash("Album Name.") -> "416c62"   // identical
```

Both end up as `album-name-416c62`. The `onConflict: 'artist_id,slug'` upsert in `db.ts` then does exactly the thing this function was written to prevent: the second release silently overwrites the first `artist_releases` row (title, release_date, artwork all get replaced), and any `release_links` already pointing at that release_id now describe the wrong release. Collisions are most likely to share a prefix (that's usually *why* they collide after slugification strips trailing punctuation/whitespace), so this misses the common case, not just a rare corner.

Verdict: **confirmed** — reproduced the exact docstring example above by hand.

---

api/functions/db.ts
```diff
@@ -24,6 +25,133 @@ export function artistSlug(name: string): string {
...
+    .from('artist_releases')
+    .select('slug,title')
+    .eq('artist_id', artistId);
+  const existingTitlesBySlug = new Map<string, string>(
+    (existingReleases || []).map((r: { slug: string; title: string }) => [r.slug, r.title])
+  );
```
brandonlucasgreen, line 66: @brandonlucasgreen **Regression / edge case — confirmed**

`existingTitlesBySlug` is snapshotted once, before the `for (const [, entries] of byNormTitle)` loop below. If two colliding titles (e.g. "Album Name" and "Album Name.") show up for the artist in the *same* `persistReleasesForArtist` call — which is exactly the first-time-discovery case a fresh search would hit — the second one won't see the first one's slug in `existingTitlesBySlug` at all (it was only populated from what's already in the DB before this call started), so `releaseSlugWithCollision` won't even detect a collision on this pass. Both get inserted as plain `album-name`, and the second upsert clobbers the first. This compounds the hash-prefix issue flagged in `release-utils.ts` — even if that were fixed, same-batch collisions still slip through here because the lookup map is never updated after each upsert in the loop.

Same pattern is duplicated in `scripts/migrate-releases.ts` (`existingTitlesBySlug` fetched once per artist, before its own inner loop).

Verdict: **confirmed** from reading the control flow — `existingTitlesBySlug` has no write-back after each iteration's upsert.

---

brandonlucasgreen (commented):
@brandonlucasgreen Automated security + regression pass on this PR. Three inline findings, all **confirmed** from reading the code (no live exploitation attempted):

1. **SSRF via redirect** (`check-releases.ts`) — the new `isUrlHostnameAllowed` gate only checks the URL the caller supplies; `fetch()` follows redirects by default, and `checkBandcamp`'s second-hop fetch + `checkFaircamp`'s RSS fetch never re-validate the resolved destination. Combined with the pre-existing faircamp label check in `middleware.ts` (`labels.indexOf('faircamp') > 0`, which any domain owner can satisfy by naming a subdomain "faircamp"), this is an exploitable path to internal services/cloud metadata, not just a defense-in-depth gap.
2. **Slug "hash" is a prefix, not a hash** (`release-utils.ts`) — `Buffer.from(title).toString('hex').slice(0,6)` is the first 3 bytes of the title. Colliding titles almost always share a prefix (that's usually *why* they collide after slugification), so the disambiguation suffix is often identical for exactly the titles it's meant to distinguish. Reproduced with the PR's own docstring example ("Album Name" / "Album Name.").
3. **Same-batch collisions aren't checked at all** (`db.ts`) — `existingTitlesBySlug` is fetched once before the loop and never updated as releases are upserted within the same call, so two colliding titles discovered together in one search never even reach the (broken) collision check. Same pattern repeated in `scripts/migrate-releases.ts`.

#2 and #3 together mean the collision-avoidance code added specifically to fix "issue #3" doesn't fix it in the case described in its own comments, and will silently clobber `artist_releases` rows (and orphan/misattribute `release_links`) for artists with similarly-named releases.

Things that looked solid: ownership/auth is unaffected by this PR (this endpoint was and remains intentionally unauthenticated, per the code's own comment, for the Mac app/extension — rate limiting is a good addition here); RLS policies on both new tables follow the existing public-read/service-write pattern correctly; no raw SQL/string interpolation, all writes go through the Supabase client; the new `releaseType` field on `check-releases.ts`'s response is additive and shouldn't break existing consumers.

One thing worth a manual look rather than a code-only verdict: the PR description already flags that `persistReleasesForArtist` adds up to N upserts per search per artist as fire-and-forget — worth confirming that doesn't show up as elevated function duration/DB load once merged.

---

PR #338, opened 2026-07-29
Search: find artists from partial queries (Mirlo API, MB discovery, Bandcamp name feedback loop)

brandonlucasgreen:
## Why

Searching **"argent"** returned only the 1970s band Argent and missed **The Argent Grub** entirely — even though Bandwagon's search returns them for that query and Mirlo's API returns them as its *first* result. The pipeline was doing name *lookups* derived from the query spelling, not searches. This PR is Tier 1 of the search-improvement plan: recover the recall that upstream platforms already offer.

## What changed

**Mirlo: slug guess → real search API.** Production fetched `mirlo.space/<query-with-spaces-removed>` and could only ever find an artist whose slug equals the query. It now calls `api.mirlo.space/v1/artists?name=` (host already covered by the `*.mirlo.space` allowlist), re-filters Mirlo's loose substring matching (`parseMirloArtistSearch`, tested), and caches with the standard never-cache-failures predicate.

**MusicBrainz: limit 1 → 5, partial matches become discovery candidates.** The top hit still owns enrichment behind the existing strict gates. Lower-ranked hits whose name contains the query (score ≥ 70) are returned as `suggestedNames` — e.g. "Goodnight Argent" for "argent".

**New Phase 1.5 — discover a name, verify the presence.** Artist names surfaced by Mirlo, Bandwagon, or MB suggestions are fed through the existing Bandcamp probe. "argent" → Mirlo says "The Argent Grub" → probe verifies `theargentgrub.bandcamp.com` (a subdomain the query spelling could never derive) with the same identity + release checks as always. Capped at 3 candidate probes per search, all cached in Supabase.

**Name-only hits keep their real display names.** Bandwagon/Faircamp/Jam.coop/Patreon/Beatport/EVEN matches now carry the name the platform displayed (`NameOnlyEntry`), and attachment tolerates leading-article variance ("Argent Grub" ↔ "The Argent Grub") but nothing looser — substring matching would cross-attach homonyms ("Argent" vs "Rod Argent").

## Bug fixes in the same paths

- Result names can no longer be opaque account ids: searching "argent" surfaced a result literally named `@695d15c12f0f56fdced0a5e6` (a Bandwagon account id reconstructed from the URL).
- Mastodon/PeerTube instance detection now matches hostnames, not URL substrings. The instance `c.im` matched inside `...%2C.imageEncoding...` in a Wix asset URL, which shipped as the band Argent's "Mastodon profile".

## Compatibility & risk

- Old name-only cache entries (bare URL strings) are coerced on read; they age out with the 30-min TTL.
- Worst-case added latency: one parallel round of ≤3 Bandcamp probes (≤2.5s) on cold searches that discover new names; zero once cached. The latency-focused PR comes next.
- Ampwall remains a stub (their site rejects non-browser clients; needs a product decision or partner ask).

## Verification

- `parseMirloArtistSearch` validated against the live Mirlo API: "argent" → The Argent Grub at `mirlo.space/the-argent-grub`.
- New unit tests for the parser, `collectMbSuggestions`, `namesEqualIgnoringArticles`, `looksLikeOpaqueId`, and the mastodon hostname fix. Full unit (419) + api (110) suites and `npm run build` green.
- `tsc` over the changed api files shows only the three pre-existing errors (api/ isn't build-typechecked; baseline compared via stash).

### DIFF
--- a/api/functions/db.ts
+++ b/api/functions/db.ts
@@ -483,18 +483,30 @@ export async function findKnownArtistSlugsByName(term: string, limit = 6): Promi
 
   const { data, error } = await client
     .from('artists')
-    .select('slug, match_confidence')
+    .select('slug, name, match_confidence')
     .ilike('name', `%${cleaned}%`)
     .in('match_confidence', ['claimed', 'verified'])
-    .limit(limit * 2);
+    .limit(40);
 
   if (error) {
     console.error('[DB] findKnownArtistSlugsByName error:', error);
     return [];
   }
 
+  // Rank before capping — the query itself returns arbitrary rows, and for a
+  // common name fragment ("patrick") the artist someone is typing toward must
+  // not lose their slot to whichever rows Postgres happened to emit first.
+  const termLower = cleaned.toLowerCase();
   return (data ?? [])
-    .sort((a, b) => (a.match_confidence === 'claimed' ? 0 : 1) - (b.match_confidence === 'claimed' ? 0 : 1))
+    .sort((a, b) => {
+      const aClaimed = a.match_confidence === 'claimed' ? 0 : 1;
+      const bClaimed = b.match_confidence === 'claimed' ? 0 : 1;
+      if (aClaimed !== bClaimed) return aClaimed - bClaimed;
+      const aPrefix = a.name.toLowerCase().startsWith(termLower) ? 0 : 1;
+      const bPrefix = b.name.toLowerCase().startsWith(termLower) ? 0 : 1;
+      if (aPrefix !== bPrefix) return aPrefix - bPrefix;
+      return a.name.length - b.name.length;
+    })
     .slice(0, limit)
     .map(r => r.slug)
     .filter(Boolean);
@@ -533,7 +545,12 @@ export async function persistSearchResults(results: ArtistResult[]): Promise<voi
         return;
       }
 
-      // Upsert artist — always store as unverified until claimed
+      // Upsert artist, keeping the pipeline's verdict. This used to hardcode
+      // 'unverified', which made the stored confidence meaningless — every
+      // non-claimed row was 'unverified' forever, so quality filters over the
+      // table (partial-name discovery, typeahead) could never distinguish a
+      // release-corroborated artist from name-match junk. Rows refresh on
+      // every search, so the stored verdict tracks the latest pipeline run.
       const { data: artist, error: artistError } = await client
         .from('artists')
         .upsert(
@@ -541,7 +558,7 @@ export async function persistSearchResults(results: ArtistResult[]): Promise<voi
             slug,
             name: result.name,
             image_url: result.imageUrl || null,
-            match_confidence: 'unverified',
+            match_confidence: result.matchConfidence === 'verified' ? 'verified' : 'unverified',
             source: 'auto',
             updated_at: new Date().toISOString(),
           },
--- a/api/functions/db.ts
+++ b/api/functions/db.ts
@@ -330,7 +330,10 @@ export async function putBandcampProbe(
 /**
  * Look up an artist by slug. Returns null if not found or Supabase is not configured.
  */
-export async function getArtistBySlug(slug: string): Promise<ArtistResult | null> {
+export async function getArtistBySlug(
+  slug: string,
+  opts?: { allowStale?: boolean },
+): Promise<ArtistResult | null> {
   const client = getClient();
   if (!client) return null;
 
@@ -376,8 +379,11 @@ export async function getArtistBySlug(slug: string): Promise<ArtistResult | null
 
     const row = artist as ArtistRow;
 
-    // Claimed artists are always fresh; auto-discovered artists expire
-    if (row.match_confidence !== 'claimed') {
+    // Claimed artists are always fresh; auto-discovered artists expire.
+    // allowStale skips the expiry: the partial-name discovery channel would
+    // rather show a known artist with slightly old links than hide them —
+    // platform URLs are stable, and an exact search refreshes the row anyway.
+    if (row.match_confidence !== 'claimed' && !opts?.allowStale) {
       const updatedAt = new Date(row.updated_at).getTime();
       const now = Date.now();
       if (now - updatedAt > FRESHNESS_TTL_MS) {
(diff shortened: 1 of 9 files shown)


brandonlucasgreen (commented):

## Security

No confirmed or plausible vulnerabilities found in this diff. Two of the changes are themselves security-relevant bug fixes worth calling out as correctly done:

- `isMastodonInstance`/`isPeerTubeInstance` moving from whole-URL substring matching to hostname comparison (`hostMatchesInstance`) closes a real misclassification bug (a Wix asset URL's encoded query string matched `c.im` as a substring) — verified against the new test coverage in `enrichment-social.test.ts`.
- `looksLikeOpaqueId` prevents reconstructing a display name from a Bandwagon account-id slug — correctly gates the URL-slug fallback path in `attachNameOnlyPlatforms`.
- The new `mirlo.space` fetch target (`api.mirlo.space/v1/artists`) is covered by the existing `*.mirlo.space` SSRF allowlist entry, and the query is `encodeURIComponent`-escaped before use.
- `searchMirlo` correctly adopts the `cacheGetOrFetch` + `shouldCache` pattern so upstream failures aren't cached as negatives, per the project's "never cache uncertainty" rule.

One thing flagged for awareness rather than as a bug in *this* PR: the new Phase 1.5 Bandcamp candidate-probing (`probeBandcampForCandidates`) and the loosened MusicBrainz suggestion collection both feed into the same `namesMatch`-based identity check inside `findBandcampArtist` that PR #339 (stacked on this one) later relies on for its exact/fuzzy split. That check is loose by design for fuzzy discovery, which is correct here — see the PR #339 review for where that looseness becomes a problem once an "exact" mode is layered on top.

## Regression

Nothing confirmed. Test coverage for the new pure functions (`parseMirloArtistSearch`, `namesEqualIgnoringArticles`, `looksLikeOpaqueId`, `collectMbSuggestions`) is thorough, including the specific edge cases called out in the PR description (opaque IDs, Mastodon substring collision, leading-article variance vs. true cross-artist substrings like "Argent" vs "Rod Argent"). `coerceNameOnlyCache` correctly handles the cache-shape migration (old bare-URL-string entries vs. new `NameOnlyEntry` objects) so a deploy doesn't need a cache flush.

---

brandonlucasgreen (commented):

**Security**
- Bandcamp-probe candidate names sourced from Mirlo/MusicBrainz (third-party data) flow into `probeBandcampForCandidates` → `findBandcampArtist`, but `bandcampSlugCandidates`/`normalizeForComparison` strip everything down to `[a-z0-9]` before it becomes part of a hostname, so there's no SSRF/injection path there — checked against `api/functions/search-utils.ts`, confirmed safe.
- The new Mirlo search hits `api.mirlo.space`, which is already covered by the `*.mirlo.space` entry in `ALLOWED_OUTBOUND_HOSTNAMES` (`api/functions/middleware.ts`) — confirmed safe, no allowlist gap.
- No secret handling, auth, or unsafe-default issues found in this diff.

**Regression**
- `api/search/enrichment.ts` — **plausible**: `isMastodonInstance`/`isPeerTubeInstance` moved from whole-URL substring matching to `new URL(url).hostname` comparison, which correctly fixes the `c.im`-inside-a-Wix-asset-URL false positive. But every *other* platform check in the same `parseSocialUrl` (instagram.com, facebook.com, tiktok.com, etc., lines ~78-104) still does plain substring matching without requiring a URL scheme — implying `url` isn't guaranteed to be an absolute `https://…` string at every call site in this file. A protocol-relative or bare-host social link that the old substring check would still have matched will now silently fail to classify as Mastodon/PeerTube, since `hostnameOf()` returns `null` (and thus `false`) on a parse failure. Worth confirming every caller of `parseSocialUrl`/`isMastodonInstance` only ever passes fully-qualified URLs.
- Everything else (opaque-account-id display-name fix, Mirlo API vs. slug-guessing swap, MB `suggestedNames`/Phase 1.5 probe gating, `namesEqualIgnoringArticles` strictness) matches its stated intent and is well covered by the added tests. No other regressions found.

---

brandonlucasgreen (commented):

No blocking findings. Summary of what was checked:

**Security**
- Mirlo query is `encodeURIComponent`-escaped before URL construction against a hardcoded host (`api.mirlo.space`), covered by the existing `*.mirlo.space` allowlist entry — no injection/SSRF risk from the query param.
- **[needs-manual-verification, low severity]** The Mirlo fetch itself doesn't call `isUrlHostnameAllowed()` directly, unlike `bandcamp-probe.ts`'s `probeCandidate` — this matches the pre-existing pattern for the other name-only platform fetchers in this file (Bandwagon/Beatport/EVEN), so it's not a regression introduced here, but it is a deviation from the CLAUDE.md rule that any code fetching an external URL should pass through the allowlist check. Worth a follow-up pass across `search-sources.ts` rather than blocking this PR.
- New Bandcamp candidate slugs discovered via Mirlo/Bandwagon/MB suggestions flow through the same `bandcampSlugCandidates()` character restriction and `isUrlHostnameAllowed()` gate as query-derived slugs — the "verify before trusting" invariant is preserved for the new discovery path.
- The Mastodon/PeerTube hostname fix (`api/search/enrichment.ts`) now does proper `URL().hostname` equality/suffix comparison instead of whole-URL substring matching, correctly closing the `c.im`-inside-a-Wix-URL false positive.
- No `dangerouslySetInnerHTML` use found on result names in `apps/web/src`; edge SSR functions already route through `escapeHtml`, unaffected by this diff.

**Regression risk**
- MB `limit=1→5`: enrichment still strictly gates on `artists[0]`; the extra 4 hits only feed `collectMbSuggestions` via a separate score/name-exclusion gate. The "top hit" identity used for enrichment is unchanged.
- `collectMbSuggestions`'s `score < 70` boundary correctly keeps `score === 70`, matching its test coverage.
- The 3-candidate-probe cap is applied once over the merged candidate list, not per-source — correctly capped in aggregate.
- **[plausible]** More Bandcamp requests per search (up to 3 extra) raises the odds of hitting Bandcamp's rate limiting; existing 429 handling correctly avoids caching a false negative, but real-world exposure is higher than before — worth monitoring, not blocking.
- Old bare-URL name-only cache entries are safely coerced on read and age out within the existing 30-minute TTL.

---

PR #339, opened 2026-07-29
Search: separate fuzzy human search from exact playback detection (mode param)

brandonlucasgreen:
> **Stacked on #338** — merge that first; this diff includes only the mode split once #338 lands.

## Why

Search and playback detection are different problems that shared one algorithm. A human typing "argent" wants discovery; the extension detecting **Argent** playing on Spotify has the exact artist name and must never be shown The Argent Grub. With #338 adding fuzzy discovery, detection needs an explicit opt-out.

## What changed

**Server** — `/api/search/sources?mode=exact|fuzzy` (default `fuzzy`; unknown values → fuzzy, so it's fully backward compatible). Exact mode:
- filters Mirlo results and name-only platform hits (Bandwagon/Faircamp/Jam.coop/Patreon/Beatport/EVEN) to names that ARE the query, tolerating only a leading article ("Argent Grub" ↔ "The Argent Grub"),
- skips the Phase 1.5 discover→verify Bandcamp probes entirely.

Filtering happens *after* the platform caches, so exact and fuzzy share every upstream cache entry — no double fetching. The v1 public API passes `mode` through and it's documented in `docs/openapi.yaml`.

**Clients that detect send exact; humans typing stay fuzzy:**

| Caller | Mode |
|---|---|
| Extension background detection + now-playing popup | exact |
| Extension popup manual search box | fuzzy |
| Mac app now-playing observer, saved-artist refresh, share extension | exact |
| Mac app typed search field | fuzzy |
| Discord bot, web app | fuzzy (default, unchanged) |

Client-side result caches (extension `chrome.storage`, Mac in-actor cache) are keyed by mode so a fuzzy superset never answers a detection lookup.

## Verification

- Unit tests for `isExactNameMatch` / `filterNameOnlyMapToExact`; full unit (423) + api (110) suites green.
- Mac app builds clean (`xcodebuild`, signing disabled).
- `tsc` over changed api files: only the three pre-existing errors (baseline compared).

### DIFF
--- a/api/functions/db.ts
+++ b/api/functions/db.ts
@@ -483,18 +483,30 @@ export async function findKnownArtistSlugsByName(term: string, limit = 6): Promi
 
   const { data, error } = await client
     .from('artists')
-    .select('slug, match_confidence')
+    .select('slug, name, match_confidence')
     .ilike('name', `%${cleaned}%`)
     .in('match_confidence', ['claimed', 'verified'])
-    .limit(limit * 2);
+    .limit(40);
 
   if (error) {
     console.error('[DB] findKnownArtistSlugsByName error:', error);
     return [];
   }
 
+  // Rank before capping — the query itself returns arbitrary rows, and for a
+  // common name fragment ("patrick") the artist someone is typing toward must
+  // not lose their slot to whichever rows Postgres happened to emit first.
+  const termLower = cleaned.toLowerCase();
   return (data ?? [])
-    .sort((a, b) => (a.match_confidence === 'claimed' ? 0 : 1) - (b.match_confidence === 'claimed' ? 0 : 1))
+    .sort((a, b) => {
+      const aClaimed = a.match_confidence === 'claimed' ? 0 : 1;
+      const bClaimed = b.match_confidence === 'claimed' ? 0 : 1;
+      if (aClaimed !== bClaimed) return aClaimed - bClaimed;
+      const aPrefix = a.name.toLowerCase().startsWith(termLower) ? 0 : 1;
+      const bPrefix = b.name.toLowerCase().startsWith(termLower) ? 0 : 1;
+      if (aPrefix !== bPrefix) return aPrefix - bPrefix;
+      return a.name.length - b.name.length;
+    })
     .slice(0, limit)
     .map(r => r.slug)
     .filter(Boolean);
@@ -533,7 +545,12 @@ export async function persistSearchResults(results: ArtistResult[]): Promise<voi
         return;
       }
 
-      // Upsert artist — always store as unverified until claimed
+      // Upsert artist, keeping the pipeline's verdict. This used to hardcode
+      // 'unverified', which made the stored confidence meaningless — every
+      // non-claimed row was 'unverified' forever, so quality filters over the
+      // table (partial-name discovery, typeahead) could never distinguish a
+      // release-corroborated artist from name-match junk. Rows refresh on
+      // every search, so the stored verdict tracks the latest pipeline run.
       const { data: artist, error: artistError } = await client
         .from('artists')
         .upsert(
@@ -541,7 +558,7 @@ export async function persistSearchResults(results: ArtistResult[]): Promise<voi
             slug,
             name: result.name,
             image_url: result.imageUrl || null,
-            match_confidence: 'unverified',
+            match_confidence: result.matchConfidence === 'verified' ? 'verified' : 'unverified',
             source: 'auto',
             updated_at: new Date().toISOString(),
           },
--- a/api/functions/db.ts
+++ b/api/functions/db.ts
@@ -330,7 +330,10 @@ export async function putBandcampProbe(
 /**
  * Look up an artist by slug. Returns null if not found or Supabase is not configured.
  */
-export async function getArtistBySlug(slug: string): Promise<ArtistResult | null> {
+export async function getArtistBySlug(
+  slug: string,
+  opts?: { allowStale?: boolean },
+): Promise<ArtistResult | null> {
   const client = getClient();
   if (!client) return null;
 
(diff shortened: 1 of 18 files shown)


brandonlucasgreen (commented):

## Security

**1. [CONFIRMED] `mode=exact` does not actually constrain the primary Bandcamp result — the failure mode this PR exists to prevent can still occur.**

`searchAllPlatforms(query, mode)` gates Mirlo (`isExactNameMatch`) and the six name-only platform maps (`filterNameOnlyMapToExact`) on `mode === 'exact'`, and skips Phase 1.5 candidate probing entirely in exact mode. But the *primary* Bandcamp result is untouched by any of this:

```ts
// search-sources.ts — runs identically in both modes, before `exact` is even computed
if (bandcampResults.status === 'fulfilled') allResults.push(...bandcampResults.value.filter(r => r.type === 'artist'));
```

`bandcampResults` comes from `searchBandcamp(query)` → `findBandcampArtist(query, 3000)`, whose identity check is `namesMatch(identity.name, query)` (`search-utils.ts`). `namesMatch` accepts substring containment either direction plus an 80%-character-match fallback for short names — the exact looseness `isExactNameMatch` (added in this PR) was built to replace everywhere else. `aggregateResults()` applies no relevance filter (by design, so fuzzy mode can surface partial matches), so whatever `searchBandcamp` returns reaches the client verbatim.

Concretely: for a short/common artist name, `bandcampSlugCandidates(query)` can land on an unrelated Bandcamp account whose actual band name merely contains the query as a substring, and `namesMatch` will accept it — in **exact** mode, i.e. for a playback-detection query. That's exactly the case the PR description calls out as unacceptable ("detecting **Argent** playing on Spotify ... must never be shown The Argent Grub") — it's just not closed for the one platform that skips the new candidate-probing path and goes straight to a direct probe.

This also isn't just a display bug: `persistSearchResults` is awaited unconditionally at the end of the handler regardless of mode, and (per #336) now writes release data into `artist_releases`/`release_links`. A misattributed exact-mode match from automated detection traffic can persist into production data, not just render once in a UI.

Files: `api/functions/search-sources.ts` (`searchBandcamp` ~L94-109; unguarded push ~L1667-1669; mode gating starts ~L1680); `api/functions/search-utils.ts` (`namesMatch` vs the new `isExactNameMatch`).

Suggest either filtering `bandcampResults` through `isExactNameMatch` in exact mode (mirroring what's done for Mirlo), or routing the exact-mode Bandcamp probe through a stricter identity check than `namesMatch`.

## Regression

**2. [plausible] Unrecognized/missing `mode` silently falls back to the looser mode, with no signal to the caller.**

`event.queryStringParameters?.mode === 'exact' ? 'exact' : 'fuzzy'` treats a typo, wrong case, or a param stripped by some intermediary as `'fuzzy'` — the strictness the detection clients depend on fails open, silently, with no error surfaced. Given finding #1, worth double-checking this is the intended fail-direction for both the extension background listener and the Mac app's now-playing observer, since a dropped `mode=exact` degrades straight into the same class of bug as finding #1, invisibly.

`api/functions/search-sources.ts` (mode resolution in `handler`); same pass-through gap in `api/functions/search-sources-v1.ts`.

---

brandonlucasgreen (commented):

**Security**
- The `mode` query param is validated with a strict equality check (`=== 'exact' ? 'exact' : 'fuzzy'`) server-side in `search-sources.ts` before it's used anywhere; the unsanitized passthrough in `search-sources-v1.ts` is safe because nothing downstream treats it as more than that two-valued enum. No injection/auth/secret issues found.

**Regression**
- `apps/extension/background/service-worker.js` — **plausible, low severity**: the new exact-mode path reuses the pre-existing cache key `cache:${artist}`, which before this PR was the *only* cache and was populated under what is now called "fuzzy" behavior (the historical default included full Bandcamp-discovery probing). Any entry cached in the `CACHE_TTL` window (5 minutes) immediately before this deploys will be read back and served as an "exact" result for a playback-detection lookup — i.e. the exact wrong-artist scenario this PR exists to prevent could still surface for a narrow post-deploy window. Self-heals within 5 minutes, so low impact, but flagging since it's precisely the bug class the PR targets.
- Mode default resolution, forward cache-key partitioning (`cache:fuzzy:*` vs `cache:*`), the Mac app's `SearchMode` default-parameter plumbing (other call sites like the now-playing observer and saved-artist refresh correctly fall through to `.exact` via the default), and the exact-mode Bandcamp-probe skip all check out against the stated design. No other regressions found.

---

brandonlucasgreen (commented):

**Correctness / unsafe defaults**

1. **[confirmed]** Exact mode doesn't actually guarantee "never show the wrong artist" — Bandcamp's primary result bypasses the new filter entirely. `api/functions/search-sources.ts:1673` pushes `bandcampResults` straight into `allResults` without running it through `isExactNameMatch`/`filterNameOnlyMapToExact`, unlike Mirlo and the name-only platform maps. The underlying `searchBandcamp` identity check (`namesMatch` in `api/search/bandcamp-probe.ts:225`, via `search-utils.ts`'s substring-based `namesMatch`) uses `normalizeForComparison`, which strips all non-alphanumeric characters — so `namesMatch("The Argent Grub", "Argent")` returns `true` because `"theargentgrub".includes("argent")`. A slug candidate derived from a short/partial query in `mode=exact` can therefore still surface a same-substring artist unfiltered on Bandcamp — the exact scenario ("detecting Argent must never surface The Argent Grub") this PR exists to prevent, on the one platform that isn't gated by the new logic.

2. **[confirmed by code reading]** `isExactNameMatch`/`filterNameOnlyMapToExact` collapse all non-Latin-script names to the empty string. `normalizeForComparison` (`search-utils.ts:242-245`) strips every character outside `a-z0-9`, so two unrelated artist names written entirely in Japanese, Korean, Cyrillic, etc. both normalize to `""`, and `"" === ""` reads as an exact match. In exact mode this can attach an unrelated non-Latin-named artist's Mirlo/name-only hit to a different playing artist — the opposite of what exact mode promises.

3. **[plausible]** `namesEqualIgnoringArticles` (`search-utils.ts:321-326`) requires a trailing space after `the|a|an` to strip it, so a name that's literally "The", or "The The", both normalize to the same 3-letter token — a low-likelihood but real gap in the new exact-match guarantee.

4. **[plausible, low severity]** Extension cache-key collision on upgrade: `apps/extension/background/service-worker.js`'s `getResults` reuses the pre-PR unprefixed key `cache:${artist}` for exact mode while fuzzy gets a new prefixed key. Entries written before this ships get read back as "exact" until they expire — bounded by the existing 5-minute `CACHE_TTL`, so self-healing.

**Not a finding:** the `mode` param validation is a strict allowlist with safe default (fuzzy) and isn't bypassable; v1 API passthrough preserves backward compatibility for callers who omit `mode`; no auth/injection/secret-handling issues in this diff.

---

PR #340, opened 2026-07-29
Search: instant typeahead from the local artist index (/api/suggest + pg_trgm)

brandonlucasgreen:
## Why

Every search today is submit-only and pays for the full 10-platform fan-out (10–20s cold). But Unstream already has a local index it never uses for search: the Supabase `artists` table accumulates every artist ever resolved. This PR makes known artists findable in milliseconds while typing — the incremental first step toward search backed by our own index.

## What changed

- **Migration** `20260729071000_artist-name-trgm.sql`: `pg_trgm` + GIN trigram index on `artists.name`. Applied automatically by the supabase-migrate workflow on merge. (Heads-up: `npm run migrate:dry-run` is broken independently of this PR — the script passes a `--project-ref` flag the current Supabase CLI doesn't accept.)
- **`/api/suggest`** (new `search-suggest.ts`): one indexed DB read, rate-limited, Upstash-cached 5 min. Only verified/claimed artists are suggested — `unverified` rows are where junk from arbitrary searches lives. A DB failure returns empty but is *not* cached ("never cache uncertainty").
- **SearchBar typeahead**: 250 ms debounce, accessible combobox pattern (arrow keys / Enter / Escape / mouse), aborts stale requests, degrades to nothing when the endpoint has no answers. Selecting a suggestion runs the normal search for that exact name.
- Dev server gets an inert stub (dev doesn't talk to Supabase).

## Verification

- Live in the running app: typing "argent" produced exactly one debounced request; the dropdown rendered; picking **The Argent Grub** navigated to `?q=The+Argent+Grub` and ran the search.
- jsdom tests for keyboard + mouse selection and the min-length gate; endpoint tests for validation, ranking, and the failure-vs-empty caching distinction. Full build green.
- The suggest quality improves as PR #338 lands and verified artists accumulate — the index is populated by real searches.

### DIFF
--- a/api/functions/__tests__/search-suggest.test.ts
+++ b/api/functions/__tests__/search-suggest.test.ts
@@ -1,0 +1,111 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+const mocks = vi.hoisted(() => ({
+  mockSuggestArtists: vi.fn(),
+  mockCheckRateLimit: vi.fn(() => Promise.resolve({ limited: false })),
+  mockGetClientIp: vi.fn(() => '127.0.0.1'),
+  // Pass-through cache: always calls the fetcher, records the shouldCache verdict.
+  mockCacheGetOrFetch: vi.fn(),
+  lastShouldCache: { value: null as boolean | null },
+}));
+
+vi.mock('../db', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('../db')>();
+  return {
+    ...actual,
+    suggestArtists: mocks.mockSuggestArtists,
+  };
+});
+vi.mock('../ratelimit', () => ({
+  checkRateLimit: mocks.mockCheckRateLimit,
+  getClientIp: mocks.mockGetClientIp,
+}));
+vi.mock('../cache', () => ({
+  cacheGetOrFetch: mocks.mockCacheGetOrFetch,
+}));
+
+import { handler } from '../search-suggest';
+import { rankArtistSuggestions } from '../db';
+
+beforeEach(() => {
+  mocks.mockSuggestArtists.mockReset();
+  mocks.lastShouldCache.value = null;
+  mocks.mockCacheGetOrFetch.mockImplementation(
+    async (_key: string, fetchFn: () => Promise<unknown>, _ttl: number, shouldCache?: (d: unknown) => boolean) => {
+      const data = await fetchFn();
+      mocks.lastShouldCache.value = shouldCache ? shouldCache(data) : true;
+      return { data, cached: false };
+    },
+  );
+});
+
+describe('search-suggest handler', () => {
+  it('returns suggestions for a valid query', async () => {
+    mocks.mockSuggestArtists.mockResolvedValue([
+      { slug: 'the-argent-grub', name: 'The Argent Grub', imageUrl: null },
+    ]);
+    const res = await handler({ queryStringParameters: { query: 'argent' } });
+    expect(res.statusCode).toBe(200);
+    const body = JSON.parse(res.body);
+    expect(body.suggestions).toHaveLength(1);
+    expect(body.suggestions[0].name).toBe('The Argent Grub');
+  });
+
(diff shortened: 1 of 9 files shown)


brandonlucasgreen (commented):

**Security**
- `suggestArtists` (`api/functions/db.ts`) strips ILIKE wildcard/meta characters from user input before querying and uses the service-role client with an explicit `match_confidence`/`source` filter. I checked the `match_confidence = 'claimed'` clause against `supabase/migrations/20260401000000_artist-profiles.sql`, which does widen the check constraint to allow `'claimed'` — so that clause isn't dead code, `schema.sql`'s baseline comment is just stale on this point (it doesn't claim to reflect that migration). Not a bug.
- The cache correctly distinguishes "DB unreachable" (`null`, not cached) from "genuinely no matches" (`[]`, cached) per the project's never-cache-uncertainty convention — confirmed correct in both `search-suggest.ts` and the test file.
- **needs-manual-verification**: `search-suggest.ts` hand-rolls `Access-Control-Allow-Origin: '*'` instead of the shared `buildCorsHeaders`/`buildPublicCorsHeaders` helpers in `middleware.ts`. This matches existing precedent in `search-sources.ts` and `check-releases.ts`, so it isn't a new deviation introduced by this PR, but it continues a repo-wide gap versus the "use the shared middleware rather than hand-rolling headers" principle in CLAUDE.md. Impact looks minimal since the data returned is already public (verified/claimed artist names), but flagging for awareness.

**Regression**
- None found. Debounce/abort handling in `SearchBar.tsx` correctly cancels stale in-flight requests, and the suggestion list's `onMouseDown`-before-`onBlur` ordering (to land the click before the blur-triggered close) is handled correctly.

---

brandonlucasgreen (commented):

**Security / correctness**

1. **[confirmed]** Cache-key collision produces wrong cached results. `api/functions/search-suggest.ts` builds the cache key from `normalizeForComparison(query)`, which strips *all* non-alphanumeric characters, while the actual DB query in `api/functions/db.ts`'s `suggestArtists()` uses `cleaned` (only `% _ , ( )` stripped, hyphens/periods/apostrophes preserved). `"sufjan-stevens"`, `"sufjan.stevens"`, and `"sufjan stevens"` all collapse to the same cache key but generate different `ILIKE` patterns against different data — whichever variant is queried first gets cached for 5 minutes and served to the others. This is the exact "cache keys must not collide across inputs that behave differently" failure class the codebase's own conventions call out. Suggest deriving the cache key from `cleaned` instead of a separately-normalized string.

2. **[confirmed]** Reset doesn't cancel a pending debounce/fetch. In `apps/web/src/components/SearchBar.tsx`, `handleReset` clears `query`/`suggestions` and closes the dropdown but never calls `clearTimeout(debounceRef.current)` or `abortRef.current?.abort()` (unlike `submitSearch`, which does both). Typing "arg" then immediately hitting Reset lets the stale fetch complete ~250ms later and unconditionally call `setSuggestions`/`setShowSuggestions`, reopening the dropdown against an empty input box.

3. **[confirmed, minor]** `search-suggest.ts` hand-builds CORS headers inline instead of calling `buildPublicCorsHeaders()` from `middleware.ts`. Functionally equivalent for this public wildcard-origin GET, but it silently drops `X-Content-Type-Options: nosniff` and the shared `Access-Control-Allow-Methods/Headers` values.

4. **[confirmed sharing / plausible impact]** `checkRateLimit(ip, 'standard', …)` reuses the same per-IP `'standard'` bucket (30/min, 1000/day) as `me-settings.ts`, `me-username.ts`, `saved-artists.ts`, `user-sharing.ts`. Typeahead fires on every ~250ms debounced keystroke pause, so a user actively refining a query could burn the 30/min quota and start getting 429s on unrelated account/settings actions in the same minute. Worth a dedicated tier.

**Not a finding:** the `match_confidence='claimed'`-only filter in `suggestArtists` is correctly scoped, ILIKE-injection is defended against by stripping `% _ , ( )`, and the never-cache-uncertainty gate (`shouldCache: (data) => data !== null`) is correctly wired and tested.

---

PR #341, opened 2026-07-29
Search: cache MB enrichment, fix false hasPendingEnrichment, zero-result fallback card

brandonlucasgreen:
> **Stacked on #338** (second in the stack alongside #339 — both are independent of each other and based on #338).

## Why

Three related problems found while auditing search:

1. **Major artists rendered bare.** Live repro: Radiohead's Phase 1 response was missing its official site and socials while claiming `hasPendingEnrichment: false` — so the client never called Phase 2 to fill the gap, even though the Phase 2 endpoint had everything. Root cause: Phase 1 counted enrichment as "applied" whenever MusicBrainz matched a *name*, even when the follow-up url-rels fetch failed.
2. **The slowest leg of every cold search was uncached.** The in-request MB enrichment costs 2+ mandatory rate-limit delays plus several fetches, repeated for every cold search of the same artist.
3. **Zero-result searches returned literally nothing** — no official site, no fallback — the worst possible first impression when someone searches a famous artist.

## What changed

- `EnrichedMusicBrainzResult` now tracks `searchFailed` and `enrichmentComplete`; enrichment counts as applied only when the data actually arrived. Partial fetches → `hasPendingEnrichment: true` → the client retries via Phase 2 and self-heals.
- MB enrichment is Upstash-cached (30 min) behind a tested `isCacheableMbResult` predicate: complete results and genuine no-matches cache; failures and partials never do ("never cache uncertainty").
- The client fires Phase 2 even with zero Phase 1 results and renders an MB fallback card (official site, Discogs, socials, platform search links) mirroring the server's own fallback — so "Radiohead" always shows *something* useful.
- CDN caching is bounded and state-aware: complete responses `s-maxage=300, swr=3600`; pending ones `s-maxage=60, swr=300`. The old header had an **unbounded** `stale-while-revalidate`, which let stale buggy responses keep being served indefinitely — likely how the Radiohead response survived past its causes.
- Dev server now always reports `hasPendingEnrichment: true` (it never enriches server-side), so the client Phase 2 path is exercised in dev.

## Verification

- Unit (426) + api (110) suites and full build green; tsc over changed api files shows only the three pre-existing errors.
- Dev-server smoke test: Phase 1 + Phase 2 fire, results render, no new console errors.
- Tests cover the cacheability matrix (complete / no-match / failed / partial) and the fallback card builder.

### DIFF
--- a/api/functions/db.ts
+++ b/api/functions/db.ts
@@ -483,18 +483,30 @@ export async function findKnownArtistSlugsByName(term: string, limit = 6): Promi
 
   const { data, error } = await client
     .from('artists')
-    .select('slug, match_confidence')
+    .select('slug, name, match_confidence')
     .ilike('name', `%${cleaned}%`)
     .in('match_confidence', ['claimed', 'verified'])
-    .limit(limit * 2);
+    .limit(40);
 
   if (error) {
     console.error('[DB] findKnownArtistSlugsByName error:', error);
     return [];
   }
 
+  // Rank before capping — the query itself returns arbitrary rows, and for a
+  // common name fragment ("patrick") the artist someone is typing toward must
+  // not lose their slot to whichever rows Postgres happened to emit first.
+  const termLower = cleaned.toLowerCase();
   return (data ?? [])
-    .sort((a, b) => (a.match_confidence === 'claimed' ? 0 : 1) - (b.match_confidence === 'claimed' ? 0 : 1))
+    .sort((a, b) => {
+      const aClaimed = a.match_confidence === 'claimed' ? 0 : 1;
+      const bClaimed = b.match_confidence === 'claimed' ? 0 : 1;
+      if (aClaimed !== bClaimed) return aClaimed - bClaimed;
+      const aPrefix = a.name.toLowerCase().startsWith(termLower) ? 0 : 1;
+      const bPrefix = b.name.toLowerCase().startsWith(termLower) ? 0 : 1;
+      if (aPrefix !== bPrefix) return aPrefix - bPrefix;
+      return a.name.length - b.name.length;
+    })
     .slice(0, limit)
     .map(r => r.slug)
     .filter(Boolean);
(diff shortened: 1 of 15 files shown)


brandonlucasgreen (commented):

**Security:** no auth, secret-handling, or injection issues found in this diff.

**Regression risk**

1. **[confirmed]** `enrichmentComplete` only reflects the `inc=url-rels` fetch (`api/functions/search-sources.ts:713`, `enrichmentComplete: artistResponse.ok`), not the other enrichment fetches that feed the same cached object (`fetchOfficialSiteSocialLinks`, `fetchDiscogsSocialLinks`, `fetchWikipediaSummary`, `fetchBandcampLocation`, `fetchMirloLocation`, `searchPeerTubeChannels`, and the release-group fetch behind `hasPre2005Release`, all around lines 494–650). Each of those catches its own timeouts/failures internally and returns empty defaults without affecting `enrichmentComplete`.
   - Failure scenario: MB search + url-rels succeed, but the official-site fetch times out. `enrichmentComplete` stays `true`, so `isCacheableMbResult` treats the partial result as complete and caches "no socials found" for the full 30-minute TTL, and the response also gets the long CDN header (`s-maxage=300, swr=3600`) since `hasPendingEnrichment` is false. This reproduces the "Radiohead shipped bare" bug this PR sets out to fix, one layer downstream of the specific fetch it patches.
   - Suggest folding the other fetches' own success/failure into `enrichmentComplete`, not just the url-rels response.

2. **[needs-manual-verification]** Low-confidence/name-mismatch MB branches (`api/functions/search-sources.ts:440-470`) now get cached for the first time, inheriting `enrichmentComplete: true` from `emptyResult` — including whatever `enrichLocationFallback` returned. A transient failure inside that helper would get cached as "no location" for 30 minutes. Severity depends on `enrichLocationFallback`'s own error handling, which isn't part of this diff.

No issues found in the fallback-card trigger condition (correctly gated on prior Phase-1 results being empty), the CDN header wiring, or the dev-server change.

---

PR #343, opened 2026-07-29
Make Redis failures fast and loud; stop the SW hijacking edge routes

brandonlucasgreen:
Started from "why does `/u/brandon` take 5 seconds" and found two independent bugs compounding, plus a much larger one behind them.

## What was wrong

Upstash deleted the free-tier database after 14 days of inactivity. Both Redis callers — `cache.ts` and `ratelimit.ts` — read the same two env vars, so rate limiting *and* response caching died together. Every Redis command then spent the SDK's default retry budget (5 retries, exponential backoff, `50+136+370+1004+2730 = ~4.3s`) before the caller caught the error and failed open.

Measured in production before the fix:

| Endpoint | Time |
|---|---|
| `/api/search/sources?query=radiohead` | **13–30s** |
| `/api/public/saved-artists/brandon` | 4.8–6.2s |
| `/api/artist-page`, `/api/artist` | 4.5–5.8s |
| `/api/analytics/stats` (returns 401!) | 4.5s |
| `/api/status`, `/api/platforms` (not rate limited) | **0.13–0.23s** |

The 401 row is the tell — no real work, still 4.5s. The delay sat entirely in front of the handler.

Two consequences past latency:

- **`cacheGet` returned `null` on error, indistinguishable from a miss.** Every lookup cost 4.3s *and* reported a miss, so every repeat query re-scraped the partner instead of doing one Redis read. That's a politeness problem, not just a slow one.
- **Rate limiting was entirely unenforced.** 45 parallel requests against a 30/min endpoint returned 45 × 404, zero 429s.

None of it reached Sentry. Every catch block only did `console.error`.

## What this changes

**`api/functions/redis.ts`** (new) — one shared client for both callers, with the two guards that were missing:

- `retry: { retries: 1, backoff: () => 50 }` — enough for a transient blip, cheap enough to swallow when Redis is genuinely gone.
- A 30s circuit breaker, so a dead Redis costs one failed command per container instead of one per call site. Capping retries alone still costs ~60ms per call site, and one search touches many.

Failures go to Sentry, throttled in memory — deliberately *not* via `checkSentryDedup`, which is itself Redis-backed and would need the thing that's broken.

**Worth a careful look in review:** the `getRedis()` re-checks in `ratelimit.ts` look redundant but are load-bearing. The limiter getters memoize (`if (standardLimiter) return standardLimiter`), so a cached limiter can never observe the circuit opening — without those checks the breaker would silently do nothing for rate limiting.

**`apps/web/vite.config.ts`** — `navigateFallbackDenylist` only listed `/a/` and `/artist/`, so returning visitors with the service worker installed got the cached SPA shell for `/u/*`, `/guides/*` and `/search`, and the edge function never ran. That's why `/u/:handle` was slow for signed-in users while curl and incognito saw ~200ms. Verified in the built output:

```
denylist:[/^\/a\//,/^\/artist\//,/^\/u\//,/^\/guides\//,/^\/search/]
```

**`.github/workflows/upstash-keepalive.yml`** — twice-weekly ping (max 4-day gap vs the 14-day threshold). Standalone rather than a step in `schedule-social-posts.yml` on purpose: a Buffer failure there must not take the keepalive down with it. Needs `UPSTASH_REDIS_REST_URL` / `UPSTASH_REDIS_REST_TOKEN` as repo secrets.

**`api/tsconfig.json`** — added the touched `api/` files to the include, per the CLAUDE.md note about that narrow list.

## Verification

- 118 API tests pass (12 files), 8 new in `redis-resilience.test.ts` covering the retry cap, breaker open/close, Sentry tagging and report throttling
- 390 web unit tests pass
- `npm run typecheck:api` clean
- `npm run lint` shows no new problems (77 pre-existing, none in touched files)
- Built `sw.js` confirmed to contain all five routes

## Note

The code here is inert until the env vars point at a live database — that part is done separately in Netlify and GitHub. Post-deploy, the check that actually proves it is the COMMANDS counter climbing in the Upstash dashboard; `COMMANDS 0` is what gave away the original breakage.

Expect the first few searches after deploy to still be slow while the cache refills, then drop sharply.

### DIFF
--- a/.github/workflows/upstash-keepalive.yml
+++ b/.github/workflows/upstash-keepalive.yml
@@ -1,0 +1,46 @@
+name: Upstash keepalive
+
+# Upstash deletes free-tier databases after 14 days of inactivity. In July 2026 the Unstream
+# database was reaped, which silently disabled rate limiting AND response caching — pushing
+# production search to 13-30s and re-scraping partner sites on every repeat query.
+#
+# This runs twice a week (max 4-day gap) so the database never approaches that threshold, and
+# it is a separate workflow rather than a step in schedule-social-posts.yml on purpose: a
+# failure over there must not take the keepalive down with it.
+#
+# It is a backstop, not the primary signal. Real traffic keeps the database warm on its own;
+# if this workflow is the only thing touching Redis, that is itself worth knowing.
+
+on:
+  schedule:
+    - cron: '0 6 * * 1,4' # Mondays and Thursdays, 06:00 UTC
+  workflow_dispatch:
+
+jobs:
+  ping:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Ping Upstash Redis
+        env:
+          UPSTASH_REDIS_REST_URL: ${{ secrets.UPSTASH_REDIS_REST_URL }}
+          UPSTASH_REDIS_REST_TOKEN: ${{ secrets.UPSTASH_REDIS_REST_TOKEN }}
+        run: |
+          if [ -z "$UPSTASH_REDIS_REST_URL" ] || [ -z "$UPSTASH_REDIS_REST_TOKEN" ]; then
+            echo "::error::UPSTASH_REDIS_REST_URL / UPSTASH_REDIS_REST_TOKEN secrets are not set"
+            exit 1
+          fi
+
+          # A SET with an expiry is a write, so it counts as activity and leaves nothing behind.
+          response=$(curl -sS -w '\n%{http_code}' \
+            -H "Authorization: Bearer $UPSTASH_REDIS_REST_TOKEN" \
+            "$UPSTASH_REDIS_REST_URL/set/keepalive/$(date -u +%Y-%m-%dT%H:%M:%SZ)?EX=86400")
+
+          status=$(echo "$response" | tail -n1)
+          body=$(echo "$response" | sed '$d')
+
+          if [ "$status" != "200" ]; then
+            echo "::error::Upstash ping failed with HTTP $status: $body"
+            exit 1
+          fi
+
+          echo "Upstash ping OK: $body"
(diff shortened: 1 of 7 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Automated staff-eng review.

## Security

- **SSRF/injection/auth**: no changes in scope here — this PR only touches the Redis client and CI workflow, no new external-input handling. Nothing to flag.
- **Secret handling**: `upstash-keepalive.yml` reads `UPSTASH_REDIS_REST_URL`/`TOKEN` from `secrets.*` into env vars and never echoes them; the curl response body it does print is Upstash's own JSON reply, not the token. Confirmed clean.
- **Data exposure — needs-manual-verification**: `reportRedisFailure(context, error)` now forwards the cache/rate-limit key (e.g. `cacheGet(artist:bandcamp:radiohead)`) to Sentry as `extra.context`. Previously this only ever hit `console.error`, never left the process. Per this repo's own convention ("cache keys derived from user input should hold normalized search terms, not PII"), that should be safe, but it's worth a conscious check that no cache key anywhere ever embeds an email/IP/user id rather than a normalized search term, since this PR is the first thing that ships those keys to a third party (Sentry).

## Regression

- The `getRedis()` re-check inside `checkRateLimit`/`checkApiRateLimit` (in addition to the memoized limiter) looked redundant at first read, but it's correct: the limiter getters memoize, so without the re-check a cached limiter could never observe the circuit breaker opening. Verified against `redis.ts` — correct.
- Circuit-breaker state (`circuitOpenUntil`, `lastReportedAt`) is module-level, so it's scoped per warm Lambda container, not cluster-wide. That's called out in the PR description as an accepted trade-off — agreed, no action needed.
- Checked for a double-report race when many parallel `cacheGet` calls fail at once (e.g. one search fanning out to ~17 platforms with Redis down): since Node is single-threaded per invocation, the first failure synchronously sets `lastReportedAt` before the next `catch` block runs, so there's no double Sentry report. Correct.
- Test coverage (8 new tests) covers retry cap, breaker open/close, Sentry tagging, and report throttling — matches the behavior actually being relied on. No gaps found in `cache.ts`/`ratelimit.ts` themselves; both are now thin delegations to the tested module.

Nothing blocking. Nice catch tracing this back to the SW denylist gap too — that's a real, separate bug (`/u/*`, `/guides/*`, `/search` silently served stale SPA shell) and the fix is correct.

---

PR #344, opened 2026-07-29
Admin: remove individual platform links from search results

brandonlucasgreen:
## Why

Taylor Swift's search results include a `*.bandcamp.com` page that looks legitimate but isn't her. Removing the artist is the wrong fix; there was no way to remove just the one link. Merge overrides can't express this — they merge and re-create whole results — and the offending links are often auto-generated (slug probe, MusicBrainz relations) rather than stored anywhere editable.

## What changed

**Admin UI**
- A ✕ next to each platform badge and social icon on a search result (admins only). It opens a dialog showing artist / platform / URL with a scope choice:
  - **This artist only** (default) — the URL is hidden under that name; a homonym who genuinely owns the page keeps it.
  - **Every artist** — the URL never appears in any result.
- Optional reason field. The card drops the link immediately.
- New `/admin/links` page listing every removal (platform, artist, URL, reason, who, when) with a **Restore** button. Linked from the admin header nav.

**Backend**
- New `platform_link_suppressions` table — server-only, RLS enabled with no policies (service-role only), two partial unique indexes so one global row and per-artist rows coexist.
- `applyLinkSuppressions` (`search-utils.ts`) runs as the **last** step of the Phase 1 pipeline, after the probe, enrichment, merge overrides, and deferred name-only platforms. A suppression applied mid-pipeline would be undone by whichever phase ran next. Results left with only search links are dropped by the existing `filterAndSort`.
- `search-musicbrainz.ts` strips suppressed links from its own response, applied *after* the cache read so a new suppression takes effect on the next request. Phase 2 links are merged in client-side, so without this pass a removed link reappears the moment enrichment lands.
- `deleteStoredLinksForUrl` clears matching `artist_links` rows on save, so the link also disappears from `/a/:slug` and DB-backed result cards instead of only from freshly fetched results.
- `GET/POST/DELETE /api/admin/link-suppression`, admin-authenticated, CORS via `buildCorsHeaders`.

## Reviewer notes

- **Not** gated on `isUrlHostnameAllowed`: that allowlist covers hosts we *fetch*, and the links most in need of removal (socials, official sites, Qobuz, MB-supplied URLs) aren't on it and are never fetched. Validation is scheme + hostname shape + length.
- Insert rather than upsert: Postgres can't infer `ON CONFLICT` against a partial index, so a duplicate (`23505`) is reported as success — "already suppressed".
- Suppressing a Bandcamp link also clears a `bcbits.com` result photo. Results carry no image provenance, and the fake page would otherwise keep supplying the picture after its link is gone.
- Suppression reads fail open (`[]` on error), matching `getMergeOverrides` — a suppression list that can't be read must not take the search response down with it.
- The migration deploys automatically on merge. Search responses carry `s-maxage=300`, so a removal can take a few minutes to reach cached queries. The dev server implements no admin endpoints (same as `/admin/merge` today), so removals only work against a deploy.

## Testing

`npm run build` green: 143 API tests, 453 web unit tests. New coverage:
- `api/functions/__tests__/link-suppressions.test.ts` — URL normalization, per-artist vs global scope, homonym isolation, photo clearing, interaction with `filterAndSort`.
- `apps/web/tests/unit/AdminRemoveLink.test.tsx` — control is admin-only, default vs global scope reaches the server, failure keeps the dialog open.

### DIFF
--- a/api/functions/__tests__/link-suppressions.test.ts
+++ b/api/functions/__tests__/link-suppressions.test.ts
@@ -6,6 +6,7 @@ import {
   applyLinkSuppressions,
   isUrlSuppressed,
   normalizeUrlForMatch,
+  urlMatchPrefilter,
   filterAndSort,
   type LinkSuppression,
 } from '../search-utils';
@@ -29,9 +30,43 @@ describe('normalizeUrlForMatch', () => {
       .toBe(normalizeUrlForMatch('https://taylorswift.bandcamp.com'));
   });
 
+  // The same page arrives as http from older stored rows and some MusicBrainz
+  // relations, and www-prefixed from official-site scrapes. A suppression saved
+  // from one spelling has to match the others.
+  it('ignores the scheme', () => {
+    expect(normalizeUrlForMatch('http://taylorswift.bandcamp.com'))
+      .toBe(normalizeUrlForMatch('https://taylorswift.bandcamp.com'));
+  });
+
+  it('ignores a leading www.', () => {
+    expect(normalizeUrlForMatch('https://www.discogs.com/artist/1024240'))
+      .toBe(normalizeUrlForMatch('https://discogs.com/artist/1024240'));
+  });
+
   it('does not conflate different paths', () => {
     expect(normalizeUrlForMatch('https://x.bandcamp.com/music'))
       .not.toBe(normalizeUrlForMatch('https://x.bandcamp.com'));
+  });
+
+  it('keeps the query string, which is what distinguishes search links', () => {
+    expect(normalizeUrlForMatch('https://duckduckgo.com/?q=site:ko-fi.com+a'))
+      .not.toBe(normalizeUrlForMatch('https://duckduckgo.com/?q=site:ko-fi.com+b'));
+  });
+
+  it('does not treat a lookalike host as the same page', () => {
+    expect(normalizeUrlForMatch('https://bandcamp.com.evil.test/x'))
+      .not.toBe(normalizeUrlForMatch('https://bandcamp.com/x'));
+  });
+
+  it('falls back to a plain key for an unparseable URL', () => {
+    expect(normalizeUrlForMatch('  Not A Url/ ')).toBe('not a url');
+  });
+});
+
+describe('urlMatchPrefilter', () => {
+  it('drops the query so one SQL pattern covers every spelling', () => {
+    expect(urlMatchPrefilter('http://WWW.Discogs.com/artist/1024240/?foo=1'))
+      .toBe('discogs.com/artist/1024240');
   });
 });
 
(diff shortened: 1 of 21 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Staff-engineer review pass (security + regression), based only on the diff and the current state of the branch.

## Security

- **URL matching normalizes case and trailing slash but not scheme/host variants — plausible**, `normalizeUrlForMatch()` in `api/functions/search-utils.ts`:
  ```ts
  export function normalizeUrlForMatch(url: string): string {
    return url.trim().replace(/\/+$/, '').toLowerCase();
  }
  ```
  This is the single comparison function behind `isUrlSuppressed` (used in both `applyLinkSuppressions` in Phase 5 and `stripSuppressedLinks` in `search-musicbrainz.ts`) and `deleteStoredLinksForUrl`. It doesn't normalize `http://` vs `https://`, or a `www.` prefix. If an admin suppresses `https://x.bandcamp.com` but a later fetch (or an older stored `artist_links` row) has the same page as `http://x.bandcamp.com`, the suppression silently fails to match on all three call sites — the bad link can reappear in a live search, in MusicBrainz enrichment, and stay visible on `/a/:slug` despite an admin having explicitly removed it. Worth confirming whether any upstream source is inconsistent about scheme (older `artist_links` rows in particular, since those predate this feature).

- Auth is solid: every method (`GET`/`POST`/`DELETE`) on `admin-link-suppression.ts` goes through `authenticateAdmin` before touching the DB, `/admin/links` has no client-side-only gate (server independently re-checks), and `getLinkSuppressions()` returns only `url`/`artist_name_norm` — no admin email or reason text leaks into the public search response. `isStorableLinkUrl`'s scope (scheme + hostname shape, not the SSRF allowlist) is correctly reasoned since this URL is only ever string-compared, never fetched — confirmed by reading the full call graph, nothing in this diff fetches the submitted URL.
- No rate limiting on `admin-link-suppression.ts`, but that matches the existing convention (`admin-verify.ts` also has none) — not a new gap introduced by this PR.

## Regression

- Suppression ordering is correct: `applyLinkSuppressions` runs as Phase 5, after the probe/enrichment/merge-override/name-only-platform phases and *before* `persistSearchResults`, so a freshly-confirmed suppressed link is filtered out of the response and never re-persisted to `artist_links` on the next search — verified by tracing `searchAllPlatforms` → `finalResults` → the `handler`'s persist call, not just by the code comments' own claims.
- `stripSuppressedLinks`'s early return on `!result.artistName` looked like a gap at first (a global suppression wouldn't get applied without an artist name to match against) — but every MusicBrainz-miss path returns `emptyResult` with all link fields already null/empty, so there's nothing to strip in that case. Not a bug.
- No off-by-one or API-contract-drift issues found; the new `onRemoveLink` prop threading through `ResultCard` → `ResultCardPlatforms`/`ResultCardSocial` is additive and gated correctly on `isAdmin && onLinkRemoved`.

---

PR #345, opened 2026-07-29
Search: persistent search bar in the header on every page

brandonlucasgreen:
Search was only reachable from the homepage and the artist dashboard. This puts it in the shared `Header` so you can start a search from anywhere in the app.

## What changed

**Header search everywhere.** `HeaderSearch` renders in `Header`, centered from `md` up. Below `md` there isn't room for the wordmark, an input and the nav at once, so a magnifier sits left of the hamburger and expands the same component as a full-width second row (autofocused, Escape-dismissed).

**Homepage drops its big search bar.** Download links become the sole CTA, with a line pointing at the header. The hero heading, byline and download links now hide once a search runs — they were pushing the first result below the fold on a phone.

**PWA mode gets its own minimal app home** instead: heading, prominent search bar, Login CTA. Marketing sections, the FAQ link and the third-party Letterbird contact embed are all dropped. It keeps a visible input on purpose — the installed app has no competing CTA and shouldn't open onto a screen with nothing to type in. Only signed-out users ever see it, since signed-in PWA users are already redirected to `/dashboard`.

**Pure-SSR edge pages** (`/a`, `/artist`, `/u`) get a plain `<form action="/" method="get">`, since they can't render React. Works with JS off.

## Bug fixed along the way

`App`'s `?q=` deep-link effect was guarded on `!hasSearched`, so **only the first search of a page's lifetime ever ran**. A second search from the header changed the URL and silently did nothing. The guard is now "is this a different query than the one we ran" — the URL is the source of truth. Side benefit: the back button now works between two searches.

The second commit completes that rule in the other direction: clicking the Unstream wordmark cleared the URL but left `results`/`hasSearched` set, so the list stayed on screen under a bare `/`. No `?q=` now means nothing to show.

## Notes for review

**Deliberately not a `/search` SPA route.** That URL belongs to the `noscript-search` edge function and is in the service-worker denylist. A second renderer there is the bifurcation trap from `docs/retros/UNS-100-bifurcation-retro.md`. Header search navigates to `/?q=…`, the path `DashboardPage` already used.

Three non-obvious constraints, all commented in the code:

- `SuggestionList` takes an `idPrefix`. The dashboard renders **both** search inputs, so a hardcoded `id="search-suggestions"` would give that page two elements with the same id and misdirect every `aria-activedescendant`.
- Header search input state stays **inside** `HeaderSearch`. Every page renders its own `Header`, so lifting it into `Header` or a context would re-render the whole mounted page on every keystroke.
- `resetSearchState()` is split out of `handleGoHome` because the URL-driven reset must **not** write to `searchParams` — setting an already-empty query can no-op, which would leave `justWentHomeRef` armed and swallow the next real search.

**Performance:** +1.0 kB gzip on the entry chunk (383.97 → 387.86 kB raw), no new chunk — `SearchBar` already shipped there via eager `App.tsx`. `/api/suggest` was already `lenient`-rate-limited and cached.

## Verification

- `npm run build` green; 466 unit + 150 API tests pass; 9 new `HeaderSearch` tests; `tsc` clean.
- Lint unchanged at its pre-existing 77-problem baseline (confirmed by stash-baseline). The 3 `deno check` errors in the touched edge functions (`.abortSignal` on `PostgrestBuilder`) are also pre-existing — same check.
- Browser-verified: cross-page search from `/faq` → `/?q=…`; two consecutive searches both hitting the API (the fixed bug); back button between searches; logo click and Clear both resetting fully, including re-searching the same term straight after; mobile magnifier toggle with no horizontal overflow.

## Two open questions

1. `DashboardPage` still renders its own `SearchBar` for the saved-artists block — arguably redundant with the header now. Left in place as out of scope.
2. The marketing sections below the results still show after a search. Only the hero was hidden; they don't compete for the fold, but happy to hide them too.

### DIFF
--- a/api/edge/artist-page-static.ts
+++ b/api/edge/artist-page-static.ts
@@ -94,6 +94,11 @@ const CSS = `
   .site-header .brand { display: flex; align-items: center; gap: 8px; font-size: 20px; font-weight: 700; color: var(--text); text-decoration: none; flex-shrink: 0; }
   .site-header .brand:hover { opacity: 0.8; }
   .site-header .brand svg { flex-shrink: 0; }
+  .site-header .header-search { display: flex; gap: 8px; flex: 1; max-width: 420px; margin: 0 auto; }
+  .site-header .header-search input { flex: 1; min-width: 0; background: var(--bg2); border: 1px solid var(--border); border-radius: 8px; padding: 8px 12px; font-size: 14px; color: var(--text); font-family: inherit; }
+  .site-header .header-search input:focus { outline: none; border-color: var(--accent); }
+  .site-header .header-search button { border: 0; border-radius: 8px; padding: 8px 14px; background: var(--accent); color: #fff; font-size: 14px; font-weight: 500; cursor: pointer; font-family: inherit; }
+  @media (max-width: 640px) { .site-header .header-search { display: none; } }
   .site-header .nav-right { display: flex; align-items: center; gap: 12px; font-size: 14px; }
   .site-header .nav-link { color: var(--muted); text-decoration: none; }
   .site-header .nav-link:hover { color: var(--text); }
@@ -277,6 +282,14 @@ export default async function handler(request: Request, context: Context) {
 <body>
   <header class="site-header">
     <a href="/" class="brand">${LOGO_SVG} Unstream</a>
+    <!-- Plain GET form, no JS: this page is pure SSR, so it can't render the
+         React HeaderSearch. Submitting lands on /?q=… which is where the SPA
+         renders results. Never point this at /search — that URL belongs to the
+         noscript-search edge function. -->
+    <form class="header-search" action="/" method="get" role="search">
+      <input type="text" name="q" placeholder="Search artists..." aria-label="Search artists" enterkeyhint="search">
+      <button type="submit">Search</button>
+    </form>
   </header>
 
   <div class="page-content">
@@ -380,6 +393,14 @@ export default async function handler(request: Request, context: Context) {
 <body>
   <header class="site-header">
     <a href="/" class="brand">${LOGO_SVG} Unstream</a>
+    <!-- Plain GET form, no JS: this page is pure SSR, so it can't render the
+         React HeaderSearch. Submitting lands on /?q=… which is where the SPA
+         renders results. Never point this at /search — that URL belongs to the
+         noscript-search edge function. -->
+    <form class="header-search" action="/" method="get" role="search">
+      <input type="text" name="q" placeholder="Search artists..." aria-label="Search artists" enterkeyhint="search">
+      <button type="submit">Search</button>
+    </form>
   </header>
 
   <div class="page-content">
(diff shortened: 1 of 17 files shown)


api/edge/artist-page-static.ts
```diff
@@ -94,6 +94,11 @@ const CSS = `
   .site-header .brand { display: flex; align-items: center; gap: 8px; font-size: 20px; font-weight: 700; color: var(--text); text-decoration: none; flex-shrink: 0; }
   .site-header .brand:hover { opacity: 0.8; }
   .site-header .brand svg { flex-shrink: 0; }
+  .site-header .header-search { display: flex; gap: 8px; flex: 1; max-width: 420px; margin: 0 auto; }
+  .site-header .header-search input { flex: 1; min-width: 0; background: var(--bg2); border: 1px solid var(--border); border-radius: 8px; padding: 8px 12px; font-size: 14px; color: var(--text); font-family: inherit; }
+  .site-header .header-search input:focus { outline: none; border-color: var(--accent); }
+  .site-header .header-search button { border: 0; border-radius: 8px; padding: 8px 14px; background: var(--accent); color: #fff; font-size: 14px; font-weight: 500; cursor: pointer; font-family: inherit; }
+  @media (max-width: 640px) { .site-header .header-search { display: none; } }
```
brandonlucasgreen, line 101: **Regression — edge case, needs-manual-verification**

On the pure-SSR pages (`/a`, `/artist`, and the matching rule in `u-handle.ts`), the header search form is hidden entirely below 640px with no fallback control — these pages have no JS, so there's no mobile magnifier icon to expand it the way `Header.tsx` does for the React app. Not a regression against prior behavior (these pages never had search before), but it means the PR's "search reachable from anywhere" doesn't actually reach mobile visitors landing on a no-JS artist/handle page — likely the most common entry point for a shared link. Worth a conscious call on whether that's acceptable for this PR or a fast-follow.

@brandonlucasgreen

---

brandonlucasgreen (commented):
@brandonlucasgreen Security + regression pass on the header search PR.

**Security** — clean. No auth/authz surface touched, no new persistence or secret handling. The `q` param is consistently `encodeURIComponent`-ed before building URLs (`HeaderSearch.tsx`), and the new plain `<form method="get">` on the SSR pages (`artist-page-static.ts`, `u-handle.ts`) doesn't interpolate any user input into the served HTML — it's a static form, so no injection/XSS surface. `/api/suggest` was already rate-limited and cached before this PR.

**Regression** — traced the reworked `App.tsx` deep-link effect carefully since it changes a load-bearing guard (`!hasSearched` → `queryParam !== lastSearchedRef.current`): the URL-resolution branch, the new bare-`/`-reset branch, and `handleGoHome`'s ordering with `justWentHomeRef` all check out against the scenarios that matter (repeat search of the same query, back/forward between two searches, logo-click reset, resolved-URL search) — no confirmed bugs found there, and it's backed by the new `HeaderSearch` tests plus the existing suite.

One `needs-manual-verification` item (inline): the SSR fallback search form on `/a`, `/artist`, and `/u` pages is hidden entirely below 640px with no alternative control, so mobile visitors landing on a no-JS artist/handle page — plausibly the single most common entry point — don't get the "search from anywhere" benefit this PR is about. Not a regression against prior behavior, just an incomplete edge for this rollout; worth a conscious call rather than a silent gap.

---

PR #346, opened 2026-07-29
Fix mobile Safari zoom-on-focus on header search bar

brandonlucasgreen:
## Summary
- The persistent header search input (`HeaderSearch.tsx`) used `text-sm` (14px), and iOS Safari auto-zooms in on focus for any input under 16px font-size, breaking the page's responsive width until the user manually zoomed back out.
- Bumped the input to `text-base` (16px) below the `lg` breakpoint (mobile/tablet), keeping `text-sm` on desktop to match the surrounding nav text where zoom isn't a concern.

## Test plan
- [x] Verified in the dev server at a 375x812 mobile viewport that the input's computed `font-size` is now `16px`
- [x] Confirmed the mobile search row layout/visual size is unchanged (screenshot before/after)
- [x] `npm run lint` — no new errors introduced (file is clean; pre-existing unrelated test errors elsewhere)

### DIFF
--- a/apps/web/src/components/HeaderSearch.tsx
+++ b/apps/web/src/components/HeaderSearch.tsx
@@ -84,7 +84,14 @@ export function HeaderSearch({
         aria-label="Search artists"
         enterKeyHint="search"
         autoFocus={autoFocus}
-        className="w-full bg-bg-secondary border border-border rounded-lg pl-9 pr-3 py-2 text-sm text-text-primary placeholder-text-muted focus:outline-none focus:border-accent-primary transition-colors"
+        // text-base (16px) below lg: iOS Safari zooms in on focus for any input
+        // under 16px, which also breaks the page's responsive width until the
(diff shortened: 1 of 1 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Security/regression review.

**Security** — none applicable. Pure CSS class change on a static input, no user input, no external calls, no auth/data paths touched.

**Regression** — none found. The class string is built via string concatenation (`+`) instead of a template literal, but every class name stays intact and whitespace-delimited within its own literal chunk (no class name is split across a `+` boundary), so Tailwind's static content scanner will still pick up `text-base`, `lg:text-sm`, etc. correctly — this won't get purged from the production build. Only visual/layout effect is the intended one (16px font on mobile prevents iOS Safari's auto-zoom-on-focus; desktop unchanged at `text-sm`).

No blocking issues — looks safe to merge.

---

PR #347, opened 2026-07-29
Artist pages: let claimed artists add dividers between their links

brandonlucasgreen:
An artist with a verified profile asked for dividers in the list of their platform links, so they can group them on their artist page.

## What changed

**Editor** (`/artist-edit/:slug`) — a third button, **+ Add divider**, next to "Add platform" and "Add other link". A divider appears as a row labelled "Divider" with a rule through it, reorders with the same ▲▼ arrows, and removes with the same ✕.

**Public page** (`/a/:slug`) — renders as a horizontal line between links in the "Support directly" list.

## Why a position, not a row

A divider is stored as a position in the artist's link order — a new `artist_profiles.link_dividers integer[]`, where each element is "the number of links before this divider".

The obvious alternative, a row in `artist_links`, was rejected: it would need a fake URL to satisfy `url text not null`, and every reader of that table would have to learn to filter it out — search results (`db.ts` maps all rows into `platforms`), the public v1 API, the Apple app, the embed widget. Positions keep the change contained to the artist page.

This is safe because the two non-artist writers to `artist_links` both bail on claimed artists (`persistSearchResults`, `persistEnrichment` check `match_confidence === 'claimed'`), and `PUT /api/artist-profile` deletes all rows and re-inserts with `display_order = index` — so the order the positions count against is exactly the artist's own list.

## Reviewer notes

- **Two renderers.** `/a/:slug` is server-rendered by `api/edge/artist-page-static.ts`, but a client-side navigation into it never hits the edge — React `ArtistPage` fetches `/api/artist-page` and renders `RichArtistProfile.tsx`. Both are updated, and both call the same shared helper.
- `api/shared/link-dividers.ts` is the one module both Deno (edge) and Node (functions) can import. It translates stored positions into indexes in the *main* (non-social) list, and drops leading, trailing, and repeated dividers — a rule with nothing on one side reads as a rendering bug, not as grouping.
- **Socials are a separate section.** A divider placed next to a social link falls into the nearest gap in the main list rather than vanishing. The editor says so rather than hiding it.
- **Migration** `20260729160000_artist-link-dividers.sql` adds one nullable column; it auto-applies on merge. No RLS change — `artist_profiles` policies already cover it.
- **Not verified end-to-end locally.** `npm run dev` implements no artist endpoints (`apps/web/server/` is search only), so the editor's load/save round-trip only exists in production. Rendering is covered by tests; the save path was reviewed by reading, not by running. Worth a click-through on the deploy preview.

## Tests

`npm run build` green. New: 7 cases for the position→index translation (`api/functions/__tests__/link-dividers.test.ts`) and 2 for divider rendering in `RichArtistProfile.test.tsx`. Full suites: 468 web unit, 157 API.

### DIFF
--- a/api/functions/artist-profile.ts
+++ b/api/functions/artist-profile.ts
@@ -32,11 +32,20 @@ const CORS_HEADERS = {
   'Access-Control-Allow-Methods': 'PUT, OPTIONS',
 };
 
+// A `platform: 'divider'` entry is a position marker in the artist's link
+// order, not a link — it carries no URL and is stored on the profile
+// (artist_profiles.link_dividers) rather than as an artist_links row.
 interface LinkUpdate {
   platform: string;
-  url: string;
+  url?: string;
   displayName?: string;
 }
+
+export const DIVIDER_PLATFORM = 'divider';
+
+// Cap dividers per profile. Every gap in a link list is a legitimate divider
+// position, so this only rules out a payload that isn't a real link list.
+const MAX_DIVIDERS = 20;
 
 // Allowed embed domains for featured releases
 export const ALLOWED_EMBED_DOMAINS = [
@@ -329,8 +338,11 @@ export async function handler(event: { httpMethod: string; headers: Record<strin
       return { statusCode: 500, headers: CORS_HEADERS, body: JSON.stringify({ error: 'Failed to update links' }) };
     }
 
-    // Insert new links
-    const validLinks = (body.links || []).filter(l => {
+    // Split the submitted order into links and divider positions. The editor
+    // sends one ordered list containing both, so the artist's arrangement of
+    // links and dividers stays a single source of truth.
+    const entries = (body.links || []).filter(l => {
+      if (l.platform === DIVIDER_PLATFORM) return true;
       if (!l.platform || !l.url) return false;
       try {
         const parsed = new URL(l.url);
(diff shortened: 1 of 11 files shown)


api/functions/artist-profile.ts
```diff
@@ -339,6 +351,35 @@ export async function handler(event: { httpMethod: string; headers: Record<strin
...
+    for (const entry of entries) {
+      if (entry.platform !== DIVIDER_PLATFORM) {
+        linksSoFar++;
+      } else if (linksSoFar > 0 && !dividerPositions.includes(linksSoFar)) {
+        dividerPositions.push(linksSoFar);
+      }
+    }
+    const finalDividers = dividerPositions
+      .filter(p => p < validLinks.length)
+      .slice(0, MAX_DIVIDERS);
```
brandonlucasgreen, line 371: @brandonlucasgreen Minor, not a security issue: past `MAX_DIVIDERS` (20), extra dividers are dropped silently — `finalDividers.slice(0, MAX_DIVIDERS)` with no error surfaced, so an artist who adds a 21st divider will save fine and then quietly lose it (or whichever ones sort past the cap) with no client-side feedback. Given 20 is generous for a link list this is unlikely to matter in practice — flagging as **needs-manual-verification** rather than a bug, since it's a product-severity call, not a correctness one.

---

brandonlucasgreen (commented):

**Security:** No confirmed issues. Ownership/authorization is intact — the divider write path sits inside the existing `profile.user_id !== userId` / `!profile.verified_at` checks in `artist-profile.ts` (lines 180-186), so only the verified owner can set dividers. Divider positions are server-computed from the position of `platform: 'divider'` entries in the submitted list, not taken as raw client-supplied integers, and are capped at 20 before being written — no injection surface there. The edge-rendered HTML for the divider (`<hr>` with static inline styles, no interpolated user data) doesn't add any new XSS surface; existing `escapeHtml` usage on link URLs/names is untouched. Both renderers (`api/edge/artist-page-static.ts` and `api/functions/artist-page.ts` → `RichArtistProfile.tsx`) gate dividers behind claimed/verified status correctly.

**Regression:** Traced `mainLinkDividerIndexes` (`api/shared/link-dividers.ts`) by hand against all 7 test cases plus a few not covered (dividers adjacent to a social link at the start/end of the list, dividers spanning multiple social links) — the pending/mainCount bookkeeping handles all of them consistently with the stated "leading/trailing/repeated dividers are dropped" rule. The one edge case worth a second pair of eyes (inline comment) is silent truncation past `MAX_DIVIDERS`. Also worth noting per the PR's own callout: the save path (`PUT /api/artist-profile` divider write) was reviewed by reading only, since `npm run dev` doesn't implement this endpoint — worth an actual click-through on the deploy preview before merge, as the author already flagged.

---

PR #349, opened 2026-07-30
Mac app: make it a Mac-assed Mac app

brandonlucasgreen:
Audits the macOS app against the [mac-assed criteria](https://github.com/HunterHillegas/mac-assed-mac-app-skill) and closes the gaps. Before this, greps across the whole app returned **zero** hits for `contextMenu`, `keyboardShortcut`, `CommandGroup`, `NSPasteboard`, `draggable`, `Transferable`, `accessibilityLabel`, and `List` — it was a SwiftUI app that ran on macOS rather than a Mac app.

macOS-only except for shared views under `Views/Shared/`, where iOS font sizes are deliberately left byte-identical.

## Keyboard and focus

- **Focus the search field when the popover opens.** `@FocusState` was declared and bound but never set, so summoning the app with the global hotkey needed a click before typing. Focus is deferred one run-loop turn because the popover's window only becomes key *after* `show()`.
- ⌘F focus search, ⌘1/⌘2 tabs, ⌘W close.

## Settings is now the real Settings scene

Deletes the hand-rolled `NSWindow`, the ⌘, local event monitor that was hijacking the `Settings…` item SwiftUI already provides, and a click-outside global event monitor that did nothing (it only observes *other* apps' events; `.transient` already dismisses). Panes become `Form`/`Section`, so toggles render as Mac checkboxes and the Plex fields get native right-aligned labels.

`openSettings()` performs the real menu item rather than hardcoding the private `showSettingsWindow:` selector, since `SettingsLink` is macOS 14+ and we support 13.

## Links behave like links

One shared `.linkActions(url:openTitle:onOpen:)` modifier gives every platform, saved, and release badge drag-to-Safari/Notes/Finder plus a context menu. Artist cards are draggable with a full menu. Saved rows and support entries get menus too — their refresh/remove buttons are hover-dimmed and had no non-hover path.

`NSSharingServicePicker` becomes `ShareLink` passing a real `URL`, so link-aware targets get a link instead of a sentence. **This also changes iOS**, which previously shared a bare string.

## System integration

- **Services menu** — select an artist name in any app → Services ▸ Find on Unstream. The plist's `NSMessage` and the provider's selector must agree or the item silently does nothing, so a Debug assertion at launch reads `NSMessage` back out of the bundle and checks the selector exists.
- **App Intents / Shortcuts** — `FindArtistIntent` returns platforms and payout shares as a value so it composes in a Shortcut; it calls `UnstreamAPI` directly rather than `AppState` because an intent can run before any popover has existed. `SearchInUnstreamIntent` opens the popover with the search already run.

## Brand color legibility

Five platform brand colors were unreadable in one appearance: Ampwall `#1E1E24` and Discogs `#333333` vanish in dark, Subvert `#D9DBDD` and TikTok/Threads `#E0E0E0` vanish in light. The search-results badge had **no** protection; two other components had an `if hex == "#000000" || hex == "#E0E0E0"` hack that missed three of the five and tested a value no longer in the catalog.

`BrandColor.legible(hex:isDark:)` clamps Rec. 709 luminance into a legible band and keeps hue. It blends toward white rather than scaling channels — scaling clips and shifts hue, turning Ampwall's near-black pure blue. Verified numerically: Ampwall lifts to `#8C8C8F` in dark, Subvert drops to `#9D9EA0` in light, Patreon stays red, Bluesky stays blue.

## Distribution: staying on the GitHub release path

Assessed Mac App Store readiness and decided against it. No private-API blocker (`MediaObserver` uses AppleScript, not `MediaRemote`), but the Apple Events entitlements are a review risk and Apple would take 15–30% of tips — hard to square with a product arguing platforms shouldn't skim from creators.

- **Tip jar → Liberapay on macOS.** iOS deliberately keeps StoreKit: it ships through the App Store where guideline 3.1.1 requires IAP for tipping and an external donation link is a rejection risk. `TipJarStore` is now iOS-only, so the Mac binary no longer links StoreKit.
- **Radiccio detection was silently broken.** `com.radiccio.mac` was scripted but in neither `scripting-targets` nor the exception, so sandbox failed it with `errAEEventNotPermitted` and the error was swallowed. Added it; removed `com.parachord.desktop`, which is read over a WebSocket and never needed it. Also: a denied *Music* prompt used to `return` before `startPolling()`, disabling detection for **every** source.
- **`checkForUpdatesAutomatically` did nothing** — `checkForUpdatesIfNeeded()` was never called. Wired to launch, gated on the setting, `lastCheckTime` persisted so "once a day" isn't "every launch". It also only `print`ed; an update now posts a notification that opens the download.

## Accessibility and type

Labels on every icon-only button, `.isSelected` on tabs, decorative icons hidden, artist/track text selectable. 55 hardcoded font sizes down to 7 (all SF Symbol glyph sizes).

## Also

Popover tabs become a segmented `Picker` (the hand-rolled pills read as a heavy blue slab in dark mode); the signed-in email moves out of the prime slot above the search field to sit beside Sign Out; a result with no platforms says "No links listed yet." instead of rendering as a bare name; welcome window drops `requestUserAttention(.criticalRequest)` and makes launch-at-login an explicit checkbox instead of registering it silently; removed a dead Roadmap link.

## Verification

- macOS Debug and Release builds clean.
- Every introduced API typechecked against the iOS 17 SDK. **A full iOS build could not run locally** — no simulator runtime installed — so the iOS target needs one real build before any TestFlight/App Store submission.
- Services registration confirmed via `pbs -dump`; both App Intents and the shortcut phrases confirmed in the built bundle's `Metadata.appintents`.
- `XCTest` fails with `test runner hung before establishing connection` — **reproduced identically on the unmodified baseline**, so it is pre-existing (accessory app as test host + a CoreSimulator version mismatch), not from this branch.
- **The Radiccio entitlement fix is unverified.** Ad-hoc local builds apply no entitlements and run unsandboxed, where Apple Events are permitted anyway. It needs a signed build to confirm.

## Note for review

Two known non-blocking items: there is still no undo anywhere (remove-from-saved syncs a server deletion irreversibly), and the status item has no right-click menu, so Quit is only reachable through the popover's ellipsis menu. Both were logged rather than fixed here.

### DIFF
--- a/apps/mac/Unstream.xcodeproj/project.pbxproj
+++ b/apps/mac/Unstream.xcodeproj/project.pbxproj
@@ -29,8 +29,10 @@
 		461E318C0B5687BD71C42E75 /* ShortcutRecorderView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1CDD15E71A80A31952A7720C /* ShortcutRecorderView.swift */; platformFilters = (macos, ); };
 		47ADDB550C3A545314574336 /* MirloReleaseChecker.swift in Sources */ = {isa = PBXBuildFile; fileRef = 397ABE663954D6016159C13E /* MirloReleaseChecker.swift */; };
 		5669C87F79BF30EBCC921AB0 /* UnstreamShareExtension.appex in Embed Foundation Extensions */ = {isa = PBXBuildFile; fileRef = C2ACE9E0FA5F270656066467 /* UnstreamShareExtension.appex */; platformFilters = (ios, ); settings = {ATTRIBUTES = (RemoveHeadersOnCopy, ); }; };
+		5708291E100158963205FAA6 /* FindArtistIntent.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3522A2E5601A58A396FDF7C4 /* FindArtistIntent.swift */; };
 		584FBF6DA0057BC30EB4C786 /* BrandIcons.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7FEFD60988639A1ECC95D56A /* BrandIcons.swift */; };
 		5ABBCE0981251D2D775F6DFF /* IndieArtistDirectoryService.swift in Sources */ = {isa = PBXBuildFile; fileRef = A4CEC5CEC07C87673B1F5A61 /* IndieArtistDirectoryService.swift */; };
+		5AF682136B14D6BDCBF42E4A /* UnstreamShortcuts.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1B86644D0E7917D893B1E7A3 /* UnstreamShortcuts.swift */; };
 		5B170D71E9A7E9A471424F16 /* iOSSettingsTab.swift in Sources */ = {isa = PBXBuildFile; fileRef = 359150E4043D5233395827D3 /* iOSSettingsTab.swift */; platformFilters = (ios, ); };
 		5FFFE71A1135656E81B34BDB /* DeviceIDManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = BEFF951B15732D3AD206D247 /* DeviceIDManager.swift */; };
 		640AB7A350949FAA23C72A53 /* iOSContentView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5033CFF636FB4C7C07D315A6 /* iOSContentView.swift */; platformFilters = (ios, ); };
@@ -45,6 +47,7 @@
 		8961565967B20B3F28A7877E /* GlobalHotkeyManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6B52C116B5CC91BC40D028AF /* GlobalHotkeyManager.swift */; platformFilters = (macos, ); };
 		8A4181366EB270C58A708C65 /* SearchBarView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 45AC0E531396E8C63A2001CF /* SearchBarView.swift */; platformFilters = (macos, ); };
 		8C94768621D55383151FD59F /* SearchResponse.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1581BE8FAF07C7C9D3CD99F5 /* SearchResponse.swift */; };
+		8CD55AA9054F310B1DBD51DA /* ServicesProvider.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1DEF52280B036F1EB34F4127 /* ServicesProvider.swift */; platformFilters = (macos, ); };
 		917C407BBA300BE70B9BDEFB /* ScrobbleManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = EEEE83E95439F19BFC679216 /* ScrobbleManager.swift */; platformFilters = (macos, ); };
 		93956240B6028E760C791028 /* SavedArtistsSyncTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 16504D9BC91FD976CD10D549 /* SavedArtistsSyncTests.swift */; };
 		A3526462820B7494F5463BE2 /* UnstreamApp.swift in Sources */ = {isa = PBXBuildFile; fileRef = 913EEE8809001930E4F88F8C /* UnstreamApp.swift */; };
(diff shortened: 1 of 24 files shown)


brandonlucasgreen (commented):
Reviewed the full diff. Scope note up front: this PR touches only the Mac/SwiftUI client (`apps/mac/`) — no `api/`, Supabase, or auth-service changes — so several of the standing security categories (RLS, middleware/CORS, SSRF allowlist, secret storage in Supabase) don't apply to this diff at all. Findings below reflect that scope.

### Security

- **needs-manual-verification** — `UpdateChecker.notify()` (`Platform/macOS/UpdateChecker.swift`) puts the `downloadUrl` from the `/api/desktop/version` response into `content.userInfo["updateUrl"]`. `AppDelegate.userNotificationCenter(_:didReceive:)` (`UnstreamApp.swift`) then does `NSWorkspace.shared.open(url)` with no scheme/host check — it's merged with the pre-existing `artistUrl`/`releaseUrl` keys in the same line, so this isn't a new pattern, but it is a new sink for server-returned data with no allowlist, unlike the explicit-hostname approach the backend uses for outbound fetches. Low severity (first-party HTTPS endpoint), but a `url.scheme == "https"` guard before `open(url)` would be cheap insurance.
- **needs-manual-verification** — `Unstream-macOS.entitlements` swaps the `com.apple.security.temporary-exception.apple-events` target from `com.parachord.desktop` to `com.radiccio.mac`. Well-documented and justified in the PR body, but any Apple-Events exception widens what the sandboxed app can automate — flagging per usual entitlement-diff scrutiny, not because anything looks wrong. The PR itself notes this specific fix is unverified against a signed build.

### Regression

- **plausible** — `AppDelegate.openSettings()` (`UnstreamApp.swift`, ~L707) now locates the Settings command by scanning `NSApp.mainMenu` for the item with `keyEquivalent == ","` and re-sending its action, replacing the old hand-built `NSWindow`. If that lookup ever fails, the fallback is `assertionFailure`, which no-ops in Release — so in production the ellipsis-menu "Settings" item would silently do nothing, with no user-facing error. Since the app is `LSUIElement` (no visible menu bar), this is the *only* path to Settings. The PR's verification notes mention Debug/Release builds compiling cleanly but not specifically exercising this lookup in a signed Release build.
- **plausible** — `searchFromExternalRequest` is called as `AppDelegate.shared?.searchFromExternalRequest(...)` from both `ServicesProvider.findOnUnstream` and `SearchInUnstreamIntent.perform()`. The optional chaining means that if `AppDelegate.shared` is nil at call time (e.g. an App Intent invoked in a narrow window before `applicationDidFinishLaunching` completes), the Service/Shortcut reports success while doing nothing visible — no error surfaced back to the user or the Shortcut.
- **confirmed, non-blocking** — `MediaObserver.checkPermissionsAndStart()`'s new early `startPolling(); return` on Music's `-1743` denial does not double-start polling; the unconditional `startPolling()` at the end of the function is only reached on the non-error paths. Calling this out only because it's the change most likely to look like a state-transition bug at a glance — it isn't.

Everything else (Plex token still behind `SecureField`, no new network parsing of untrusted input, `ServicesProvider`'s 100-char cap + newline collapsing on pasteboard input, `BrandColor.legible`'s luminance math) checked out clean.

@brandonlucasgreen

---

brandonlucasgreen (commented):
@brandonlucasgreen Automated security/regression pass on this PR. No backend/API/secret-handling code is touched here — this is client-only Swift, so most security categories don't apply; noted below where relevant.

## Security

**auth/authorization** — n/a, no change to how the app authenticates to Supabase/UnstreamAPI.

**input validation** — no issues found (confirmed). `ServicesProvider.findOnUnstream` bounds the pasteboard-derived query to 100 chars and rejects empty selections before handing it to search, consistent with how typed search input is already handled.

**secret handling** — no issues found (confirmed). The Plex token field stays a `SecureField`; nothing new is logged (`print` statements added are status/diagnostic text, not credentials).

**injection risks** — no issues found (confirmed). `MediaObserver`'s AppleScript strings are unchanged static literals, not built from external/user input. No `Process`/shell invocation added.

**unsafe defaults / data exposure** — **plausible, needs-manual-verification**: `UpdateChecker.notify()` adds `updateUrl` (server-supplied `downloadUrl` from `GET https://unstream.stream/api/desktop/version`) to the same notification-tap sink (`NSWorkspace.shared.open(url)`) that already handles `artistUrl`/`releaseUrl`. No scheme/host validation is applied before opening — this matches the pre-existing pattern for the other two keys, so it's not a new class of risk, but it does mean a compromised or MITM'd version-check response could get `NSWorkspace.open()` called on an arbitrary URL/scheme. Since the endpoint is first-party HTTPS and this mirrors existing accepted behavior, flagging for awareness rather than as a regression this PR introduced.

- **Entitlements**: swaps the `temporary-exception.apple-events` target from `com.parachord.desktop` to `com.radiccio.mac` (net removal of one exception, addition of another — least-privilege maintained, not widened). PR description already flags this as **unverified** (ad-hoc local builds run unsandboxed, so the entitlement change can't be exercised locally) — agreed this needs a signed build before shipping.

## Regression

- **Plausible, needs-manual-verification**: `AppDelegate.openSettings()` no longer builds its own window — it now locates the SwiftUI-provided "Settings…" menu item at runtime (`NSApp.mainMenu?.items.first?.submenu?.items.first(where: { $0.keyEquivalent == "," ... })`) and synthesizes the click. If that lookup fails (menu not populated yet, item not found), the guard hits `assertionFailure` and returns — but `assertionFailure` compiles out in Release, so in production this becomes a silent no-op: the user clicks Settings (menu, hotkey, or ⌘,) and nothing happens, with no error surfaced. Worth a Release-build smoke test of Settings opening via all three entry points (status item menu, `SettingsLink`-style access, ⌘,) since this is exactly the kind of "no runtime symptom" failure the PR's own `ServicesProvider` comment warns about for the Services-menu wiring.
- **Changed assumption, verified correct**: `MediaObserver` used to `return` on a denied Music automation prompt, disabling polling for every source. It now calls `startPolling()` and continues. Read `fetchNowPlaying()`: it tries Music → Spotify → Radiccio per poll tick regardless of earlier permission state, so a denied Music prompt just means the Music branch keeps failing silently each tick while Spotify/Radiccio still work — matches the intended fix, no repeated permission-dialog risk (macOS doesn't re-prompt after a denial, it just keeps returning error -1743).
- **State transition**: `WelcomeWindowLauncher` now sets `hasLaunchedBefore = true` both on explicit dismiss and via `windowWillClose` (title-bar close button). Previously the title-bar close path didn't mark the flag, so the window would reappear next launch. This is called out as intentional in the PR description — confirmed as an intended behavior change, not an accidental one.
- **API contract drift**: n/a — no server-facing contract changes; `FindArtistIntent`/`SearchInUnstreamIntent` are new client-only entry points into the existing `UnstreamAPI`/`AppState` surface.
- Two known gaps are self-disclosed by the author (no undo on saved-artist removal, no status-item right-click menu) — not blocking, already logged rather than silently left out.

No confirmed security findings. One regression item (`openSettings()` silent-failure path) worth a manual Release-build check before merge.

---

PR #350, opened 2026-07-30
Save artist links in one transaction so a failed save can't wipe them

brandonlucasgreen:
## What happened

Kid Lightbulbs (verified) lost all 13 platform links on 2026-07-29 at 18:15:08Z. Root cause is confirmed, not inferred:

| Time (UTC) | Event |
|---|---|
| 18:07:47 | Deploy preview for #347 goes live — **has** the `link_dividers` write |
| **18:15:08** | Profile saved against that preview → links deleted, then 500 |
| 18:15:46 | #347 merged |
| 18:15:47 | Production deploy |
| **18:16:17** | Migration adds the `link_dividers` column |

The preview's functions use the production Supabase, so the divider write hit a database that didn't have the column yet. Inside `artist-profile.ts` the order was:

1. `DELETE FROM artist_links` → succeeded
2. `UPDATE artist_profiles SET link_dividers` → failed (`PGRST204`, confirmed against the live DB)
3. `return 500` → **the re-insert never ran**

## Blast radius: one artist

Every wipe requires a save through the editor, and every save bumps `artist_profiles.updated_at`. Exactly one profile was updated inside the exposure window (18:07:47–18:16:17): Kid Lightbulbs. Confirmed independently by a manual check of ~40 profiles. The 123 claimed artists with zero links are artists who claimed but never added links — that's the normal state, not damage.

The links have been restored from the only Wayback snapshot of the page (2026-05-21) and are live. Note a `click:mastodon` event on 2026-06-16 means at least one link added after that snapshot is unrecoverable.

## The fix

The generalizable problem isn't the missing column — it's that a destructive delete was followed by two more fallible writes with no transaction. `replace_artist_links()` does all three in one transaction: either the new arrangement is stored or the old one is untouched.

Verified against a real Postgres, not just reasoned about:

- Happy path stores links, display names, ordering, and dividers
- **A duplicate-platform insert failure rolls back with the previous links intact** — this is the other live path into the same wipe, since `artist_links` has `unique(artist_id, platform)`
- Clearing every link still works
- `anon` and `authenticated` are denied; `service_role` succeeds

That last point matters: PostgREST publishes public functions as RPC endpoints, so without the explicit `REVOKE` the anon key could wipe any artist's links.

This change is also self-protecting against the exact ordering problem that caused the incident. If the code deploys before the migration applies, the RPC call fails and **nothing is deleted** — a save that doesn't work, instead of data loss.

## Also

- Divider/ordering math moves to `api/shared/link-dividers.ts` as `buildLinkRows`, unit-tested (8 new cases) instead of inline in the handler
- Failed saves now report to Sentry — this incident was invisible to monitoring, `console.error` only
- The delete stays deliberately unscoped rather than limited to `source = 'claimed'`; a claimed artist's editor list is the complete intended contents of their page, and scoping it would resurface links they'd removed

## Checks

`npm run build` passes (includes `tsc -b`, 164 API tests, 468 web unit tests). Lint unchanged at the same 77 pre-existing `apps/web` problems. `api/` isn't covered by the build's typecheck, so `artist-profile.ts` and `link-dividers.ts` were typechecked explicitly under `--strict`.

Heads up for `semantic-revert-check`: this relocates divider logic out of `artist-profile.ts`, which that bot tends to flag as a partial revert. The behavior is preserved — `buildLinkRows` keeps the leading/trailing/repeated-divider rules and the `other_N` assignment, and the existing `mainLinkDividerIndexes` tests still pass.

Unrelated: `npm run migrate:dry-run` is currently broken — it passes `--project-ref`, which the current Supabase CLI rejects for `db push`. Not touched here.

### DIFF
--- a/api/functions/__tests__/link-dividers.test.ts
+++ b/api/functions/__tests__/link-dividers.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from 'vitest';
-import { mainLinkDividerIndexes } from '../../shared/link-dividers';
+import { mainLinkDividerIndexes, buildLinkRows, MAX_DIVIDERS } from '../../shared/link-dividers';
 
 // The artist page splits links into "Support directly" (main) and "Follow"
 // (social), but divider positions are stored against the full link order — so
@@ -44,5 +44,60 @@ describe('mainLinkDividerIndexes', () => {
   it('handles several dividers in order', () => {
     const links = [link('bandcamp'), link('mirlo'), link('patreon'), link('ko-fi')];
     expect(mainLinkDividerIndexes(links, isMain, [1, 3])).toEqual([1, 3]);
+  });
+});
+
+// buildLinkRows turns the editor's single ordered list into what gets stored.
+// It feeds replace_artist_links, so a mistake here rewrites an artist's page.
+const entry = (platform: string, url = `https://${platform}.example`) => ({ platform, url });
+const divider = { platform: 'divider' };
+
+describe('buildLinkRows', () => {
+  it('stores links in submitted order and keeps dividers out of the rows', () => {
+    const { links, dividers } = buildLinkRows([entry('bandcamp'), divider, entry('patreon')]);
+    expect(links.map(l => l.platform)).toEqual(['bandcamp', 'patreon']);
+    expect(links.map(l => l.display_order)).toEqual([0, 1]);
+    expect(dividers).toEqual([1]);
+  });
+
+  it('gives each "other" link a unique platform id', () => {
+    // artist_links has unique(artist_id, platform), so two bare "other" rows
+    // would abort the whole save.
+    const { links } = buildLinkRows([entry('other'), entry('bandcamp'), entry('other')]);
+    expect(links.map(l => l.platform)).toEqual(['other_0', 'bandcamp', 'other_1']);
+  });
+
+  it('trims and truncates a display name, and stores a blank one as null', () => {
+    const { links } = buildLinkRows([
+      { platform: 'other', url: 'https://a.example', displayName: `  ${'x'.repeat(80)}  ` },
+      { platform: 'other', url: 'https://b.example', displayName: '   ' },
+    ]);
+    expect(links[0].display_name).toBe('x'.repeat(50));
+    expect(links[1].display_name).toBeNull();
+  });
+
+  it('drops leading, trailing, and repeated dividers', () => {
+    expect(buildLinkRows([divider, entry('bandcamp')]).dividers).toEqual([]);
+    expect(buildLinkRows([entry('bandcamp'), divider]).dividers).toEqual([]);
+    expect(buildLinkRows([entry('bandcamp'), divider, divider, entry('mirlo')]).dividers).toEqual([1]);
+  });
+
+  it('caps dividers so a runaway payload cannot balloon the stored array', () => {
+    const entries = [entry('p0')];
+    for (let i = 1; i <= MAX_DIVIDERS + 5; i++) {
+      entries.push(divider, entry(`p${i}`));
+    }
+    expect(buildLinkRows(entries).dividers).toHaveLength(MAX_DIVIDERS);
+  });
+
+  it('returns an empty set when the artist removed every link', () => {
+    // The caller still runs the replace, which is how a full clear is saved.
+    expect(buildLinkRows([])).toEqual({ links: [], dividers: [] });
+  });
+
+  it('drops a divider that only had links removed from one side', () => {
+    // Everything after the divider was invalid and filtered out upstream, so the
+    // position now equals the link count — a rule with nothing under it.
+    expect(buildLinkRows([entry('bandcamp'), entry('mirlo'), divider]).dividers).toEqual([]);
   });
 });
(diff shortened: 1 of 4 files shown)


brandonlucasgreen (commented):

## Security

**auth/authorization** — no issues found (confirmed). Ownership (`profile.user_id !== userId`) and `verified_at` checks in `artist-profile.ts` run before the links block and gate it, unchanged by this diff. The new `replace_artist_links` RPC does `REVOKE ALL ... FROM public/anon/authenticated` + `GRANT EXECUTE ... TO service_role` in the same migration — matches the existing "server-only function" pattern, and since only the service-role client calls it (after the ownership check above), there's no path for an unauthenticated caller to reach it.

**input validation** — no issues found (confirmed). URL/platform validation happens before `buildLinkRows` (unchanged filter: real platform + parseable `http(s)` URL); `buildLinkRows` itself only reorders/truncates/dedupes already-validated entries and is unit-tested for the edge cases (dedup `other_N`, display-name truncation/blank→null, divider capping at `MAX_DIVIDERS`).

**injection risks** — no issues found (confirmed). The SQL function uses `jsonb_array_elements`/`->>` extraction, not string-built dynamic SQL — not injectable via link platform/url/display_name content.

**secret handling** — n/a, no secrets touched.

**unsafe defaults** — no issues found (confirmed). `SET search_path = public` on the function is good practice; empty `p_dividers` correctly falls through to `NULL` via the `array_length(...) > 0` check (empty Postgres arrays report `NULL` length, same outcome as the old `finalDividers.length > 0 ? ... : null`).

**data exposure** — no issues found (confirmed). Sentry `extra` payload is `artistId`/`linkCount`/`dividerCount` only, no PII/URLs/tokens.

## Regression

- **Changed assumption verified, not a bug**: the RPC's `WHERE artist_id = p_artist_id` replaces the old `WHERE id = profile.id` update target. These are different columns — worth flagging since a mismatch here would silently no-op the divider write on every save. Checked `supabase/schema.sql`: `artist_profiles.artist_id` is `unique not null`, so this is equivalent to the old lookup, just via the FK instead of the PK. **Confirmed correct**, no regression.
- **Failure-mode change (intended, verified by description)**: previously a mid-save failure could zero out an artist's links (the incident this PR fixes); now a failed `replace_artist_links` call rolls back the whole transaction, so the artist keeps their prior links and the client sees a 500. This flips "silent data loss" into "loud save failure," consistent with the "never cache uncertainty / fail loudly" principle in CLAUDE.md.
- **Deploy-ordering edge case**: if the function code deploys before this migration applies (the same race that caused the original incident), `client.rpc('replace_artist_links', …)` will 404/error since the function doesn't exist yet — but nothing gets deleted first, so it degrades to a failed save rather than data loss. This is called out in the PR description and matches what the migration comment says; **confirmed** by reading the flow (RPC call is atomic — no separate delete happens client-side anymore).
- **API contract drift**: none — response shape/status codes on the `/api/artist-profile` links block are unchanged (500 + `{error}` on failure, same success path).
- No off-by-one or null/empty-state issues found in `buildLinkRows`; the test suite explicitly covers leading/trailing/repeated dividers and the "links removed from one side of a divider" case.

Overall: solid, well-tested fix for the incident it describes. No confirmed or plausible security findings.

---

brandonlucasgreen (commented):

**Security**

- Authorization: no change. `replace_artist_links` grants `EXECUTE` to `service_role` only, revoking `public`/`anon`/`authenticated` — the RPC can't be reached with the anon key even though PostgREST auto-publishes it. The handler still authenticates the bearer token and checks `profile.user_id === userId` + `verified_at` before ever calling the RPC. **confirmed** — correctly scoped, least privilege.
- Injection: `p_links`/`p_dividers` are passed as bound RPC parameters (jsonb / integer[]), not interpolated SQL, and the function unpacks them with `jsonb_array_elements` + `->>`. **confirmed** — no injection path.
- `SET search_path = public` on the function guards against search-path hijacking. **confirmed** — good practice, matches the existing convention in this codebase's other functions.
- Error handling: failures now report to Sentry with only `artistId`/`linkCount`/`dividerCount` in `extra` — no PII/secrets logged, and the client still only sees a generic `{ error: 'Failed to update links' }`. **confirmed** — no data exposure.
- One thing worth a manual look, not a defect in this diff: the `entries` filter in `artist-profile.ts` (unchanged by this PR) accepts any non-empty `platform` string with a valid http(s) URL — it doesn't check the value against the platform registry. Pre-existing, not introduced here. **needs-manual-verification**.

**Regression**

- Verified the `WHERE artist_id = p_artist_id` in the new SQL function is equivalent to the old `.eq('id', profile.id)`: `profile` is fetched via `.eq('artist_id', artist.id).single()`, so the two predicates identify the same row. No behavior change.
- Divider off-by-one/edge rules (leading/trailing/repeated dropped, `MAX_DIVIDERS` cap) are preserved verbatim in `buildLinkRows` and covered by the new unit tests, including the empty-links and "divider left dangling after upstream filtering" cases.
- Empty-links save: `array_length(p_dividers, 1)` returns `NULL` (not `0`) for an empty Postgres array — worth knowing, but the `CASE WHEN ... > 0 THEN ... ELSE NULL` still resolves to `NULL` in that case, so the empty-array path behaves correctly. Flagging only so it doesn't look like a footgun in future edits to that function.
- The core state-transition fix (partial delete → failed write → 500 with no re-insert) is closed: the new function runs delete+insert+divider-update in one Postgres transaction, so any exception (including the unique-constraint case on duplicate "other" platforms) rolls back everything, including the delete.

No confirmed or plausible vulnerabilities found in this diff. Nice root-cause fix — closing the actual bug class (destructive delete followed by fallible writes) rather than just patching the one missing-column trigger.

---

PR #351, opened 2026-07-30
Release: Mac app 3.3.0

brandonlucasgreen:
Version bump for the Mac 3.3.0 release, plus a stale endpoint that this release makes load-bearing.

## Version bump

3.2.0 → **3.3.0**, build 11 → **12**, across `Info-macOS.plist`, `Info-iOS.plist`, and the share extension in `project.yml`.

## The desktop version endpoint had drifted

`api/functions/desktop-version.ts` still advertised **2.1.0** while the app shipped 3.2.0. That went unnoticed because nothing ever called the update checker — `checkForUpdatesIfNeeded()` was defined and never invoked.

#349 wired it to run at launch, so a stale value here now actively tells users they're up to date when they aren't. Updated to 3.3.0 with real release notes and a comment tying it to `Info-macOS.plist` so the two stay in step on future releases.

## Build status

The DMG is already built and notarized from this commit:

- Developer ID Application: Brandon Green (CV926C8KS8)
- Hardened runtime enabled (`flags=0x10000(runtime)`)
- Notarization submission `cc9921cd-f89b-4c80-b6e7-b87b59a9e43a` — **Accepted**
- Stapled, and verified from inside the mounted image: `spctl` reports `accepted / source=Notarized Developer ID`

Verified in the signed bundle rather than assumed: version 3.3.0 (12), the `com.radiccio.mac` Apple Events entitlement present with `com.parachord.desktop` removed, `NSServices` carrying "Find on Unstream", and `Metadata.appintents` present.

That last one closes a caveat from #349 — the Radiccio entitlement fix could not be verified in an ad-hoc build, because unsigned builds apply no entitlements and run unsandboxed.

## Tests

`test:api` 157 passed, `test:unit` 468 passed. `npm run lint` reports 69 errors, all pre-existing in `apps/web/server/*` and `App.tsx` — none in touched files, and lint is not part of `npm run build`.

A GitHub release is staged as a **draft** with the DMG attached; it needs merging here first so the tag lands on `main`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #352, opened 2026-07-30
Release: Mac app 3.3.1 — fix saved artists not loading

brandonlucasgreen:
Hotfix for a bug reported immediately after 3.3.0 shipped: **Saved Artists rendered empty after signing in, with no error shown.**

The artists were present on the web app, which ruled out the data and the server straight away.

## Root cause

In `apps/mac/Unstream/Services/SavedArtistsSync.swift`:

- `lastSyncTime` is persisted in **UserDefaults**, so it survives quit, relaunch, sign-out and sign-in.
- `syncedArtists` is **in-memory** and starts empty every launch.
- Every caller passes `force: false` except the poll loop's every-fifth iteration (~5 min).

So each launch asked the server *"what changed since `<cursor>`"*, correctly got nothing, merged nothing into an empty list, and showed an empty tab. `pull()` only sets `syncError` on an HTTP failure, so there was nothing for the user to see.

The cursor is then advanced **even on an empty response**, which made the state self-perpetuating: once empty, always empty, until the poll loop's five-minute forced pull.

Confirmed rather than inferred — `defaults read lol.bgreen.Unstream sync.lastSyncTime` returned a live cursor on the affected machine.

**This is pre-existing, not a 3.3.0 regression.** `SavedArtistsSync.swift` is untouched by #349; 3.2.0 had the identical bug. Upgrading and signing in fresh is simply the shape that exposes it, because you arrive holding a cursor and your first pull is a delta.

## Fix

```swift
let needsFullPull = force || syncedArtists.isEmpty
```

A delta pull is only meaningful when there is something to merge into, so an empty list forces a full pull regardless of the cursor — covering launch, sign-in and sign-out uniformly. The response handler uses the same decision, so a full pull gets full-pull semantics (tombstones filtered, list replaced wholesale) rather than falling through to the merge path.

**The general rule:** a durable cursor must never outlive the ephemeral data it indexes. Persist both or neither, and don't advance a cursor on an empty response when the local set is empty — that turns a transient miss into a permanent one.

## Workaround for anyone who reports it before updating

```bash
defaults delete lol.bgreen.Unstream sync.lastSyncTime
```

## Release

3.3.0 → **3.3.1**, build 12 → **13**, across both plists and `project.yml`. `desktop-version.ts` updated so the automatic update check offers it.

DMG built from this commit and ready to upload:

- Developer ID Application: Brandon Green (CV926C8KS8), hardened runtime (`flags=0x10000(runtime)`)
- Notarization `36eb6296-951b-494c-b289-a74236653f14` — **Accepted**, stapled
- Verified from inside the mounted image: `accepted / source=Notarized Developer ID`, version 3.3.1 (13)

`test:api` 164 passed.

### Note on the v3.3.0 tag

The `v3.3.0` tag points at `5321950` (#350), which predates the version bump in #351 — so the tag's tree still says 3.2.0. The DMG that was uploaded is genuinely 3.3.0, so only the tag is misplaced. The 3.3.1 release should be tagged after this merges so its tag contains the matching version.

### DIFF
--- a/api/functions/desktop-version.ts
+++ b/api/functions/desktop-version.ts
@@ -7,9 +7,9 @@ import type { Handler } from '@netlify/functions';
 // v3.3.0 it runs at launch, so a stale value here means users are told they're up to
 // date when they aren't. Keep it in step with Info-macOS.plist.
 const VERSION_INFO = {
-  latestVersion: '3.3.0',
+  latestVersion: '3.3.1',
   downloadUrl: 'https://github.com/brandonlucasgreen/unstream/releases/latest',
-  releaseNotes: 'Keyboard shortcuts, right-click menus, drag-and-drop links, a real Settings window, and "Find on Unstream" in the system Services menu',
+  releaseNotes: 'Fixes saved artists not loading after signing in',
 };
 
 export const handler: Handler = async (event) => {
--- a/apps/mac/Unstream/Info-iOS.plist
+++ b/apps/mac/Unstream/Info-iOS.plist
@@ -21,7 +21,7 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>3.3.0</string>
+	<string>3.3.1</string>
 	<key>CFBundleURLTypes</key>
 	<array>
 		<dict>
@@ -34,7 +34,7 @@
 		</dict>
 	</array>
 	<key>CFBundleVersion</key>
-	<string>12</string>
+	<string>13</string>
 	<key>ITSAppUsesNonExemptEncryption</key>
 	<false/>
 	<key>LSRequiresIPhoneOS</key>
--- a/apps/mac/Unstream/Info-macOS.plist
+++ b/apps/mac/Unstream/Info-macOS.plist
@@ -17,7 +17,7 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>3.3.0</string>
+	<string>3.3.1</string>
 	<key>CFBundleURLTypes</key>
 	<array>
 		<dict>
@@ -30,7 +30,7 @@
 		</dict>
 	</array>
 	<key>CFBundleVersion</key>
-	<string>12</string>
+	<string>13</string>
 	<key>ITSAppUsesNonExemptEncryption</key>
 	<false/>
 	<key>LSApplicationCategoryType</key>
--- a/apps/mac/Unstream/Services/SavedArtistsSync.swift
+++ b/apps/mac/Unstream/Services/SavedArtistsSync.swift
@@ -40,8 +40,18 @@ class SavedArtistsSync: ObservableObject {
         isSyncing = true
         syncError = nil
 
+        // A delta pull is only meaningful when there is something to merge into.
+        // `lastSyncTime` lives in UserDefaults and so outlives quit/relaunch,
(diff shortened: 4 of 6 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Security/regression pass on this hotfix.

## Security

**1. Cross-account data exposure via stale `syncedArtists`/`lastSyncTime` surviving sign-out — CONFIRMED**

`AuthService.signOut()` (`apps/mac/Unstream/Services/AuthService.swift:206-213`) only clears `session = nil`. It never touches `SavedArtistsSync.shared.syncedArtists` or `lastSyncTime` (the latter is a single global `UserDefaults` key, not scoped per user). I checked every write site of both — `syncedArtists =` only appears inside `pull()`, and `lastSyncTime =` only ever gets set from server responses; nothing clears either on sign-out.

The new `needsFullPull = force || syncedArtists.isEmpty` heuristic (`SavedArtistsSync.swift:50`) assumes an empty `syncedArtists` is a reliable signal that "no data has been loaded for the current session." That assumption breaks exactly in the sign-out → sign-in case this PR's own comment claims to cover ("covering launch, sign-in and sign-out uniformly"):

1. User A signs in (no relaunch) → `pull()` sees an empty list, does a full pull, populates `syncedArtists` with A's artists, sets the global `lastSyncTime`.
2. User A signs out. `syncedArtists` still holds A's list; `lastSyncTime` still holds A's cursor.
3. User B signs in on the same running instance. `SyncedArtistsView.task` (`Views/Shared/SyncedArtistsView.swift:22-24`) fires `await sync.pull()` — unforced — as soon as `auth.isSignedIn` flips true. `syncedArtists.isEmpty` is now **false** (still A's data), so `needsFullPull` is false, and the request goes out as a delta pull (`?since=<A's cursor>`) authenticated as B.
4. If the server (scoped to B) has nothing changed since that unrelated timestamp, the merge path leaves A's stale entries in place, merged with whatever of B's own data changed. **B's Saved Artists tab shows A's saved artists** until the 5-minute forced poll (`PopoverView.swift:301`) eventually does a real full pull.

This is a real path on a shared/family Mac using the menu bar app without quitting between accounts, and it's the same class of "cursor outlives the data it indexes" bug this PR is fixing — just not fully closed. Fix: clear `syncedArtists = []` and `lastSyncTime = nil` in `AuthService.signOut()` (or have `SavedArtistsSync` observe sign-out and reset itself), rather than relying on emptiness as a proxy for "fresh session."

**2. Auth/authorization, input validation, secret handling, injection, unsafe defaults (other than #1)** — no issues found in this diff. Token handling, `since` percent-encoding, and CORS/auth boundaries are unchanged; the diff only touches version strings and the pull-mode decision. *(confirmed)*

## Regression

**3. Users with zero saved artists always take the full-pull path — plausible, low severity.** Since `syncedArtists.isEmpty` is permanently true for such a user, every 60s poll iteration now does a full pull instead of ever using the `since` cursor, collapsing the documented 60s-delta/5min-full split for that population. Functionally harmless (same empty result either way) but worth a one-line note if it's not intentional.

**4. No test added for the new `needsFullPull` branch or the sign-out/sign-in transition — confirmed.** `SavedArtistsSyncTests.swift` still only covers `SyncResponse`/`SyncedArtist` decoding; nothing exercises the `force`/`isEmpty`/merge-vs-replace decision this PR changes, so #1 above would not have been caught by CI.

**5. State-transition note (not a regression, just confirming scope):** the merge-vs-replace branch selection (`needsFullPull` at both the request-building call site and the response-handling call site) is read from the same local `let`, so there's no window where the URL built reflects one decision and the response handling reflects another — that part is internally consistent.

---

PR #353, opened 2026-07-31
Scroll to the top of the page on navigation

brandonlucasgreen:
Browsers keep the current scroll position across SPA pushState navigations
and React Router's <BrowserRouter> has no built-in scroll reset, so clicking
a link halfway down a page landed you halfway down the next page.

A small ScrollToTop component mounted inside the router resets scroll when
the pathname changes. It deliberately stays out of the way for back/forward
(the browser restores position itself), same-pathname navigations (the search
page rewrites ?q= as you type), and URLs with a hash anchor.

Claude-Session: https://claude.ai/code/session_01DXZi629BJrChcBr9o53ukb

### DIFF
--- a/apps/web/src/components/ScrollToTop.tsx
+++ b/apps/web/src/components/ScrollToTop.tsx
@@ -1,0 +1,34 @@
+import { useEffect, useRef } from 'react'
+import { useLocation, useNavigationType } from 'react-router-dom'
+
+/**
+ * Scrolls to the top of the page when the user navigates to a new page.
+ *
+ * Browsers keep the current scroll position across SPA (pushState) navigations,
+ * so clicking a link halfway down a page used to drop you halfway down the next
+ * page. React Router has no built-in scroll reset when using <BrowserRouter>.
+ *
+ * Three cases are deliberately left alone:
+ * - POP (back/forward): the browser restores the previous scroll position itself.
+ * - Same pathname: the search page rewrites `?q=` as you search, and yanking the
+ *   page to the top mid-search would be worse than the bug this fixes.
+ * - URLs with a hash: the target anchor decides where the page lands.
+ */
+export function ScrollToTop() {
+  const { pathname, hash } = useLocation()
+  const navigationType = useNavigationType()
+  const previousPathname = useRef(pathname)
+
+  useEffect(() => {
+    const pathnameChanged = previousPathname.current !== pathname
+    previousPathname.current = pathname
+
+    if (!pathnameChanged) return
+    if (navigationType === 'POP') return
+    if (hash) return
+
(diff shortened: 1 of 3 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Automated staff-eng review of this PR.

**Scope:** `ScrollToTop.tsx` (new), `main.tsx` (mounts it inside `BrowserRouter`), `ScrollToTop.test.tsx` (new). Purely client-side, no server/API/DB touched.

### Security
No findings. This component only reads `useLocation()`/`useNavigationType()` from react-router and calls `window.scrollTo(0, 0)` — no external input, no DOM injection (`innerHTML`/`dangerouslySetInnerHTML`), no network calls, no auth/session/secret handling, nothing that touches `api/functions/middleware.ts`'s SSRF/CORS/auth surface. **confirmed: no security-relevant code paths introduced.**

### Regression / correctness
- **plausible — doc comment vs. actual browser behavior:** The comment says hash URLs are skipped because "the target anchor decides where the page lands." For a `history.pushState`-driven navigation (which is what React Router `<Link>` does), browsers do **not** auto-scroll to the fragment the way they do on a full page load — that behavior is specific to real navigations / `location.hash` assignment. So navigating from a scrolled-down page to `/artist/x#releases` likely leaves the viewport at the old scroll offset rather than landing on `#releases`. This isn't a regression from this PR (that case was equally broken before — scroll position was preserved everywhere), but the comment overstates what actually happens, which could mislead the next person who reads it. Worth a manual check in a real browser before treating hash-anchor deep links as working. **needs-manual-verification.**
- **confirmed — initial mount is correctly excluded twice over:** `previousPathname` is seeded from the initial `pathname`, so `pathnameChanged` is already `false` on first render; the `navigationType === 'POP'` check is redundant for the mount case but correct and harmless for real back/forward navigation. No off-by-one/state issue here.
- **confirmed — same-pathname query changes correctly ignored:** covered by test, matches the stated intent (search page rewriting `?q=`).
- **plausible — scope of the "same pathname" exemption is broader than just search:** any other paginated/filterable page that changes only `?query` params (e.g. future `/artists?sort=`) will also keep scroll position by this same rule. Not a bug today (no such page exists yet per the routes in `main.tsx`), just noting the exemption is global, not search-specific, in case that surprises someone later.

### Test coverage
Good: initial mount, pathname change, same-pathname query-only change, hash present, and POP (back) are all covered. CI `check` is green.

Overall: low risk, self-contained, no security concerns. The only thing I'd want eyes on before merging is whether hash-anchor navigation to a *different* route actually needs to scroll to that anchor (currently it won't, on this PR or before it) — worth deciding if that's in scope for this fix or a follow-up.

---

PR #354, opened 2026-07-31
Homepage: screenshot carousel and a display font for headings

brandonlucasgreen:
The homepage showed a single static macOS screenshot and set every heading in the body face. This adds a carousel, a display font, and some accent colour in the headings.

## Carousel

`ScreenshotCarousel` replaces the static teaser image in the "Keep track of artists you want to support" section.

- Prev/next buttons and five dots, crossfade between slides
- 5s autoplay that **stops permanently** on any hover, focus, tap, swipe or button press — it never fights someone who is looking
- Arrow keys, touch swipe, `aria-live` "Screenshot N of 5", and no autoplay under `prefers-reduced-motion`
- Fixed `aspect-[640/783]` frame, so slides never resize the layout

The five source PNGs were 5.6 MB total and not all the same aspect ratio (`macos-saved` is 956×1170, the rest 960×1254). They're cropped to a common box and converted to WebP — **173 KB total**, first slide eager, rest lazy. The `magick` recipe is in a comment at the top of the component so future shots match. Confirmed they are *not* pulled into the PWA precache manifest.

`unstream-mac-teaser.png` stays — `MacAppPromo` still uses it on the 404 page.

## Typography

**Darker Grotesque** carries the headings; Stack Sans Headline stays the body face. Both load in one combined Google Fonts request.

It needed a size bump. Measured via canvas `actualBoundingBoxAscent` at 100px/800, its cap height is **56.3** against Stack Sans's **70** — it renders about 20% smaller at the same `font-size`, so dropped in at the old sizes it looked undersized next to the body copy. Hero went `md:text-5xl` → `md:text-6xl`, section h2s `text-2xl md:text-3xl` → `text-3xl md:text-4xl`.

**Mobile deliberately keeps `text-4xl` on the hero.** At 48px it runs to four lines and pushes the first download button from y=407 to y=479 in an 812px viewport, which would put two of the four CTAs at the fold. There's a comment in `App.tsx` recording why.

Two bits of colour:
- `.marker-word` picks out "right now" in accent orange with a teal wavy rule
- `.heading-accent` puts all three accents in a short bar under each section h2

A highlighter-style fill behind the word was tried first and dropped: translucent orange over `#0d0d0d` just reads muddy brown. Highlighters need a light substrate.

## Login button

`NavItem` gains a `button` flag, set on the logged-out `/login` entry, so it renders as a filled accent button instead of muted grey text. The mobile drawer ignores the flag on purpose — its rows are already full-width tap targets in the primary text colour.

## Reviewer notes

- **The font is declared in five places, not one.** `index.html`, `index.css`, and the three SSR edge functions (`artist-page-static.ts` has the `<link>` twice, once per render function). They're mirrored here so `/a`, `/u` and the no-JS search page don't drift from the SPA.
- **Contrast on the Login button, for the record.** The button block is 6.85:1 against the page where the old muted link was 3.38:1, but white on `#ff6b35` is only **2.84:1**, under WCAG AA. That's the site's existing CTA idiom — every Download button does the same — so it was matched rather than fixed in isolation. There is no AA-passing way to put the brand orange behind white text (`#e55a2b` in light theme is 3.60:1); fixing it means a darker orange token or dark text on the accent, sitewide. Worth a separate decision.
- **The source PNGs aren't committed.** They live untracked in `/docs/screenshots` in the main checkout; only the WebP output is here.

## Testing

`npm run build` passes. 480 unit tests (7 new, covering wrap-around, dot jumps, autoplay handover, reduced motion and keyboard) and 164 API tests green. Lint unchanged from baseline (77 pre-existing).

`npm run test` also reports 4 failures in `tests/integration/search-accuracy.test.ts` ("Matt Young" ×2, "Ben-G!", "French Montana"). That's the live-API suite, excluded from the build by design and unrelated to this change.

Verified by hand in the browser at desktop and mobile widths, in both light and dark themes.


apps/web/src/components/ScreenshotCarousel.tsx
```diff
@@ -0,0 +1,146 @@
...
+    stopAutoplay();
+    go(index + (event.key === 'ArrowRight' ? 1 : -1));
+  };
+
+  const handleTouchEnd = (event: React.TouchEvent) => {
+    const startX = touchStartX.current;
+    touchStartX.current = null;
+    if (startX === null) return;
+    const deltaX = event.changedTouches[0].clientX - startX;
+    if (Math.abs(deltaX) < 40) return;
```
brandonlucasgreen, line 66: **Regression — confirmed:** a plain tap (no drag) returns here before `stopAutoplay()` runs, so it doesn't stop autoplay. That contradicts the PR description ("stops permanently on any hover, focus, tap, swipe or button press") and the module comment above (line 34–35: "any hover, focus, touch or button press"). On a touchscreen, `onMouseEnter`/`onFocus` typically don't fire from a tap, so a visitor who taps the carousel without swiping ≥40px gets no feedback and autoplay silently resumes — the slide can change under their thumb a few seconds later. The new test suite doesn't cover a tap-without-swipe case either.

Suggest calling `stopAutoplay()` unconditionally in `handleTouchEnd` before the `deltaX` threshold check (or in `onTouchStart`), so any touch contact — not just a successful swipe — hands off control.

---

brandonlucasgreen (commented):
@brandonlucasgreen Staff-eng review of #354 (homepage carousel + display font). Scope: security and regression only, per the checklist.

## Security

No findings. This diff has no auth/authorization surface, no new external input handling, no secrets, and no new outbound fetches — it's static markup, CSS, hardcoded font URLs, and bundled screenshot assets. Specifically checked:
- **Auth/authorization** — `Header.tsx`'s new `button` flag on `NavItem` is purely presentational (styling for the existing `/login` entry); no gating logic changed. `needs-manual-verification: no` — confirmed by reading `Header.tsx`/`MobileNav.tsx` in full.
- **Input validation / injection** — the Google Fonts `<link href>` additions in the three edge functions (`artist-page-static.ts`, `noscript-search.ts`, `u-handle.ts`) and `index.html` are static strings, not built from request/query data, so no template-injection risk alongside the existing `escapeHtml`-guarded dynamic fields in those files. `ScreenshotCarousel`'s `SLIDES` array is a hardcoded module constant, not user-controlled.
- **Secrets** — none touched.
- **Data exposure** — none; new screenshots show existing public/marketing app UI.
- **Unsafe defaults** — n/a.

## Regression

1. **Confirmed** — inline comment on `ScreenshotCarousel.tsx` line 66: a plain tap (drag distance < 40px) returns from `handleTouchEnd` before `stopAutoplay()` runs, so autoplay is not disabled by "any tap," contradicting both the PR description and the in-code comment. Low severity (touch-only, cosmetic), but worth a one-line fix since the PR explicitly advertises tap as a stop trigger and the new test suite has no case for it.
2. **Needs-manual-verification** — the font family is now declared in five separate places (`index.html`, `index.css`, and three edge functions), as the PR description itself flags. Nothing enforces they stay in sync, so a future SSR surface (or a `netlify.toml` edge-route change) could silently drift from the SPA. No action needed now — just flagging the maintenance risk the author already called out, since nothing in CI would catch it.
3. **Needs-manual-verification** — PR body claims the new WebP screenshots are "confirmed not pulled into the PWA precache manifest." Checked `apps/web/vite.config.ts`: `VitePWA`'s `workbox` config doesn't set `globPatterns`, and WebP isn't in Workbox's default precache extension list, so the claim checks out against the current config — but it's worth eyeballing the built `sw.js`/manifest once on the deploy preview, since a future bump to Workbox defaults (or an explicit `globPatterns` addition elsewhere) could silently start precaching all ~173 KB of them.

No off-by-one or wraparound issues found in the carousel index math (`(next + SLIDES.length) % SLIDES.length` correctly handles negative steps), and `MobileNav`'s handling of the new `button` field matches its documented intent (ignored in the drawer).

---

PR #355, opened 2026-07-31
Harden check-releases: validate every fetch hop, meter the endpoint

brandonlucasgreen:
## Why

`check-releases` on `main` is an **unauthenticated POST endpoint with no URL validation, no allowlist check, and no rate limit.** It fetches URLs from the request body, then follows a link found *inside* the fetched page, and reflects parsed fields back to the caller.

That's SSRF with partial content reflection, plus an anonymous outbound-request amplifier running from Netlify's IPs. This is independent of the Releases feature work and was flagged as "ship this anyway" during that review.

## The part that isn't obvious

**Validating the URL we are *given* says nothing about the URL we actually *retrieve*.** Node's `fetch` follows redirects transparently, and this endpoint already makes a deliberate second hop to a URL parsed out of fetched HTML.

This isn't theoretical. Bandcamp Pro artists point custom domains at their stores, so the redirect path is exercised by legitimate traffic every day:

```
sufjanstevens.bandcamp.com/music  ->  301  ->  music.sufjan.com/music
```

So an allowlist check at the boundary is redirect-bypassable, *and* a naive allowlist-only fix would break real Bandcamp and self-hosted Faircamp artists.

## What changed

**`middleware.ts`** — split the dangerous-target checks out of `isUrlHostnameAllowed` into a new exported `isSafePublicHostname`, so callers that legitimately fetch off-allowlist hosts can still refuse internal ones. `isUrlHostnameAllowed` now calls it first, so an allowlisted name can never override a private-address check. Two gaps closed while I was in there: the full `127/8` loopback range (only `127.0.0.1` was blocked) and `.local` / `.internal` suffixes.

**`check-releases.ts`**
- A single `safeFetch` path resolves redirects **manually** (`redirect: 'manual'`) and re-validates **every hop**, capped at 5.
- The second-hop album URL is resolved against the page we actually landed on (which may be a custom domain) and re-validated before use.
- Incoming URLs must be allowlisted **or** already stored as a link for that artist.
- Rate limited at the **`lenient`** tier (120/min, 5000/day).
- Refusing to look now returns an explicit error instead of an empty result.

**`db.ts`** — new `isStoredArtistLink(url, artistName)`. Faircamp is self-hosted on arbitrary domains, so `isUrlHostnameAllowed` (which needs a literal `faircamp` label) rejects most genuine Faircamp sites. "Have we already discovered this link for this artist?" gives the same containment without breaking the platform. Fails closed on any error.

## Why `lenient` and not `strict`

`strict` is 10/min. Both shipped clients loop **once per saved artist with no delay and swallow errors**:

- [`service-worker.js`](apps/extension/background/service-worker.js) — `for (const artistName of artistNames)`, weekly
- [`ReleaseAlertManager.swift:203`](apps/mac/Unstream/ReleaseAlertManager.swift:203) — `for entry in entries`

Both then commit their "last checked" state regardless. So `strict` would silently stop release alerts for every artist past the tenth, for anyone tracking more than 10 artists — trading a security hole for a broken feature. `lenient` fits the caller that actually exists.

**A batch request shape is the real fix** and is the natural follow-up once clients can ship one. I deliberately did *not* add unused batch surface here.

## Backward compatibility

Response shape, status codes, and the wildcard CORS origin are unchanged — Mac 3.3.x and extension 2.5.x are already released against this contract. The wildcard is deliberate and now carries a comment saying so: neither native client sends an `Origin` that `buildCorsHeaders()` would accept.

## Testing

**35 new tests** in `api/functions/__tests__/check-releases-ssrf.test.ts`, covering metadata endpoints (AWS/GCP/Alibaba), private ranges, loopback, IPv4-mapped IPv6, decimal/octal-encoded IPs, `.local`/`.internal`, non-HTTP schemes, **the redirect bypass**, custom-domain following, relative-redirect resolution, the redirect cap, the tier choice, and the shipped-client contract.

- `npm run test:api` — 199 passed (16 files)
- `npm run build` — passes end to end
- Explicit `tsc --strict` over the three touched files: only the pre-existing `mirloRssCache is possibly null`, confirmed identical on `origin/main` by checking out that version and re-running. `api/` isn't in the build's typecheck, so this was verified by hand.

`checkRateLimit` is mocked rather than Redis, so no test can reach the real Upstash instance CI has credentials for.

## Known limitation, stated deliberately

This does not defend against **DNS rebinding** — a public hostname whose DNS resolves to a private address. That needs resolution-time checks Node's `fetch` doesn't expose. The docstring on `isSafePublicHostname` says so rather than implying full coverage.

### DIFF
--- a/api/functions/check-releases.ts
+++ b/api/functions/check-releases.ts
@@ -76,8 +76,10 @@ async function safeFetch(url: string, timeoutMs: number = 5000): Promise<Respons
 
     if (response.status < 300 || response.status >= 400) return response;
 
+    // A 3xx with no Location has nowhere to go. Returning it as-is is deliberate rather
+    // than unhandled: callers check `.ok`, which is false for a 3xx, so it fails closed.
     const location = response.headers.get('location');
-    if (!location) return response; // 3xx without a target — nothing to follow
+    if (!location) return response;
     current = new URL(location, current).toString();
   }
 
@@ -209,10 +211,31 @@ async function checkBandcamp(artistUrl: string): Promise<ReleaseResult | null> {
 
     if (!href || !title) return null;
 
-    // Build full URL. `href` comes out of fetched HTML, so it is untrusted input — it is
-    // resolved against the page we actually landed on (which may be a custom domain after
-    // a redirect) and then re-validated by safeFetch before the second request.
-    const fullUrl = href.startsWith('http') ? href : new URL(href, response.url || baseUrl).toString();
+    // `href` comes out of fetched HTML, so it is untrusted input. Resolve it against the
+    // page we actually landed on — which may be a custom domain after a redirect — and then
+    // confine it to that same host.
+    //
+    // The host check matters because safeFetch only asks "is this target safe to fetch",
+    // not "is this target allowlisted or stored for this artist" the way mayFetch does for
+    // the entry URL. Without it, any absolute href appearing in fetched markup becomes a
+    // URL this endpoint will request, which is a narrower trust boundary than the one
+    // established at the entry point and leaves a "fetch an arbitrary third-party URL on
+    // our behalf" primitive. Bandcamp's own templates always link same-origin (verified
+    // against a live /music page: every grid href is root-relative), so this costs nothing
+    // real and still supports the custom-domain redirect this function is built around.
+    const landedUrl = response.url || musicUrl;
+    let fullUrl: string;
+    try {
+      fullUrl = new URL(href, landedUrl).toString();
+      if (new URL(fullUrl).host !== new URL(landedUrl).host) {
+        console.warn(
+          `[check-releases] refused cross-host album link: ${safeHostname(fullUrl)} from ${safeHostname(landedUrl)}`
+        );
+        return null;
+      }
+    } catch {
+      return null;
+    }
 
     // Fetch the album page to get release date
     const albumResponse = await safeFetch(fullUrl);
--- a/api/functions/check-releases.ts
+++ b/api/functions/check-releases.ts
@@ -1,4 +1,13 @@
 import { parse } from 'node-html-parser';
+import { isSafePublicHostname, isUrlHostnameAllowed } from './middleware';
+import { checkRateLimit, getClientIp } from './ratelimit';
+import { isStoredArtistLink } from './db';
+
+const USER_AGENT = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36';
+
+// Bandcamp Pro artists point custom domains at their store, so a single request can
+// legitimately redirect off *.bandcamp.com. Follow a few hops, validating each one.
+const MAX_REDIRECTS = 5;
 
 interface PlatformUrls {
   bandcamp?: string;
@@ -24,20 +33,72 @@ interface CheckReleasesResponse {
   error?: string;
 }
 
-// Helper to fetch with timeout
-async function fetchWithTimeout(url: string, options: RequestInit = {}, timeoutMs: number = 5000): Promise<Response> {
-  const controller = new AbortController();
-  const timeoutId = setTimeout(() => controller.abort(), timeoutMs);
+/**
+ * Fetch with a timeout, validating **every** hop against `isSafePublicHostname`.
+ *
+ * Node's fetch follows redirects transparently, which means a check on the URL we were
+ * given says nothing about the URL we actually retrieve. That matters here for two
+ * reasons: this endpoint fetches caller-supplied URLs, and it then follows a link found
+ * *inside* the page it fetched. Redirects are resolved manually so each destination is
+ * re-validated before another request goes out.
+ *
+ * Returns null when a target is refused or the redirect chain is too long — callers treat
+ * that the same as an unreachable platform.
+ */
+async function safeFetch(url: string, timeoutMs: number = 5000): Promise<Response | null> {
+  let current = url;
 
-  try {
-    const response = await fetch(url, {
-      ...options,
-      signal: controller.signal,
-    });
-    return response;
-  } finally {
-    clearTimeout(timeoutId);
+  for (let hop = 0; hop <= MAX_REDIRECTS; hop++) {
+    if (!isSafePublicHostname(current)) {
+      // Hostname only — the full URL is caller-supplied and shouldn't land in logs.
+      console.warn(`[check-releases] refused unsafe fetch target: ${safeHostname(current)}`);
+      return null;
+    }
+
+    const controller = new AbortController();
+    const timeoutId = setTimeout(() => controller.abort(), timeoutMs);
+    let response: Response;
+    try {
+      response = await fetch(current, {
+        headers: { 'User-Agent': USER_AGENT },
+        redirect: 'manual',
+        signal: controller.signal,
+      });
+    } finally {
+      clearTimeout(timeoutId);
+    }
+
+    if (response.status < 300 || response.status >= 400) return response;
+
+    const location = response.headers.get('location');
+    if (!location) return response; // 3xx without a target — nothing to follow
+    current = new URL(location, current).toString();
   }
+
+  console.warn(`[check-releases] too many redirects from ${safeHostname(url)}`);
+  return null;
+}
+
+function safeHostname(urlString: string): string {
+  try {
+    return new URL(urlString).hostname;
+  } catch {
+    return '<unparseable>';
+  }
+}
+
+/**
+ * May we fetch this URL on an anonymous caller's behalf?
+ *
+ * Allowlisted hosts pass outright. Anything else must already be stored as a link for
+ * this artist — which is how self-hosted Faircamp (arbitrary domains like
+ * music.someartist.com) and Bandcamp custom domains stay reachable without turning the
+ * endpoint into an open fetch proxy.
+ */
+async function mayFetch(url: string, artistName: string): Promise<boolean> {
+  if (!isSafePublicHostname(url)) return false;
+  if (isUrlHostnameAllowed(url)) return true;
+  return isStoredArtistLink(url, artistName);
 }
 
 // Parse various date formats into ISO string
@@ -86,13 +147,8 @@ async function checkBandcamp(artistUrl: string): Promise<ReleaseResult | null> {
       .replace(/\/$/, '');
     const musicUrl = `${baseUrl}/music`;
 
-    const response = await fetchWithTimeout(musicUrl, {
-      headers: {
-        'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36',
-      },
-    });
-
-    if (!response.ok) return null;
+    const response = await safeFetch(musicUrl);
+    if (!response || !response.ok) return null;
 
     const html = await response.text();
     const root = parse(html);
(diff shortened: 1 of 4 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Automated staff-eng review of this PR's security- and regression-relevant surface. Solid hardening overall (redirect-hop revalidation, `redirect: 'manual'`, fail-closed `isStoredArtistLink`, escaped `ilike` pattern, stricter input typing) — findings below are things worth a deliberate decision, not blockers I'd assume are oversights.

## Security

**1. `isSafePublicHostname` never resolves DNS, so the "validate every hop" guarantee doesn't hold against a hostname that is textually public but resolves to a private address — `confirmed` (code-level), exploit path `needs-manual-verification`**
`safeFetch` (api/functions/check-releases.ts:48-80) revalidates every redirect hop with `isSafePublicHostname`, but that function (api/functions/middleware.ts:358-412) is a pure string/regex check on the hostname — it blocks literal IP addresses and a short list of known metadata hostnames, but never does a DNS lookup. I confirmed with Node's own `URL` parser that hex/octal/decimal-encoded IPs already get normalized to dotted-decimal by `new URL()` before this code even sees them (so those specific checks are redundant-but-harmless), but a hostname like `something.attacker-domain.com` — or a wildcard-DNS service like `169-254-169-254.nip.io` — passes every check here and is only actually resolved when Node's `fetch()` connects.

The redirect path this PR was built to support is the exact way to reach this: `*.bandcamp.com` is allowlisted outright (api/functions/check-releases.ts:98-102 `mayFetch`), so a caller doesn't need `isStoredArtistLink` at all — just a Bandcamp Pro artist page whose custom domain the attacker controls DNS for. If that custom domain resolves to `169.254.169.254` or another internal address, `safeFetch`'s hop check (which is `isSafePublicHostname` only, not the allowlist) waves it through, and the underlying request goes wherever DNS actually points.

The docstring on `isSafePublicHostname` (middleware.ts:354-356) calls this out as a "DNS rebinding" limitation, but it isn't really a rebinding/TOCTOU race — it's that no resolution-time check exists at all, and it's reachable in one shot, without a stored link or an authenticated account, via infrastructure (Bandcamp custom domains) this PR explicitly treats as legitimate. Worth a real decision (resolve-then-check-then-connect via an agent/dispatcher that pins the IP, or accept the documented risk explicitly rather than as an aside) rather than leaving it as a caveat line.

**2. `lenient` rate-limit tier is designed for cheap endpoints; this one isn't — `plausible`**
`checkRateLimit(ip, 'lenient', ...)` (check-releases.ts handler) is commented in ratelimit.ts as "120 req/min for cheap high-frequency endpoints (typeahead)". `check-releases` does up to 2 outbound fetches per platform (up to 6 for bandcamp+faircamp+mirlo) to third-party origins per call. 120/min per IP therefore permits up to ~720 outbound third-party fetches/min from a single caller, and nothing stops rotating IPs since there's no global or per-target cap. That only partially closes the "anonymous outbound-request amplifier" problem this PR's own description opens with — worth confirming the tier choice was weighed against amplification risk specifically, not just against not breaking shipped clients.

**3. `isStoredArtistLink` trusts whatever is in `artist_links` — `needs-manual-verification`**
`mayFetch` (check-releases.ts:98) authorizes fetching any URL already stored for the named artist, via `isStoredArtistLink` (db.ts, new). That's a good containment idea, but its safety is only as strong as the write path into `artist_links`. This diff doesn't touch that write path, so I can't confirm from here whether it's populated exclusively by server-side discovery (search/enrichment) or whether a claimed-artist self-edit flow can also write arbitrary URLs into it. If the latter, an attacker who claims/edits an artist profile could store an attacker-controlled URL there and later get it fetched anonymously via this endpoint. Worth confirming the write side has equivalent (or stronger) validation before relying on this as an SSRF boundary.

## Regression

- **Partial fix, self-acknowledged** — the `lenient` tier avoids breaking the Mac app / extension's un-batched per-artist loop for the common case, but doesn't fully solve it for anyone tracking >120 artists in one check cycle; consistent with the PR calling the real fix "a batch request shape" as follow-up. Not a blocker, just flagging it's a mitigation, not closure.
- **Verified, not a regression**: `OPTIONS` response changed 200→204. Neither shipped client depends on the preflight status code (the Mac app doesn't issue browser preflights at all; the extension's own fetch code only checks `response.ok` on the actual POST, not the preflight) — confirmed via `apps/mac/Unstream/ReleaseAlertManager.swift` and `apps/extension/background/service-worker.js`.
- **Verified, not a regression**: stricter `typeof === 'string' && length > 0` checks on `artistName`/platform values only reject inputs that were never valid for shipped clients (both send plain string URLs); the 400-with-`error` early-return path is unchanged in shape for well-formed requests.

---

api/functions/check-releases.ts
```diff
@@ -111,17 +167,14 @@ async function checkBandcamp(artistUrl: string): Promise<ReleaseResult | null> {
...
 
     // Fetch the album page to get release date
-    const albumResponse = await fetchWithTimeout(fullUrl, {
-      headers: {
-        'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36',
-      },
-    });
-
-    if (!albumResponse.ok) return null;
+    const albumResponse = await safeFetch(fullUrl);
```
brandonlucasgreen, line 241: **[unsafe-defaults] plausible** — `mayFetch` (line 98) gates the *initial* platform URL against the allowlist-or-stored-link check, but `safeFetch` itself (used here for the second-hop album page, and for the derived `musicUrl` on line 150) only re-validates each hop against `isSafePublicHostname` — i.e. "not internal", not "on the allowlist or previously stored for this artist."

So once an initial URL clears `mayFetch`, `href` parsed out of the fetched HTML (line 173, explicitly called out as untrusted) can point at *any* public host and it will be fetched, no allowlist/stored-link check applied. That's a narrower trust boundary than the one `mayFetch` establishes for the entry point.

Exploitability today is limited — Bandcamp's own page templates generate `.music-grid-item a[href]`, and a Bandcamp Pro custom-domain redirect still serves Bandcamp-rendered HTML, not attacker-controlled markup. But it's an inconsistency worth closing deliberately rather than leaving implicit: e.g. restrict the second hop to the same hostname as the (already-approved) first hop's final `response.url`, rather than accepting any safe-but-arbitrary public host. That would still support the custom-domain-redirect case this PR is built around, without the "arbitrary public URL embedded in fetched content becomes fetchable" side door.

Distinct from the DNS-rebinding gap, which is already disclosed in the `isSafePublicHostname` docstring — this one isn't currently called out anywhere.

---

brandonlucasgreen (commented):
@brandonlucasgreen Reviewed this against the actual bypass classes it's meant to close, not just the test names.

## Security

**Auth/authorization** — endpoint was and remains anonymous by design (native clients, no user session); the new gate is "may we fetch this on a stranger's behalf," which is the right question for an unauthenticated fetcher. `isStoredArtistLink` correctly binds URL-to-artist via a match-key comparison (not just an `ILIKE` prefilter) so a stored `.../music` link can't authorize an unrelated deeper path — verified the prefilter-escape logic mirrors the existing `deleteStoredLinksForUrl` pattern rather than being freshly (and riskily) hand-rolled.

**Input validation**
- Redirect-hop revalidation is the core fix and it's correctly wired: `redirect: 'manual'`, every hop re-parsed with `new URL(location, current)` and re-checked, capped at 5. I confirmed independently (not just via the test suite) that Node's URL parser already canonicalizes decimal/octal/hex/shorthand IPv4 (`127.1`, `0177.0.0.1`, `0x7f000001`) to dotted-decimal *before* `isSafePublicHostname`'s regex runs, and fullwidth-digit homoglyphs too — so those bypass classes are closed by parser normalization, not just the explicit decimal/octal regexes, which only need to catch the fully-numeric and leading-zero-single-octet forms they target.
- **[unsafe-defaults] plausible** — see inline comment on `check-releases.ts:176`: the second-hop album fetch (and derived `musicUrl`) validates against `isSafePublicHostname` only, not the `mayFetch` allowlist/stored-link gate applied to the entry URL. Narrower trust boundary than the one this PR establishes at the boundary; low exploitability today but worth closing explicitly.
- Request body validation additions (`typeof artistName !== 'string'`, per-key `typeof platforms[p] === 'string'`) correctly handle non-object/array `platforms` without throwing, since JS property access on primitives doesn't crash.

**Secret handling** — n/a, no secrets touched.

**Injection risks** — none; Supabase calls are parameterized (`.eq`, `.ilike` with escaped wildcards).

**Data exposure** — parsed fields returned to the caller are narrow (title/date/url per platform), unchanged from before. `console.warn` on refusal logs hostname only, not full URL — correct call given the URL is caller-supplied.

## Regression

- **API contract**: verified in both shipped clients, not just asserted — `ReleaseAlertManager.swift:274-277` catches on any thrown error (including a 400 from `!response.ok`) and returns `nil` silently; `service-worker.js:643-645` throws on `!response.ok` and the caller already treats any failure as "no release." So the new 400-for-"nothing could be verified" response (replacing what used to reach `checkAllPlatforms` unfiltered) is not a functional regression for either shipped client today — confirmed rather than assumed.
- **State transitions**: `mirloRssCache` module-level cache is untouched by this diff; fine.
- **Off-by-one**: redirect loop is `hop <= MAX_REDIRECTS` → 6 fetch attempts max for 5 configured redirects, matches the test's expectation and the docstring.
- **Null/empty behavior**: 3xx response with no `Location` header now returns the response as-is (falls through the `if (!location) return response`) rather than treating it as an error — reasonable, but note it means a bare `.ok` check downstream on a 3xx-with-no-Location response will just fail closed (`!response.ok` → null), which is the correct outcome, just flagging it was a deliberate path worth the one-line mention since it's easy to misread as unhandled.

Nothing else stood out. The `lenient`-tier rationale is well-argued and matches what the client code actually does (loop-per-artist, no batching) rather than an assumption about it.

---

brandonlucasgreen (commented):
**Security review**

1. **[plausible] Second-hop fetch bypasses the origin allowlist/stored-link gate, not just the private-IP block.** `safeFetch` (`check-releases.ts`) validates every hop with `isSafePublicHostname` only — it never re-runs `mayFetch`'s allowlist-or-stored-link check. In `checkBandcamp`, the album `href` comes straight out of the first page's HTML (`fullUrl = href.startsWith('http') ? href : ...`), so if that HTML contains an absolute `href` to any *public* host:port, the endpoint will issue a second GET there — no longer restricted to the artist's own domain or the platform allowlist. Internal targets are still blocked (`isSafePublicHostname` covers metadata/private ranges), so this isn't the internal-network SSRF the PR set out to close, but it is a residual "fetch an arbitrary third-party public URL on our server's behalf" primitive, gated only by needing a URL that's already allowlisted-or-stored for *some* artist — which is exactly the self-hosted Faircamp/custom-domain case this PR added support for. Whether that's a meaningful barrier depends on how `artist_links` rows get written (auth-gated claim/edit flow vs. auto-ingested from external metadata); I didn't trace that far, hence "plausible" rather than "confirmed." Worth a look before calling the endpoint fully hardened.

2. **[confirmed, already disclosed]** DNS rebinding remains out of scope, as the docstring says — validation is on the literal hostname string, not the resolved address. No action needed beyond what's documented; restating for the record since this function is explicitly the SSRF boundary.

**Regression**

- No breaks found in the preserved contract (response shape, wildcard CORS, `lenient` rate-limit tier) — all covered by tests. One shape change: the new "no platform could be verified" 400 now returns `{artistName, release: null, error}` instead of the old bare `{error}`, but that's a *different* failure mode (previously-missing-platform vs. now-refused-platform) so it's additive, not a break. **[needs-manual-verification]**: didn't check whether the Mac app (3.3.x) or extension (2.5.x) branch on `Object.keys(body)` in a way that would notice the new fields on an error response.
- Off-by-one on `MAX_REDIRECTS` is fine: loop runs hop 0..5 inclusive = 6 total fetch attempts, matching the test's `toBeLessThanOrEqual(6)`.

Solid fix overall — the redirect-revalidation insight (allowlist-the-input says nothing about what you retrieve) is the right one, and the test suite is unusually thorough for an SSRF fix.

@brandonlucasgreen

---

PR #356, opened 2026-07-31
Releases: schema and pure utils

brandonlucasgreen:
Step 1 of Unstream Releases. **Schema plus the pure helper module** every later phase depends on. No application code reads or writes these tables yet — deployable and reversible on its own.

Scope per the agreed plan: three sources (Bandcamp → Discogs → MusicBrainz, no iTunes), demand-driven cataloging, three fan-facing pillars.

## Three tables

A release is one thing that exists in several places, each of which may sell it in several formats at different prices:

| Table | Holds |
|---|---|
| `releases` | one row per actual release |
| `release_sources` | where it can be found — one row per platform |
| `release_offers` | what you can buy there — format, price, availability |

`release_offers` exists in V1 even though the UI comes later, because it's the differentiator — *"vinyl $30 · Bandcamp · ≈$25 to the artist"* is a page nobody else has — and retrofitting it after pages ship costs more than including it now.

## Design decisions that are load-bearing

**`external_id`, globally unique per platform.** The platform's own stable id — Bandcamp's `data-item-id="album-1891263657"`, Mirlo's trackGroup id, a Discogs release id. This is what makes re-ingest idempotent: a title can change upstream, an id can't, so we update in place instead of minting a duplicate. The original spec had no external-id concept at all.

**`curated_fields` + `source`, shipped now for a UI that comes later.** Cataloging re-runs on a schedule. Without field-level provenance, every crawl silently reverts an artist's corrections — data loss that is slow, repeated, and *by design*, which makes it much harder to notice than a one-off bad write. Adding these columns later needs a migration **plus a backfill plus determining which values a human authored**, which is unknowable after the fact. They're nearly free today. Mirrors the existing `artist_links.source` convention.

**`date_precision`.** MusicBrainz returns year-only and month-only dates. Rendering "1 January" for a year-only date invents a fact, and it poisons any date-proximity dedup that assumes day precision.

**`status` (announced/released).** Stored, not computed per request, so feeds filter cheaply — and because a list that silently mixes future and past releases under "newest first" reads as broken.

**No `is_streaming` boolean.** The platform registry already carries `category` and payout, which is strictly more information in one place. (The spec's version also listed four platform ids Unstream doesn't index, and defaulted `true` for a catalog that's overwhelmingly not streaming.)

**Granular `release_type`.** Matching *within* a type is the strongest dedup signal available for artists with no shared identifier.

## release-utils.ts

Pure — no network, DB, or cache. Carries regressions for the three bugs that sank the first attempt ([#336](https://github.com/brandonlucasgreen/unstream/pull/336), which shipped zero tests):

1. **The "hash" wasn't one.** `Buffer.from(title).toString('hex').slice(0, 6)` is the first three bytes of the *input*, so `Album Name.`, `Album Name!` and `Album Name?` all produced the same suffix — precisely the case it existed to separate. Now `createHash('sha1')` (node:crypto, not `crypto.subtle`, which Netlify Functions lack).
2. **Non-Latin titles were deleted.** A bare `[^a-z0-9]` strip reduced `東京`, `Привет`, and `Ⅱ` to `""`, and the ingest then did `if (!norm) continue` — silently excluding every CJK/Cyrillic/Greek title from the catalog with no log line. The match key now keeps any Unicode letter or number while still folding accents via the existing `normalizeAccents`.
3. **Same-run slug collisions clobbered.** The taken-slug set was read once per artist and never updated as rows were inserted.

Worth noting on (2): the obvious fix — reuse `normalizeForComparison` from search-utils — is *also* wrong here. It strips to ASCII, so `Tokyo 東京` and `Tokyo 大阪` would both normalize to `tokyo` and **merge two different records.** Under-merging is recoverable; a wrong merge asserts two albums are one and nobody notices. So this builds on `normalizeAccents` and changes only the character-class step.

**Date bounds are about live data, not hypotheticals.** Mirlo currently carries a release dated **2925-11-02**. Unbounded, one typo sorts to the top of every chronology and lands in every calendar subscriber's feed permanently. Genuine future dates are preserved — Mirlo really does schedule to 2027-09-07.

## Testing

`npm run migrate:dry-run` is broken (passes `--project-ref`, rejected by the current CLI), so the migration was validated by **applying it twice to a throwaway `postgres:15-alpine` container** and asserting each constraint rejects what it should:

- date lower bound; per-artist slug uniqueness; same slug allowed under a *different* artist
- partial unique indexes on MBID and Discogs master (multiple NULLs allowed, duplicates rejected)
- **global `(platform, external_id)` uniqueness across different releases** — the constraint that makes re-crawls safe
- `(release_id, platform)` uniqueness; enum CHECKs; uppercase-only currency; negative prices; name-your-price nulls accepted
- cascade deletes; the `updated_at` trigger; `curated_fields` array round-trip
- second application is a clean no-op (idempotent)

Plus: **34 unit tests**, `npm run test:api` 198 passing, `npm run build` clean, explicit `tsc --strict` over `release-utils.ts` clean.

## Two choices worth a second opinion

- **`schema.sql` deliberately untouched.** It only covers `artists`, `artist_links`, `artist_profiles`; every table added since (suppressions, probes, usernames, saved_artists, merge overrides) lives in migrations only. Editing it would recreate the schema.sql/migration disagreement that #336 shipped. Happy to add it if you'd rather keep that file current.
- **Table named `releases`, not `artist_releases`.** It reads as a first-class entity (which is the point of the feature) and keeps the child tables coherent as `release_sources`/`release_offers`. The counter-argument is consistency with `artist_links`/`artist_profiles`. Cheap to rename now, expensive later — say the word.

### DIFF
--- a/supabase/migrations/20260731120000_releases.sql
+++ b/supabase/migrations/20260731120000_releases.sql
@@ -1,0 +1,239 @@
+-- Migration: releases, release_sources, release_offers
+--
+-- First-class release entities under an artist, for Unstream Releases. Three tables,
+-- because a release is one thing that exists in several places, each of which may sell it
+-- in several formats at different prices:
+--
+--   releases         one row per actual release (album, EP, single…)
+--   release_sources  where that release can be found — one row per platform
+--   release_offers   what you can buy there — one row per format, with price
+--
+-- Schema only. No application code reads or writes these yet.
+--
+-- Deliberately NOT here: the `artist_links.latest_release` jsonb lift proposed in the
+-- original spec. That column holds one release per platform per artist, so lifting it
+-- would produce a "discography" of length 1. Catalog ingest supersedes it. The jsonb
+-- column stays as-is — it remains load-bearing for search disambiguation
+-- (allReleaseTitles, release-overlap merging), which is a separate consumer with a
+-- separate shape.
+
+-- ---------------------------------------------------------------------------
+-- releases
+-- ---------------------------------------------------------------------------
+
+CREATE TABLE IF NOT EXISTS releases (
+  id uuid PRIMARY KEY DEFAULT gen_random_uuid(),
+  artist_id uuid NOT NULL REFERENCES artists(id) ON DELETE CASCADE,
+
+  -- Display-quality title, as the source presents it (accents and punctuation intact).
+  title text NOT NULL,
+
+  -- URL segment for /a/{artist}/{slug}. Unique per artist.
+  slug text NOT NULL,
+
+  -- Normalized title used for dedup only — never displayed. Built with
+  -- normalizeForComparison() from search-utils.ts, which NFD-folds accents rather than
+  -- deleting them, so non-Latin titles survive instead of collapsing to an empty string.
+  match_key text NOT NULL,
+
+  -- Source-native type. Kept granular on purpose: matching *within* a type is the single
+  -- most useful dedup signal we have, and collapsing everything to album-or-single throws
+  -- it away. 'other' is the escape hatch for source values we don't recognize.
+  release_type text NOT NULL DEFAULT 'other'
+    CHECK (release_type IN ('album', 'ep', 'single', 'compilation', 'live', 'remix', 'other')),
+
+  -- Sanity-bounded on the way in. The lower bound is a CHECK; the upper bound
+  -- (roughly today + 3 years) can't be, because now() isn't immutable and CHECK
+  -- constraints must be. Enforced in release-utils instead. This matters: Mirlo has a
+  -- live release dated 2925-11-02, which unbounded would sort to the top of every
+  -- chronology and land in every subscriber's calendar.
+  release_date date CHECK (release_date IS NULL OR release_date >= DATE '1900-01-01'),
+
+  -- How much of release_date is real. MusicBrainz returns year-only and month-only dates,
+  -- and rendering "January 1" for a year-only date invents a fact — it also poisons any
+  -- date-proximity dedup that assumes day precision.
+  date_precision text NOT NULL DEFAULT 'day'
+    CHECK (date_precision IN ('day', 'month', 'year', 'unknown')),
+
+  -- 'announced' = dated in the future or flagged as a pre-order upstream. Derived on
+  -- ingest, stored so feeds and chronologies can filter without recomputing per request.
+  -- A list that silently mixes future and past releases under "newest first" reads as broken.
+  status text NOT NULL DEFAULT 'released'
+    CHECK (status IN ('announced', 'released')),
+
+  artwork_url text,
+
+  -- Cross-source identity anchors. Both are pre-existing groupings we get for free:
+  -- MusicBrainz release groups, and Discogs masters (which already collapse every
+  -- pressing of a record into one entity). Where either is present, dedup is exact
+  -- rather than a title guess.
+  musicbrainz_release_group_id text,
+  discogs_master_id text,
+
+  -- Suppression and review, shipped with the schema rather than added later. Auto-created
+  -- releases inherit the Bandcamp probe's residual wrong-artist rate, and a wrong release
+  -- here becomes a durable URL asserting that an artist made a record they didn't.
+  -- Ingest must never write is_hidden.
+  is_hidden boolean NOT NULL DEFAULT false,
+  needs_review boolean NOT NULL DEFAULT false,
+
+  -- Provenance. 'claimed' means a verified artist authored or corrected this row.
+  -- Mirrors artist_links.source so there's one convention in the codebase.
+  source text NOT NULL DEFAULT 'auto'
+    CHECK (source IN ('auto', 'claimed')),
+
+  -- Field-level provenance: names of columns a verified artist has edited, which a
+  -- re-crawl must leave alone. Scheduled ingest re-runs forever, so without this every
+  -- crawl quietly reverts an artist's corrections — data loss that is slow, repeated, and
+  -- by design, which makes it far harder to notice than a one-off bad write.
+  -- e.g. '{title,release_date}'
+  curated_fields text[] NOT NULL DEFAULT '{}',
+
+  created_at timestamptz NOT NULL DEFAULT now(),
+  updated_at timestamptz NOT NULL DEFAULT now(),
+
+  UNIQUE (artist_id, slug)
+);
+
+CREATE INDEX IF NOT EXISTS idx_releases_artist_id ON releases (artist_id);
+
+-- The primary read for both the artist-page section and the feeds: one artist's releases,
+-- newest first. release_date is nullable and V1 data often lacks it, so created_at is a
+-- stable tiebreaker — otherwise "newest first" is nondeterministic between requests.
+CREATE INDEX IF NOT EXISTS idx_releases_artist_chrono
+  ON releases (artist_id, release_date DESC NULLS LAST, created_at DESC);
+
+-- Dedup lookups happen within (artist, type) — see the release_type comment.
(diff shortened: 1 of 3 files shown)


supabase/migrations/20260731120000_releases.sql
```diff
@@ -0,0 +1,239 @@
...
+ALTER TABLE release_sources ENABLE ROW LEVEL SECURITY;
+ALTER TABLE release_offers ENABLE ROW LEVEL SECURITY;
+
+DROP POLICY IF EXISTS "Public read access" ON releases;
+DROP POLICY IF EXISTS "Public read access" ON release_sources;
+DROP POLICY IF EXISTS "Public read access" ON release_offers;
+
+CREATE POLICY "Public read access" ON releases FOR SELECT USING (true);
+CREATE POLICY "Public read access" ON release_sources FOR SELECT USING (true);
+CREATE POLICY "Public read access" ON release_offers FOR SELECT USING (true);
```
brandonlucasgreen, line 223: **[data-exposure] confirmed** — `is_hidden`/`needs_review` are defined on `releases` (lines 71-77) specifically so a wrong auto-generated release can be suppressed without deleting it, but the `"Public read access" ... USING (true)` policy here doesn't filter on either column. The anon/publishable key that powers the web app, extension, and Mac app can read `releases` (and `release_sources`/`release_offers` via join) directly through PostgREST once RLS admits it — no app code needs to exist for that path to work, that's the point of RLS-as-boundary. So the suppression flag this migration was written to carry (per the PR description: *"a wrong release here becomes a durable URL asserting an artist made a record they didn't"*) currently has zero enforcement at the data layer; it only works if every future reader remembers to add `WHERE is_hidden = false` in application code.

Compare with `platform_link_suppressions` (`20260729130000_platform-link-suppressions.sql`), the existing suppression table for search results: that one is server-only with *no* public policies at all, so there's nothing to forget. `releases` took the opposite shape (suppression flag alongside publicly-readable rows), which needs the policy itself to enforce it.

Suggest narrowing the SELECT policies now, before any client is built against "public read = everything":
```sql
CREATE POLICY "Public read access" ON releases FOR SELECT USING (NOT is_hidden);
CREATE POLICY "Public read access" ON release_sources FOR SELECT USING (
  EXISTS (SELECT 1 FROM releases r WHERE r.id = release_sources.release_id AND NOT r.is_hidden)
);
CREATE POLICY "Public read access" ON release_offers FOR SELECT USING (
  EXISTS (
    SELECT 1 FROM release_sources rs JOIN releases r ON r.id = rs.release_id
    WHERE rs.id = release_offers.release_source_id AND NOT r.is_hidden
  )
);
```
`needs_review` is a softer signal (unverified, not necessarily wrong) so gating on it too is a product call, not a security one — flagging `is_hidden` as the one that matters here.

@brandonlucasgreen

---

brandonlucasgreen (commented):
@brandonlucasgreen Security + regression pass on this one (schema is new, no application code reads/writes it yet, so blast radius is limited — but RLS is the thing that has to be right on day one since it doesn't get re-reviewed later).

## Security

**Auth/authorization**
- Table-level: `releases`/`release_sources`/`release_offers` are public-read (`USING (true)`), service-role write-only (no insert/update/delete policies) — correct shape, matches `artists`/`artist_links`.
- **[confirmed] Data exposure** — see inline comment on the RLS policies (lines 217-223): `is_hidden`/`needs_review` exist to suppress wrong auto-generated releases, but the public SELECT policy doesn't filter on either, so the suppression has no enforcement at the DB layer. Recommend narrowing the policy now.

**Input validation / injection risks**
- No SQL injection surface — this PR is schema + pure functions, no query construction.
- CHECK constraints are sound: `release_date >= 1900-01-01`, enum CHECKs on `release_type`/`format`/`availability`/`source`, `currency ~ '^[A-Z]{3}$'`, `price >= 0`. The upstream-year-typo problem (Mirlo's `2925-11-02`) is correctly handled in `release-utils.ts` (`parseReleaseDate`'s `MAX_YEARS_AHEAD` bound) since a CHECK constraint can't reference `now()`.

**Secret handling** — none introduced.

**Unsafe defaults**
- `set_releases_updated_at()` sets `SET search_path = ''` — correct, avoids search_path hijacking on the trigger function.
- Partial unique indexes on `musicbrainz_release_group_id`/`discogs_master_id` correctly handle NULL-as-distinct via `WHERE ... IS NOT NULL`, so multiple no-MBID rows don't collide.
- `curated_fields`/`is_hidden` are documented as "ingest must never write" but that's a comment, not an enforced constraint (e.g. a trigger or column-level grant) — acceptable for now since no ingest code exists yet, but worth remembering when that code lands.

## Regression (vs. abandoned PR #336, which this replaces)

Verified each of the three named regressions actually has a passing test and correct logic, not just a test that happens to pass:
- `shortHash` — real `sha1` digest vs. #336's byte-prefix; confirmed `Album Name.`/`!`/`?` produce distinct hashes.
- `releaseMatchKey` — keeps `\p{L}\p{N}` (any Unicode letter/number) rather than ASCII-only, so CJK/Cyrillic/Greek titles don't collapse to `""` (under #336, that meant either silent drop or false-merge). Also confirmed it doesn't over-merge (`Tokyo 東京` vs `Tokyo 大阪` stay distinct) — worth calling out because the fix for the drop bug is exactly the kind of change that could accidentally introduce an over-merge bug, and it doesn't here.
- `uniqueReleaseSlug` — takes `taken` as a parameter the caller must update per-insert; #336's version read the set once per artist. The contract (caller must `taken.add()` the result) is easy to violate at a future call site since nothing enforces it structurally — not a bug today, just a sharp edge worth a one-line comment at the call site when ingest is written.

**Edge cases checked:** `isCalendarDate` correctly rejects Feb 31 and non-leap Feb 29 via UTC round-trip rather than trusting `Date`'s rollover; `parseReleaseDate` keeps `precision` distinct from a padded `date` rather than inventing a day for year/month-only input; `deriveStatus` special-cases undated releases as `released` rather than `announced`, with the reasoning documented.

No regressions found in state transitions or API contract — nothing here is read or written by existing code yet, so there's no existing contract to drift from.

Nice catch pulling `release-utils.ts` out as pure/tested-first given what sank #336. The one item above (RLS on `is_hidden`) is the only thing I'd want fixed before this merges, since retrofitting it after a client is built against "public read = everything" is exactly the kind of migration-plus-backfill cost this PR's own design section (curated_fields) argues against paying later.

---

brandonlucasgreen (commented):
**Security review**

No material findings. This PR only adds a migration (schema, no read/write code path yet) and a pure, network-free utility module (`release-utils.ts`), so there's no attacker-controlled input to validate yet.

- RLS follows the repo's established convention correctly: public read policies on all three tables, no insert/update/delete policies (service-role bypasses RLS; anon/authenticated get nothing), and the migration comments say so explicitly. **[confirmed]** matches the "server-only writes" pattern used elsewhere in the schema.
- `set_releases_updated_at()` sets `SET search_path = ''`, the right hardening against search-path hijacking in a trigger function. **[confirmed]** good practice, no gap.
- `shortHash` uses SHA-1, but only for slug-disambiguation suffixes, not for anything security-sensitive (no auth, no integrity check) — not a finding.
- The upper date bound (`now + 3 years`) is enforced only in `parseReleaseDate`, not as a DB CHECK (can't be, since `now()` isn't immutable) — the migration comment says so plainly. Worth remembering once ingest code lands: any writer that bypasses `release-utils.ts` and inserts `release_date` directly skips this bound. Not a bug in this PR — **[needs-manual-verification]** for whoever writes the ingest path next.

**Regression**

None — this is a net-new, fully isolated table set and utility module; nothing existing reads or writes it yet, and the 34 new unit tests plus the described double-apply migration test cover it well.

@brandonlucasgreen

---

PR #357, opened 2026-07-31
Bandcamp grid parser: keep the release id, URL and artwork it already reads

brandonlucasgreen:
Prerequisite for Bandcamp catalog ingest. Pure parser change — **no existing behaviour changes**, and it's independent of the schema in #356, so the two can land in either order.

## What and why

`parseBandcampReleaseTitles` already walks the `.music-grid-item` DOM on a `/music` page, but keeps only accent-folded titles, because it exists to answer *"is this the right artist?"*. Building a release catalog needs what that discards — and it's all in the node the parser is already standing in:

| Kept now | Why it matters |
|---|---|
| **`data-item-id`** (`"album-1891263657"`) | A stable per-release id **and** an album-vs-track type prefix |
| **`href`** | The release URL |
| **artwork** | `src`, or `data-original` when the grid lazy-loads |

The id is the valuable part. It's worth more than the title for identity: **an artist can rename a release, but the id doesn't move**, so re-reading a page updates a row instead of minting a duplicate. `parseBandcampReleaseCounts` already reads this same attribute for its album/track split, so this follows an established pattern rather than introducing one.

## What is deliberately not here: dates

Release dates are **not in the grid at all** — only on individual release pages. So a catalog *with* dates costs one extra request per release rather than coming free with this parse.

This is documented on the function itself, because an earlier review of this feature got it wrong and concluded full-catalog capture was free. Anything that needs dates has to opt into that cost knowingly.

## Verified against the real page

Not just the fixture. Run against the live 168KB `sufjanstevens.bandcamp.com/music` (fetched 2026-07-31):

```
grid releases parsed: 16
with externalId:      16
with artwork:         16
types:                { album: 15, track: 1 }
duplicate ids?        false
any empty titles?     false
```

`{ album: 15, track: 1 }` matches `parseBandcampReleaseCounts` on the same HTML exactly, and titles come through entity-decoded and display-quality (`Carrie & Lowell (10th Anniversary Edition)`).

The fixture markup in the tests is **copied from that page**, including its real whitespace and attribute ordering, so a Bandcamp markup change fails a test rather than silently emptying catalogs in production.

## Edge cases covered

- Single-release **`#discography` sidebar** layout — handled for the same reason `parseBandcampReleaseCounts` handles it: otherwise a one-release artist is indistinguishable from a parked squatter. Grid wins when both are present.
- Missing `data-item-id` → falls back to the URL path for type
- Lazy-loaded artwork → reads `data-original`, not the `data:` placeholder
- Non-release grid entries (merch/packages) skipped
- Accents and non-Latin titles preserved (`Takk... — Sigur Rós`, `東京`)
- Malformed/empty HTML doesn't throw; parked account returns `[]`
- Duplicates deduped; items with no href or title skipped

Two extra tests assert the **existing** parsers still behave identically, since the new one reads the same DOM.

## Testing

- 17 tests added
- `npm run test:api` — 181 passing (16 files)
- `npm run build` — clean
- explicit `tsc --strict` over `search-parsers.ts` — clean (`api/` isn't in the build's typecheck, so verified by hand)

### DIFF
--- a/api/functions/__tests__/bandcamp-grid-releases.test.ts
+++ b/api/functions/__tests__/bandcamp-grid-releases.test.ts
@@ -1,0 +1,221 @@
+// parseBandcampGridReleases — the catalog-building read of a Bandcamp /music page.
+//
+// The grid markup below is copied from a live page (sufjanstevens.bandcamp.com/music,
+// fetched 2026-07-31) rather than invented, including its real whitespace and attribute
+// order, so a Bandcamp markup change shows up here as a failure instead of as silently
+// empty catalogs in production.
+
+import { describe, it, expect } from 'vitest';
+import {
+  parseBandcampGridReleases,
+  parseBandcampReleaseCounts,
+  parseBandcampReleaseTitles,
+} from '../search-parsers';
+
+/** Real grid-item markup, verbatim in shape. */
+function gridItem(opts: {
+  itemId?: string;
+  href: string;
+  title: string;
+  img?: string;
+  lazyImg?: string;
+}): string {
+  const img = opts.lazyImg
+    ? `<img data-original="${opts.lazyImg}" src="data:image/gif;base64,R0lGOD" alt="" />`
+    : opts.img === undefined
+      ? '<img src="https://f4.bcbits.com/img/a1059506682_2.jpg" alt="" />'
+      : opts.img === ''
+        ? ''
+        : `<img src="${opts.img}" alt="" />`;
+
+  return `
+<li ${opts.itemId ? `data-item-id="${opts.itemId}"` : ''}
+    data-band-id="203035041"
+    class="music-grid-item square first-four
+
+    "
+    data-bind="css: {'featured': featured()}"
+>
+    <a href="${opts.href}">
+        <div class="art">
+            ${img}
+        </div>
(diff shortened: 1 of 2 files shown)


brandonlucasgreen (commented):
**Security review**

No material findings — this is a pure HTML parser (`parseBandcampGridReleases`) with no network calls, no DB access, and per the PR description, no consumer yet.

One forward-looking note rather than a bug in this PR: `artworkFrom` returns the raw `src`/`data-original` attribute value verbatim, with no scheme or host validation. **[needs-manual-verification]** — fine today since nothing renders it, but whoever wires `artworkUrl` into an `` later should treat it as untrusted (a compromised/malicious Bandcamp-alike page could point it at an arbitrary URL). Same applies to `title`/`href` — plain extracted text/URLs, not yet escaped for display, which is the right call for a pure parser but worth flagging so the eventual consumer doesn't assume these are pre-sanitized.

**Regression**

None — the two "existing parsers are unaffected" tests directly assert `parseBandcampReleaseCounts` and `parseBandcampReleaseTitles` are unchanged on the same fixture, and the function is entirely additive/unused elsewhere in the codebase.

@brandonlucasgreen

---

PR #358, opened 2026-07-31
Extract the hardened fetch path into a shared module

brandonlucasgreen:
**Pure relocation, no behaviour change.** Prerequisite for release catalog ingest, kept as its own PR so the diff is easy to read and the revert-check flag is easy to dismiss.

## Why share it rather than let ingest do its own fetching

Ingest has to fetch Bandcamp URLs **read out of `artist_links`** — and those aren't trustworthy (a claimed artist can store any URL there; that's the separate task already spun out). So ingest needs the same protections `check-releases` just got.

Reimplementing them is the failure mode worth designing against, because neither check is obvious and each exists because the obvious version was wrong:

1. **Validating the URL you were *given* says nothing about the URL you *retrieve*.** Node's `fetch` follows redirects transparently, so redirects are resolved manually and every hop re-validated.
2. **A hostname string check is not an SSRF boundary.** `169-254-169-254.nip.io` is an ordinary public name that resolves to `169.254.169.254`. Every hop is also resolved, and every answer must be public.

Both lessons now sit at the top of the shared module rather than buried inside one endpoint.

## What moved, and what didn't

**Moved** to `api/functions/safe-fetch.ts`: `safeFetch`, `safeHostname`, `resolvesToPublicAddress`, plus the `FETCH_USER_AGENT` / `MAX_REDIRECTS` constants.

**Deliberately stayed** in `check-releases.ts`: `mayFetch`. It answers *"may we fetch this on an **anonymous caller's** behalf"* — allowlisted, or already stored for this artist — which is specific to that endpoint. Ingest has no anonymous caller, and its URLs come *from* `artist_links`, so the stored-link half would be trivially true and tell it nothing. Ingest will use `isUrlHostnameAllowed` directly to reject a stored URL that isn't actually a Bandcamp link.

Signature is unchanged to keep this a pure move; the only functional difference is the log prefix (`[check-releases]` → `[safe-fetch]`).

One docstring addition worth noting: `safeFetch` now says plainly that it answers **"safe to fetch", not "allowed to fetch"**. Conflating those two questions is exactly what produced the second-hop finding in #355, so it's recorded where the next caller will read it.

## Regression proof

- **All 292 pre-existing API tests pass untouched** — including the 77 that exercise this code through the `check-releases` handler. For a pure move, that's the evidence that matters.
- **13 new tests** pin the now-public exported surface directly, since ingest won't go through that handler: every-hop re-resolution, refusal-by-default on resolver failure / empty answers / junk hostnames, the private-AAAA-behind-public-A case, the redirect bound, relative `Location` resolution, and the deliberate 3xx-without-`Location` fall-through.
- `npm run test:api` — 305 passing (19 files)
- `npm run build` — clean
- Explicit `tsc --strict` over both files — only the pre-existing `mirloRssCache is possibly null`, which reproduces on `main`

## For the semantic-revert check

This is a **relocation, not a revert.** The moved code is byte-identical apart from the log prefix, and the DNS validation added in #355 (825df10) is preserved in full — see the `resolvesToPublicAddress` docstring, which still carries the `nip.io` rationale and the undici-pinning follow-up note.

### DIFF
--- a/api/functions/__tests__/safe-fetch.test.ts
+++ b/api/functions/__tests__/safe-fetch.test.ts
@@ -1,0 +1,146 @@
+// Direct tests for the shared safe-fetch module.
+//
+// check-releases-ssrf.test.ts already exercises this heavily *through* the handler; these
+// pin the exported surface itself, because it is now shared and the next caller (release
+// catalog ingest) won't go through that handler. The properties worth locking are the two
+// that are easy to reimplement wrongly: every hop gets re-validated, and refusal is the
+// default whenever a check can't be completed.
+
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+const mocks = vi.hoisted(() => ({ lookup: vi.fn(), fetch: vi.fn() }));
+vi.mock('dns/promises', () => ({ lookup: mocks.lookup }));
+
+import { safeFetch, safeHostname, resolvesToPublicAddress, MAX_REDIRECTS, FETCH_USER_AGENT } from '../safe-fetch';
+
+function res(status: number, opts: { location?: string; url?: string } = {}): Response {
+  return {
+    status,
+    ok: status >= 200 && status < 300,
+    url: opts.url ?? '',
+    headers: { get: (h: string) => (h.toLowerCase() === 'location' ? opts.location ?? null : null) },
+    text: async () => '',
+  } as unknown as Response;
+}
+
+const PUBLIC = [{ address: '93.184.216.34', family: 4 }];
+
+beforeEach(() => {
+  vi.clearAllMocks();
+  mocks.lookup.mockResolvedValue(PUBLIC);
+  vi.stubGlobal('fetch', mocks.fetch);
+});
+
+describe('safeHostname', () => {
+  it('extracts the hostname', () => {
+    expect(safeHostname('https://example.com/a/b?c=d')).toBe('example.com');
+  });
+
+  it('returns a sentinel rather than throwing on junk', () => {
+    expect(safeHostname('not a url')).toBe('<unparseable>');
+    expect(safeHostname('')).toBe('<unparseable>');
+  });
+});
+
+describe('resolvesToPublicAddress', () => {
+  it('accepts a host whose every answer is public', async () => {
+    mocks.lookup.mockResolvedValue([
+      { address: '93.184.216.34', family: 4 },
+      { address: '2606:4700::1111', family: 6 },
+    ]);
+    expect(await resolvesToPublicAddress('example.com')).toBe(true);
+  });
+
+  it('rejects a host resolving into private space', async () => {
+    mocks.lookup.mockResolvedValue([{ address: '169.254.169.254', family: 4 }]);
+    expect(await resolvesToPublicAddress('169-254-169-254.nip.io')).toBe(false);
+  });
+
+  it('rejects when any single answer is private', async () => {
+    mocks.lookup.mockResolvedValue([
+      { address: '93.184.216.34', family: 4 },
+      { address: 'fd00::1', family: 6 },
+    ]);
+    expect(await resolvesToPublicAddress('dual-stack.example.com')).toBe(false);
+  });
+
+  it('refuses on resolver failure, empty answers, or a junk hostname', async () => {
+    mocks.lookup.mockRejectedValue(new Error('ENOTFOUND'));
+    expect(await resolvesToPublicAddress('nope.example.com')).toBe(false);
+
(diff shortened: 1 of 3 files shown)


brandonlucasgreen (commented):

**Summary: clean relocation, no findings that block merge.** I diffed the moved code line-by-line against what was removed from `check-releases.ts`, and it checks out as claimed — `safeFetch`, `safeHostname`, `resolvesToPublicAddress`, `MAX_REDIRECTS`, `FETCH_USER_AGENT` are byte-identical apart from the log-tag rename. `mayFetch` and the cross-host confinement logic in `checkBandcamp` (the parts that actually decide what's safe/allowed) are untouched.

### Security

- **Auth/authorization** — no change. `mayFetch`'s allowlist-or-stored-link gate is untouched; the new module doesn't add a new entry point.
- **Input validation** — no change. `isSafePublicHostname` and `resolvesToPublicAddress` are preserved verbatim, including the "refuse on any failure/empty answer" default and the all-DNS-answers-must-be-public check that closes the `nip.io`/metadata-endpoint gap. **confirmed** — verified by direct diff, and locked by 13 new unit tests in `safe-fetch.test.ts` that exercise the private-AAAA-behind-public-A case, resolver failure, and empty-answer cases independently of the handler.
- **Secret handling** — n/a, nothing touches secrets here.
- **Injection risks** — n/a, no new parsing/templating introduced.
- **Unsafe defaults** — no change. Refusal-by-default on redirect targets and on resolver failure is preserved (`resolvesToPublicAddress` still returns `false` on any exception).
- **Data exposure** — no change. The full caller-supplied URL still never reaches logs; only `safeHostname()` output does, same as before.

One **needs-manual-verification** note, not a code defect: `safe-fetch.ts` isn't added to `api/tsconfig.json`'s narrow `include` list, so per this repo's own build gotcha (CLAUDE.md) it won't be typechecked by `tsc -b` or caught in CI on a future edit — same exposure `check-releases.ts` already had, just now duplicated onto a second file. Worth adding both to the include list given this is SSRF-security-critical code, but that's a follow-up, not something this PR regresses.

### Regression

- **Changed assumptions** — the only observable behavior change is the log line prefix, `[check-releases]` → `[safe-fetch]`, called out explicitly in the PR description. **needs-manual-verification**: if any external log-based alerting filters on the literal `[check-releases]` string for SSRF-refusal warnings, it will go quiet for this code path until updated. Low severity, easy to miss because nothing in the diff or tests would catch it.
- **Edge cases / null/empty behavior** — unchanged; the 3xx-with-no-`Location` fail-closed behavior and the `<unparseable>` hostname sentinel path are preserved and still covered by tests.
- **Off-by-one logic** — the `hop <= MAX_REDIRECTS` bound (6 attempts total) is unchanged, and the new test `bounds a redirect loop at MAX_REDIRECTS + 1 attempts` pins it directly, which is new coverage `check-releases.ts` didn't have in isolation before.
- **State transitions / API contract drift** — none. `checkBandcamp`/`checkFaircamp`/`checkMirlo`/the handler all keep calling `safeFetch`/`safeHostname` with the same signatures, just from a new import path.

Nothing here rises to confirmed/plausible security risk — the two items above are process notes, not exploitable gaps. `npm run test:api` claim (305 passing) is consistent with the diff; didn't independently re-run CI.

---

brandonlucasgreen (commented):
@brandonlucasgreen Security + regression review.

**Confirmed clean.** Verified this is a pure relocation: `safeFetch`, `safeHostname`, and `resolvesToPublicAddress` in the new `api/functions/safe-fetch.ts` are logically identical to what was removed from `check-releases.ts` (only the log prefix and the `USER_AGENT`→`FETCH_USER_AGENT` export rename changed). `check-releases.ts`'s remaining imports (`isSafePublicHostname`, `isUrlHostnameAllowed`, plus `safeFetch`/`safeHostname` from the new module) match every call site left in that file — no stray references to the removed local copies.

- **Auth/authorization, input validation, injection, secret handling:** N/A — no behavior surface changed.
- **Unsafe defaults / data exposure:** N/A — the SSRF guarantees (per-hop re-validation, all-DNS-answers-public, refuse-on-resolver-failure, redirect-bound) are carried over unchanged, including the documented residual risk (DNS-answer-changes-between-check-and-connect) and its rationale for not pinning to a resolved address.
- **Regression:** none found. The new direct unit tests (`safe-fetch.test.ts`) cover the exported surface, and the PR notes all 292 pre-existing tests (including the 77 exercising this code through the `check-releases` handler) pass unmodified — consistent with a behavior-preserving move.

No blocking issues. My review of the stacked #359 (which depends on this) is posted there.

---

PR #359, opened 2026-07-31
Demand-driven release cataloging from Bandcamp

brandonlucasgreen:
> **Stacked on #358** (shared safe-fetch extraction) — base is `refactor/shared-safe-fetch`, so the diff shows only this work. Merge #358 first and GitHub will retarget this to `main` automatically.

Builds an artist's release catalog when a fan **saves** them or a **search resolves** them. No bulk backfill, no scheduler — per your decisions: *"the pages are generated solely based on demand"*, save **and** search, and no cron yet.

```
save / search  →  requestArtistCatalog()  →  catalog-artist-background
                                              →  one Bandcamp /music fetch
                                              →  releases + release_sources
```

## The three guards

Search is unauthenticated and far higher volume than saving, so this isn't left open-ended:

| Guard | Value | Why |
|---|---|---|
| **Cooldown** | 7 days | Repeat searches for the same artist cost nothing upstream |
| **Hourly cap** | `searched` 60/h, `saved` 240/h | A traffic spike can't turn into us hammering Bandcamp |
| **Priority** | falls out of the two budgets | When searches are being dropped, a deliberate save still gets through |

Plus per-artist exponential backoff on consecutive failures, so a permanently broken artist isn't retried on every single search forever.

## Nothing is on a user's critical path

Both call sites pay **one HTTP handshake** to a background function (returns 202, then runs up to 15 min on its own). All gating happens *inside* that function, so a search never does a DB round trip per artist just to learn that artist was crawled last week.

This is the specific mistake the abandoned first attempt made — it did release writes inline in `persistSearchResults`, which is awaited during search, adding an estimated **0.6–1.6s to every query** after a long effort to get search down to ~1.8–3.0s.

The search hook collects artist IDs during the existing fan-out and makes **one** call afterwards, rather than one per artist.

## Details that are load-bearing

- **A bot challenge is reported distinctly from an empty catalog**, and *throws* rather than recording success. Recording "0 releases" for a blocked fetch would poison the 7-day cooldown with a confident wrong answer. This is the never-cache-uncertainty rule applied to a crawl.
- **`persistReleases` merges and never writes null over a stored value**, so a partial run can't erase artwork or a date an earlier run found. It also **skips any column listed in `curated_fields`** — that's the provenance mechanism from #356 doing its job, so a re-crawl can't revert an artist's edits.
- **Identity is `(artist_id, release_type, match_key)`, not the slug.** A slug derives from a title and moves when the title does.
- **The stored Bandcamp URL isn't trusted just because it's ours.** A claimed artist can save any http(s) URL to their profile links, so it's checked against the allowlist (*"is this really Bandcamp"*) **in addition to** `safeFetch` (*"is this safe to fetch"*). Different questions — conflating them is what produced the second-hop finding in #355.
- **Grid hrefs are confined to the host we landed on.** A stored source URL is later shown to fans as a place to buy the record, so an href out of fetched markup must not point elsewhere. Works after a custom-domain redirect, where same-host means the custom domain.
- **The background function requires a shared secret** and refuses everything when `INTERNAL_TASK_SECRET` is unset (closed, not open). The Discord background function in this repo has no auth; copying that would create an open "make Unstream crawl Bandcamp" endpoint — the exact amplifier #355 existed to close.
- **Dates are left null.** They aren't in the grid at all — one request per release — so guessing would put a fabricated date into a chronology and into subscribers' calendars. A later pass fills them in, lazily and prioritised.

## Not a job queue, deliberately

Cataloging is triggered by a user action and invoked immediately, so there's no scheduler to feed and nothing to drain. Dropping a *searched* artist at the cap is fine — the next person to search or save them triggers it again. A queue is worth adding when something needs guaranteed eventual execution; this doesn't.

## Testing

- **13 tests** on the pure mapping: bot-challenge-vs-empty, cross-host href rejection, custom-domain acceptance, same-page dedup, album-vs-single *not* merging (under-merge, never over-merge), non-Latin titles surviving, colliding slugs, and never inventing a date.
- **Migration validated** by applying it twice to a throwaway `postgres:15-alpine` container: the `last_trigger` CHECK, the rate-cap query shape, the shared `updated_at` trigger, **RLS enabled with zero policies** (server-only, deliberate), and cascade from `artists`.
- `npm run test:api` — 318 passing (20 files)
- `npm run build` — clean
- The 8 `client is possibly null` errors in `saved-artists.ts` are **pre-existing** — verified identical against `origin/main` with a stash baseline, since `api/` isn't in the build's typecheck.

## Needs from you before this does anything

**A new env var: `INTERNAL_TASK_SECRET`** (any long random string) in Netlify. Without it the feature is simply off — `requestArtistCatalog` logs and returns, and the background function refuses all requests. Nothing breaks; cataloging just never starts.

## What this does not do yet

Release pages, feeds, alert rewiring, dates/formats/prices, and Discogs/MusicBrainz. This is the data foundation those sit on — the catalog now fills in as people use the site.

### DIFF
--- a/api/functions/catalog-artist-background.ts
+++ b/api/functions/catalog-artist-background.ts
@@ -29,10 +29,15 @@ const MAX_ARTISTS_PER_RUN = 25;
 const DELAY_BETWEEN_ARTISTS_MS = 1_000;
 
 function isAuthorized(header: string | undefined): boolean {
-  const secret = process.env.INTERNAL_TASK_SECRET;
+  // Reuses the secret this repo already has for internal function-to-function calls
+  // (resolve-url, search-sources, and both v1 wrappers). A second near-identically-named
+  // variable would be a footgun, and the blast radius of this one is small: cooldown and
+  // hourly caps are enforced inside this function regardless of who calls it, and only
+  // artists already in our database can be named.
+  const secret = process.env.INTERNAL_FUNCTION_SECRET;
   // No secret configured means the endpoint is closed, not open.
   if (!secret) {
-    console.error('[catalog] INTERNAL_TASK_SECRET is not set — refusing all requests');
+    console.error('[catalog] INTERNAL_FUNCTION_SECRET is not set — refusing all requests');
     return false;
   }
   if (!header?.startsWith('Bearer ')) return false;
--- a/api/functions/catalog-artist-background.ts
+++ b/api/functions/catalog-artist-background.ts
@@ -1,0 +1,156 @@
+// Netlify Background Function: build an artist's release catalog from Bandcamp.
+//
+// Invoked by requestArtistCatalog() when a fan saves an artist or a search resolves one.
+// Runs off the request path — a `-background` function returns 202 to its caller immediately
+// and then runs for up to 15 minutes — so nothing here is on a user's critical path. That
+// matters: cataloging one artist costs one Bandcamp request, and the search that triggered it
+// must not wait for it.
+//
+// Authenticated with a shared secret. The Discord background function in this repo has no
+// auth, and copying that here would be a mistake: an open endpoint that makes Unstream crawl
+// Bandcamp on demand is exactly the amplifier the check-releases hardening existed to close.
+
+import { timingSafeEqual } from 'crypto';
+import {
+  claimArtistForCatalog,
+  getArtistBandcampUrl,
+  persistReleases,
+  recordCatalogOutcome,
+  type CatalogTrigger,
+} from './db';
+import { isUrlHostnameAllowed } from './middleware';
+import { safeFetch, safeHostname } from './safe-fetch';
+import { bandcampMusicUrl, ingestBandcampGrid } from './release-ingest';
+
+/** Ceiling on artists per invocation, so one call can't become an unbounded crawl. */
+const MAX_ARTISTS_PER_RUN = 25;
+
+/** Pause between artists. One Bandcamp request each, spaced out rather than in a burst. */
+const DELAY_BETWEEN_ARTISTS_MS = 1_000;
+
+function isAuthorized(header: string | undefined): boolean {
+  const secret = process.env.INTERNAL_TASK_SECRET;
+  // No secret configured means the endpoint is closed, not open.
+  if (!secret) {
+    console.error('[catalog] INTERNAL_TASK_SECRET is not set — refusing all requests');
+    return false;
+  }
+  if (!header?.startsWith('Bearer ')) return false;
+
+  const provided = Buffer.from(header.slice(7));
+  const expected = Buffer.from(secret);
+  // timingSafeEqual throws on length mismatch, so compare lengths first.
+  if (provided.length !== expected.length) return false;
+  return timingSafeEqual(provided, expected);
+}
+
+export async function handler(event: {
+  httpMethod?: string;
+  headers?: Record<string, string | undefined>;
+  body?: string;
+}) {
+  if (event.httpMethod !== 'POST') {
+    return { statusCode: 405, body: '' };
+  }
+
+  if (!isAuthorized(event.headers?.authorization ?? event.headers?.Authorization)) {
+    return { statusCode: 401, body: '' };
+  }
+
+  let body: { artistIds?: unknown; trigger?: unknown };
+  try {
+    body = JSON.parse(event.body || '{}');
+  } catch {
+    return { statusCode: 400, body: '' };
+  }
+
+  const artistIds = Array.isArray(body.artistIds)
+    ? body.artistIds.filter((id): id is string => typeof id === 'string').slice(0, MAX_ARTISTS_PER_RUN)
+    : [];
+  const trigger: CatalogTrigger = body.trigger === 'saved' ? 'saved' : 'searched';
+
+  if (artistIds.length === 0) return { statusCode: 400, body: '' };
+
+  let catalogued = 0;
+  let skipped = 0;
+
+  for (const [index, artistId] of artistIds.entries()) {
+    // Cooldown, hourly cap and claim all happen here rather than at the call site, so the
+    // search and save paths pay one cheap invocation instead of a DB round trip per artist.
+    if (!(await claimArtistForCatalog(artistId, trigger))) {
+      skipped++;
+      continue;
+    }
+
+    if (index > 0) await sleep(DELAY_BETWEEN_ARTISTS_MS);
+
+    try {
+      const found = await catalogArtist(artistId);
+      if (found === null) {
+        skipped++;
+      } else {
+        catalogued++;
+      }
+    } catch (error) {
+      const message = error instanceof Error ? error.message : String(error);
+      console.error(`[catalog] artist ${artistId} failed:`, message);
+      await recordCatalogOutcome(artistId, { error: message });
+    }
+  }
+
+  console.log(`[catalog] trigger=${trigger} catalogued=${catalogued} skipped=${skipped}`);
+  return { statusCode: 200, body: JSON.stringify({ catalogued, skipped }) };
+}
+
+/**
+ * Catalog one artist. Returns the release count, or null when there was nothing to do.
+ *
+ * @throws on a genuine failure, so the caller records it against the backoff counter.
+ */
+async function catalogArtist(artistId: string): Promise<number | null> {
+  const storedUrl = await getArtistBandcampUrl(artistId);
+  if (!storedUrl) {
+    await recordCatalogOutcome(artistId, { error: 'no bandcamp link stored' });
+    return null;
+  }
+
+  // The stored URL is not automatically trustworthy: a claimed artist can save any http(s)
+  // URL to their profile links, so a row labelled 'bandcamp' may not be Bandcamp at all.
+  // The allowlist answers "is this really Bandcamp"; safeFetch separately answers "is this
+  // safe to fetch". Both are needed — they are different questions.
+  if (!isUrlHostnameAllowed(storedUrl)) {
+    await recordCatalogOutcome(artistId, { error: `stored bandcamp url not allowlisted: ${safeHostname(storedUrl)}` });
+    return null;
+  }
+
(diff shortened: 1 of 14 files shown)


api/functions/db.ts
```diff
@@ -852,10 +1178,21 @@ export async function persistSearchResults(results: ArtistResult[]): Promise<voi
...
     } catch (error) {
       console.error(`[DB] Error persisting "${result.name}":`, error);
     }
   }));
+
+  // One request for the whole result set, after the writes, and only a 202 handshake is
+  // awaited. Cooldown and rate-cap checks happen inside the background function, so a search
+  // never pays a database round trip per artist to find out an artist was crawled last week.
+  if (catalogCandidates.length > 0) {
+    await requestArtistCatalog(catalogCandidates, 'searched');
```
brandonlucasgreen, line 1194: **Regression (confirmed) — this reintroduces latency on the search critical path, contrary to the PR's stated design.**

`requestArtistCatalog()` does a real outbound `fetch()` to the background function (up to a 3s `AbortController` timeout in `request-catalog.ts`). This call is `await`ed here, and `persistSearchResults()` is itself `await`ed in `search-sources.ts` (line ~2035) *before* the search response object is built and returned — the comment two lines up even says so explicitly ("this function is awaited before the search response is sent").

That directly contradicts the PR description's central claim: *"Both call sites pay one HTTP handshake... nothing is on a user's critical path."* It is on the critical path. In the typical case this is a fast 202 ack, but:
- worst case (Netlify background-function cold start, DNS/TLS hiccup, self-referential HTTP round trip) adds up to 3s to *every* search that resolves an artist with a stored Bandcamp link — likely the majority of searches, since Bandcamp is this product's primary target platform.
- this is close to the exact mistake the PR says the "abandoned first attempt" made (0.6–1.6s added to every query), just moved from an inline DB write to an inline HTTP handshake.

Suggest firing this without awaiting the network call on the response path — e.g. `void requestArtistCatalog(...)` guarded with its own `.catch()`, or moving the call after the response is constructed/sent rather than inside the awaited `persistSearchResults`. If Lambda-freeze-kills-floating-promises is the reason it's awaited (per the comment in `request-catalog.ts`), that's a real constraint worth stating in this file too, but it means the "not on critical path" claim needs to be corrected, not that the await is safe.

---

api/functions/db.ts
```diff
@@ -366,6 +367,326 @@ export async function isStoredArtistLink(url: string, artistName: string): Promi
...
+    if ((count ?? 0) >= CATALOG_HOURLY_CAP[trigger]) {
+      console.log(`[catalog] hourly cap reached for trigger=${trigger} (${count}) — skipping`);
+      return false;
+    }
+
+    const { error: claimError } = await client
+      .from('release_catalog_state')
+      .upsert(
+        { artist_id: artistId, last_attempted_at: new Date(now).toISOString(), last_trigger: trigger },
+        { onConflict: 'artist_id' }
```
brandonlucasgreen, line 456: **Regression (plausible) — state transition / TOCTOU race in `claimArtistForCatalog`.**

The claim is read-then-write, not atomic: SELECT cooldown/backoff state → SELECT hourly count → UPSERT `last_attempted_at`. The docstring above (`claimArtistForCatalog`) says "two near-simultaneous triggers for the same artist won't both do the work, because the second sees a fresh attempt timestamp" — but that only holds if the second request's SELECT happens *after* the first request's UPSERT commits. Two truly concurrent triggers for the same artist (plausible for a trending/newly-searched artist hit by many users within the same second, which is exactly the "search is unauthenticated and high-volume" scenario this function's own comments worry about) can both read the same pre-claim state, both pass the cooldown/backoff/cap checks, and both proceed to `catalogArtist()` — duplicate concurrent Bandcamp fetches for the same artist, and the hourly cap can be oversold by however many requests raced through the count check together.

Not exploitable as a security hole on its own (bounded by `MAX_ARTISTS_PER_RUN`/hourly cap eventually catching up), but it undermines the two guarantees this table exists for (single-claim, hard hourly ceiling). A conditional upsert (e.g. `upsert(...).select()` gated by a `WHERE last_attempted_at < threshold` via a Postgres function/RPC, or a unique constraint + catching the conflict) would make the claim atomic instead of relying on timing.

---

api/functions/catalog-artist-background.ts
```diff
@@ -0,0 +1,156 @@
...
+  if (!storedUrl) {
+    await recordCatalogOutcome(artistId, { error: 'no bandcamp link stored' });
+    return null;
+  }
+
+  // The stored URL is not automatically trustworthy: a claimed artist can save any http(s)
+  // URL to their profile links, so a row labelled 'bandcamp' may not be Bandcamp at all.
+  // The allowlist answers "is this really Bandcamp"; safeFetch separately answers "is this
+  // safe to fetch". Both are needed — they are different questions.
+  if (!isUrlHostnameAllowed(storedUrl)) {
```
brandonlucasgreen, line 135: **Security / input validation (plausible) — `isUrlHostnameAllowed` doesn't actually verify "is this really Bandcamp".**

The comment above this line says the allowlist check answers "is this really Bandcamp"; `safeFetch` separately answers "is this safe to fetch." But `isUrlHostnameAllowed` checks against `ALLOWED_OUTBOUND_HOSTNAMES` in `middleware.ts`, which is the *general* SSRF allowlist — it also contains ~20 non-Bandcamp hosts (patreon.com, ko-fi.com, buymeacoffee.com, discogs.com, linktr.ee, discord.com, beatport.com, …), plus a `faircamp`-label wildcard rule. A row stored under `platform: 'bandcamp'` in `artist_links` (which per the comment above can be "any http(s) URL" a claimed artist saved) pointing at, say, `patreon.com` would pass this check, and the code would then fetch `patreon.com/music` believing it's scraping Bandcamp — not an SSRF (the target is itself an allowlisted, safe host), but it's a real authenticity gap: it scrapes a third-party site under a false assumption, stores a mislabeled `platform: 'bandcamp'` release-source row if anything parses, and does so without checking that site's `robots.txt` for `/music` (the codebase's own stated principle in CLAUDE.md).

Separately — and in the other direction — this same check likely produces a false negative for **legitimate** Bandcamp Pro custom-domain artists (e.g. `music.sufjan.com`, which `ingestBandcampGrid`'s own tests explicitly account for downstream). `isUrlHostnameAllowed` only admits `bandcamp.com` / `*.bandcamp.com`, so if a custom domain is ever the value stored under `platform: 'bandcamp'` (rather than the canonical `*.bandcamp.com` URL), it gets rejected here before `safeFetch`'s per-hop redirect handling ever runs — meaning those artists could be silently and permanently skipped by cataloging. Worth confirming (needs-manual-verification) whether `artist_links.platform = 'bandcamp'` rows are ever populated with a custom domain directly, or always the canonical subdomain — if the former is possible, this line is the wrong check in both directions and should instead assert the hostname is actually `bandcamp.com`/`*.bandcamp.com` (or otherwise resolve custom domains via the canonical URL before the allowlist gate).

---

brandonlucasgreen (commented):
@brandonlucasgreen Security + regression review of this PR (companion review posted on #358).

**Summary:** This is a well-guarded feature — real SSRF protections reused from #358, a real shared-secret auth check with a constant-time compare, sane input bounds (25 artists/run, hourly caps, cooldown), and careful never-cache-uncertainty handling of bot challenges vs. genuine empty catalogs. Three findings below, inline on the relevant lines.

### Security
- **`api/functions/catalog-artist-background.ts:121`** — plausible. `isUrlHostnameAllowed(storedUrl)` is used as "is this really Bandcamp," but it's the general SSRF allowlist (~20 hosts including Patreon, Ko-fi, Discogs, Linktree, Discord, a `faircamp`-label wildcard, etc.), not a Bandcamp-specific check. A mislabeled `platform: 'bandcamp'` link pointing at another allowlisted host would pass and get scraped as if it were Bandcamp. In the other direction, this same check plausibly rejects legitimate Bandcamp Pro custom domains (e.g. `music.sufjan.com`) before `safeFetch`'s redirect handling ever runs, even though `ingestBandcampGrid`'s own tests account for that case downstream. Needs-manual-verification on whether custom domains ever land directly in `artist_links.platform='bandcamp'`.
- Auth/secret handling (`isAuthorized` in the same file): confirmed sound — refuses when `INTERNAL_TASK_SECRET` is unset (closed by default), constant-time comparison via `timingSafeEqual` with a length check first, no info leak on 401/400/405 (empty bodies). No issue.
- Input validation on the background function's body (`artistIds` filtered to strings, capped at 25; `trigger` defaults to the lower-priority `'searched'`): confirmed sound.
- SQL/injection: confirmed no raw SQL — all Supabase calls in `db.ts` use parameterized `.eq`/`.insert`/`.update`/`.upsert`. No injection risk.

### Regression
- **`api/functions/db.ts:1194`** — confirmed. `requestArtistCatalog()` is awaited inside `persistSearchResults()`, which is itself awaited before the search response is built in `search-sources.ts` (~line 2035). The PR description states "nothing is on a user's critical path," but the HTTP handshake (up to a 3s timeout) is on the critical path for every search that resolves a Bandcamp-linked artist. Worth fixing before merge given the product's explicit ~1.8–3.0s search latency target — see inline comment for a suggested fix (fire without awaiting the network leg, or move the call after the response is sent).
- **`api/functions/db.ts:449-456`** — plausible. `claimArtistForCatalog`'s claim (cooldown check → hourly-cap count → upsert) is read-then-write with no atomicity, so genuinely concurrent triggers for the same artist can both pass the checks and both proceed — undermining the single-claim and hard-hourly-cap guarantees the table exists for. Low real-world likelihood but plausible under burst load on a trending artist, which is the exact scenario the hourly cap was designed for.
- Everything else checked out: bot-challenge-vs-empty-catalog handling is correct and tested, `persistReleases`'s never-null-over-a-value / never-touch-curated-fields merge logic is correct, cross-host href confinement in `release-ingest.ts` is correct and tested, the new migration has RLS enabled with the documented no-policy pattern matching `bandcamp_slug_probes`.

---

PR #360, opened 2026-07-31
Read Bandcamp release pages for dates, formats and prices

brandonlucasgreen:
Step 3 landed the catalog with identity and artwork only: Bandcamp's `/music` grid carries **no dates, formats or prices**, so every release stored so far has `release_date = null` and an empty `release_offers`. The differentiator the whole feature rests on — *"vinyl $25 · CD $12 · digital $10, and roughly this much reaches the artist"* — had a schema and no data. This closes that, and it belongs before the release page rather than after it.

## What it does

- **`parseBandcampReleaseDetail`** reads the release page's `application/ld+json` — the documented schema.org graph Bandcamp publishes for machines — rather than the private `data-tralbum` blob sitting next to it in the markup. Stabler contract, and one we're plainly invited to read.
- **`ingestBandcampDetail`** maps it to a sanity-bounded date, one offer per format, an availability state, and a pre-order becoming `announced`.
- **Variants of one format collapse to the cheapest available.** `release_offers` is unique on `(source, format)` and Bandcamp routinely sells a standard LP and a £60 deluxe box. Quoting the box as "the price of the vinyl" is a wrong number in front of someone deciding whether to buy.
- **Sold-out formats are kept, not dropped.** "Vinyl — sold out" serves a fan better than a page that silently omits the format, which reads as "never pressed".

## The crawl budget

This is the one part of ingest whose cost scales with catalog size rather than artist count, so it's metered rather than swept: newest-first, **20 pages per artist**, **100 per invocation**, ~**1 request/second**, and a **9-minute deadline** inside the 15-minute background function. Offers refresh after **30 days** — prices change and vinyl sells out, but re-reading weekly would triple the crawl for data that rarely moves. When the budget runs out it says so in the log; a silent cap reads as *"we checked and there were no prices"*.

`release_sources.detail_checked_at` (new column) records what's been read, so a bounded run resumes where the last one stopped instead of re-reading the same newest few forever.

## Failure modes, kept apart

A bot challenge on a release page stops the pass and leaves `detail_checked_at` unset so those releases retry; an unreadable page is logged and skipped; **neither is ever written down as "this release has no price"**. `release_catalog_state.releases_detailed` (new column) sits beside `releases_found` because the two fail independently — a grid can parse perfectly while every release page is challenged, and without the second number that looks like a healthy run.

## Write safety

Upsert offers → prune formats no longer offered → stamp `detail_checked_at`, in that order and never delete-first (PR #350's lesson). Pruning only runs when the page offered *something*: zero offers is the normal state of a standalone track page, not a withdrawal. A curated release date takes both the date and the derived status out of ingest's hands.

## Verified against live pages

`npm run ingest:try -- explosionsinthesky --detail=3` — real fetch, real parse, no writes:

```
explosionsinthesky.bandcamp.com/album/end
  date: 2023-09-15 (day) · status: released
  digital   10 USD       available
  vinyl     25 USD       available
  cd        12 USD       available

explosionsinthesky.bandcamp.com/album/how-strange-innocence-anniversary-edition
  date: 2019-08-16 (day) · status: released
  digital   10 USD       available
  vinyl     35 USD       sold_out
```

Robots checked: `/album/` is permitted on both `bandcamp.com` and artist subdomains; only `/search`, `/api/`, `/stream`, `/cart/` and friends are disallowed.

## Testing

26 tests on the ingest path (11 new), 353 across `api/functions`. Full `npm run build` green.

Nothing is user-visible yet — this fills the tables the release page will read.

### DIFF
--- a/api/functions/db.ts
+++ b/api/functions/db.ts
@@ -471,7 +471,7 @@ export async function claimArtistForCatalog(
 /** Record the outcome of a catalog run. Failures increment the backoff counter. */
 export async function recordCatalogOutcome(
   artistId: string,
-  outcome: { releasesFound: number } | { error: string }
+  outcome: { releasesFound: number; releasesDetailed?: number } | { error: string }
 ): Promise<void> {
   const client = getClient();
   if (!client) return;
@@ -482,6 +482,7 @@ export async function recordCatalogOutcome(
       : {
           last_catalogued_at: new Date().toISOString(),
           releases_found: outcome.releasesFound,
+          releases_detailed: outcome.releasesDetailed ?? 0,
           consecutive_failures: 0,
           last_error: null,
         };
@@ -522,6 +523,21 @@ interface ReleaseToPersist {
 }
 
 /**
+ * A release row and its source after a write, which is what the detail pass needs to decide
+ * whether to spend a request on this release's own page.
+ */
+export interface PersistedRelease {
+  releaseId: string;
+  sourceId: string;
+  /** The release page for this source — where the date, formats and prices live. */
+  url: string;
+  /** When that page was last read. Null means never. */
+  detailCheckedAt: string | null;
+  /** Columns on the release row a verified artist has authored. Ingest leaves these alone. */
+  curatedFields: string[];
+}
+
+/**
  * Write a catalog run's releases, merging rather than replacing.
  *
  * Three rules that matter more than they look:
@@ -535,14 +551,16 @@ interface ReleaseToPersist {
  *    artist authored; ingest re-runs forever, so without this every crawl silently reverts
  *    their corrections.
  *
- * Returns the number of releases written or updated.
+ * Returns the releases written or updated, in the order they were given — which for a
+ * Bandcamp grid is newest first, so a budgeted detail pass reading the front of the list
+ * spends its requests on the releases a fan is most likely to be looking at.
  */
 export async function persistReleases(
   artistId: string,
   releases: ReleaseToPersist[]
-): Promise<number> {
+): Promise<PersistedRelease[]> {
   const client = getClient();
-  if (!client || releases.length === 0) return 0;
+  if (!client || releases.length === 0) return [];
 
   try {
     const { data: existingRows, error: readError } = await client
@@ -552,7 +570,7 @@ export async function persistReleases(
 
     if (readError) {
       console.error('[DB] persistReleases read failed:', readError.message);
-      return 0;
+      return [];
     }
 
     type ExistingRow = {
@@ -568,7 +586,7 @@ export async function persistReleases(
     const byIdentity = new Map(existing.map(r => [`${r.release_type}:${r.match_key}`, r]));
     const takenSlugs = new Set(existing.map(r => r.slug));
 
-    let written = 0;
+    const written: PersistedRelease[] = [];
 
     for (const release of releases) {
       const identity = `${release.releaseType}:${release.matchKey}`;
@@ -637,29 +655,145 @@ export async function persistReleases(
         });
       }
 
-      const { error: sourceError } = await client.from('release_sources').upsert(
-        {
-          release_id: releaseId,
-          platform: release.source.platform,
-          url: release.source.url,
-          external_id: release.source.externalId,
-          last_seen_at: new Date().toISOString(),
-        },
-        { onConflict: 'release_id,platform' }
-      );
+      // detail_checked_at is read back rather than written: it belongs to the detail pass, and
+      // a grid re-crawl must not reset it or every crawl would re-fetch every release page.
+      const { data: sourceRow, error: sourceError } = await client
+        .from('release_sources')
+        .upsert(
+          {
+            release_id: releaseId,
+            platform: release.source.platform,
+            url: release.source.url,
+            external_id: release.source.externalId,
+            last_seen_at: new Date().toISOString(),
+          },
+          { onConflict: 'release_id,platform' }
+        )
+        .select('id, url, detail_checked_at')
+        .single();
 
-      if (sourceError) {
-        console.error('[DB] persistReleases source upsert failed:', sourceError.message);
+      if (sourceError || !sourceRow) {
+        console.error('[DB] persistReleases source upsert failed:', sourceError?.message);
         continue;
       }
 
-      written++;
+      const source = sourceRow as { id: string; url: string; detail_checked_at: string | null };
+      written.push({
+        releaseId,
+        sourceId: source.id,
+        url: source.url,
+        detailCheckedAt: source.detail_checked_at,
+        curatedFields: [...curated],
+      });
     }
 
     return written;
   } catch (error) {
     console.error('[DB] persistReleases error:', error);
-    return 0;
+    return [];
+  }
+}
+
+interface DetailToPersist {
+  releaseDate: string | null;
+  datePrecision: string;
+  status: string;
+  offers: Array<{
+    format: string;
+    price: number | null;
+    currency: string | null;
+    availability: string;
+  }>;
+}
+
+/**
+ * Write what a release's own page told us: its date, and what you can buy there.
+ *
+ * The order of operations is the interesting part. Offers are written **before** stale ones
+ * are pruned, and `detail_checked_at` is stamped **last**:
+ *
+ * - Writing before pruning means a failure part-way through leaves the old offers in place
+ *   rather than an empty offer list. This project has already destroyed an artist's links
+ *   once with a delete-before-a-fallible-write (PR #350); an empty release page is the same
+ *   shape of mistake.
+ * - Pruning only happens when the page actually offered something. Zero offers is the normal
+ *   state of a standalone track page, so treating it as "everything was withdrawn" would
+ *   erase real prices.
+ * - Stamping last means an interrupted run is retried next cycle instead of being recorded as
+ *   done.
+ *
+ * Returns false when nothing could be written, so the caller can count it as a failure.
+ */
+export async function persistReleaseDetail(
+  release: PersistedRelease,
+  detail: DetailToPersist
+): Promise<boolean> {
+  const client = getClient();
+  if (!client) return false;
+
+  try {
+    // A date the artist authored outranks anything upstream says, and status is derived from
+    // the date, so a curated date takes both columns out of ingest's hands.
+    if (!release.curatedFields.includes('release_date')) {
+      const patch: Record<string, unknown> = { status: detail.status };
+      if (detail.releaseDate) {
+        patch.release_date = detail.releaseDate;
+        patch.date_precision = detail.datePrecision;
+      }
+
+      const { error } = await client.from('releases').update(patch).eq('id', release.releaseId);
+      if (error) {
+        console.error('[DB] persistReleaseDetail release update failed:', error.message);
+        return false;
+      }
+    }
+
+    if (detail.offers.length > 0) {
+      const capturedAt = new Date().toISOString();
+      const { error: offerError } = await client.from('release_offers').upsert(
+        detail.offers.map(offer => ({
+          release_source_id: release.sourceId,
+          format: offer.format,
+          price: offer.price,
+          currency: offer.currency,
+          availability: offer.availability,
+          captured_at: capturedAt,
+        })),
+        { onConflict: 'release_source_id,format' }
+      );
+
+      if (offerError) {
+        console.error('[DB] persistReleaseDetail offer upsert failed:', offerError.message);
+        return false;
+      }
+
+      const { error: pruneError } = await client
+        .from('release_offers')
+        .delete()
+        .eq('release_source_id', release.sourceId)
+        .not('format', 'in', `(${detail.offers.map(o => o.format).join(',')})`);
+
+      // A failed prune leaves a stale format listed, which is worth logging but not worth
+      // failing the whole detail write over — the prices we did just refresh are still good.
+      if (pruneError) {
+        console.error('[DB] persistReleaseDetail offer prune failed:', pruneError.message);
+      }
+    }
+
+    const { error: stampError } = await client
+      .from('release_sources')
+      .update({ detail_checked_at: new Date().toISOString() })
+      .eq('id', release.sourceId);
+
+    if (stampError) {
+      console.error('[DB] persistReleaseDetail stamp failed:', stampError.message);
+      return false;
+    }
+
+    return true;
+  } catch (error) {
+    console.error('[DB] persistReleaseDetail error:', error);
+    return false;
   }
 }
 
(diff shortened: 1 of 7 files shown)


api/functions/db.ts
```diff
@@ -637,29 +655,145 @@ export async function persistReleases(
...
+  detail: DetailToPersist
+): Promise<boolean> {
+  const client = getClient();
+  if (!client) return false;
+
+  try {
+    // A date the artist authored outranks anything upstream says, and status is derived from
+    // the date, so a curated date takes both columns out of ingest's hands.
+    if (!release.curatedFields.includes('release_date')) {
+      const patch: Record<string, unknown> = { status: detail.status };
```
brandonlucasgreen, line 738: @brandonlucasgreen **Regression — state transition / edge case (plausible, no verdict tag: needs a decision, not a fix I'd stamp confirmed on without seeing it against a live page).**

`patch.status = detail.status` is set unconditionally here, but `patch.release_date` / `patch.date_precision` are only set `if (detail.releaseDate)` (line 739-742). `detail.status` comes from `deriveStatus(date, hasPreorder, now)` in `release-ingest.ts`, and per `release-utils.ts`'s own documented default, an undated release "counts as released."

That's the right default for the *grid* pass, where `releaseDate` is always null. But here it also fires on every **detail refresh** (30-day re-read, `needsDetail` in `catalog-artist-background.ts`). Walk the sequence:

1. First detail read succeeds: page has a future `datePublished` → `release_date` = future, `status` = `announced`. Both columns written and consistent.
2. 30+ days later, refresh re-fetches the same page. `ingestBandcampDetail` returns `ok: true` (no bot challenge, JSON-LD present) but this run's `parseReleaseDate` fails to produce a date this time — e.g. the implausible-date guard trips, or Bandcamp's date string doesn't match the parser this time (`release-utils.ts` has explicit tests for both). `detail.releaseDate` is `null`.
3. `detail.status` is recomputed via `deriveStatus(null, hasPreorder, now)`. If the offers on this same read also didn't get flagged `preorder` (a partial/independent parse failure, not necessarily the same one), the default kicks in → `'released'`.
4. The `if (detail.releaseDate)` guard means `release_date` is **not** touched — the DB keeps the real future date from step 1. But `status` **is** written unconditionally → `'released'`.

Net result: a row with `release_date` in the future and `status = 'released'` — the exact contradiction `deriveStatus`'s own invariant (`date > today ? 'announced' : 'released'`) is supposed to prevent, produced by a partial parse failure on refresh rather than a genuine state change. Any "upcoming releases" view filtering on `status = 'announced'` would silently drop a release that's still legitimately unreleased.

Suggest gating the status write the same way the date write is gated: only overwrite `status` when either `detail.releaseDate` is present, or `hasPreorder` is explicitly true — otherwise leave the existing status alone, since "we didn't get a date this time" isn't evidence the release status changed.

---

scripts/ingest-releases.ts
```diff
@@ -28,12 +34,17 @@ config({ path: resolve(import.meta.dirname ?? '.', '../.env') });
...
+const { ingestBandcampGrid, ingestBandcampDetail, bandcampMusicUrl } =
+  await import('../api/functions/release-ingest.js');
 
 const args = process.argv.slice(2);
 const asJson = args.includes('--json');
 const target = args.find(a => !a.startsWith('--'));
 
+/** How many release pages to read. 0 = none, which is the default: they cost a request each. */
+const detailArg = args.find(a => a === '--detail' || a.startsWith('--detail='));
+const detailCount = !detailArg ? 0 : Number(detailArg.split('=')[1] ?? 3) || 3;
```
brandonlucasgreen, line 46: @brandonlucasgreen Minor — off-by-one/edge-case (dev script only, low impact): `Number(detailArg.split('=')[1] ?? 3) || 3` treats an explicit `--detail=0` the same as "no count given," because `0` is falsy and falls through the `|| 3`. So `npx tsx scripts/ingest-releases.ts <artist> --detail=0` silently reads 3 release pages instead of zero. Not a security issue and it's a local dry-run tool, but worth a `Number.isFinite` check if `--detail=0` should ever mean "none."

---

brandonlucasgreen (commented):
@brandonlucasgreen Reviewed PR #360 (Bandcamp release-detail ingest). Summary below; inline comments have the specifics.

**Security**

- **SSRF (confirmed safe, no finding).** `isFetchableReleaseUrl` in `catalog-artist-background.ts` allows a release URL through if its host matches the page we just landed on (bypassing the static allowlist), which sounds risky in isolation. But `resolveReleaseUrl` in `release-ingest.ts` already constrains every stored `release_sources.url` to the host of the page it was scraped from, at http/https only, at capture time — and `landedUrl` itself only ever comes back from a `safeFetch` call that enforces the allowlist on the initial request and (per its own docs) on every redirect hop. So the same-host branch is re-validating a value that was already scoped safely; it's how Bandcamp Pro custom domains (`music.artist.com`) get covered at all, not a new hole. No finding.
- **Injection (confirmed safe, no finding).** `persistReleaseDetail`'s `.not('format', 'in', (${...}))` in `db.ts` builds a PostgREST filter by string-joining `offer.format` values without quoting. Looked like a candidate for filter injection, but `format` is only ever produced by `mapOfferFormat()`, a closed 7-value enum — never raw scraped text — so there's nothing attacker-controlled reaching that interpolation.
- **Data exposure / RLS (confirmed safe, no finding).** The new `release_sources.detail_checked_at` and `release_catalog_state.releases_detailed` columns land on tables whose RLS posture is already established (public-read for `release_sources`, server-only/no-policy for `release_catalog_state`) and neither new column is sensitive. No new policy needed, and the migration correctly adds none.
- **Auth, secrets, unsafe defaults:** unchanged in this diff (`isAuthorized` untouched); nothing new to flag.

**Regression**

1. **`api/functions/db.ts:738`** (inline comment) — `persistReleaseDetail` unconditionally overwrites `status` from `deriveStatus(null, ...)` even when this run failed to extract a date, while leaving the existing `release_date` untouched. On a 30-day refresh where the date parse fails but the page otherwise reads fine (`ok: true`), this can flip a real future release from `announced` back to `released` while its `release_date` is still in the future — an inconsistent row, and a release that silently vanishes from any "upcoming" filter. Plausible, not confirmed against a live page; recommend gating the status write the same way the date write is already gated.
2. **`scripts/ingest-releases.ts:46`** (inline comment) — `--detail=0` is swallowed by `|| 3` and silently becomes `--detail=3`. Dev-only, cosmetic.

Everything else — the offer aggregation (cheapest-price/most-available merge), the bot-challenge vs. unreadable distinction, the write-before-prune/stamp-last ordering, and the budget/backoff plumbing — matches the PR description and looks correct against the test suite.

---

brandonlucasgreen (commented):

**Security**
- auth/authz: no new endpoint surface; `catalog-artist-background.ts`'s existing shared-secret + production-only gates are untouched.
- SSRF: traced the new detail-page fetch path (`isFetchableReleaseUrl` → `safeFetch`). It reuses the same trust boundary the existing grid ingest already relies on (`resolveReleaseUrl`'s same-host constraint on parsed hrefs), and every hop still goes through `safeFetch`'s `isSafePublicHostname` + DNS-resolution private-IP check regardless of which branch of `isFetchableReleaseUrl` admitted the URL. **needs-manual-verification** — didn't find a bypass, but it's a three-layer check (`isUrlHostnameAllowed`, `isFetchableReleaseUrl`, `safeFetch`) and worth a second pair of eyes given how load-bearing SSRF protection is here.
- injection: the new `.not('format', 'in', (...))` Supabase filter is built from `format` values, but those only ever come from `mapOfferFormat`'s fixed enum return type — never from raw upstream text — so there's no way for page content to reach that filter string. **confirmed** safe.
- JSON-LD parsing (`parseBandcampReleaseDetail`) uses `JSON.parse` only, no eval/Function construction. **confirmed** safe.
- secrets/data exposure: logs consistently use `safeHostname(url)` rather than the full URL, matching the existing convention. No issues found.

**Regression**
- **plausible**, `api/functions/db.ts` `persistReleaseDetail`: when `detail.offers.length === 0`, both the offer upsert *and* the prune are skipped. That's correct for a standalone track page (which never has offers), but the same zero-offers result is also what you'd get if a full album/EP page legitimately loses all its packages (e.g. an artist pulls a release from sale). In that case stale `release_offers` rows (price/availability) survive indefinitely with no automatic way to age out beyond a consumer checking `captured_at`. Might be worth distinguishing "standalone track, no offers expected" from "album page, zero offers found" before deciding whether to prune.
- **plausible**, same function: if the `release_offers` upsert fails a DB constraint (e.g. `currency` not matching `^[A-Z]{3}$`), `persistReleaseDetail` returns `false` *before* stamping `detail_checked_at`. Since `needsDetail()` treats an unstamped source as always due, that specific release will be retried — and fail identically — on every future catalog run, silently occupying one of the run's limited `MAX_DETAIL_FETCHES_PER_RUN` slots each time rather than backing off.
- Checked for API contract drift: `persistReleases` changed its return type from `Promise<number>` to `Promise<PersistedRelease[]>`. Only one caller (`catalog-artist-background.ts`), and it was updated in the same PR to use `.length`. **confirmed** no drift.
- State transitions around `curated_fields` look correct: a curated `release_date` correctly withholds both `release_date` and `status` from the ingest patch (per the code comment), and offer writes are unconditionally allowed since there's no curation concept for `release_offers` yet — consistent with the current schema (no `curated_fields` column on that table).
- Note (not a bug, flagging for awareness): `catalog-artist-background.ts`, `db.ts`, `release-ingest.ts`, and `search-parsers.ts` are all outside `api/tsconfig.json`'s narrow `include`, so a type error here wouldn't fail the build — relying on the new unit tests (which look thorough) rather than the typechecker for these files, per the project's own documented gotcha.

---

api/functions/release-ingest.ts
```diff
@@ -107,6 +112,168 @@ export function ingestBandcampGrid(html: string, pageUrl: string, now: Date = ne
...
+      continue;
+    }
+
+    // A missing price loses to a real one; between two real prices the lower wins.
+    if (offer.price !== null && (existing.price === null || offer.price < existing.price)) {
+      existing.price = offer.price;
+      existing.currency = offer.currency;
+    }
+    if (AVAILABILITY_RANK[availability] > AVAILABILITY_RANK[existing.availability]) {
+      existing.availability = availability;
```
brandonlucasgreen, line 200: **[confirmed] Regression / correctness: price and availability are maximized independently across variants of the same format, and can end up describing two different packages.**

`aggregateOffers` tracks the cheapest price and the "most available" state as two separate running values, not as a single pick tied to one source offer. When a cheaper variant is sold out and a pricier variant of the same format is in stock — a completely normal Bandcamp pattern (standard LP vs. deluxe box) — the result combines the cheap price with the in-stock status from the *other* package.

Failure scenario, traced through the code:
```
packages: [
  { format: 'VinylFormat', price: 25, availability: 'SoldOut' },
  { format: 'VinylFormat', price: 60, availability: 'InStock' },
]
```
`existing.price` ends at `25` (line 196, lower real price wins) while `existing.availability` ends at `'available'` (line 199-200, higher rank wins) — so the aggregated offer reads `{ format: 'vinyl', price: 25, availability: 'available' }`. That's a false claim: the $25 pressing is the one that's sold out; only the $60 one can actually be bought. On a page whose entire premise is "here's what you can buy and for how much," this is the one kind of error the feature can't afford.

The existing test (`collapses variants of one format to the cheapest available`) only covers the case where the cheap variant happens to be the available one, so this doesn't get caught.

Suggest tying availability to whichever single offer supplies the reported price (e.g. prefer the cheapest *available* variant's price+availability pair as a unit; only fall back to a sold-out price when nothing is available), rather than maximizing the two fields independently.

---

api/functions/db.ts
```diff
@@ -637,29 +655,145 @@ export async function persistReleases(
...
+      if (offerError) {
+        console.error('[DB] persistReleaseDetail offer upsert failed:', offerError.message);
+        return false;
+      }
+
+      const { error: pruneError } = await client
+        .from('release_offers')
+        .delete()
+        .eq('release_source_id', release.sourceId)
+        .not('format', 'in', `(${detail.offers.map(o => o.format).join(',')})`);
```
brandonlucasgreen, line 774: **[plausible, low severity] Injection risk: raw string interpolation into a PostgREST filter.**

`.not('format', 'in', \`(${detail.offers.map(o => o.format).join(',')})\`)` builds the filter by string concatenation rather than passing a parameterized value. Not exploitable today — `format` only ever comes from `mapOfferFormat()`, which returns one of a fixed 7-value enum, so no external string reaches this call. Flagging because the pattern itself doesn't enforce that invariant at the call site; if a future change ever fed this from a less-constrained source (an admin override, a wider format mapping), it would become a filter-syntax injection point with no warning. Worth a comment noting the invariant it depends on, or switching to whatever parameterized "not in" form the installed postgrest-js version supports.

---

brandonlucasgreen (commented):

**Security** (auth, input validation, secrets, injection, unsafe defaults, data exposure)
- SSRF: the new detail-fetch path (`isFetchableReleaseUrl` in `catalog-artist-background.ts`) correctly follows the codebase's established pattern — same-host-as-landed-page or allowlist, with `safeFetch` doing the actual per-hop DNS/scheme validation. No new bypass. `confirmed` clean.
- Auth/secrets: no changes to `isAuthorized`, no new secret handling, nothing logged beyond hostnames (`safeHostname`). `confirmed` clean.
- Injection: one low-severity note left inline on `db.ts` (raw string built for a `.not(...,'in',...)` filter) — currently safe due to a closed enum, flagged as `plausible`/hardening-only.
- Data exposure: write ordering (upsert offers → prune → stamp `detail_checked_at` last) correctly avoids the delete-before-fallible-write shape from PR #350. Bot-challenge vs. unreadable-page failure modes are kept apart per the "never cache uncertainty" rule and neither is written down as a negative result. `confirmed` clean.

**Regression** (changed assumptions, edge cases, off-by-one, state transitions, API contract drift)
- `persistReleases` return type changed `number` → `PersistedRelease[]`; verified the only call site (`catalogArtist`) and `scripts/ingest-releases.ts` (read-only, doesn't call it) are consistent.
- `deriveStatus`/`parseReleaseDate` reused from `release-utils.ts` with the correct argument order for the new caller; date sanity-bounding (min year, +3y ahead) and textual-date parsing ("15 Sep 2023 00:00:00 GMT") already handle Bandcamp's format — verified against the new tests.
- One `confirmed` correctness bug left inline: `aggregateOffers` in `release-ingest.ts` can combine the price of one variant with the availability of a different variant of the same format, producing a self-inconsistent "in stock" claim for a price that's actually sold out. Concrete repro in the inline comment.

Nice discipline overall on the write-ordering and failure-mode separation — the two items above are the only things I'd want resolved before merge, and only the first one is release-blocking in my view.

---

PR #361, opened 2026-07-31
Quote the MusicBrainz artist query as a Lucene phrase

brandonlucasgreen:
Reported as: **"viagra boys" returns no results in search.**

## Root cause

MusicBrainz is Lucene-backed, and the artist search sent an **unquoted** field term — `artist:viagra boys`. Lucene reads that as `artist:viagra` OR a bare `boys`, so the loose token drags in whichever popular artist matches it best. Measured against the live MB API:

| Query sent | Top hit |
|---|---|
| `artist:viagra boys` (before) | **The Beach Boys**, score 100 |
| `artist:"viagra boys"` (after) | **Viagra Boys**, score 100, MBID `9df0c0ec…` |

The wrong hit is then *correctly* rejected by the downstream name-similarity guard, so the artist silently loses all enrichment — official site, socials, Discogs, location, and Qobuz, which MusicBrainz is now the only source of. The failure presents as "MusicBrainz has nothing" when MB actually has a score-100 match: a bad **query** masquerading as a missing **record**.

Phase 1 (`limit=5`) had the right artist at index 1 (score 99) but only reads `artists[0]` for enrichment, so it lost too.

## Blast radius

Sampled 20 multi-word names from `data/artists-manifest.json`:

- **4/20 wrong** unquoted — `Ian McDonald`→Michael McDonald, `Bradley Joseph`→Joseph Haydn, `Dinah Jane`→[unknown], `Red One`→[unknown]
- **4/4 fixed** by quoting
- 543 of ~790 catalog artists are multi-word, so this was roughly **14% of the catalog** silently losing MB enrichment.

## The change

Adds `musicBrainzArtistQuery` to `search-utils.ts` (pure, unit-tested, per the "keep new logic out of search-sources.ts" guidance) and uses it at both live call sites. Embedded `"` and `\` are stripped rather than escaped — no artist name needs them, and a stripped character costs a slightly looser match while a broken phrase costs the whole query.

`apps/web/server` is the dev-only shim and cannot import from `api/functions`, so it gets a small mirrored helper with a pointer comment. Without it, `npm run dev` would keep showing the bug and mislead whoever debugs this next. The dead Vercel-era copies in `api/search/sources.ts` and `api/search/musicbrainz.ts` are deliberately untouched.

## Verification

Against the dev API — "viagra boys" goes from `artistName: null` / 0 results to:

```
artistName : Viagra Boys
officialUrl: https://vboysstockholm.com/
discogsUrl : https://www.discogs.com/artist/4861285
socials    : twitter, facebook, instagram, youtube, tiktok
```

UI renders "Found 1 result" with Discogs (marketplace), Official Site, and socials. No console errors.

- `Ian McDonald` and `Bradley Joseph` now resolve correctly; `Radiohead` unchanged (no regression on previously-working queries)
- `npm run build` passes (347 API tests, 22 files)
- Lint baseline unchanged: 77 problems before and after, none in touched files
- Explicit `tsc --strict` on the three `api/` files (which the build does **not** typecheck): same 3 pre-existing errors before and after, shifted one line by the added import

## Not a bug, for the record

The Bandcamp miss was correct. `viagraboys.bandcamp.com` exists and name-matches, but the account is genuinely empty — zero grid items and zero release links on both `/music` and root — so the probe's `rejected_empty` verdict is right. Controls fetched identically: radiohead 15 grid items, sufjanstevens 16. Their catalog isn't on their label's page either, and MB lists no Bandcamp relation.

### DIFF
--- a/api/functions/__tests__/musicbrainz-artist-query.test.ts
+++ b/api/functions/__tests__/musicbrainz-artist-query.test.ts
@@ -1,0 +1,67 @@
+import { describe, it, expect } from 'vitest';
+
+// The bug this guards: MusicBrainz is Lucene-backed, and the artist search used to send
+// an unquoted field term — `artist:viagra boys`. Lucene reads that as `artist:viagra` OR
+// a bare `boys`, so the loose token pulls in whichever popular artist matches it best.
+// Measured against the live API on 2026-07-31:
+//
+//   artist:viagra boys    -> "The Beach Boys" (score 100)  <- wrong artist wins
+//   artist:"viagra boys"  -> "Viagra Boys"    (score 100)
+//
+// The wrong hit is then correctly rejected by the downstream name-similarity guard, so
+// the artist silently loses ALL enrichment (official site, socials, Discogs, location,
+// and Qobuz, which MB is now the only source of) and the search returns nothing.
+//
+// On a 20-artist sample from data/artists-manifest.json, 4 got the wrong top hit
+// unquoted and quoting fixed all 4. 543 of ~790 catalog artists are multi-word.
+
+import { musicBrainzArtistQuery } from '../search-utils';
+
+describe('musicBrainzArtistQuery', () => {
+  it('wraps multi-word names in a Lucene phrase', () => {
+    // The regression itself: without the quotes this is `artist:viagra boys`.
+    expect(musicBrainzArtistQuery('viagra boys')).toBe('artist:"viagra boys"');
+    expect(musicBrainzArtistQuery('Ian McDonald')).toBe('artist:"Ian McDonald"');
+    expect(musicBrainzArtistQuery('the mountain goats')).toBe('artist:"the mountain goats"');
+  });
+
+  it('quotes single-word names too, so there is one code path', () => {
+    expect(musicBrainzArtistQuery('Radiohead')).toBe('artist:"Radiohead"');
+  });
+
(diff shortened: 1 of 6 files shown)


brandonlucasgreen (commented):

**Security**
- auth/authz: n/a, no endpoint surface changed.
- input validation: the new `musicBrainzArtistQuery` strips `"` and `\` before wrapping the query in a Lucene phrase — the two characters that would let a caller break out of the quoted phrase and re-inject a loose/boolean Lucene expression. **confirmed** this closes that class of injection into the MusicBrainz query (checked against `stripping` regex and the "breaks out of phrase" tests in the new test file).
- injection risk: none into our own systems (query goes to MusicBrainz's public API only, via `encodeURIComponent`), and this PR narrows what a user-supplied artist name can do to that query rather than widening it.
- secret handling / unsafe defaults / data exposure: n/a.

**Regression**
- Verified all three call sites (`search-musicbrainz.ts`, `search-sources.ts`, and the dev-only `apps/web/server/search/musicbrainz.ts`) were updated together, so there's no leftover unquoted path. **confirmed**.
- Edge cases (empty-after-stripping input, embedded quotes/backslashes, accents, whitespace collapsing, URL-encoding) are covered by the new test file and look correct by inspection.
- No API contract drift — `musicBrainzArtistQuery` is a pure new export, nothing else changes shape.

No findings. This is a well-scoped, well-tested fix.

---

PR #362, opened 2026-07-31
Add the release page at /a/{artist}/{release}

brandonlucasgreen:
**Stacked on [#360](https://github.com/brandonlucasgreen/unstream/pull/360)** — it reads `release_sources.detail_checked_at`, which that PR's migration adds. Review #360 first; this diff is only the two new files plus the route.

Step 4 of the V1 scope. The first user-visible piece of Releases: a buying guide for one record — what formats exist, what they cost, and roughly how much reaches the artist. Odesli, the distributor landing pages and feature.fm all answer *"where can I stream this"*, which is the opposite of what Unstream is for.

```
Vinyl     $25   ≈$20–$21.25 to artist
CD        $12   ≈$9.60–$10.20 to artist
Digital   $10   ≈$8–$8.50 to artist
Cassette  $14   Sold out
```

## Decisions worth stating

**Pure SSR, and only SSR.** This URL renders here and nowhere else — the SPA never takes it over. One route with two renderers is the UNS-100 bifurcation class. Nothing hydrates; the only script on the page is an inline image-error fallback (the same one `artist-page-static` uses, because hotlinked CDN artwork failing is normal, and an empty bordered box on a buying page reads as broken).

**Route declared before `/a/*`**, since that glob also matches `/a/artist/release` and `artist-page-static` would otherwise answer first — it'd still work, but only by accident, and at the cost of a DB query per release page view. A regex rather than a path glob because what matters is the *number of path segments*, which Netlify's globs can't express. Netlify's docs confirm both that `pattern` is supported in `netlify.toml` and that declarations run top-to-bottom. The service worker's navigation denylist already covers `/^\/a\//`, so no change needed there.

**`noindex, follow`, and deliberately no JSON-LD.** These pages are generated from a crawl, so a wrong one mints a durable URL asserting an artist made a record they didn't — and a catalog of thin auto-generated pages is textbook scaled-content abuse, whose penalty is *site-wide*, putting the artist-page and guide rankings already earned at risk. Structured data on a noindex page does nothing except publish a machine-readable claim we aren't yet confident enough to make. Indexation stays a later, deliberate decision.

**Payouts are ranges, never point estimates.** The registry's figures are strings (`'80-85%'`, `'~70%'`), which makes false precision impossible by construction — keep it that way. Bandcamp Friday is honoured, because on those days the usual range is simply wrong.

**Sources order by payout; artist-paying options lead.** Moot with one source today, but it's the rule that stops Discogs' secondhand marketplace outranking the artist's own store once step 7 lands.

**Honest states, not empty ones.** *"Still gathering formats and prices"* is a different statement from *"there is nothing to buy"* — and it's what **every release currently in the catalog will show**, since grid ingest has no prices. A release that doesn't exist redirects to the artist page rather than falling through to a blank SPA route (there's no catch-all route for a two-segment `/a/` path).

**Prices are claims with an age.** The footer reports the *oldest* capture on the page; saying "checked today" because one source was refreshed would overstate the rest.

## Testing

The formatting — money, payout ranges, precision-aware dates, freshness, source ordering — lives in `api/shared/release-display.ts` rather than inside the edge function. Edge functions can't run under vitest, and the existing workaround has been to copy the function into the test file and keep the copy in sync by hand; code that makes claims about someone's money shouldn't be verified against a copy of itself. 14 tests, including the ones that matter most: a year-only date never renders a fabricated day, an unknown payout renders nothing rather than a guess, and an unknown currency degrades to a code rather than a wrong symbol.

Verified by running the real edge module against canned data and rendering it at desktop and mobile widths in both themes, including the no-offers state. Mobile initially put a full-bleed square of artwork above everything and wrapped the payout onto two lines — both fixed and measured, not eyeballed.

Full `npm run build` green; 367 tests across `api/functions`.

## Still not wired up

Nothing links here yet. Alerts still open the raw platform URL — that's step 5, and it's what makes this page reachable.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -172,8 +172,9 @@ have your laptop writing production data.
 Use the dry run instead:
 
 ```bash
-npm run ingest:try -- sufjanstevens          # table of what would be written
-npm run ingest:try -- sufjanstevens --json   # full row shapes
+npm run ingest:try -- sufjanstevens              # table of what would be written
+npm run ingest:try -- sufjanstevens --json       # full row shapes
+npm run ingest:try -- sufjanstevens --detail=3   # + dates, formats and prices for the newest 3
 ```
 
 It runs the real path — the same SSRF-safe fetcher, the same allowlist check, the same parser
@@ -183,6 +184,22 @@ request per run; don't loop it.
 There is deliberately no `--write` flag. Everything with a decision in it lives upstream of the
 database, and `persistReleases` is covered by unit tests plus a migration validated against a
 real Postgres. To test the write path, point `SUPABASE_URL` at a branch database on purpose.
+
+### Seeing the release page locally
+
+`npm run dev` **cannot** show you `/a/{artist}/{release}` — the Vite dev server doesn't run edge
+functions at all. `netlify dev` does, but it reads the production Supabase, where a release only
+exists once demand-driven cataloging has run for that artist. So the page you most want to look
+at is the one neither of those can render.
+
+```bash
+npm run preview:release -- explosionsinthesky    # then open http://localhost:8788
+```
+
+Fetches the real `/music` grid, lists the discography, and renders any release through the
+**real** `api/edge/release-page.ts` — same template, same payout maths — fetching that release's
+page from Bandcamp on demand. Only the two database reads are stubbed; there is no database
+connection, so nothing can be written. One Bandcamp request per release page you open.
 
 Once ingest is live, `release_catalog_state` is the observability surface:
 `last_attempted_at`, `releases_found`, `last_error`, `consecutive_failures`. A run that suddenly
--- a/api/edge/release-page.ts
+++ b/api/edge/release-page.ts
@@ -23,7 +23,7 @@ import {
   AVAILABILITY_LABELS,
   AVAILABILITY_ORDER,
   FORMAT_LABELS,
-  formatMoney,
+  formatOfferPrice,
   formatReleaseDate,
   payoutEstimate,
   payoutRank,
@@ -64,9 +64,12 @@ const CSS = `
   .source { border: 1px solid var(--border); border-radius: 12px; overflow: hidden; margin-bottom: 12px; }
   .source-head { display: flex; align-items: center; gap: 10px; padding: 12px 16px; background: var(--bg2); }
   .offer { display: flex; align-items: baseline; gap: 12px; padding: 10px 16px; border-top: 1px solid var(--border); font-size: 14px; }
-  .offer-format { width: 88px; flex-shrink: 0; font-weight: 500; text-transform: capitalize; }
-  .offer-price { width: 88px; flex-shrink: 0; }
-  .offer-payout { color: var(--muted); font-size: 13px; }
+  .offer-format { flex: 0 0 88px; font-weight: 500; text-transform: capitalize; }
+  /* Sized to content with a floor, not fixed: "$25" and "Name your price" both live in this
+     column, and a width that suits one wraps the other onto two lines. Rows within a card still
+     line up, because one artist's prices are all the same shape. */
+  .offer-price { flex: 0 0 auto; min-width: 88px; }
+  .offer-payout { flex: 1; color: var(--muted); font-size: 13px; }
   .offer.gone { color: var(--muted); }
   .buy { display: block; padding: 12px 16px; border-top: 1px solid var(--border); text-align: center; font-weight: 600; text-decoration: none; color: var(--accent); }
   .buy:hover { text-decoration: underline; }
@@ -83,8 +86,8 @@ const CSS = `
     .release-art, .release-art-fallback { width: 200px; height: 200px; }
     /* Narrower fixed columns so the payout estimate — the line this page exists for — stays on
        one line instead of wrapping under the price. */
-    .offer-format { width: 64px; }
-    .offer-price { width: 72px; }
+    .offer-format { flex-basis: 64px; }
+    .offer-price { min-width: 72px; }
   }
 `;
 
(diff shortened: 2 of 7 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Security + regression pass on this PR (reviewed together with #360, which it's stacked on).

**Security** (auth, input validation, secrets, injection, unsafe defaults, data exposure)
- XSS: every DB-derived value that lands in HTML text or an attribute goes through `escapeHtml` — checked `title`, `artistName`, `statusText`, `typeLabel`, `dateText`, `artworkUrl` (both `src` and `og:image`), and `source.url` in the `href`. `confirmed` clean.
- `href="${escapeHtml(source.url)}"` doesn't itself block a `javascript:` scheme, but the URL can only reach `release_sources.url` via #360's `resolveReleaseUrl`, which requires `protocol === 'https:'/'http:'` and same-host-as-landed-page — so a `javascript:` URI can never get stored there in the first place. `confirmed` clean, verified the upstream constraint rather than assuming it.
- Query params (`artistSlug`, `releaseSlug`) go through Supabase's query builder (`.eq(...)`), not string-built SQL — no injection surface. `confirmed` clean.
- `is_hidden` is filtered in the query itself rather than post-fetch, consistent with the stated intent that a suppressed release must be indistinguishable from one that never existed. `confirmed` clean.
- Route precedence (`pattern` entry placed before the existing `path = "/a/*"` → `artist-page-static` entry in `netlify.toml`) — verified against the base branch's file; ordering is correct, so `artist-page-static` won't shadow this route. `confirmed` clean.
- No new secret handling beyond the existing `SUPABASE_SERVICE_KEY` pattern already used by sibling edge functions.

**Regression** (changed assumptions, edge cases, state transitions, contract drift)
- Empty/null states: no sources → "haven't found anywhere to buy this yet"; a source with zero offers → distinguishes "still gathering" vs. "no formats listed" based on `detail_checked_at`; missing artwork → SVG fallback via `onerror`. All exercised in the PR's own testing notes and consistent with the code.
- `formatReleaseDate`/`relativeDays`/`payoutEstimate` edge cases (year-only dates, unknown currency, unknown payout, future-dated capture) are unit-tested and I traced the logic — no gaps found beyond what's already tested.
- No functional bug found in this diff. The one thing worth flagging isn't in this PR's code: since this page directly renders whatever #360's `aggregateOffers` computes, the price/availability-mismatch bug I found there (see #360 review) will surface here as an incorrect "in stock" badge next to a price that isn't actually available — so I'd land the #360 fix before merging this one, not because this PR did anything wrong.

No release-blocking findings in this diff itself.

---

brandonlucasgreen (commented):

## Security

**1. [needs-manual-verification] `Cache-Tag` header built from decoded URL path segments without validation.**
`api/edge/release-page.ts` sets `'Cache-Tag': release-${artistSlug}-${releaseSlug}` where both values come from `decodeURIComponent(segments[1])` / `decodeURIComponent(segments[2])` on the raw request path — i.e. attacker-controlled input, decoded before use, and never passed through `escapeHtml` (that function is only applied to values interpolated into the HTML body). If a crafted path could decode to characters invalid in an HTTP header value (e.g. CR/LF), the question is only what happens next: Deno's `Headers`/`Response` implementation generally rejects invalid header values, which here would be caught by the function's own top-level `try/catch` and fall through to `context.next()` (SPA) rather than actually injecting anything — so this is likely inert, not exploitable. I can't fully rule it out from the diff alone. Note this exact pattern (`Cache-Tag: artist-${slug}` unescaped) already exists in `artist-page-static.ts`, so this isn't a new class of risk introduced here, just propagated to a second file — worth a shared helper if it's ever tightened.

**2. [plausible] `release_type.toUpperCase()` called on a DB value with no runtime null check, only a TS type assumption.**
`const typeLabel = release.release_type === 'other' ? '' : release.release_type.toUpperCase();` — `release_type` is typed as `string` at compile time but nothing at runtime verifies the Supabase row actually has a non-null value (the interface annotation doesn't enforce anything at the boundary). If a `releases` row is ever inserted with a null/missing `release_type` (bad ingest run, manual DB edit, future schema drift), this throws, gets swallowed by the outer `catch { return context.next(); }`, and the request silently falls through to the SPA's blank fallback (there's no catch-all route for a two-segment `/a/` path, per the PR's own description) instead of the graceful "still gathering formats" state the rest of the page is designed around. Low severity since it degrades to a blank page rather than a crash, but it's the one field on the row read without a fallback (`title`, dates, and offers all have `?? '...'`/optional handling). Worth confirming the `release_type` column is `NOT NULL` with no legacy nullable path (migration is in the stacked #360, not in this diff, so I couldn't verify from what's here).

## Regression

**1. [confirmed] Route precedence and the two-segment regex are correct.** `pattern = "^/a/[^/]+/[^/]+/?$"` is declared before the `path = "/a/*"` → `artist-page-static` entry in `netlify.toml`, matching the PR's stated requirement. Verified by reading the full file, not just the diff context.

**2. [plausible] No handling for a percent-encoded slash inside a slug segment.** `[^/]+` matches a literal `%2F` sequence (it's not a `/` until decoded), so a path like `/a/artist/%2Fsomething` reaches the handler and `decodeURIComponent` turns the second segment into a string containing `/`. It's then used only as an opaque `.eq('slug', ...)` value (parameterized, no injection) and in `encodeURIComponent()` calls for outbound links, so the practical effect is just "no matching row → 302 to the artist page" — not a security issue, but worth a conscious "yes, this is fine" rather than an implicit one, since it's the kind of input the regex was specifically trying to constrain.

Everything else — the SSRF-safe fetch reuse in `preview-release-page.ts`, the `AbortController` timeouts on both Supabase reads, the `is_hidden` filter applied in the query rather than post-filtered, the escaping of all user/DB-derived strings interpolated into the HTML body, and the payout/date formatting edge cases (year-only dates, unknown currency, zero/null price) — looks correct and is backed by the new `release-display.test.ts` coverage.

---

PR #364, opened 2026-07-31
Treat a MusicBrainz identity match as verification

brandonlucasgreen:
Searching **nine inch nails** returned a card showing the band's real official site, Discogs, Qobuz and five socials — under a yellow warning:

> **Unverified match:** We couldn't confirm this is the same "Nine Inch Nails" as the other results.

Every link in that card came from the MusicBrainz record that supposedly couldn't confirm it. And for **viagra boys** the card was the only result on the page, so there were no "other results" to be the same as.

Two separate defects, both in how confidence is decided and described.

## 1. A MusicBrainz identity match didn't count as verification

`matchConfidence: 'verified'` only ever meant "release titles overlap across platforms" (or a curated platform — mirlo/faircamp/jamcoop — listed them). So an artist none of our parseable platforms carried stayed unverified no matter how well MusicBrainz knew them. Since MB is now our only source of Qobuz links and socials, that's a lot of cards showing real, correct links under a warning.

But MB naming the artist and handing over their official site, Discogs, Qobuz and socials answers the question the badge is actually asking — more directly than release-title overlap does.

`musicBrainzConfirmsIdentity()` now gates a `musicBrainzConfirmed` flag, set wherever MB identity data lands on a result: the Phase 2.2 fallback card, Phase 2.1 enrichment, and both client-side equivalents. It vouches for a result the same way a curated platform does.

A **name-only** MB hit deliberately does not qualify — that card is nothing but search links, and calling it verified would overclaim. "Rich" means MB gave us at least one real destination.

## 2. The warning claimed a comparison that often never happened

Two unrelated situations both produced `'unverified'`:

| Situation | Is it a warning? |
|---|---|
| Bandcamp page split off because its releases contradicted a verified sibling | Yes — may be a different artist with the same name |
| No release data to compare at all | No — absence of evidence, not evidence of a mismatch |

The copy was written for the first and shown for both. `unverifiedReason` now records which, and the banner renders only for `'conflicting-releases'`. Absence of evidence gets the quiet header chip instead — whose tooltip made the same false claim and no longer does. Results restored from the DB carry no reason (only the confidence itself persists) and take the quiet path, which is the safe default.

An MB match does **not** override a conflicting-releases split, in either the server or client path: that split came from comparing actual release titles, and a name-level match is weaker evidence than the thing it would erase.

## Verification

Against the dev API on this branch (rebased onto #361, which is what makes these two queries resolve at all):

- `?q=viagra+boys` → 1 result, no banner, no chip
- `?q=nine+inch+nails` → 1 result, no banner, no chip

492 web unit tests + 347 API tests pass. Lint and an explicit `tsc` over the touched `api/` files are byte-identical to a stashed baseline (77 pre-existing lint problems, 3 pre-existing type errors — `api/` isn't in the typecheck include, so this was checked by hand).

New coverage: `apps/web/tests/unit/UnverifiedBanner.test.tsx`, plus cases in `disambiguation.test.ts` and `sources-service.test.ts` for the confidence rule, the reason split, and the conflicting-releases guard.

## Worth a second look

This widens what the public "Verified" badge means, and `db.ts` persists `match_confidence` — so newly-verified MB artists propagate to stored artists, the Apple app, and the artist pages. That's the intended effect, but it is the part with reach beyond search.

## Relationship to #361

Independent. #361 fixed the MusicBrainz Lucene quoting so these queries resolve; this fixes how the resulting card is labelled. The mislabelling predates #361 and affects any artist our platforms miss.

### DIFF
--- a/api/functions/search-sources.ts
+++ b/api/functions/search-sources.ts
@@ -1353,6 +1353,7 @@ async function attachNameOnlyPlatforms(
         type: 'artist',
         platforms: toAttach.map(p => ({ sourceId: p.sourceId as SourceId, url: p.url })),
         matchConfidence: 'unverified',
+        unverifiedReason: 'no-release-data',
       };
 
       // If we have Faircamp, fetch releases to seed the result
@@ -1362,6 +1363,7 @@ async function attachNameOnlyPlatforms(
           const fcPlatform = newResult.platforms.find(p => p.sourceId === 'faircamp');
           if (fcPlatform) fcPlatform.allReleaseTitles = titles;
           newResult.matchConfidence = 'verified';
+          newResult.unverifiedReason = undefined;
         }
       }
 
@@ -1435,6 +1437,7 @@ async function attachNameOnlyPlatforms(
           type: 'artist',
           platforms: toAttach.map(p => ({ sourceId: p.sourceId as SourceId, url: p.url })),
           matchConfidence: 'unverified',
+          unverifiedReason: 'no-release-data',
         };
         merged.push(newResult);
         console.log(`[Deferred Attach] Ambiguous "${displayName}" with no Faircamp data — created separate result with ${toAttach.map(p => p.sourceId).join(', ')}`);
@@ -1443,6 +1446,26 @@ async function attachNameOnlyPlatforms(
       }
     }
   }
+}
+
+// ---------------------------------------------------------------------------
+/**
+ * Did MusicBrainz give us a rich answer for this artist — a confirmed name plus at
+ * least one real destination?
+ *
+ * A name-only hit isn't enough: the card it produces is nothing but search links,
+ * and calling that verified would overclaim. But once MB hands over an official
+ * site, a Discogs page, a Qobuz page or a social profile, it has told us who the
+ * artist is and where to find them, which is the question "verified" answers.
+ */
+function musicBrainzConfirmsIdentity(mbData: EnrichedMusicBrainzResult): boolean {
+  if (!mbData.artistName) return false;
+  return Boolean(
+    mbData.officialUrl ||
+    mbData.discogsUrl ||
+    mbData.qobuzUrl ||
+    (mbData.socialLinks && mbData.socialLinks.length > 0)
+  );
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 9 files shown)


brandonlucasgreen (commented):

## Security

**1. [plausible] Unsafe default — a crowd-editable data source now single-handedly earns the public "Verified" badge, and that verdict is persisted and propagated.**
`musicBrainzConfirmsIdentity()` treats *any* MusicBrainz relation (official site, Discogs, Qobuz, or a single social link) as sufficient to flip `matchConfidence` to `'verified'` — no release-title cross-check required. MusicBrainz is a public, community-editable database. A vandalized or simply incorrect relation on an artist's MB entry (e.g. a wrong "official site" URL attached to a popular artist) would now surface as a "Verified" badge, and per `db.ts` that `match_confidence` is persisted — so it propagates to stored artists, the Apple app, and static artist pages, not just the one search response. The PR description already flags this ("this widens what the public 'Verified' badge means... it is the part with reach beyond search"), so this isn't a surprise to you — flagging it here so it's a deliberate accept rather than an assumed one. Worth confirming MB's practical vandalism turnaround is fast enough, and/or considering whether a badge this consequential should require corroboration (e.g. two independent MB relation types) rather than any single one.

## Regression

**1. [needs-manual-verification] `mergeByReleaseOverlap` (search-utils.ts, untouched by this diff) doesn't propagate the new `matchConfidence` / `unverifiedReason` / `musicBrainzConfirmed` fields when folding a duplicate result into an existing one.**
When two same-name results with no shared platform and overlapping (or absent) release titles get merged (`search-utils.ts` around line 994-1003), only `platforms` and `imageUrl` are copied from the discarded result onto the surviving one — confidence-related fields are not. Today this appears safe only because of pipeline *ordering*: MB-enriched/confirmed results are aggregated in Phase 2.1/2.2 and already occupy the "existing" slot in the merge map by the time `attachNameOnlyPlatforms` (Phase 4) pushes new, less-confident duplicates later in the array, so the more-confident record always wins the map slot. That's an implicit invariant of the current phase order, not something the merge function enforces — a future reordering (or a new phase inserted before Phase 4) could silently downgrade an MB-verified card back to unverified, or drop `musicBrainzConfirmed`, with no test catching it. `disambiguation.test.ts`'s `mergeByReleaseOverlap` suite doesn't currently cover this interaction with the new fields. Worth either a test locking in the current ordering assumption, or having the merge explicitly prefer the higher-confidence side regardless of array order.

No other issues found — the reason-split logic (`conflicting-releases` vs `no-release-data`), the "don't override a conflicting-releases split" guard in `mergeWithMusicBrainzData`, and the claimed-profile protection all look correct and are covered by the new tests.

---

PR #366, opened 2026-08-01
Show an artist's releases on their page

brandonlucasgreen:
Release pages have been reachable only by typing the URL — nothing in the site linked to one. The artist page is their natural parent, so the section goes there.

```
RELEASES
  [art]  fruit is year 3 (honeycrush)              COMING
         Single · 5 September 2026
  [art]  infinite normal
         Album · 23 October 2025 · Name your price
  [art]  RUINED CASTLE
         Album · 1 June 2024 · from $7
  [art]  SOLO PIANO
         Album
  and 5 more
```

Placed **below Follow**: Support directly and Follow together form the links region whose order the artist controls via link dividers, and releases are a different kind of thing that shouldn't interrupt that ordering.

## Three calls worth recording

**Shown for every artist with a catalogue, not only verified ones.** The releases derive from the artist's stored Bandcamp link — which the page is *already* displaying as a place to buy. Listing what's behind that link asserts nothing the page doesn't already assert. Gating on verification would hide the feature from nearly every artist, since almost none are claimed.

**Nothing renders when there are no releases.** Not an empty box. An empty "Releases" heading reads as broken; its absence reads as "nothing here yet" — which is the truth for any artist nobody has saved or searched, because cataloguing is demand-driven.

**Sold-out formats don't set the price.** "from $8" pointing at a record nobody can buy is the one genuinely misleading number this list could show — a fan clicks expecting $8 vinyl and finds it gone. A release whose only offer is sold out summarises as nothing at all.

## Smaller things

- `count: 'exact'` rides along with the limit in the same round trip, so "and N more" is a real number rather than a guess. One extra query on the page total.
- Ordering matches `idx_releases_artist_chrono`. Many releases still have no date — grid ingest gets identity and artwork but no dates — and without the `created_at` tiebreaker those undated rows would shuffle between requests.
- Capped at 24. Most catalogues fit (16 for Sufjan Stevens, 13 for Explosions in the Sky); the cap is for the artists who don't, where an unbounded list would bury the platform links the page exists to show.
- An undated release shows no date rather than a fabricated one — `date_precision` earning its place again.
- Capitalising the release type happens in code, not via `text-transform: capitalize`, which applies to the whole line and turned "from $7" into "From $7" and "Name your price" into "Name Your Price".

## Testing

The price summary moved to `api/shared/release-display.ts` with tests rather than living inside the edge function — same reason as the rest of that module: code that makes claims about someone's money shouldn't be unverifiable. 21 tests there now, 365 across `api/functions`, full `npm run build` green.

Verified by rendering both templates locally against the real Kid Lightbulbs profile: claimed and unclaimed, section ordering (Support directly → Follow → Releases → claim CTA), the "Coming" badge on a future-dated release, an undated release, and the empty-catalogue case rendering no heading at all.

## What this doesn't do

Search results are unchanged — deliberately. Search is the hot path, and a release query per result would add database reads for data that's absent for nearly every artist today. Worth revisiting once the catalogue has real coverage.

### DIFF
--- a/api/edge/artist-page-static.ts
+++ b/api/edge/artist-page-static.ts
@@ -3,6 +3,27 @@ import { createClient } from "https://esm.sh/@supabase/supabase-js@2";
 import { PLATFORMS } from "../shared/platform-registry.ts";
 import { isBandcampFriday } from "../shared/bandcamp-friday.ts";
 import { mainLinkDividerIndexes } from "../shared/link-dividers.ts";
+import { cheapestOfferSummary, formatReleaseDate } from "../shared/release-display.ts";
+
+/**
+ * How many releases to list before summarising the rest.
+ *
+ * Most catalogues fit: 16 for Sufjan Stevens, 13 for Explosions in the Sky. The cap exists for
+ * the artists who don't — one live Mirlo artist has 33 releases including *Inaction I–V* — where
+ * an unbounded list would bury the platform links this page exists to show.
+ */
+const MAX_RELEASES_SHOWN = 24;
+
+interface ReleaseRow {
+  slug: string;
+  title: string;
+  release_type: string;
+  release_date: string | null;
+  date_precision: string | null;
+  status: string;
+  artwork_url: string | null;
+  release_sources: { release_offers: { price: number | null; currency: string | null; availability: string }[] | null }[] | null;
+}
 
 // Allowed embed domains for featured releases (must match api/functions/artist-profile.ts)
 const ALLOWED_EMBED_DOMAINS = [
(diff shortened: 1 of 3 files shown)


api/edge/artist-page-static.ts
```diff
@@ -165,6 +229,29 @@ export default async function handler(request: Request, context: Context) {
...
+          'slug, title, release_type, release_date, date_precision, status, artwork_url,' +
+          ' release_sources ( release_offers ( price, currency, availability ) )',
+          { count: 'exact' }
+        )
+        .eq('artist_id', artist.id)
+        .eq('is_hidden', false)
+        .order('release_date', { ascending: false, nullsFirst: false })
+        .order('created_at', { ascending: false })
+        .limit(MAX_RELEASES_SHOWN)
+        .abortSignal(controller.signal);
```
brandonlucasgreen, line 251: **Regression — changed assumption, needs-manual-verification.**

The artist/profile/links/releases fetches are sequential `await`s sharing one 5-second `AbortController` timeout (set for the original 3 queries). This adds a 4th round trip, and it's the heaviest one — a two-level nested join (`release_sources → release_offers`) plus `count: 'exact'`, which does a full count scan on top of the row fetch.

If this pushes some requests past the 5s budget, the *whole* SSR response falls back to `context.next()` (client SPA) — not just the releases section — for artists who previously rendered fine within budget. That's a regression in SSR/SEO coverage for the artist page generally, not just a missing Releases block. Worth checking real p95/p99 latency for this function after deploy (Netlify function logs), since the failure mode is silent (falls through to SPA, no error surfaced).

Label: **plausible**, needs-manual-verification against production latency data.

---

api/edge/artist-page-static.ts
```diff
@@ -40,6 +61,47 @@ const SOCIAL_ICONS: Record<string, string> = {
...
+  // line and would turn "from $7" into "From $7" and "Name your price" into "Name Your Price".
+  const type = release.release_type === 'other'
+    ? ''
+    : release.release_type.charAt(0).toUpperCase() + release.release_type.slice(1);
+
+  const offers = (release.release_sources || []).flatMap(s => s.release_offers || []);
+  const meta = [type, date, cheapestOfferSummary(offers)].filter(Boolean).join(' · ');
+
+  const art = release.artwork_url
+    ? `<img src="${escapeHtml(release.artwork_url)}" alt="" loading="lazy" referrerpolicy="no-referrer" style="width:48px;height:48px;border-radius:6px;object-fit:cover;flex-shrink:0;background:var(--bg2)">`
```
brandonlucasgreen, line 86: **Security — data exposure / injection risk, needs-manual-verification (low severity).**

`release.artwork_url` is HTML-escaped before going into `src="..."`, but there's no scheme allowlist — a `javascript:` or `data:` URL would pass through as an escaped-but-otherwise-untouched attribute value. Modern browsers don't execute script via ``'&lt;img src="javascript:..."&gt;'``, so exploitability is very low, and this mirrors the identical pre-existing pattern in `api/edge/release-page.ts:322-323` (also escape-only, no scheme check) — so this isn't new risk introduced by this PR, just the same accepted pattern applied to a second surface.

Flagging for awareness rather than as a blocking issue, since `artwork_url` provenance is a scrape from Bandcamp's own CDN via the probe, not raw user input. If you want to close this off for both surfaces at once, an `isUrlHostnameAllowed`-style check (or at minimum enforcing `https:`) on `artwork_url` at ingest time (`api/functions/db.ts`) would cover both render sites in one place.

Label: **plausible**, low severity, pre-existing pattern.

---

api/edge/artist-page-static.ts
```diff
@@ -40,6 +61,47 @@ const SOCIAL_ICONS: Record<string, string> = {
...
+function renderReleaseRow(release: ReleaseRow, artistSlug: string): string {
+  const href = `/a/${encodeURIComponent(artistSlug)}/${encodeURIComponent(release.slug)}`;
+  const title = escapeHtml(release.title);
+  const date = formatReleaseDate(release.release_date, release.date_precision);
+
+  // Capitalized here rather than with `text-transform: capitalize`, which applies to the whole
+  // line and would turn "from $7" into "From $7" and "Name your price" into "Name Your Price".
+  const type = release.release_type === 'other'
+    ? ''
+    : release.release_type.charAt(0).toUpperCase() + release.release_type.slice(1);
```
brandonlucasgreen, line 80: **Regression — edge case, confirmed (cosmetic, non-blocking).**

`release_type.charAt(0).toUpperCase() + release_type.slice(1)` turns `'ep'` into `"Ep"` rather than `"EP"`. Small display inconsistency for EP releases specifically (album/single/compilation/live/remix all capitalize correctly under this scheme). Not a functional bug, just worth a one-line special case if you want it clean:

```ts
const type = release.release_type === 'other' ? ''
  : release.release_type === 'ep' ? 'EP'
  : release.release_type.charAt(0).toUpperCase() + release.release_type.slice(1);
```

Label: **confirmed**, cosmetic only.

---

brandonlucasgreen (commented):
@brandonlucasgreen Reviewed this against the security + regression checklist (auth/authz, input validation, secrets, injection, unsafe defaults, data exposure; changed assumptions, edge cases, off-by-one, state transitions, API contract drift). No blocking issues.

**Summary**

- **Auth/authorization** — no change; releases are gated the same way the page already gates platform links (public, `is_hidden`-filtered), consistent with the PR's stated design (shown for unverified artists too, since it derives from an already-displayed Bandcamp link). No new privilege surface.
- **Input validation / injection** — all Supabase reads go through the query builder (no raw SQL, no string-built filters), and `artist_id` comes from a prior trusted lookup. No injection risk.
- **Secrets** — untouched.
- **Unsafe defaults** — `MAX_RELEASES_SHOWN = 24`, `count: 'exact'` fallback to `releases.length` — reasonable.
- **Data exposure** — checked the `releases` schema for a status/visibility field this query might be missing: `is_hidden` is filtered correctly. `needs_review` exists in the schema but isn't written or filtered anywhere yet (not by this PR, not by ingest) — dead column, not a live issue. Left one low-severity, non-blocking note on `artwork_url` escaping (inline).
- **XSS** — verified every DB-sourced string that reaches the HTML template (`title`, `type`, `date`, price summary, `artwork_url`) goes through `escapeHtml`, either directly or via the `meta` join. `href` uses `encodeURIComponent` on both path segments. Looks solid; the one inline note is about scheme validation, not escaping.
- **Regression** — two inline notes: a latency/timeout-budget concern from adding a 4th sequential query to a shared 5s abort window (plausible, worth checking real latency post-deploy), and a cosmetic `"Ep"` vs `"EP"` capitalization edge case (confirmed, non-blocking).

Nice attention to the "never cache uncertainty" / honest-data instincts here (sold-out exclusion from price summary, no fabricated dates, no heading when the catalogue is empty) — matches the codebase's existing conventions well.

---

brandonlucasgreen (commented):
@brandonlucasgreen

Reviewed the diff (`api/edge/artist-page-static.ts`, `api/shared/release-display.ts`, `api/functions/__tests__/release-display.test.ts`), the `releases` schema/RLS (migration `20260731120000_releases.sql`), and how the artist page is reached from the rest of the app. CI is green.

## Security

- **XSS (injection)** — No issues found (confirmed). `release.title` and the joined meta line go through `escapeHtml`. `release_type` is DB-`CHECK`-constrained to a fixed enum, so it can't carry markup. `artwork_url` is `escapeHtml`'d into the `img src` attribute, same treatment as the existing `imageUrl` a few lines up — consistent with the file's existing pattern. Release row `href`s are built with `encodeURIComponent` on both `artistSlug` and `release.slug` before landing inside a double-quoted attribute, which neutralizes `"`, `<`, `>`, `&`.
- **SSRF** — No issues found (confirmed). No new outbound `fetch` is introduced; `artwork_url` is only embedded for the *browser* to load (with `referrerpolicy="no-referrer"`), not fetched server-side, so `isUrlHostnameAllowed()` doesn't apply here.
- **Auth/authorization** — No issues found (confirmed). Page stays public-read; releases are gated by the existing `is_hidden = false` filter, matching the sibling `release-page.ts` query and the migration's "public read, service write" RLS policy. No new tables/columns, so no new migration was needed here.
- **Input validation** — No issues found (confirmed), per the XSS point above.
- **Secret handling** — N/A, not touched.
- **Data exposure** — No issues found (confirmed). Only `price`/`currency`/`availability` are selected from `release_offers`; nothing beyond what's rendered. One thing worth a manual look (needs-manual-verification): the query doesn't filter on `needs_review`, so a release flagged for review is still listed — but that matches `release-page.ts`'s existing behavior, not a gap this PR introduces.

## Regression

- **API contract drift / two renderers for one URL (confirmed, high-value finding)** — This is genuinely new SSR-only content. The comment at line ~296 confirms `artist-page-static` output is "pure SSR" with no hydration script. But the primary way users reach an artist page is client-side `<Link to={`/a/${slug}`}>` navigation (`ResultCardHeader.tsx`, `ArtistDirectoryPage.tsx`, `DashboardPage.tsx`, `PublicSavedArtistsPage.tsx`, `ClaimDoneStep.tsx`, …), which React Router intercepts and routes to `ArtistPage.tsx` → `GET /api/artist-page` (`api/functions/artist-page.ts`) → `RichArtistProfile`/`UnclaimedQuietCard`. None of those three were touched by this PR. So the new Releases section only appears on a hard reload, a deep link, or a crawler hit — the normal in-app click-through from search results won't show it at all. This is exactly the "one route, one renderer" bifurcation the project's own retro (UNS-100) warns about; this PR adds a feature to only one side of an already-split route rather than closing the split.
- **Error handling coupling (plausible)** — The releases query was added inside the *same* `try`/`catch` and shares the single 5s `abortSignal` timeout with the artist/profile/links queries, and all four run sequentially. A slow or erroring releases query (two joins + `count: 'exact'`) can now cause the *entire* SSR page to fall through to `context.next()`, not just the releases section — coupling an optional supplementary feature to the core page's SSR/SEO eligibility.
- **Failure vs. empty conflation (plausible)** — Neither the new releases query nor the pre-existing artist/profile/links queries check the Supabase `error` field, only `data`. A genuine query failure is indistinguishable from "this artist has zero releases" and silently renders as no Releases section with no Sentry signal — the same "silent empty result on failure" pattern the project explicitly calls out elsewhere, applied here to a case that predates this PR but that this PR extends to a new query.
- **Off-by-one / pagination** — No issues found. `releaseCount > releases.length` correctly avoids an "and 0 more" line exactly at the 24-item cap.
- **Null/missing fields** — No issues found. `release_sources`/`release_offers` are null-guarded (`|| []`), undated releases render without a date, missing artwork falls back to a placeholder, and sold-out-only offers correctly summarize to nothing (`cheapestOfferSummary` — well covered by the new tests).

---

PR #368, opened 2026-08-01
Stop silently discarding rejected featured-release embeds

brandonlucasgreen:
## Summary
- A pasted featured-release embed that failed the domain allowlist was saved as `null` while the endpoint still returned `success: true`, so the artist edit page showed "Changes saved!" for an embed that had actually been discarded — only visible on reload.
- `sanitizeEmbed()` now returns a 400 with a clear error instead of silently saving `null` when a non-empty embed is rejected.
- `sanitizeEmbed()` also accepts a domain that matches one of the artist's own already-linked platforms (via a new `getOwnedLinkHostnames()` helper), so self-hosted embeds — like Faircamp, which has no single fixed domain — work once that platform is linked to the profile. The fixed `ALLOWED_EMBED_DOMAINS` allowlist for major platforms (Bandcamp, YouTube, Spotify, SoundCloud, Apple Music, Vimeo, Tidal) is unchanged.

## Test plan
- [x] Added `api/functions/__tests__/artist-profile-embed.test.ts` (8 tests) covering: trusted-platform acceptance, owned-hostname acceptance, https-only enforcement on owned hostnames, rejection of unrelated domains, and empty/malformed input.
- [x] `npm run test:api` — 385/385 passing
- [x] `npm run test:unit` — 492/492 passing
- [x] `npm run lint` — no new failures (pre-existing failures confirmed unchanged via `git stash` baseline)

### DIFF
--- a/api/functions/__tests__/artist-profile-embed.test.ts
+++ b/api/functions/__tests__/artist-profile-embed.test.ts
@@ -1,0 +1,50 @@
+import { describe, it, expect } from 'vitest';
+import { sanitizeEmbed } from '../artist-profile';
+
+// A rejected embed used to be saved as null and reported as a success, so a
+// pasted Faircamp/unsupported embed silently vanished on reload. These pin
+// down: trusted platforms still pass, an artist's own linked domain (e.g. a
+// self-hosted Faircamp site) is now accepted, and everything else is still
+// rejected rather than silently accepted.
+
+describe('sanitizeEmbed', () => {
+  it('accepts a Bandcamp embed', () => {
+    const html = '<iframe style="border: 0; width: 350px; height: 470px;" src="https://bandcamp.com/EmbeddedPlayer/album=123/size=large/" seamless></iframe>';
+    expect(sanitizeEmbed(html)).toContain('src="https://bandcamp.com/EmbeddedPlayer/album=123/size=large/"');
+  });
+
+  it('accepts a Spotify embed', () => {
+    const html = '<iframe src="https://open.spotify.com/embed/track/abc123" width="100%" height="152"></iframe>';
+    expect(sanitizeEmbed(html)).toContain('open.spotify.com');
+  });
+
+  it('rejects a self-hosted domain with no owned-link match', () => {
+    const html = '<iframe src="https://cult.example-artist.com/embed/release/"></iframe>';
+    expect(sanitizeEmbed(html)).toBeNull();
+  });
+
+  it('accepts a self-hosted domain that matches one of the artist\'s own links', () => {
+    const html = '<iframe src="https://cult.example-artist.com/embed/release/"></iframe>';
+    expect(sanitizeEmbed(html, ['cult.example-artist.com'])).toContain('cult.example-artist.com');
+  });
+
+  it('does not let an owned hostname bypass the https requirement', () => {
+    const html = '<iframe src="http://cult.example-artist.com/embed/release/"></iframe>';
+    expect(sanitizeEmbed(html, ['cult.example-artist.com'])).toBeNull();
+  });
+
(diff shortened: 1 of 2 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen

Reviewed the full diff (2 files, +110/-5): `api/functions/artist-profile.ts` (the `sanitizeEmbed`/`getOwnedLinkHostnames` change plus the silent-null-to-400 fix) and the new `artist-profile-embed.test.ts`.

## Security

- **Auth/authorization**: No issues found — endpoint still requires `authenticateRequest`, profile ownership (`profile.user_id !== userId` → 403), and `profile.verified_at` (unchanged by this PR).
- **Input validation / unsafe defaults / data exposure — CONFIRMED**: `getOwnedLinkHostnames()` treats any URL the artist has ever listed as a profile "link" as a trusted, embeddable domain. Nothing in this codebase actually verifies that an artist *owns* a link's hostname — the "Update links" block a few lines down (~line 377 of the updated file) only checks that the URL parses as `http:`/`https:`, not domain ownership. Combined with `sanitizeEmbed` now accepting any `ownedHostnames` entry, and `getOwnedLinkHostnames` reading `body.links` from the *same request* when present, a verified/claimed artist account can submit one `PUT` with both a brand-new link (`{ platform: 'website', url: 'https://attacker.example' }`) and `featuredEmbed: '<iframe src="https://attacker.example/...">'`, and the embed passes validation and gets persisted to `featured_embed`. That's a complete bypass of the fixed `ALLOWED_EMBED_DOMAINS` allowlist this endpoint exists to enforce — it converts a curated, trusted-platform-only list into an artist-self-declared one, in a single request, with no review step. The public artist page then renders that iframe directly into the page — a much stronger trust signal than an outbound link (it's visually part of the unstream.stream page, i.e. phishing/clickjacking surface), which is exactly the risk class `ALLOWED_EMBED_DOMAINS` is meant to contain. Blast radius is bounded to claimed/verified profiles rather than anonymous users, but that's the same trust tier the allowlist was built to constrain.
  - Suggested fix: only treat a hostname as "owned" if it was already persisted *before* this request (drop or gate the `bodyLinks` same-request path), and/or require an actual ownership check (DNS/file verification, or manual review) before trusting a self-hosted domain for iframe embedding.
- **Secret handling**: No issues found.
- **Injection risks**: No issues found — no SQL/command construction changed; Supabase client calls remain parameterized.

## Regression

- **Changed assumptions — needs-manual-verification**: The comment "must match `api/functions/artist-profile.ts`" on `ALLOWED_EMBED_DOMAINS` in `api/edge/artist-page-static.ts` is now stale — that file's `isEmbedDomainAllowed` only checks the static list, not `ownedHostnames`. A self-hosted (Faircamp-style) embed this PR is meant to unlock will validate and save via `artist-profile.ts`, then likely fail `isEmbedDomainAllowed` in the edge renderer and not display — reproducing a variant of the exact "accepted but not shown" bug this PR sets out to fix, just moved from write-time to render-time. Worth confirming before merge.
- **Edge cases / null-empty behavior**: Confirmed handled correctly — empty/whitespace `featuredEmbed` still clears the embed (`embed` stays `null`) without hitting the new 400 path; covered by the new tests.
- **API contract drift**: New 400 `{ error: string }` response on a rejected non-empty embed is a behavior change from the prior always-200 response; worth a quick check that the web editor already surfaces this string (out of scope of this diff to confirm from the API side alone).
- **State transitions / off-by-one**: None of substance.

**Test coverage**: The new tests cover the intended owned-hostname behavior well but don't cover the same-request `bodyLinks`-supplies-a-brand-new-domain scenario above — worth adding a test asserting an embed is *not* accepted purely because it was submitted alongside a never-before-seen link in the same request, once the fix lands.

---

PR #369, opened 2026-08-01
Fix artist pages rendering two different UIs depending on entry point

brandonlucasgreen:
## Summary

`/a/:slug` and `/artist/:slug` had two renderers for one URL — the exact "two renderers for one URL" anti-pattern called out in CLAUDE.md and `docs/retros/UNS-100-bifurcation-retro.md`:

- **Direct load** (typed URL, bookmark, external link, page refresh): the `artist-page-static` edge function unconditionally served a hand-written static page — 640px container, bare header with no Login link or auth state.
- **Client-side navigation** (clicking an artist from `/artists` while already in the app): React Router matched the SPA's own `/a/:slug` route, rendering the real `ArtistPage.tsx` — wide (`max-w-4xl`) container, full `Header` with Login.

Same URL, two different UIs depending on how you arrived — which is exactly what was reported: opening `/a/kid-lightbulbs` directly gives a narrower page missing the Login link.

## Fix

`api/edge/og-metadata.ts` (the `/` route) already solves this correctly: it checks the User-Agent and only serves crawlers a static page, letting real browsers (and JS-executing indexing crawlers) fall through to the SPA via `context.next()`. This PR applies the same gate to `artist-page-static.ts`:

- Real browsers → `context.next()` → Netlify's SPA catch-all → `index.html` → React Router → the same `ArtistPage.tsx` a client-side click would render. Direct load and in-app navigation are now identical.
- Social crawlers (Facebook, Twitter, Slack, etc., which don't execute JS) and indexing crawlers (Googlebot, etc.) still get the fully-populated static HTML with real OG tags and JSON-LD — unchanged for SEO/link previews.

Extracted the shared bot-detection lists/helpers (previously duplicated only in `og-metadata.ts`) to `api/shared/crawler-detection.ts` so both edge functions share one implementation.

## Verification

`npm run dev` doesn't run edge functions, and `netlify dev` in this environment fails to bundle unrelated functions (`jose` resolution error, pre-existing). Verified instead by executing the real `artist-page-static.ts` handler with its Deno-only imports rewritten to local stubs (same technique as `scripts/preview-release-page.ts`), covering:

- Regular browser UA → falls through to SPA ✅
- No UA header → falls through to SPA ✅
- Googlebot UA → still gets full static HTML, contains artist name ✅
- `facebookexternalhit` UA → still gets full static HTML ✅

Also confirmed via `deno check --allow-import` that no new type errors were introduced (the 3 pre-existing `.abortSignal`/Supabase typing errors are unrelated and predate this change).

## Test plan

- [x] `npm run test:unit` — 492 passed
- [x] `npm run test:api` — 389 passed
- [x] `npm run lint` — no new errors (77 pre-existing, confirmed identical via stash baseline)
- [x] Manual handler execution confirms routing behavior for both human and crawler UAs
- [ ] Spot-check `/a/kid-lightbulbs` on the deploy preview to confirm the Login link and wide layout now appear on a direct load

### DIFF
--- a/api/edge/artist-page-static.ts
+++ b/api/edge/artist-page-static.ts
@@ -4,6 +4,7 @@ import { PLATFORMS } from "../shared/platform-registry.ts";
 import { isBandcampFriday } from "../shared/bandcamp-friday.ts";
 import { mainLinkDividerIndexes } from "../shared/link-dividers.ts";
 import { cheapestOfferSummary, formatReleaseDate } from "../shared/release-display.ts";
+import { isSocialCrawler, isIndexingCrawler } from "../shared/crawler-detection.ts";
 
 /**
  * How many releases to list before summarising the rest.
@@ -181,6 +182,16 @@ export default async function handler(request: Request, context: Context) {
     const slug = url.pathname.replace(/^\/(a|artist)\//, '').replace(/\/$/, '');
 
     if (!slug) return context.next();
+
+    // Real browsers get the SPA, same as a client-side <Link> navigation from /artists would
+    // render — so a direct load and an in-app click produce the same UI (see
+    // docs/retros/UNS-100-bifurcation-retro.md). Only crawlers, which either can't run JS
+    // (social previews) or benefit from content that doesn't wait on a client fetch (indexing),
+    // get this static render.
+    const userAgent = request.headers.get('user-agent');
+    if (!isSocialCrawler(userAgent) && !isIndexingCrawler(userAgent)) {
+      return context.next();
+    }
 
     const supabaseUrl = Deno.env.get("SUPABASE_URL");
     const supabaseKey = Deno.env.get("SUPABASE_SERVICE_KEY");
--- a/api/edge/og-metadata.ts
+++ b/api/edge/og-metadata.ts
@@ -1,51 +1,5 @@
 import { Context } from "https://edge.netlify.com";
-
-// Social media crawler user agents (for OG tag previews only)
-const SOCIAL_CRAWLER_USER_AGENTS = [
-  'facebookexternalhit',
-  'Facebot',
-  'Twitterbot',
-  'LinkedInBot',
-  'Pinterest',
-  'Slackbot',
-  'TelegramBot',
-  'WhatsApp',
-  'Discordbot',
-  'Applebot',
-  'Mastodon',
-  'Pleroma',
-  'Misskey',
-  'Akkoma',
-  'Pixelfed',
-  'PeerTube',
-  'Lemmy',
-  'Bluesky',
-  'bsky.app',
-  'redditbot',
-];
-
-// Indexing crawlers (need actual page content, not just OG tags)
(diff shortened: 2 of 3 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Automated staff-eng review of this PR (security + regression pass).

**Scope reviewed:** `api/edge/artist-page-static.ts`, `api/edge/og-metadata.ts`, `api/shared/crawler-detection.ts` (the full diff — no other open PRs).

## Security

- **Auth/authorization** — not applicable; no auth/session logic touched.
- **Input validation** — `confirmed`, no risk. `User-Agent` is only read and matched with case-insensitive `.includes()` to pick a control-flow branch (`context.next()` vs. static render). It is never reflected into HTML, logs, or a query, so there's no injection surface here.
- **Secret handling** — unchanged; `SUPABASE_SERVICE_KEY` usage is untouched by this diff.
- **Injection risks** — none introduced. All existing output in the untouched parts of the file still goes through `escapeHtml`/`isEmbedDomainAllowed`.
- **Unsafe defaults** — `plausible`, low severity, pre-existing pattern (not new to this PR): both crawler lists match on a simple substring, so any client can spoof e.g. `User-Agent: Googlebot` to force the Supabase-backed render path (up to 4 queries) instead of the free SPA passthrough. Since this route was *unconditional* before this PR (every request already hit Supabase), the change is a net reduction in exposure, not an increase — flagging for awareness only, not blocking.
- **Data exposure** — none. Static and SPA paths render the same public artist data.

## Regression

- **Changed assumption (intended)** — direct-loaded `/a/:slug` and `/artist/:slug` requests from real browsers now fall through to the SPA instead of getting the static HTML. This is the fix and matches the PR description.
- **Edge case: missing `User-Agent`** — a request with no UA header now falls through to the SPA (previously got the static page). Test plan calls this out as intentional; worth a quick sanity check that no internal tooling, health checks, or legacy clients relied on the old no-UA static-page response.
- **Edge case: `needs-manual-verification`** — a few entries in `SOCIAL_CRAWLER_USER_AGENTS` (`WhatsApp`, `Applebot`, `Pinterest`) are known to also appear in real end-user in-app-browser UAs, not just the bot that fetches link previews. If a human taps a link inside one of those apps and the UA contains the same substring, they'd get the static/narrow page again — reproducing the exact bug this PR fixes, just for a different UA class. Worth a real tap-test from WhatsApp/Pinterest on the deploy preview before calling this fully closed.
- **Test coverage gap** — `plausible`, minor: `isSocialCrawler`/`isIndexingCrawler` moved into a new shared module now used by two routes, but no dedicated unit test was added for the module itself. A regression here now has 2x blast radius; a small test would be cheap insurance.
- **Off-by-one / state transitions / API contract drift** — none found. Handler is stateless per-request and the crawler-facing response is byte-identical to before.

Overall: small, well-scoped diff, and the manual UA-matrix verification in the description is solid. Nothing here blocks merge.


---

PR #370, opened 2026-08-01
Make releases work on the React artist page (list + admin catalog button)

brandonlucasgreen:
Two commits, both fixing the same root cause: **the Releases feature was built on the artist-page renderer #369 retired for humans.**

#366 put the releases list in `artist-page-static`, and #369 — landing hours later — made that edge function crawler-only. So Googlebot has been seeing artists' discographies with prices and no human ever has. My first version of the admin button had the same flaw.

## 1. The releases list, on the renderer real browsers get

- `getArtistReleases` in `db.ts` — one shape, used once. `count: 'exact'` rides along with the limit so "and N more" is a real number rather than a guess, and the ordering matches `idx_releases_artist_chrono` (many releases still have no date; without the `created_at` tiebreaker those rows shuffle between requests).
- `/api/artist-page` now carries `releases` and `releaseCount`. Both optional in the type, since cached responses predate them.
- `ReleasesSection` renders in `RichArtistProfile` and `UnclaimedQuietCard`, **below Follow** — Support directly and Follow are one artist-ordered region that releases shouldn't interrupt.

```
RELEASES
  [art]  fruit is year 3 (honeycrush)              COMING
         Single · 5 September 2026
  [art]  RUINED CASTLE
         Album · 1 June 2024 · from $7
  and 8 more
```

**One implementation of the formatting.** `ReleasesSection` imports `cheapestOfferSummary` and `formatReleaseDate` from `api/shared/release-display.ts` — the same module the crawler-side renderer uses, so the two can't drift on what a price or a partial date looks like. This is the first import from `api/shared` into the web app; it builds and typechecks. The alternative was a third hand-maintained copy of code that makes claims about someone's money, which `sources.ts` already shows the cost of.

The edge function keeps its own copy for crawlers. That's what it's for now.

## 2. The admin "Catalog releases now" button

Also React, for the same reason. `isAdmin` from `useAuth()` decides whether it renders, matching how the rest of the app gates admin UI.

**The internal secret never reaches the browser** — you authenticate with your ordinary Supabase token, and `/api/admin/catalog-artist` holds `INTERNAL_FUNCTION_SECRET` server-side. `isAdmin` is presentation only; the endpoint checks `ADMIN_EMAIL` on every request.

Two details worth keeping: the endpoint **clears the 7-day cooldown** first (otherwise the button appears to work and silently does nothing for a week), and it **polls `release_catalog_state`** rather than trusting the 202, which a background function returns even when it's about to refuse.

## Testing

Nine component tests on the things that would be quietly wrong rather than broken — a fabricated day on a year-only release, a price quoted from a sold-out format, a zero price reading as free rather than name-your-price, and a heading with nothing under it. 35 web test files and 24 API files pass; lint is at the pre-existing baseline; full `npm run build` green.

## Requires `ADMIN_EMAIL`

`authenticateAdmin` returns null when it's unset, so the endpoint 403s everyone and the button never appears. Worth confirming alongside `INTERNAL_FUNCTION_SECRET`.

### DIFF
--- a/api/functions/admin-catalog-artist.ts
+++ b/api/functions/admin-catalog-artist.ts
@@ -33,9 +33,22 @@ interface CatalogState {
   consecutive_failures: number;
 }
 
-async function readState(artistId: string): Promise<CatalogState | null> {
+/**
+ * Read this artist's catalog state.
+ *
+ * Deliberately three-valued, not two. "We couldn't ask" and "we asked and this artist has never
+ * been catalogued" are different facts, and collapsing them into one `null` makes a broken
+ * database read render as a confident "Never catalogued" — the same shape as the bug class the
+ * "never cache uncertainty" principle exists to prevent. This whole feature exists to give
+ * cataloging an observable surface, so an unreadable one has to say so.
+ */
+type StateResult =
+  | { ok: true; state: CatalogState | null }
+  | { ok: false; reason: string };
+
+async function readState(artistId: string): Promise<StateResult> {
   const client = getClient();
-  if (!client) return null;
+  if (!client) return { ok: false, reason: 'Supabase is not configured on this deploy' };
 
   const { data, error } = await client
     .from('release_catalog_state')
@@ -45,9 +58,9 @@ async function readState(artistId: string): Promise<CatalogState | null> {
 
   if (error) {
     console.error('[admin-catalog] state read failed:', error.message);
-    return null;
+    return { ok: false, reason: 'Could not read catalog state' };
   }
-  return (data as CatalogState | null) ?? null;
+  return { ok: true, state: (data as CatalogState | null) ?? null };
 }
 
 /**
@@ -106,15 +119,40 @@ export async function handler(event: {
   }
 
   if (event.httpMethod === 'GET') {
+    const result = await readState(artistId);
+    if (!result.ok) {
+      // 503, not a 200 carrying a null state: the caller must be able to tell "we couldn't ask"
+      // from "never catalogued", or it will report the second when the first is true.
+      return {
+        statusCode: 503,
+        headers: CORS_HEADERS,
+        body: JSON.stringify({ error: result.reason }),
+      };
+    }
     return {
       statusCode: 200,
       headers: CORS_HEADERS,
-      body: JSON.stringify({ admin: true, state: await readState(artistId) }),
+      body: JSON.stringify({ admin: true, state: result.state }),
     };
   }
 
   if (event.httpMethod !== 'POST') {
     return { statusCode: 405, headers: CORS_HEADERS, body: JSON.stringify({ error: 'Method not allowed' }) };
+  }
+
+  // Every reason the background function would refuse, checked here instead — because its
+  // answer never comes back. Netlify's dispatcher returns 202 to the caller the instant it
+  // accepts a background invocation and discards whatever the handler returns, so a 401 from a
+  // bad secret and a 403 from a non-production context both reach us as 202. Anything not
+  // checked up front is indistinguishable from a slow crawl.
+  if (process.env.CONTEXT !== 'production') {
+    return {
+      statusCode: 503,
+      headers: CORS_HEADERS,
+      body: JSON.stringify({
+        error: `Cataloging only runs in production (this deploy is "${process.env.CONTEXT ?? 'unset'}").`,
+      }),
+    };
   }
 
   const secret = process.env.INTERNAL_FUNCTION_SECRET;
@@ -135,20 +173,17 @@ export async function handler(event: {
 
   try {
     const trigger: CatalogTrigger = 'saved'; // the larger hourly budget — this is a deliberate act
-    const response = await fetch(`${siteUrl}/.netlify/functions/catalog-artist-background`, {
+    await fetch(`${siteUrl}/.netlify/functions/catalog-artist-background`, {
       method: 'POST',
       headers: { 'Content-Type': 'application/json', Authorization: `Bearer ${secret}` },
       body: JSON.stringify({ artistIds: [artistId], trigger }),
     });
 
-    // A background function answers 202 the moment it accepts the job — including when it is
-    // about to refuse on a bad secret. This status is a handshake, not an outcome, so the client
-    // polls the GET above for what actually happened.
-    return {
-      statusCode: 202,
-      headers: CORS_HEADERS,
-      body: JSON.stringify({ started: response.status === 202 || response.ok }),
-    };
+    // Nothing useful to report back from that response, and deliberately no `started` flag:
+    // the dispatcher's 202 says only that the request was queued, so any boolean derived from
+    // it would read as "the crawl was accepted" while being true even when it wasn't. The
+    // client polls the GET above, which reports what actually happened.
+    return { statusCode: 202, headers: CORS_HEADERS, body: JSON.stringify({ queued: true }) };
   } catch (error) {
     Sentry.captureException(error);
     return {
--- a/api/functions/admin-catalog-artist.ts
+++ b/api/functions/admin-catalog-artist.ts
@@ -1,0 +1,169 @@
+// API endpoint: GET/POST /api/admin/catalog-artist
+//
+// Admin-only. Catalogs one artist's releases on demand, from a button on their page.
+//
+// Cataloging is otherwise triggered only by a fan saving an artist or a search resolving one,
+// which is deliberate — it's what keeps the crawl small and the page count bounded. This is the
+// same escape hatch as `npm run catalog:artist`, reachable without a terminal.
+//
+// **The internal secret never leaves the server.** The browser authenticates as an admin with
+// its ordinary Supabase token; this function holds INTERNAL_FUNCTION_SECRET and uses it to
+// invoke the background function. Shipping that secret to the client would hand every visitor
+// the ability to make Unstream crawl Bandcamp on demand — the exact amplifier the check-releases
+// hardening existed to close.
+//
+//   GET  ?artistId=…  → is the caller an admin, and what is this artist's catalog state
+//   POST { artistId } → clear the cooldown and start a crawl
+//
+// The GET is what the button uses to decide whether to show itself: visibility follows the
+// server's real admin rule rather than a copy of it in page markup.
+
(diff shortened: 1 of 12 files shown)


api/functions/admin-catalog-artist.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  const { data, error } = await client
+    .from('release_catalog_state')
+    .select('last_catalogued_at, last_attempted_at, releases_found, releases_detailed, last_error, consecutive_failures')
+    .eq('artist_id', artistId)
+    .maybeSingle();
+
+  if (error) {
+    console.error('[admin-catalog] state read failed:', error.message);
+    return null;
```
brandonlucasgreen: **[regression][confirmed] `readState` conflates "no crawl yet", "DB error", and "Supabase not configured" into the same `null`**

Three distinct outcomes all collapse to `return null` here:
- line 38: `getClient()` failed (env misconfigured)
- line 48: the Supabase query itself errored
- line 50: the query succeeded but no row exists yet

The client (`AdminCatalogButton.describe()`) then renders all three as `"Never catalogued"`. If Supabase creds are misconfigured or a query transiently fails, an admin who has already cataloged an artist would see "Never catalogued" — indistinguishable from a true never-run state, and indistinguishable from an infra failure worth paging on.

This is the exact anti-pattern the codebase calls out elsewhere (`CLAUDE.md` → "Never cache uncertainty" / `bandcamp-probe.ts`'s `undecided` verdict): distinguish "the upstream answered with nothing" from "the upstream didn't answer." Worth at least a `console.error`/Sentry capture that's distinguishable from the true-empty case, or surfacing an explicit "couldn't load state" string to the admin instead of silently coercing to "never catalogued."

Not a security issue — this endpoint is admin-only — but worth fixing given how central this principle is elsewhere in the codebase.

---

apps/web/src/components/AdminCatalogButton.tsx
```diff
@@ -0,0 +1,157 @@
...
+        method: 'POST',
+        headers: { Authorization: `Bearer ${token}`, 'Content-Type': 'application/json' },
+        body: JSON.stringify({ artistId }),
+      });
+      const body = await response.json().catch(() => ({}));
+
+      // 202 is only a handshake — a Netlify background function answers it the instant it
+      // accepts the job, including when it is about to refuse. Polling below reports what
+      // actually happened.
+      if (!response.ok && response.status !== 202) {
```
brandonlucasgreen, line 162: **[regression: API contract drift][confirmed] `body.started` is parsed but never checked**

The server (`admin-catalog-artist.ts`) deliberately returns `{ started: response.status === 202 || response.ok }` in its 202 response, precisely to let the client know whether the background invocation actually succeeded (comment there: *"202 is a handshake, not an outcome"*).

But here, `response.status` is the **admin-catalog-artist** function's own status — which is `202` on every non-error path, regardless of what `body.started` says. Since `fetch`'s `response.ok` is `true` for any 2xx (202 included), the condition on line 131 (`!response.ok && response.status !== 202`) is always false whenever the outer call itself succeeded, so `poll()` always runs — even when `body.started === false` (e.g. the background function 401'd on a secret mismatch, or refused because `CONTEXT !== 'production'`).

Concretely: if the background invocation is rejected, the admin sees "Cataloguing…" → polls for up to 2 minutes → "Still running — check back shortly", instead of a clear failure message, even though nothing is running. Suggest checking `body.started` explicitly and short-circuiting to an error state when it's `false`.

Not exploitable, but it makes the one signal this feature added for diagnosing a broken deploy (bad `INTERNAL_FUNCTION_SECRET`, wrong `CONTEXT`) silently unreachable from the UI.

---

brandonlucasgreen (commented):

## Security

**auth/authorization** — needs-manual-verification → **confirmed sound.** Both `GET` and `POST` go through `authenticateAdmin`, which does a fresh Supabase `auth.getUser()` call (not a cached/local JWT check) and compares the result against `ADMIN_EMAIL` server-side (`middleware.ts`). `isAdmin` in `AdminCatalogButton` is presentation-only, as the PR description states — the endpoint doesn't trust it. Verified `ADMIN_EMAIL` unset → `authenticateAdmin` returns `null` → 403 for everyone, so the fail-closed default holds.

**input validation** — confirmed sound. `artistId` is validated against a UUID regex on both GET (query param) and POST (JSON body) before it ever reaches a Supabase query; POST body parsing (`safeParseArtistId`) is wrapped in try/catch and fails closed to a 400.

**secret handling** — confirmed sound. `INTERNAL_FUNCTION_SECRET` is read only server-side and only used to build the `Authorization` header sent server-to-server to `catalog-artist-background`; it never appears in any response body. The downstream background function verifies it with `timingSafeEqual`, not `===`.

**injection risks** — confirmed none. All DB access goes through the Supabase client's structured query builder (`.eq()`, `.update()`), no raw SQL, no string-interpolated queries.

**unsafe defaults** — confirmed sound. Missing `INTERNAL_FUNCTION_SECRET`/site URL → 503 (deny), not a silent no-op. Missing `ADMIN_EMAIL` → 403. CORS uses `buildCorsHeaders(origin, false)`, so it's restricted to `ALLOWED_ORIGIN` rather than reflecting arbitrary origins.

**data exposure** — confirmed none beyond the endpoint's own trust boundary. `GET` returns catalog state (including `last_error`) only to an authenticated admin.

No confirmed or plausible security findings on this PR.

## Regression

Two confirmed correctness issues, both left as inline comments — neither is a security hole, but both degrade the observability this feature exists to provide:

1. **`api/functions/admin-catalog-artist.ts` `readState`** (edge case / null handling) — collapses "no client configured", "query error", and "genuinely never catalogued" into the same `null`, which the UI then renders identically as "Never catalogued." Same class of bug the codebase's "never cache uncertainty" principle exists to prevent elsewhere.
2. **`apps/web/src/components/AdminCatalogButton.tsx` `start()`** (API contract drift) — the server's `started: boolean` field (meant to distinguish "background job accepted" from "background job rejected") is parsed but never read, because `fetch`'s `response.ok` is already true for the wrapping function's 202. A rejected background invocation (bad secret, non-production `CONTEXT`) is indistinguishable in the UI from a slow one.

Everything else checked out: `claimArtistForCatalog`'s cooldown-clearing math (2h backdate + `consecutive_failures: 0` correctly clears both the 7-day cooldown and the backoff), the `'saved'` trigger's larger hourly cap (240 vs 60, matches the code), the `netlify.toml` redirect (no ordering/shadowing issue with other `/api/admin/*` routes), and the background function's independent `CONTEXT === 'production'` + `timingSafeEqual` checks (defense in depth, not solely relying on this new endpoint's gating).

---

brandonlucasgreen (commented):
@brandonlucasgreen Automated security/regression pass on the new `admin-catalog-artist` endpoint + releases-on-artist-page feature.

## Security

**Auth/authorization — confirmed, no issue.** `authenticateAdmin` (checked in `api/functions/middleware.ts`) verifies the bearer token *and* matches the caller's email against `ADMIN_EMAIL` server-side; the 403 path covers anonymous and non-admin callers alike, and this happens before any input is trusted. The `isAdmin` check in `AdminCatalogButton.tsx` is correctly presentation-only per the file's own comment — matches the existing `me-*` pattern.

**Secret handling — confirmed, no issue.** `INTERNAL_FUNCTION_SECRET` stays server-side; the browser never sees it, only the admin's ordinary Supabase token. Consistent with the file's stated threat model (don't hand every visitor a Bandcamp-crawl amplifier).

**Input validation — confirmed, no issue.** `artistId` is validated against a strict UUID regex before any DB query, on both GET (query param) and POST (JSON body via `safeParseArtistId`, which fails closed to `undefined` on parse error). Test suite explicitly covers a path-traversal-shaped input (`../../etc/passwd`) rejected with 400.

**Injection risk — confirmed, no issue.** All DB access goes through the Supabase query builder (`getArtistReleases`, `readState`, `clearCooldown`) — no raw SQL, no string interpolation into queries.

**Unsafe defaults — confirmed, no issue.** The POST path fails closed: refuses outside `CONTEXT === 'production'`, refuses if `INTERNAL_FUNCTION_SECRET` is unset, and — because Netlify's background-function dispatch discards the real response — the code deliberately checks every predictable failure up front rather than trusting a 202. `buildCorsHeaders(origin, false)` also fails closed to `ALLOWED_ORIGIN` for non-API-key requests, matching every other authenticated endpoint.

**Data exposure — needs-manual-verification (low severity).** `getArtistReleases` in `db.ts` is called from the public (unauthenticated) `artist-page.ts` handler and returns price/availability data to any visitor — this looks intentional (it's the feature), but worth a human confirming that release price/availability was meant to be public-by-default rather than gated the way the catalog-*trigger* is. Filtering `is_hidden = false` in the query itself (not after) is the right pattern per the codebase's own "never cache uncertainty"-style reasoning.

No confirmed vulnerabilities.

## Regression

- **plausible** — `AdminCatalogButton.tsx`: `MAX_POLLS=24` × `POLL_INTERVAL_MS=5000` caps polling at ~2 minutes, but the same file's comment says "a full catalogue with prices takes a few minutes." If a real crawl regularly exceeds 2 minutes, the UI will report "Still running" as a give-up state prematurely on every use rather than actually catching completion. Worth confirming the real crawl duration against this constant.
- **needs-manual-verification** — `admin-catalog-artist.ts` is a new file under `api/functions/` but `api/tsconfig.json`'s `include` is scoped narrowly to `me-*` functions (per this repo's own documented gotcha). This file won't be typechecked by `tsc -b` or `npm run typecheck:api`, so a type error here would only surface at runtime in production, not in CI. Consider adding it to that include list given it's a new, non-trivial admin surface.
- **confirmed, non-blocking** — state machine for "never catalogued" vs. "couldn't read" vs. "real state" is explicitly three-valued end-to-end (DB → function → component) and the test suite (`admin-catalog-artist.test.ts`, `ReleasesSection.test.tsx`) locks the specific edge cases called out in this repo's "never cache uncertainty" principle (zero-price vs. name-your-price, sold-out offers, year-only/undated releases, failed-read vs. never-catalogued). This is good — flagging only because the same rigor didn't carry over to the polling timeout above.

Overall: solid, defensive implementation consistent with the codebase's stated security/engineering principles. One low-severity data-exposure point to confirm is intentional, one plausible UX/timing mismatch, and one typecheck-coverage gap worth closing.

---

PR #372, opened 2026-08-01
Gate cataloging on a variable that exists at function runtime

brandonlucasgreen:
**Release cataloging has never run. Not once, in production, since #359.**

All three gates tested `process.env.CONTEXT !== 'production'`. Netlify exposes **only `URL`, `SITE_NAME` and `SITE_ID`** to a serverless function at runtime — `CONTEXT` and `DEPLOY_PRIME_URL` are build-time variables ([docs](https://docs.netlify.com/build/functions/environment-variables/)). So `CONTEXT` is `undefined` in every deployed function, and every gate refused:

| Gate | Effect in production |
|---|---|
| `request-catalog.ts` | the save trigger and the search trigger never fired |
| `catalog-artist-background.ts` | refused every invocation it did receive |
| `admin-catalog-artist.ts` | the error you just saw |

So the empty `releases` table had **two** independent causes — the unset `INTERNAL_FUNCTION_SECRET`, and this. Fixing the first alone would not have helped.

## Why it hid for so long

Refusing looked exactly like working. Two of the three gates logged a line nobody reads, and an empty catalogue is precisely what a quiet week looks like — which is why I accepted "no traffic yet" as the explanation twice.

It surfaced only when the admin button reported the refusal out loud: *"this deploy is unset"*. That's the entire argument for the observability work in #370 — the same refusal had been happening silently for days.

**And the tests passed the whole time**, because they set `CONTEXT` themselves. They were testing a rule the runtime could never satisfy. That's the more uncomfortable lesson: a test that constructs the environment it asserts on can only ever confirm the author's model of it.

## The fix

`RELEASE_CATALOG_ENABLED`, a custom variable scoped to Functions and set for the **Production context only** — the documented way to get a per-context value into a function.

```ts
export function isCatalogEnabled(): boolean {
  return process.env.RELEASE_CATALOG_ENABLED === 'true';
}
```

Unset means off, and so does anything that isn't exactly `'true'`. A crawl budget that fails open is worse than one that fails closed. Deploy previews and branch deploys stay off, preserving the original intent — they run against the *production* Supabase, so an ungated one would write real releases and spend the real hourly budget.

Also drops `DEPLOY_PRIME_URL || URL` for plain `URL`, since the former is likewise undefined at runtime. The comment explaining why `DEPLOY_PRIME_URL` was preferred described build-time behaviour and was never true of this code path.

## ⚠️ Action needed

**Set `RELEASE_CATALOG_ENABLED=true` in Netlify — scoped to Functions, Production context only.** Without it cataloging stays off; same as today, but now for a visible reason.

## Testing

The gating tests now assert the real rule, including that a non-`'true'` value stays off. 32 tests across the two catalog test files, 25 API + 35 web files green, full build passes. CLAUDE.md updated with the runtime-variable constraint so nobody reaches for `CONTEXT` in a function again.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -163,11 +163,17 @@ This is the lesson behind a run of bug fixes (#317–#328) and it applies to eve
 
 ### Testing release ingest locally
 
-Release cataloging only runs when `CONTEXT === 'production'`, because deploy previews and local
-runs both point at the **production** Supabase — so an ungated preview would write real
-`releases` rows and spend the real hourly crawl budget. That means ingest cannot be exercised
-on a deploy preview, and **`CONTEXT=production` is not a valid local workaround** — it would
-have your laptop writing production data.
+Release cataloging only runs where `RELEASE_CATALOG_ENABLED=true`, a custom env var scoped to
+Functions and set for the **Production context only**. Deploy previews and local runs both point
+at the **production** Supabase, so an ungated preview would write real `releases` rows and spend
+the real hourly crawl budget. Ingest therefore cannot be exercised on a deploy preview, and
+**setting the flag locally is not a valid workaround** — it would have your laptop writing
+production data.
+
+This used to gate on `CONTEXT === 'production'`, which silently disabled cataloging entirely:
+Netlify exposes only `URL`, `SITE_NAME` and `SITE_ID` to a serverless function at runtime, so
+`process.env.CONTEXT` is `undefined` in every deployed function. Don't reach for `CONTEXT` or
+`DEPLOY_PRIME_URL` in a function — neither exists there.
 
 Use the dry run instead:
 
--- a/api/functions/__tests__/admin-catalog-artist.test.ts
+++ b/api/functions/__tests__/admin-catalog-artist.test.ts
@@ -46,7 +46,7 @@ beforeEach(() => {
   mocks.authenticateAdmin.mockResolvedValue({ userId: 'u1', email: 'admin@example.com' });
   mocks.maybeSingle.mockResolvedValue({ data: null, error: null });
   vi.stubGlobal('fetch', mocks.fetch);
-  process.env.CONTEXT = 'production';
+  process.env.RELEASE_CATALOG_ENABLED = 'true';
   process.env.INTERNAL_FUNCTION_SECRET = 'secret';
   process.env.URL = 'https://unstream.stream';
 });
@@ -54,7 +54,7 @@ beforeEach(() => {
 afterEach(() => {
   vi.unstubAllGlobals();
   vi.clearAllMocks();
-  delete process.env.CONTEXT;
+  delete process.env.RELEASE_CATALOG_ENABLED;
   delete process.env.INTERNAL_FUNCTION_SECRET;
   delete process.env.URL;
 });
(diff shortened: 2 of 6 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Security/regression pass on the final (rebased) diff — 6 files, +81/-52 against `main`, `mergeable_state: clean`, CI green. Note: this PR briefly showed a much larger diff (and tripped the `semantic-revert-check` bot) because the branch was based on a pre-`#370`/`#359` commit; it's since been rebased to a clean, minimal diff, so that concern is resolved and I reviewed the current state only.

## Security

- **Auth/authorization** — no changes. `authenticateAdmin` (admin-catalog-artist) and the shared-secret `timingSafeEqual` check (catalog-artist-background) are untouched. — *confirmed no issue*
- **Unsafe defaults** — this is the crux of the PR. `isCatalogEnabled()` returns `true` only for the exact string `'true'`; unset, `'false'`, `'1'`, `'yes'`, and `''` all stay disabled, and this is covered by `it.each` cases in both test files. Correctly fail-closed. — *confirmed no issue*
- **Secret handling** — `INTERNAL_FUNCTION_SECRET` flow is unchanged; `RELEASE_CATALOG_ENABLED` is a plain feature flag, not a secret, so no exposure concern there. — *confirmed no issue*
- **Input validation / injection / data exposure** — no new input paths or query construction in this diff. — *confirmed no issue*
- **Case/whitespace sensitivity in the flag** — `RELEASE_CATALOG_ENABLED === 'true'` is a strict match. A value like `'True'`, `'TRUE'`, or one with trailing whitespace from a copy-paste into the Netlify UI silently stays disabled, with nothing distinguishing that from "not set yet." Correct fail-closed behavior per the codebase's own principle, but worth double-checking exact casing/whitespace when you set it in Netlify, since a typo here reproduces the same silent-failure shape this PR exists to fix. — *plausible, low severity*

## Regression

- **The fix is inert without a manual step.** Cataloging stays off until `RELEASE_CATALOG_ENABLED=true` is set in Netlify (Functions scope, Production context), which the PR body already flags under "⚠️ Action needed." Worth confirming that's actually been done post-merge — otherwise this ships as a correct fix that silently doesn't fix anything, which is exactly the failure mode the PR describes. — *needs-manual-verification*
- **Two stale references to the removed `CONTEXT` gate survive outside this diff's file list:**
  - `scripts/ingest-releases.ts` (~line 21): "…which is the exact thing the `CONTEXT === 'production'` gate exists to prevent."
  - `apps/web/src/components/AdminCatalogButton.tsx` (~line 160): "the refusals it can predict (non-production context, missing secret)."
  
  Both predate this PR and weren't touched by it, but given CLAUDE.md was updated specifically so "nobody reaches for `CONTEXT` in a function again," and this codebase treats these why-comments as load-bearing, these two are now describing a mechanism that no longer exists. No functional impact — worth a follow-up cleanup pass. — *confirmed (doc drift, not a runtime issue)*
- **API response text changed but nothing parses it.** Confirmed via repo-wide grep that no code depends on the old `'non-production context'` / `Cataloging only runs in production…` strings — the background function's response body is discarded by Netlify's dispatcher, and the admin endpoint's error string is only ever displayed verbatim in the UI. — *confirmed no issue*
- **State transitions / off-by-one / null-empty behavior** — untouched by this diff; `claimArtistForCatalog`'s cooldown/backoff logic in `db.ts` is unchanged. — *confirmed no issue*

Overall: the gating fix itself is correct and well-tested (fail-closed, exact-match, both call sites covered). The only load-bearing item is the Netlify env var actually being set after merge.

---

PR #373, opened 2026-08-01
Serve real browsers the SPA on direct /u/:handle loads, not static HTML

brandonlucasgreen:
## Summary

- `/u/:handle` (public saved-artists pages) had the same two-renderers bug fixed for `/a/:slug` in #369: a direct page load hit `api/edge/u-handle.ts`'s hand-rolled static HTML with a stale header (no Login, plain search form) and footer (missing Dashboard/FAQ/Donate/Contact links) inside a 640px container, while client-side `<Link>` navigation rendered `PublicSavedArtistsPage.tsx` with the real, current `Header`/`Footer` components.
- Applied the same crawler gate used in `artist-page-static.ts`: real browsers now `context.next()` through to the SPA, so a direct load and an in-app click produce identical UI. Crawlers (Googlebot, facebookexternalhit, etc.) still get the fast static render for SEO/link previews.

## Why

Brandon reported `unstream.stream/u/brandon` was using an old header/footer and possibly the wrong layout. Root cause matched a bug already flagged (but left unfixed) after the original `/a/:slug` fix — same bifurcation pattern from `docs/retros/UNS-100-bifurcation-retro.md`.

## Test plan

- [x] `npm run test:api` — 389 tests passing
- [x] Verified in the dev server that `PublicSavedArtistsPage` (the path real browsers now take) renders with the current site-wide `Header`/`Footer` (Login, live search, full footer nav)
- [ ] Spot-check `unstream.stream/u/brandon` after deploy to confirm the header/footer now match the rest of the site

### DIFF
--- a/api/edge/u-handle.ts
+++ b/api/edge/u-handle.ts
@@ -1,11 +1,13 @@
 // Edge function: /u/{handle}
 // Static HTML renderer for public saved-artists pages.
 // Models after api/edge/artist-page-static.ts — same pattern, same RLS-aware data fetching.
-// The edge function is the single renderer per CLAUDE.md "one route, one renderer".
-// React app hydrates client-side for the Copy URL button only (progressive enhancement).
+// Real browsers get the SPA (PublicSavedArtistsPage.tsx), same as a client-side <Link>
+// navigation would render, so a direct load and an in-app click produce the same UI (see
+// docs/retros/UNS-100-bifurcation-retro.md). Only crawlers get this static render.
 
 import { Context } from "https://edge.netlify.com";
 import { createClient } from "https://esm.sh/@supabase/supabase-js@2";
+import { isSocialCrawler, isIndexingCrawler } from "../shared/crawler-detection.ts";
 
 function escapeHtml(str: string): string {
   return str.replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/>/g, '&gt;').replace(/"/g, '&quot;').replace(/'/g, '&#39;');
@@ -49,6 +51,11 @@ export default async function handler(request: Request, context: Context) {
     const handle = url.pathname.replace(/^\/u\//, '').replace(/\/$/, '');
 
     if (!handle) return context.next();
+
+    const userAgent = request.headers.get('user-agent');
+    if (!isSocialCrawler(userAgent) && !isIndexingCrawler(userAgent)) {
+      return context.next();
+    }
 
     const supabaseUrl = Deno.env.get("SUPABASE_URL");
     const supabaseKey = Deno.env.get("SUPABASE_SERVICE_KEY");


brandonlucasgreen (commented):
@brandonlucasgreen Automated staff-eng review of this PR.

Scope: the actual diff (per `merge-base`, this branch is `871c6e3` + one commit) is exactly the 11-line change in `api/edge/u-handle.ts`. I traced the downstream code it depends on (`api/functions/public-saved-artists.ts`, `apps/web/src/pages/PublicSavedArtistsPage.tsx`) even though they weren't touched, since the whole point of the change is to route real browsers to them instead of the SSR path.

## Security

- **Auth/authorization — needs-manual-verification, no issue found.** The new early return doesn't add or remove any authorization check itself — it just decides SSR-vs-SPA rendering. The actual gate (`saved_artists_public` flag) still lives in the SSR path for crawlers and, separately, in `public-saved-artists.ts` for the client fetch real browsers now take. I read `public-saved-artists.ts` and confirmed it independently checks `saved_artists_public`, returns a generic 404 (not 403) on both "unknown handle" and "private" to avoid leaking existence, and never includes email/user_id in the response. So there's no new exposure — this is the same authorization real in-app `<Link>` navigation already relied on before this PR; direct loads now just take the same path. Flagging as needs-manual-verification only because it depends on a file outside this diff, not because I found a defect.
- **Input validation — confirmed, no issue.** `isSocialCrawler`/`isIndexingCrawler` match on a raw `User-Agent` header (trivially spoofable), but spoofing it only changes *which renderer* serves the response, not what data is authorized — both renderers enforce the same sharing check independently. Not a new risk vs. the identical pattern already shipped in `artist-page-static.ts` (#369).
- **Secret handling — confirmed, no issue.** `SUPABASE_SERVICE_KEY` usage is unchanged.
- **Injection risks — confirmed, no issue.** No new interpolation; existing `escapeHtml` usage untouched.
- **Unsafe defaults / data exposure — confirmed, no issue.** Verified above.

## Regression

- **Changed assumption — confirmed, no issue.** The file's own comment previously said "React app hydrates client-side for the Copy URL button only" (single SSR renderer). I diffed the SSR HTML against `PublicSavedArtistsPage.tsx` field-by-field (owner name, `owner_location`, "Supported" badge, avatar fallback, empty-state copy, artist links) and they match, so real browsers won't see a content regression vs. what SSR was rendering.
- **Edge case: unknown/private handle — confirmed, no issue.** Both before and after this PR, a private or nonexistent handle falls through to `context.next()` → SPA for real browsers (the SSR path's own sharing check already did this pre-PR). This PR only changes behavior for the *success* case. `PublicSavedArtistsPage.tsx` renders a proper "Not found" state for both the private and nonexistent cases, matching prior behavior.
- **State transitions / API contract — confirmed, no issue.** `/api/public/saved-artists/:handle` contract is untouched.
- **Off-by-one / null handling — confirmed, no issue.** Nothing in the diff touches parsing or indexing logic.
- **Branch is 2 commits behind `main`** (missing #370 releases/admin-catalog work and #372's runtime-variable fix) — plausible source of merge noise, not a functional regression since the touched file is untouched by those commits. Worth rebasing before merge as hygiene, not a blocker.

Net: this looks like a correct, minimally-scoped follow-up to #369 that closes the same bifurcation gap for `/u/:handle`. No confirmed or plausible security/regression issues found in the diff itself.

---

brandonlucasgreen (commented):

**Security**
- Authorization/data exposure: the privacy check (`usernameRow.saved_artists_public`) is still enforced independently on both paths after this change — the edge function's own Supabase query for crawlers, and `api/functions/public-saved-artists.ts` for real browsers via `PublicSavedArtistsPage.tsx`. Both 404 on a private or unknown handle without distinguishing the two (no existence leak). **confirmed** no privacy regression — the gate that matters (sharing flag) isn't bypassed by routing more traffic to the SPA path.
- Secret handling: `SUPABASE_SERVICE_KEY` usage is unchanged and still confined to the crawler branch, which now runs for a smaller slice of traffic (only requests classified as crawlers) — reduces, not increases, exposure surface. No issue.
- Input validation: `handle` extraction and the new `userAgent` read (`request.headers.get('user-agent')`) are both defensively nullable-typed and passed straight into pattern-matching helpers (`isSocialCrawler`/`isIndexingCrawler`) that lowercase and `.includes()`-check against fixed lists — no injection surface. No issue.
- Unsafe defaults: on a missing/unrecognized user-agent, `isSocialCrawler`/`isIndexingCrawler` both return `false`, so the request falls through to `context.next()` (SPA path) rather than the static HTML path. That's fail-open toward the *already-privacy-checked* SPA route, not fail-open toward data exposure — same default behavior as the existing `artist-page-static.ts` this mirrors. No issue.

**Regression**
- This is a line-for-line port of the crawler-gate pattern already shipped in `api/edge/artist-page-static.ts` (same `isSocialCrawler`/`isIndexingCrawler` calls, same `context.next()` fallback, same Cache-Control/Netlify-CDN-Cache-Control headers). **confirmed** consistent with the established, previously-reviewed reference implementation rather than a novel code path.
- State transition: sharing-disabled and unknown-handle cases both still resolve through `context.next()` before this change's new early-return, so their behavior (SPA renders 404 via the same public API 404) is unaffected. No drift found.
- **needs-manual-verification**: neither this function nor `artist-page-static.ts` sets a `Vary: User-Agent` header on the static HTML response. If a shared/CDN cache in front of the edge function keys purely on URL, a cached crawler response could theoretically be served to a real browser (or vice versa) at the same path within the `s-maxage=300` window. This is pre-existing behavior inherited from the reference pattern, not introduced by this diff, so it's not a regression from this PR specifically — flagging because it's a real cache-key gap I can't verify without checking actual CDN behavior in front of Netlify Edge Functions.
- No test file accompanies this change, and there wasn't one for `u-handle.ts` before either — pre-existing gap, not a new regression, but worth noting since the sibling `artist-page-static.ts` crawler-gate presumably has (or should have) test coverage to compare against.

No blocking issues. LGTM.

---

PR #374, opened 2026-08-01
Navigate to release pages instead of client-side routing to them

brandonlucasgreen:
Clicking a release from an artist page rendered a **blank page**. A hard refresh loaded it, and **Back was dead** afterwards.

## Cause

`ReleasesSection` linked with react-router's `<Link>`, but `/a/{artist}/{release}` is served by the `release-page` edge function **and by nothing else**:

```
main.tsx:  <Route path="/a/:slug" element={<ArtistPage />} />   ← one segment
           (no catch-all <Route>)
```

So a client-side navigation to a two-segment path matched no route and rendered nothing. That single cause explains all three symptoms:

| Symptom | Why |
|---|---|
| Blank page on click | React Router matched nothing, and there's no catch-all to render a 404 |
| Hard refresh works | Only a real request reaches the edge function |
| Back button dead | The URL had already changed while React rendered nothing, so history had nowhere sensible to return to |

## Fix

A plain `<a>`. A real navigation reaches the one renderer that exists, and it's the honest expression of what that URL is — it also means an in-app click and a pasted link produce the same page, the property #369 established for artist pages.

**This is the UNS-100 bifurcation class wearing a different hat.** The familiar shape is two renderers for one URL; this was one renderer plus a link asserting a second. Worth stating plainly, because it isn't obvious from the retro: **a `<Link>` is an assertion that the SPA can render the target.**

## Making it catchable

The test file now renders **without a Router**. `<Link>` can't mount outside one, so that harness is the assertion. Verified by reverting the fix — 9 of the 10 tests fail, rather than passing quietly:

```
with <Link> restored:  Tests  9 failed | 1 passed (10)
with the fix:          Tests  10 passed (10)
```

## Scope

Every other `to={/a/...}` in the app is a single-segment artist link, which *does* have an SPA route. Checked all eight; they stay as they are.

## Related, not fixed here

**There is no catch-all `<Route>`**, so *any* unmatched URL renders a blank page rather than a 404 — a typo'd link, an old bookmark, anything. That's pre-existing and unrelated to Releases, and what a 404 should look like is a product decision, so I've left it. Worth a follow-up: it's the reason this bug presented as "blank" instead of "not found", which cost more diagnosis than it should have.

25 API + 35 web test files pass, lint at baseline, build green.

### DIFF
--- a/apps/web/src/components/ReleasesSection.tsx
+++ b/apps/web/src/components/ReleasesSection.tsx
@@ -1,4 +1,3 @@
-import { Link } from 'react-router-dom';
 import { cheapestOfferSummary, formatReleaseDate } from '../../../../api/shared/release-display';
 import type { ArtistPagePayload } from '../types/artist-page';
 
@@ -17,6 +16,13 @@ type Release = NonNullable<ArtistPagePayload['releases']>[number];
  *
  * The formatting helpers come from `api/shared/release-display.ts`, which the crawler-side edge
  * renderer also uses, so the two can't drift on what a price or a partial date looks like.
+ *
+ * **Rows are plain `<a>`, never react-router `<Link>`.** `/a/{artist}/{release}` is rendered by
+ * the `release-page` edge function and by nothing else — the SPA has no route for a two-segment
+ * `/a/` path, and there is no catch-all `<Route>`, so a client-side navigation there matches
+ * nothing and renders a blank page. It also strands history: the URL changes, React renders
+ * nothing, and Back has nowhere sensible to go. A real navigation reaches the one renderer that
+ * exists, which is also what makes an in-app click and a pasted link produce the same page.
  */
 export function ReleasesSection({
   releases,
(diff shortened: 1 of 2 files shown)


brandonlucasgreen (commented):

**Security**
- Input handling: `artistSlug` and `release.slug` are still passed through `encodeURIComponent` before being interpolated into the `href` — no change in escaping behavior from the previous `<Link to=...>`. **confirmed** no injection/XSS surface introduced.
- No auth/authorization, secret handling, or data-exposure surfaces touched — this is a pure client-side link-element swap. Nothing to flag.

**Regression**
- Swapping `<Link>` for `<a>` correctly forces a real navigation, matching the fact that `/a/{artist}/{release}` has exactly one renderer (the `release-page` edge function) and no SPA route — consistent with the "one route, one renderer" principle already established for `/a/:slug` and `/u/:handle`. **confirmed** fix matches root cause.
- Test regression coverage is real, not cosmetic: `show()` now renders `ReleasesSection` with no `<MemoryRouter>`, so if `<Link>` were reintroduced the suite would fail on mount (`useHref`/`useContext` invariant) rather than merely on an assertion. Verified this by reading the diff — the existing `links each release to its buying guide` test also still asserts the exact `href` value via `getByRole('link')`. **confirmed** good regression lock.
- Minor/needs-manual-verification: this changes release links from client-side transitions to full page loads for every artist-page → release-page click. That's the intended fix, but worth confirming in a deploy preview that the release page's own loading state/perf feels acceptable on a full reload (no code issue, just a UX check I can't verify statically).

No blocking issues. LGTM.

---

PR #376, opened 2026-08-01
Releases: Discogs + MusicBrainz + Faircamp sources, dedup review queues, and artist curation

brandonlucasgreen:
Picks up the Releases feature at spec steps 7–10, plus §11 curation. Bandcamp-only
release pages read as an undifferentiated Bandcamp mirror, so this adds sources, the
dedup logic to combine them safely, and the human surfaces to correct what automation
gets wrong.

## What a fan sees

The artist-page release list now shows **which platforms** a release is on and **what it
costs, with the payout estimate** — previously it showed a bare "cheapest price" with no
attribution, so a second source had nothing to show for itself until you clicked through.

The price shown is the **leading artist-paying source, not the globally cheapest**. Once
Discogs is in the mix, "cheapest" could rank a $3 secondhand listing above a $25 direct
purchase — exactly the off-mission ordering the sourcing spec warns about.

## Sources

- **Discogs** — two-tier like Bandcamp: a cheap `role=Main`/`type=master` listing, then a
  budgeted detail pass for price/format/availability. Its marketplace stats describe the
  whole release rather than per format, so only one offer is emitted; zero listings reports
  as `unknown`, never `sold_out`.
- **MusicBrainz** — enrichment only. Fills in release-group MBIDs and partial dates on
  releases that already exist; never creates one, since it has no purchase link to attach.
- **Faircamp** — verified against a real live instance. Its homepage doubles as the release
  list, and it has no reliable date or price anywhere in its markup (no JSON-LD, no `<time>`,
  no RSS `pubDate`), so this ingest honestly produces identity and artwork only.
- **Discovered links** — some platforms can't be fetched at all, but an artist's other pages
  sometimes link straight to a specific release there. This reads those out of markup already
  fetched for another reason and attaches them only on an **exact** title match.

**Mirlo is deliberately absent.** Its API is the best of any source — full discography in one
request, with prices — but `api.mirlo.space/robots.txt` disallows `/v1/` for every user agent
under a `# Disallow crawling the API` comment. Raising it with Mirlo directly instead.
**Subvert** is a hard wall: even a known-good artist URL returns HTTP 429 from a bot
checkpoint, for any path. **Bandwagon** sets `Crawl-delay: 1000` seconds, and its discography
appears to be client-rendered anyway.

## Dedup, and the two review surfaces

Three tiers, using identity columns the step-1 schema already had but nothing populated:

1. Hard identifiers (`discogs_master_id`, `musicbrainz_release_group_id`)
2. Exact normalized title within release type
3. **Fuzzy — never merges.** Flags both sides `needs_review` and records which release
   triggered it, via a new nullable self-reference.

Tier 3 needs a human, so this adds both:

- **`/admin/release-review`** — flagged pairs, deduplicated so a symmetric flag is one
  decision rather than two. Dismiss, or merge keeping either side.
- **`/artist-edit/:slug/releases`** — the artist's own version, plus hide, fix
  title/date/artwork, add a missing platform link, and add a release ingest never found.

Merges move `release_sources` onto the survivor **before** deleting the duplicate, and only
delete once that has succeeded — this project has lost data to the reverse order before
(#350). A same-platform conflict refuses the merge rather than silently dropping a source.

## "Scan my links for releases" — admin-only for now

The self-serve trigger, on the artist's curation page. **Ownership authorizes it; the admin
check is a separate rollout gate** — one function, two call sites, and a comment saying that
opening it to all verified artists means deleting exactly that, with no change to the
ownership path. Ownership is still enforced on top, so an admin can't catalog someone else's
profile through this endpoint.

Ready to test in production behind that gate.

## Notes for review

- **Ingest can no longer clobber a curated edit.** Sources an artist added or corrected are
  written `source: 'claimed'`, and every ingest path checks that before touching a source's
  URL — the same rule `curated_fields` already enforced on the release row, extended to the
  sources table. Without it a re-crawl would have silently reverted an artist's fix in 30 days.
- **Catalog state is three-valued end to end.** "We couldn't ask" never renders as a confident
  "Never catalogued".
- One migration: `releases.flagged_against_release_id` (nullable, `ON DELETE SET NULL`).
- New env requirement: none beyond what cataloging already needs.

491 API tests, 523 web unit tests, clean typecheck. No new lint findings.

### DIFF
--- a/api/edge/artist-page-static.ts
+++ b/api/edge/artist-page-static.ts
@@ -3,7 +3,7 @@ import { createClient } from "https://esm.sh/@supabase/supabase-js@2";
 import { PLATFORMS } from "../shared/platform-registry.ts";
 import { isBandcampFriday } from "../shared/bandcamp-friday.ts";
 import { mainLinkDividerIndexes } from "../shared/link-dividers.ts";
-import { cheapestOfferSummary, formatReleaseDate } from "../shared/release-display.ts";
+import { leadingOfferSummary, orderedSourcePlatforms, formatReleaseDate } from "../shared/release-display.ts";
 import { isSocialCrawler, isIndexingCrawler } from "../shared/crawler-detection.ts";
 
 /**
@@ -23,7 +23,10 @@ interface ReleaseRow {
   date_precision: string | null;
   status: string;
   artwork_url: string | null;
-  release_sources: { release_offers: { price: number | null; currency: string | null; availability: string }[] | null }[] | null;
+  release_sources: {
+    platform: string;
+    release_offers: { price: number | null; currency: string | null; availability: string }[] | null;
+  }[] | null;
 }
 
 // Allowed embed domains for featured releases (must match api/functions/artist-profile.ts)
@@ -80,8 +83,19 @@ function renderReleaseRow(release: ReleaseRow, artistSlug: string): string {
     ? ''
     : release.release_type.charAt(0).toUpperCase() + release.release_type.slice(1);
 
-  const offers = (release.release_sources || []).flatMap(s => s.release_offers || []);
-  const meta = [type, date, cheapestOfferSummary(offers)].filter(Boolean).join(' · ');
+  const sources = (release.release_sources || []).map(s => ({ platform: s.platform, offers: s.release_offers || [] }));
+  // leadingOfferSummary, not the globally cheapest price across every source — see the
+  // function's own doc for why picking the absolute cheapest could rank a Discogs secondhand
+  // listing above a Bandcamp direct purchase.
+  const meta = [type, date, leadingOfferSummary(sources)].filter(Boolean).join(' · ');
+
+  // Same artist-paying-first order as the summary above and the release page itself.
+  const platformsHtml = orderedSourcePlatforms(sources)
+    .map(p => {
+      const info = PLATFORM_INFO[p];
+      return `<span title="${escapeHtml(info?.name || p)}" style="font-size:12px">${info?.icon || '🔗'}</span>`;
+    })
+    .join('');
 
   const art = release.artwork_url
     ? `<img src="${escapeHtml(release.artwork_url)}" alt="" loading="lazy" referrerpolicy="no-referrer" style="width:48px;height:48px;border-radius:6px;object-fit:cover;flex-shrink:0;background:var(--bg2)">`
@@ -98,6 +112,7 @@ function renderReleaseRow(release: ReleaseRow, artistSlug: string): string {
     <span style="flex:1;min-width:0">
       <span style="display:block;font-weight:500;overflow:hidden;text-overflow:ellipsis;white-space:nowrap">${title}</span>
       ${meta ? `<span style="display:block;font-size:12px;color:var(--muted)">${escapeHtml(meta)}</span>` : ''}
+      ${platformsHtml ? `<span style="display:flex;gap:4px;margin-top:2px">${platformsHtml}</span>` : ''}
     </span>
     ${upcoming}
   </a>`;
@@ -251,7 +266,7 @@ export default async function handler(request: Request, context: Context) {
         .from('releases')
         .select(
           'slug, title, release_type, release_date, date_precision, status, artwork_url,' +
-          ' release_sources ( release_offers ( price, currency, availability ) )',
+          ' release_sources ( platform, release_offers ( price, currency, availability ) )',
           { count: 'exact' }
         )
         .eq('artist_id', artist.id)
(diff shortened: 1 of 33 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Automated staff-eng review of this PR (security + regression pass). Scope: `api/functions/{admin-release-review,artist-releases,db,catalog-artist-background,release-ingest,request-catalog,middleware}.ts`, the new migration, and the frontend routes/pages.

## Security

**1. `mergeReleases()` (db.ts) doesn't verify both releases belong to the same artist — plausible**
`api/functions/db.ts` — `mergeReleases(keepId, dropId)`, called from `admin-release-review.ts`.

The artist-facing merge path (`artist-releases.ts`) correctly calls `verifyReleaseOwnership(artistId, [keepId, dropId])` before merging, explicitly to stop an artist merging in someone else's release by guessing a UUID (good — nice comment calling that out). But the *admin* endpoint (`admin-release-review.ts`) calls `mergeReleases(keepId, dropId)` directly with no equivalent check that the two ids share an `artist_id`. `mergeReleases` itself never checks either. In practice the admin UI only ever submits pairs sourced from the review queue (which is always same-artist, since tier-3 fuzzy matching is scoped per-artist), so today's UI can't trigger it — but the API has no server-side invariant preventing two arbitrary release UUIDs from different artists being merged (moving one artist's `release_sources` onto another artist's release, then deleting the original). Given `mergeReleases` is also exported and reused, and the merge is genuinely hard to reverse (source rows moved, row deleted), it seems worth an explicit same-artist assertion inside `mergeReleases` itself rather than relying on "no caller currently sends a mismatched pair."

**2. `artist-releases.ts` uses hardcoded wildcard CORS instead of `buildCorsHeaders` — informational, matches existing pattern**
`api/functions/artist-releases.ts:193-198` sets `Access-Control-Allow-Origin: '*'` by hand rather than using `middleware.ts`'s `buildCorsHeaders`/`buildPublicCorsHeaders`, which every admin endpoint in this PR (and most of `api/functions/`) uses instead. CLAUDE.md calls this out directly: "Use the shared middleware rather than hand-rolling headers." That said, this exactly mirrors `artist-profile.ts`, which the PR explicitly cites as the pattern being followed, and that file has the same wildcard header already — so this isn't a new hole, just an existing one extended to a new surface with write actions (hide/merge/update/create). Not blocking, but worth a follow-up to migrate both to the shared helper.

**3. `addLink`/`create` accept any `platform` string and any http(s) URL with no domain allowlist — needs-manual-verification, matches existing pattern**
`api/functions/artist-releases.ts` — `isHttpUrl()` is explicitly commented as "same rule as artist-profile.ts's link validation: only a protocol check, no domain allowlist." A verified artist can attach an arbitrary URL under an arbitrary platform label to their public release list. This is consistent with how `artist-profile.ts` already handles bio links, so it's a known, accepted risk rather than new — flagging only because it's now extended to release rows too, in case that widening was worth a second look given release rows are more prominently surfaced (artist page release list) than a generic profile link.

## Regression / correctness

**4. `update` action doesn't validate `title`/`releaseDate` types before use — plausible**
`api/functions/artist-releases.ts` (`update` action) passes `body.title`/`body.releaseDate` straight through to `updateArtistReleaseFields` with no runtime type check (the TS types are compile-time only — a client can send `"title": 123` or `"title": {}`). Inside `db.ts`, `updateArtistReleaseFields` calls `patch.title.trim()` unconditionally when `patch.title !== undefined`. A non-string `title` throws synchronously inside that async function; the handler has no surrounding try/catch, so the rejection propagates as an unhandled function crash (bare 500, no clean CORS-headed JSON error) instead of the 400 every other malformed-body case in this file returns. Low severity (self-inflicted 500, no data exposure), but worth a `typeof title === 'string'` guard for consistency with the rest of the validation in this handler.

**5. `flagged_against_release_id` can point at the losing side of a merge — plausible**
When `mergeReleases(keepId, dropId)` runs, `dropId` is deleted and the migration's `ON DELETE SET NULL` correctly nulls out any *other* release's `flagged_against_release_id` that pointed at `dropId`. But `dismissReleaseReview`'s "clear both sides" logic isn't reused here — `mergeReleases` only clears `keepId`'s own flag at the end (`.eq('id', keepId)`), which is right since `dropId` is gone. This one looks fine on inspection, no bug — noting it because it's the same code shape as #1 (asymmetric handling between the two "resolve a flagged pair" functions) and worth a test asserting a third release flagged against `dropId` (not part of the merged pair) ends up correctly `null`'d rather than dangling. If `release-review.test.ts` already covers this, disregard.

Everything else — the three-valued catalog state, the `claimed`-source clobber guard applied uniformly across Bandcamp/Discogs/Faircamp/discovered-link ingest paths, the SSRF handling (`safeFetch` used everywhere a URL isn't hardcoded to a fixed API host; the two `api.discogs.com` raw-`fetch` call sites build the URL from a regex-extracted numeric id only, matching the existing `enrichment.ts` precedent), and the ownership checks on the artist-facing endpoints — held up well under review.

---

PR #377, opened 2026-08-01
Fix release catalog prices and Faircamp artist-page ingest

brandonlucasgreen:
Four bugs found on a real catalog run against Kid Lightbulbs.

## Bandcamp quoted the subscription fee as the album price

An album page's JSON-LD `albumRelease` list isn't only that album's packages — Bandcamp parks two whole-catalog products in it, both typed `DigitalFormat`/`Digital` and indistinguishable from the album's own download except by `item_type`:

| item_type | what it is | price on INFINITE NORMAL |
|---|---|---|
| `a` | the album | $0 (name your price) |
| `i` | **monthly subscription** | $3.33/month |
| `b` | **full digital discography** | $23.20 |

Offers collapse to the cheapest per format, so on any album priced above $3.33 the subscription won. Both cross-catalog products are now dropped. A package with no `item_type` is kept — the known intruders both carry one, and refusing everything unlabelled would silently empty the offers on any page whose markup drifts.

## Faircamp catalogued artists as releases

On a multi-artist Faircamp site, each release block carries its credited artists as relative links of exactly the same shape as the release links beside them:

```html
<div class="release">
  <a href="infinite-normal/">INFINITE NORMAL</a>
  <div class="release_artists"><a href="kl/">Kid Lightbulbs</a></div>
</div>
```

The ingest read every bare relative link, so `kl/` and `blg/` became releases. Now restricted to links inside a release block, excluding the artist-credit div; a page with no release blocks still falls back to the old scan, minus credits. Verified against the live homepage: 23 releases, no artists.

Nothing on the linked page distinguishes them afterwards — an artist page and a release page both publish an `og:title`, and neither sets a useful `og:type` — so it has to be caught at the homepage.

## Faircamp releases never got a price

Faircamp publishes no price on a release page at all: no JSON-LD, no microdata, nothing. It lives one level down on `{release}/purchase/{token}/`. That page is now fetched and read for a digital offer, budgeted and refreshed on the same 30-day rule as Bandcamp's detail pass.

Matched against Faircamp's own generator (`renderer/src/pages/multitrack_purchase.rs`) rather than guessed, so both shapes parse: a price input with `data-min` (name-your-price, "up to X", "$5–20") and a fixed price rendered as localized text. Unreadable means no offer, never a price of zero.

The empty state's copy changes from "Still gathering formats and prices." to "Price information not available." — for a release whose purchase page can't be read, nothing is coming, and the old wording was a promise we don't keep.

## Artist pages paginate releases two-up

Ten per page in two columns, collapsing to one column under 640px, paged over the payload already fetched so turning a page costs no request. The payload cap rises 24 → 60 (six pages) since it now bounds the payload rather than what a fan can reach.

## Subvert: investigated, not fixable here

`subvert.fm` sits behind a Vercel bot wall that 429s everything, including `robots.txt`, so Unstream never fetches it. Subvert links enter the catalog only when they appear in markup fetched for another reason, and `kidlightbulbs.com` links exactly one release URL. The other three albums aren't linked from any page we fetch, so there is nothing to find. No code change.

## Existing bad rows

The parser fix stops new ones; it doesn't delete the artist-as-release rows already written. Those can be hidden from the artist's own release curation UI.

## Testing

- `npm run build` green (lint, `test:api` 518, `test:unit` 523).
- 7 new tests in `release-ingest.test.ts`. The subscription test was verified to fail with the filter neutralized.
- Both Faircamp parsers run against the live pages: the homepage yields 23 releases and no artists; INFINITE NORMAL's purchase page yields `{digital, price 0, USD, available}`.
- Pagination verified in the browser at 1280px (two columns) and 375px (one).

### DIFF
--- a/api/edge/release-page.ts
+++ b/api/edge/release-page.ts
@@ -248,13 +248,15 @@ export default async function handler(request: Request, context: Context) {
         </div>`;
       }).join('');
 
-      // An honest empty state. "We know it's here, we haven't read the prices yet" is a
-      // different statement from "there is nothing to buy", and a fan can act on the first.
+      // An honest empty state. Both branches say the same thing to a fan — we have no price to
+      // show you — and neither says "this is free" or "there is nothing to buy". The wording no
+      // longer promises that a price is on its way: a Faircamp release whose purchase page we
+      // can't read has no price coming, and "still gathering" would be a promise we don't keep.
       const pendingHtml = offers.length === 0
         ? `<div class="offer"><span style="color:var(--muted)">${
             source.detail_checked_at
               ? 'No formats listed on this page.'
-              : 'Still gathering formats and prices.'
+              : 'Price information not available.'
           }</span></div>`
         : '';
 
--- a/api/functions/__tests__/release-ingest.test.ts
+++ b/api/functions/__tests__/release-ingest.test.ts
@@ -17,6 +17,7 @@ import {
   mapDiscogsFormatName,
   ingestMusicBrainzReleaseGroups,
   ingestFaircampHomeLinks,
+  ingestFaircampPurchasePage,
   ingestFaircampReleasePage,
   buildFaircampRelease,
   findDiscoveredReleaseLinks,
@@ -189,6 +190,8 @@ function detailPage(opts: {
   packages?: Array<{
     format?: string;
     typeName?: string;
+    /** Bandcamp's own product kind: 'a' album, 'p' package, 'b' discography, 'i' subscription. */
+    itemType?: string;
     price?: number | string;
     currency?: string;
     availability?: string;
@@ -201,7 +204,10 @@ function detailPage(opts: {
     graph.albumRelease = opts.packages.map(p => ({
       '@type': ['MusicRelease', 'Product'],
       musicReleaseFormat: p.format,
-      additionalProperty: [{ '@type': 'PropertyValue', name: 'type_name', value: p.typeName }],
+      additionalProperty: [
+        { '@type': 'PropertyValue', name: 'type_name', value: p.typeName },
+        ...(p.itemType ? [{ '@type': 'PropertyValue', name: 'item_type', value: p.itemType }] : []),
+      ],
       offers: {
         '@type': 'Offer',
         price: p.price,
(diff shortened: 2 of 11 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen — automated security + regression review of this PR.

## Security

1. **SSRF surface — checked, clean.** `catalogFaircampPrices`'s new purchase-page fetch (`api/functions/catalog-artist-background.ts`) goes through `safeFetch`, which re-validates hostname + DNS resolution + private-IP rejection on every redirect hop — not just `resolveSameHost`'s same-host string check, which is only a confinement layer on top. Confirmed by reading `safe-fetch.ts`. No new SSRF exposure. *(needs-manual-verification → resolved clean)*

2. **Unscoped price regex — plausible.** `ingestFaircampPurchasePage`'s fixed-price fallback (`api/functions/release-ingest.ts`) matches `/([0-9]+(?:[.,][0-9]+)?)\s+([A-Z]{3})\b/` against the *entire* stripped purchase-page text rather than a scoped price container. Any other "number + 3 uppercase letters" sequence elsewhere on that page (future Faircamp copy, a track count, a format note) would be silently read as the price instead of producing the intended `null`, which is exactly the "never cache uncertainty" failure mode this codebase is otherwise careful about. Low likelihood against today's known Faircamp markup, but worth scoping to `#confirm_price` if there's a cheap way to do it. **plausible**

No auth/authorization, secret-handling, or injection issues found — no new endpoints or auth paths in this diff, and no interpolation into raw queries.

## Regression

1. **Confirmed — empty-state message regressed for every non-Faircamp platform.** In `api/edge/release-page.ts`, the pending-offer message changed from `"Still gathering formats and prices."` to `"Price information not available."` whenever `source.detail_checked_at` is null. That branch runs for *every* release source, not only Faircamp. For Bandcamp (or any platform with a detail pass), `detail_checked_at` is genuinely null during the ordinary window between a release being cataloged and its detail pass completing — a window that can span multiple background-function invocations once `MAX_DETAIL_FETCHES_PER_RUN` (shared across a 25-artist batch) is exhausted by other artists in the same run. During that window this now tells a fan the price is "not available" when it's actually just pending — the opposite of the PR's own stated rationale, which is Faircamp-specific ("a Faircamp release whose purchase page can't be read has no price coming"). Suggest gating the new copy on the source's platform (or another signal that distinguishes "will never have a price" from "hasn't been checked yet") rather than applying it uniformly. **confirmed**

2. **Needs-manual-verification — claimed Faircamp sources can silently never get a price.** `catalogFaircampPrices` (`api/functions/catalog-artist-background.ts`) keys its `purchaseUrls` map by the just-scraped release URL (`release.externalUrl`), then looks it up later via `written[i].url`, which is `source.url` as returned by `upsertReleaseSource`. For a *claimed* source (`release_sources.source === 'claimed'`), `upsertReleaseSource` returns the existing DB row untouched rather than the freshly-scraped URL (`db.ts`, `upsertReleaseSource`/`getClaimedSourceKeys`) — so if an artist's claimed Faircamp URL ever differs from what's freshly scraped, the map lookup misses and that release's price is never fetched, with no error or log distinguishing it from "no purchase page found." I can't tell from this diff whether Faircamp release sources are actually editable via the claim/edit flow today — worth a quick check, since if they are, this is a real and silent gap. **needs-manual-verification**

No off-by-one or auth-adjacent state issues found elsewhere; the `ReleasesSection` pagination state resets correctly on artist navigation (the parent unmounts it via the `isLoading` branch on slug change), and the item-type filter / Faircamp artist-credit filter both match their new tests exactly.

## Verification performed

- Checked out the branch, ran `npm install`, `npm run test:api` (518 passed), the new `release-ingest.test.ts` (68 passed), and `ReleasesSection.test.tsx` (12 passed) — all green, matching the PR description.
- Ran `tsc --noEmit` over all of `api/**/*.ts` (not just the narrow `api/tsconfig.json` include, given this repo's known typecheck-coverage gap) to check for the "compiles but is wrong" trap. Errors surfaced only in files this PR doesn't touch (`search-sources.ts`, `saved-artists.ts`, etc.); nothing new in `release-ingest.ts`, `catalog-artist-background.ts`, `db.ts`, `search-parsers.ts`, or `artist-page.ts`.
- The `semantic-revert-check` bot flagged this PR as a possible revert of #376/#370/#360/#362. Checked it and it's a false positive: the flagged deletions are lines modified in place (the `status` type widening, the Faircamp link-scan rewrite, the release-cap bump), not features being removed — the Discogs/MusicBrainz/Faircamp functionality from #376 is present and its tests pass on this branch.

---

brandonlucasgreen (commented):
@brandonlucasgreen — follow-up review covering the second commit (`9c828fa`, "Make release-link curation reachable and its failures visible"), pushed after my first pass on `75932fd`. The findings from that earlier review still stand and aren't addressed by this commit — this is purely additive coverage of the new changes.

## Security

- **`normalizeLinkUrl` (client) + `isHttpUrl` (server)** — the client's scheme-guessing (`subvert.fm/...` → `https://subvert.fm/...`) only affects convenience; the server still independently requires `http(s)` via `isHttpUrl`, so a direct API call bypassing the client gets the same enforcement. `javascript:`/`data:` schemes are rejected either way. **checked, clean**
- **Stored release-link URLs render escaped.** Traced where a claimed link's URL ends up on the SSR'd release page (`api/edge/release-page.ts:271`) — it goes through `escapeHtml(source.url)` before landing in an `href`, so a URL containing quote/angle-bracket characters can't break out of the attribute. This isn't new in this commit, but it's the render path this commit is making more reachable (Dashboard link + reordered panel), so worth confirming it's covered — it is. **checked, clean**
- **`external_id: null` → omitted (`db.ts`)** — verified this is a real fix, not a regression: Postgres `ON CONFLICT DO UPDATE` only sets columns present in the upsert payload, so omitting `external_id` correctly preserves whatever ingest already stored on update and defaults to the column's `NULL` on insert (confirmed nullable, no default, in `20260731120000_releases.sql`). No authorization change — `addArtistReleaseLink` still calls `verifyReleaseOwnership` first, unchanged. **confirmed correct**

## Regression

- Dashboard's new "Releases" link (`/artist-edit/:slug/releases`) points at a route that already exists in `main.tsx` — not a dead link. **checked, clean**
- No state/off-by-one issues in the new `linkStatus` inline-error handling or the "keep typed URL on failure" change — both are exercised by the two new tests and match the described behavior.

## Verification performed

- Checked out `9c828fa`, ran the two new tests in `ArtistReleasesPage.test.tsx` (16 passed in that file), the full `npm run test:api` (518 passed), and `npx tsc -b` for `apps/web` (clean) to cover the files this commit touches that fall under the web tree's typecheck.

No new confirmed issues on top of the first review.

---

PR #378, opened 2026-08-01
Make a requested re-catalogue actually re-read prices

brandonlucasgreen:
Re-cataloguing Kid Lightbulbs after #377 changed nothing visible, and the crawl reported success.

## Why

Clearing the artist's cooldown lets the run happen. But dates, formats and prices come **only** from individual release pages, and both detail passes skip a source that has already been read:

```js
// catalog-artist-background.ts
function needsDetail(release) {
  if (!release.detailCheckedAt) return true;
  return Date.now() - new Date(release.detailCheckedAt).getTime() > 30 * 24 * 3600_000;
}
```

Discogs is stricter still — `if (release.detailCheckedAt) continue;`, never re-priced.

The earlier catalogue stamped every Bandcamp source, so the re-run re-read the release *list*, skipped every release page, and left every wrong price in place. The `item_type` fix from #377 was deployed and correct; it was simply never asked to parse anything.

That refresh rule is right for the demand-driven crawl it was written for — re-reading every page weekly would triple the request count for data that rarely moves. It's wrong for someone who has just been told their prices are wrong and pressed the button.

## The fix

Both explicit triggers — the admin button and the artist's own "Scan my links" — now also reset that artist's sources to unpriced (`clearReleaseDetailCooldown`). Null, not backdated: "never read" is the honest state for a price we've decided not to trust, and it's what both passes already test for.

**Per-artist detail budget 20 → 40**, as a direct consequence. With every source reset, an artist with more than 20 releases could never get their whole catalogue priced — each press re-read the same newest 20 and left the tail permanently unread. 40 covers every catalogue measured (22 Kid Lightbulbs, 16 Sufjan Stevens, 33 for the largest live Mirlo artist) and costs 40 paced seconds for one artist. The invocation-wide cap of 100 still bounds a whole batch.

## Testing

`npm run build` green (518 API, 525 web). Both trigger tests assert the reset and **both fail without it** — verified by neutralizing the call.

## What this does not fix

The Faircamp rows that were catalogued as releases (`Kid Lightbulbs`, `Brandon Lucas Green`) are still in the database. #377 stops new ones; it doesn't delete existing rows. Hide them from Manage Releases, or say the word for a one-off cleanup.

### DIFF
--- a/api/functions/artist-releases.ts
+++ b/api/functions/artist-releases.ts
@@ -23,6 +23,7 @@ import {
   mergeReleases,
   getCatalogState,
   clearCatalogCooldown,
+  clearReleaseDetailCooldown,
 } from './db';
 import { authenticateAdmin, authenticateBearer, buildCorsHeaders } from './middleware';
 import { cacheDeleteByArtist } from './cache';
@@ -219,7 +220,10 @@ export async function handler(event: {
       };
     }
 
-    await clearCatalogCooldown(artistId);
+    // Both, always: the cooldown lets the crawl run, and the detail reset is what makes it
+  // re-read prices rather than only re-confirming the release list. See db.ts.
+  await clearCatalogCooldown(artistId);
+  await clearReleaseDetailCooldown(artistId);
     const queued = await triggerCatalogNow(artistId);
     if (!queued.ok) {
       return { statusCode: queued.status, headers: CORS_HEADERS, body: JSON.stringify({ error: queued.error }) };
--- a/api/functions/db.ts
+++ b/api/functions/db.ts
@@ -578,6 +578,43 @@ export async function clearCatalogCooldown(artistId: string): Promise<void> {
   if (error) console.error('[DB] clearCatalogCooldown failed:', error.message);
 }
 
+/**
+ * Mark every one of an artist's sources as never priced, so a deliberately-requested crawl
+ * re-reads the release pages instead of only the grid.
+ *
+ * **Clearing the artist's cooldown alone is not enough**, and the gap is invisible: the crawl
+ * runs, reports success, and changes nothing an artist can see. Dates, formats and prices live
+ * only on individual release pages, and both detail passes skip a source that was already read —
+ * Bandcamp's within `DETAIL_REFRESH_DAYS`, Discogs' ever. That refresh rule is right for the
+ * demand-driven crawl it was written for, where re-reading every page weekly would triple the
+ * request count for data that rarely moves. It is wrong for someone standing in front of the
+ * button having just been told their prices are wrong: found once, in exactly that situation,
+ * when a parser fix for a bad Bandcamp price shipped and re-cataloguing didn't apply it.
+ *
+ * Null, not backdated: "never read" is the honest state for a source whose stored price we've
+ * decided not to trust, and it's what both passes already test for.
+ */
+export async function clearReleaseDetailCooldown(artistId: string): Promise<void> {
+  const client = getClient();
+  if (!client) return;
+
+  const { data, error: readError } = await client.from('releases').select('id').eq('artist_id', artistId);
+  if (readError) {
+    console.error('[DB] clearReleaseDetailCooldown read failed:', readError.message);
+    return;
+  }
+
+  const releaseIds = ((data as { id: string }[] | null) || []).map(r => r.id);
+  if (releaseIds.length === 0) return;
+
+  const { error } = await client
+    .from('release_sources')
+    .update({ detail_checked_at: null })
+    .in('release_id', releaseIds);
+
+  if (error) console.error('[DB] clearReleaseDetailCooldown failed:', error.message);
+}
+
 interface ReleaseToPersist {
   title: string;
   slug: string;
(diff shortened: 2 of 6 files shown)


api/functions/db.ts
```diff
@@ -578,6 +578,43 @@ export async function clearCatalogCooldown(artistId: string): Promise<void> {
...
+
+  const releaseIds = ((data as { id: string }[] | null) || []).map(r => r.id);
+  if (releaseIds.length === 0) return;
+
+  const { error } = await client
+    .from('release_sources')
+    .update({ detail_checked_at: null })
+    .in('release_id', releaseIds);
+
+  if (error) console.error('[DB] clearReleaseDetailCooldown failed:', error.message);
```
brandonlucasgreen, line 615: **Regression — plausible.** `clearReleaseDetailCooldown` (and the pre-existing `clearCatalogCooldown` above it) swallow a failed update: they `console.error` and return, with no signal to the caller. All three call sites (`admin-catalog-artist.ts`, `artist-releases.ts`) do:

```ts
await clearCatalogCooldown(artistId);
await clearReleaseDetailCooldown(artistId);
const queued = await triggerCatalogNow(artistId);
```

If the `release_sources` update on line 613-614 fails (RLS hiccup, transient PostgREST error, etc.) while `clearCatalogCooldown` above it succeeds, the code doesn't know — it proceeds straight to `triggerCatalogNow` and returns `202 { queued: true }`. The crawl then runs, but every source still has its old `detail_checked_at`, so both detail passes skip everything exactly as before. That's the precise bug this PR fixes ("the crawl reported success... and left every wrong price in place"), just reintroduced silently on a DB error instead of unconditionally.

Given this function exists specifically to close that failure mode, it might be worth having it return a boolean/result the caller can check (or at least Sentry-report the failure), rather than relying on best-effort logging — consistent with the "never cache uncertainty / fail loudly" pattern this codebase uses elsewhere for exactly this kind of silent-success gap.

---

api/functions/artist-releases.ts
```diff
@@ -219,7 +220,10 @@ export async function handler(event: {
       };
     }
 
-    await clearCatalogCooldown(artistId);
+    // Both, always: the cooldown lets the crawl run, and the detail reset is what makes it
+  // re-read prices rather than only re-confirming the release list. See db.ts.
+  await clearCatalogCooldown(artistId);
+  await clearReleaseDetailCooldown(artistId);
```
brandonlucasgreen, line 226: Nit — indentation drift, not a logic bug. Lines 224-226 are at 2 spaces instead of the 4 spaces the rest of this `if (action === 'catalog')` block uses:

```ts
    // Both, always: the cooldown lets the crawl run, and the detail reset is what makes it
  // re-read prices rather than only re-confirming the release list. See db.ts.
  await clearCatalogCooldown(artistId);
  await clearReleaseDetailCooldown(artistId);
    const queued = await triggerCatalogNow(artistId);
```

Functionally harmless (JS doesn't care), and `check` is green so nothing enforces it here — but worth a quick `prettier --write` pass before merge so it doesn't read as a merge artifact later.

---

brandonlucasgreen (commented):
@brandonlucasgreen Security + regression pass on this one. One open PR right now, this is it.

## Security

- **Auth/authorization** — no issues found. `admin-catalog-artist.ts` gates on `authenticateAdmin`; `artist-releases.ts`'s `catalog` action is gated on ownership (`resolveOwnedArtist`, established before the new `clearReleaseDetailCooldown` call) *and* the existing `canTriggerCatalog` admin rollout gate. Consistent with the rest of the file.
- **Input validation** — no issues found. `artistId` reaching `clearReleaseDetailCooldown` is already validated upstream (UUID regex on the admin path, ownership resolution on the self-serve path).
- **Secret handling** — not touched by this diff.
- **Injection risks** — no issues found. New query in `db.ts` uses the Supabase query builder (`.eq`, `.in`) throughout, no raw SQL / string interpolation.
- **Unsafe defaults** — *plausible, low severity.* `MAX_DETAIL_FETCHES_PER_ARTIST` doubles 20 → 40 (`catalog-artist-background.ts`), so one button press can now cost up to 40 paced Bandcamp/Discogs requests instead of 20. Well-reasoned in the comment and still bounded by the invocation-wide cap of 100, so I'd call this an accepted, documented trade-off rather than a defect — flagging only because it's a real change in third-party request volume per trigger.
- **Data exposure** — no issues found.

## Regression

- **State transitions — plausible** (inline comment on `db.ts:615`): `clearCatalogCooldown` and `clearReleaseDetailCooldown` both swallow their own failures (log + return, no signal to the caller). All three call sites fire them sequentially and proceed to `triggerCatalogNow` regardless of whether either succeeded. A failed `release_sources` update after a successful cooldown clear silently reproduces the exact bug this PR fixes — crawl runs, reports success, prices don't change — just gated on a DB hiccup instead of the missing call this PR adds. Worth a return value or a Sentry report given this function's entire purpose is closing that gap.
- **Edge cases / null-empty** — handled correctly: zero releases short-circuits before the `release_sources` update, avoiding an empty `.in()`.
- **Off-by-one** — none found; the 20→40 change is a plain constant bump, not an off-by-one.
- **API contract drift** — none. Response shapes for the `catalog` action (`202 {queued:true}`, existing 400/403/500 error paths) are unchanged.
- **Style nit** (inline comment on `artist-releases.ts:224-226`): indentation drift in the new lines inside the `catalog` block — cosmetic only, `check` is green so nothing's enforcing it, but worth a `prettier` pass before merge.

Nothing here blocks merging; the state-transition point is the one I'd actually want addressed before this ships, since it undermines the PR's own stated goal on failure.

---

PR #379, opened 2026-08-01
Show artist page support links as platform pills

brandonlucasgreen:
The releases section made the "Support directly" list feel oversized: eight platforms meant eight full-width rows before a fan reached anything else.

This renders them as the same `SourceBadge` pills the search results use — in the artist's own order, **not** grouped by category.

## Measured

Same 8-link fixture, `getBoundingClientRect()` on the section:

| | Height | Rows |
|---|---|---|
| Before | 542px | 8 |
| After | 210px | 4 |

Widest pill is 257px, so it fits a 375px phone column without horizontal overflow.

## Changes

- **`RichArtistProfile.tsx`** — stacked rows → `SourceBadge` in a wrapped flex. New pure helper `groupLinksByDivider` splits links at the divider positions the artist chose, so each group is its own wrapped row with a rule between rows. Dividers still mean what they meant.
- **`SourceBadge.tsx`** — optional `onClick`, fired alongside its own `trackPlatformClick`, so the artist page keeps `trackArtistLinkClick`. Same both-events pairing `ResultCardRelease` and `ResultCardSocial` already use.
- Tests updated, plus new coverage for the custom-link ("Link") fallback and for a leading divider not drawing a stray rule.

## Worth knowing

- Pills carry the payout badge and AI-policy chip from the platform registry, so the section shows *more* signal than the old rows did — those only showed a payout when the stored link row happened to have one.
- Bandcamp Friday: the pill's built-in `~97%` treatment covers it, and the pulsing "Bandcamp Friday!" label stays beside the pill. It now uses the `.bandcamp-friday-label` class that was already in `index.css` but unused.
- `data-track-platform` is gone from these links. Nothing outside the component tests read it.

## Deliberately unchanged

- `UnclaimedQuietCard` keeps the old rows — unclaimed pages are intentionally quieter.
- The `artist-page-static` edge function still emits the old rows. That render is crawler-only, so no human sees the mismatch.

Verified in the browser with a throwaway Vite harness (the dev API has no `/api/artist-page` mock). 527 unit tests pass; lint clean for the touched files.

### DIFF
--- a/apps/web/src/components/RichArtistProfile.tsx
+++ b/apps/web/src/components/RichArtistProfile.tsx
@@ -2,9 +2,10 @@ import { Fragment, useState, useCallback } from 'react';
 import type { ArtistPagePayload } from '../types/artist-page';
 import { sources } from '../services/sources';
 import { analytics } from '../services/analytics';
-import { PlatformIcon } from './PlatformIcon';
 import { ReleasesSection } from './ReleasesSection';
 import { SocialIcon } from './SocialIcon';
+import { SourceBadge } from './SourceBadge';
+import { getSource } from './ResultCardUtils';
 import type { SourceId } from '../types';
 
 interface RichArtistProfileProps {
@@ -17,6 +18,28 @@ interface RichArtistProfileProps {
   disabledSave?: boolean;
 }
 
+type ArtistLink = ArtistPagePayload['links'][number];
+
+/**
+ * Split the artist's links at the divider positions they chose, so each group can
+ * render as its own wrapped row of pills with a rule between rows.
+ */
+function groupLinksByDivider(links: ArtistLink[], dividerIndexes: Set<number>): ArtistLink[][] {
+  const groups: ArtistLink[][] = [];
+  let current: ArtistLink[] = [];
+
+  links.forEach((link, index) => {
+    if (dividerIndexes.has(index) && current.length > 0) {
+      groups.push(current);
+      current = [];
+    }
+    current.push(link);
+  });
+
+  if (current.length > 0) groups.push(current);
+  return groups;
+}
+
 function getLocationText(artist: ArtistPagePayload['artist']): string {
   const region = artist.country || artist.countryCode;
   if (artist.city && region) return `${artist.city}, ${region}`;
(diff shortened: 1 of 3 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Security + regression pass on this one.

## Security

Nothing to flag. The diff is a presentational refactor of already-trusted data (server-provided `link.url`/`link.platform` rendered the same way — plain `href`, React-escaped text — as before this change). No auth/authz, secrets, injection, or SSRF-relevant code is touched.

- **Injection risks — no finding.** `href={url}` and `displayName` were already rendered this way pre-PR; `getSource()` (`ResultCardUtils.ts`) safely falls back to `sources['other']` for unknown platform ids, and React escapes all text content. No `dangerouslySetInnerHTML` introduced.
- **Input validation / data exposure / unsafe defaults / auth — no finding.** No new inputs, endpoints, or trust boundaries in this diff.

## Regression

**[confirmed] Every real visitor will see a layout flash on `/a/:slug` and `/artist/:slug`, not just crawlers, contradicting the PR's own "Deliberately unchanged" note.**

The PR says:
> The `artist-page-static` edge function still emits the old rows. That render is crawler-only, so no human sees the mismatch.

That assumption doesn't hold given how the SPA mounts:

- `api/edge/artist-page-static.ts` (untouched by this PR, ~L467/L566) still server-renders the **old** full-width stacked "Support directly" rows.
- `apps/web/src/main.tsx:50` mounts with `createRoot(document.getElementById('root')!).render(...)`, **not** `hydrateRoot`. That means the client does a full replacement render rather than hydrating the existing DOM.

So on first paint every visitor (JS or not) sees the edge-rendered old-style rows, and once React mounts the DOM gets swapped to the new compact pill layout underneath them — a visible reflow/CLS jump on every page load of these routes, not a crawler-only cosmetic mismatch. This is the same "two renderers for one URL" pattern called out in `docs/retros/UNS-100-bifurcation-retro.md` and the "One route, one renderer" principle in CLAUDE.md. Worth updating `artist-page-static.ts` to emit the same pill markup (or at least matching row heights) before merging, rather than treating it as a follow-up.

**Everything else checks out:**
- `groupLinksByDivider`'s new no-leading-divider behavior (`current.length > 0` guard) is actually a fix vs. the old per-link `<hr>` check, which *would* have rendered a stray leading rule for `linkDividers: [0]`. Covered by the new test.
- Payout percent source change (per-link stored value → `source.artistPayoutPercent` from the registry) — verified server-side (`api/functions/artist-page.ts:77`) that the stored `payoutPercent` was already derived straight from the same platform registry, so this isn't a data divergence.
- `data-track-platform` removal — confirmed nothing else in the repo (tests, other components) reads it off these particular links.
- Ran the updated `RichArtistProfile.test.tsx` locally against the PR branch: 30/30 pass.


---

PR #380, opened 2026-08-01
Serve unclaimed artists from /api/artist-page instead of a 404

brandonlucasgreen:
## The bug

`https://unstream.stream/artist/funkadelic` showed **"We couldn't find that artist"** — a URL we'd just put in a social post. Searching "funkadelic" returned the same artist fine, so the data was never missing.

It isn't Funkadelic. It's **~790 of the 791 generated artist pages** — every artist who hasn't claimed their profile. Measured before the fix:

| slug | `/api/artist-page` |
|---|---|
| funkadelic, air, al-green, agalloch, aiko, sufjan-stevens | 404 |
| kid-lightbulbs (claimed) | 200 |

## Why it happened

The page fetches `/api/artist-page?slug=…`, which returned 404. `getArtistProfileBySlug` refused anything whose `match_confidence` wasn't `'claimed'` — correct when the endpoint shipped in #272 for the claimed-profile page, and harmless while the `artist-page-static` edge function rendered these URLs for everyone.

#369 then correctly stopped serving real browsers static HTML and handed them the SPA, which reads this endpoint. The claimed-only filter became a 404 for every fan running JavaScript.

**Crawlers never saw it.** The edge function has never had that filter, so social previews, Google's index and any `curl` check all kept returning a full page. Only the click broke. The SPA even shipped an `UnclaimedQuietCard` branch its own endpoint could never feed — dead since June 16.

## The change

- Drop the `match_confidence === 'claimed'` filter in `db.ts`. `artist-page.ts` is its only caller, so nothing else shifts.
- Mirror the edge function's claimed test — artist row claimed **and** profile verified — for the custom image and for the `verifiedAt` the SPA branches on. Without this, an artist mid-claim could get the rich profile in the SPA and the quiet card from the edge function: the same two-renderers-disagree class as #369.
- New `api/functions/__tests__/artist-page.test.ts`: unclaimed, claimed, mid-claim, missing slug, no slug.

## Reviewer notes

- **The test has teeth.** I reinstated the gate and confirmed 2 of the 5 tests fail, rather than trusting that they would.
- **Verified against real data, not mocks.** Ran the real handler against production Supabase (read-only): Funkadelic returns 6 links and 13 releases, Kid Lightbulbs' profile is byte-for-byte unchanged, an unknown slug still 404s. Then rendered both branches in a browser — no console errors.
- **`semantic-revert-check` may flag this.** The diff deletes a line that reads like a deliberate access check. It isn't a revert: the check was never a security boundary — the same rows are already public through the edge function, and `artist_links` for unclaimed artists are served to crawlers today.
- Rebased onto `main` after #379 landed. No conflicts — that PR touched web components, this touches `api/`. 527 web + 523 API tests pass.

### DIFF
--- a/api/functions/artist-page.ts
+++ b/api/functions/artist-page.ts
@@ -1,5 +1,6 @@
 // API endpoint: GET /api/artist-page?slug=...
-// Returns the full rich-profile payload for a claimed artist as JSON.
+// Returns an artist's page payload as JSON — the rich profile for a claimed artist, the links
+// and releases for an unclaimed one.
 // This is the data source for the React SPA artist page (UNS-102).
 
 import { getArtistProfileBySlug, getArtistReleases } from './db';
@@ -38,7 +39,7 @@ export async function handler(event: { queryStringParameters?: Record<string, st
   try {
     const bundle = await getArtistProfileBySlug(slug);
 
-    // getArtistProfileBySlug returns null if artist not found or not claimed
+    // getArtistProfileBySlug returns null only if there's no artist row for this slug
     if (!bundle) {
       return {
         statusCode: 404,
@@ -98,11 +99,16 @@ export async function handler(event: { queryStringParameters?: Record<string, st
       };
     });
 
+    // Same claimed test the artist-page-static edge function applies, so the two renderers of
+    // this URL agree on which card a fan sees: a profile row alone isn't enough, the artist row
+    // has to be claimed too.
+    const isClaimed = artistRow.match_confidence === 'claimed' && !!profile?.verified_at;
+
     // Sanitize the featured embed
     const featuredEmbed = sanitizeEmbed(profile?.featured_embed ?? null);
 
     // Build the image URL: custom > artist row > null
-    const imageUrl = profile?.custom_image_url || artistRow.image_url || null;
+    const imageUrl = (isClaimed && profile?.custom_image_url) || artistRow.image_url || null;
 
     // Releases. The page paginates these ten at a time rather than listing them all, so the cap
     // here bounds the payload, not what a fan can reach: six pages, which covers every catalogue
@@ -125,7 +131,8 @@ export async function handler(event: { queryStringParameters?: Record<string, st
         bio: profile.bio,
         customImageUrl: profile.custom_image_url,
         featuredEmbed,
-        verifiedAt: profile.verified_at,
+        // The SPA picks RichArtistProfile vs UnclaimedQuietCard off this field.
+        verifiedAt: isClaimed ? profile.verified_at : null,
       } : null,
       links,
       // Indexes into `links` above which a horizontal divider is drawn.
--- a/api/functions/__tests__/artist-page.test.ts
+++ b/api/functions/__tests__/artist-page.test.ts
@@ -1,0 +1,123 @@
+// GET /api/artist-page — the JSON the React artist page renders from.
+//
+// The thing worth locking: an *unclaimed* artist gets a 200 with their links. This endpoint
+// shipped for claimed profiles only, and once #369 made real browsers render /artist/:slug from
+// the SPA instead of the static edge HTML, that filter turned into a 404 on every one of the ~790
+// unclaimed artist pages — while crawlers, which the edge function still serves, saw a fine page.
+// Restore the `match_confidence === 'claimed'` gate anywhere in this path and the first test here
+// fails.
+
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+const mocks = vi.hoisted(() => ({
+  getArtistProfileBySlug: vi.fn(),
+  getArtistReleases: vi.fn(),
+  checkRateLimit: vi.fn(),
+}));
+
+vi.mock('../db', () => ({
+  getArtistProfileBySlug: mocks.getArtistProfileBySlug,
+  getArtistReleases: mocks.getArtistReleases,
+}));
+
+vi.mock('../ratelimit', () => ({
+  checkRateLimit: mocks.checkRateLimit,
+  getClientIp: () => '203.0.113.1',
+}));
+
+import { handler } from '../artist-page';
+
+const artistRow = (overrides: Record<string, unknown> = {}) => ({
+  id: 'artist-1',
+  slug: 'funkadelic',
+  name: 'Funkadelic',
+  image_url: 'https://f4.bcbits.com/img/0026457743_23.jpg',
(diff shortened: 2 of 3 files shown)


api/functions/artist-page.ts
```diff
@@ -125,7 +131,8 @@ export async function handler(event: { queryStringParameters?: Record<string, st
         bio: profile.bio,
         customImageUrl: profile.custom_image_url,
         featuredEmbed,
-        verifiedAt: profile.verified_at,
+        // The SPA picks RichArtistProfile vs UnclaimedQuietCard off this field.
+        verifiedAt: isClaimed ? profile.verified_at : null,
```
brandonlucasgreen, line 135: @brandonlucasgreen **Data exposure — confirmed.**

`isClaimed` gates `verifiedAt` and the `imageUrl` fallback above, but `bio`, `customImageUrl`, and `featuredEmbed` are still returned unconditionally whenever a profile row exists — only `verifiedAt` got the gate:

```js
profile: profile ? {
  bio: profile.bio,
  customImageUrl: profile.custom_image_url,
  featuredEmbed,
  verifiedAt: isClaimed ? profile.verified_at : null,
} : null,
```

Compare with `artist-page-static.ts`: there, `bio`, `featured_embed`, and the `custom_image_url`-driven `imageUrl` are only ever referenced *inside* `if (isClaimed) { ... }` — an unclaimed/mid-claim artist's crawler-facing HTML never contains them. This function was clearly written to mirror that gate (the comment above `isClaimed` says so) but only mirrors it for 2 of the 4 fields the edge function withholds.

**This state is reachable, not just theoretical:** `claim-artist.ts`'s `verify` action sets `artist_profiles.verified_at` in one update, then sets `artists.match_confidence = 'claimed'` in a separate update a few lines later — and explicitly treats failure of the second update as non-fatal (`// Non-fatal — profile is verified, just the artist table update failed`, claim-artist.ts:630-633). That leaves `verified_at` truthy while `match_confidence` stays at `'verified'` — exactly the fixture your new "mid-claim" test uses. `artist-profile.ts`'s edit endpoint only checks `profile.verified_at` before allowing bio/photo edits (line 210), not `match_confidence`, so a real artist stuck in that state can go on to populate `bio` and `custom_image_url`.

**Impact:** `GET /api/artist-page?slug=...` is public (CORS `*`, no auth). In that state it returns the real bio/custom photo/embed in the JSON even though `verifiedAt: null` tells the SPA to render `UnclaimedQuietCard` instead of `RichArtistProfile`, and even though the edge function withholds the same fields from crawlers. The SPA's own branch on `verifiedAt` happens to hide it from the rendered UI, but the endpoint itself leaks it to any direct caller.

**Test gap:** the new "does not treat a profile row on an unclaimed artist as claimed" test only asserts `profile.verifiedAt` and `artist.imageUrl` — it doesn't check `profile.bio` or `profile.customImageUrl`, which is why this doesn't fail today.

Suggested fix: gate the whole `profile` object (or at least `bio`/`customImageUrl`/`featuredEmbed`) on `isClaimed`, matching `artist-page-static.ts`. While there, `linkDividers` (line 139) is also computed unconditionally from `profile?.link_dividers` — the edge function only applies dividers `isClaimed ? ... : []` (artist-page-static.ts:313-315), so the same mid-claim state would also apply an artist-configured divider layout to a page that's supposed to be rendering as unclaimed.

Confidence: **confirmed**, verified by reading both renderers (`artist-page.ts` vs `artist-page-static.ts`) and the claim flow's non-atomic `verified_at` / `match_confidence` update in `claim-artist.ts`.

---

brandonlucasgreen (commented):
@brandonlucasgreen Reviewed as the only open PR right now. Security + regression pass below.

## Security

**Auth/authorization** — unaffected; endpoint remains unauthenticated/public by design (matches the edge function's public crawler access). Not a new concern.

**Input validation** — `slug` is passed straight into a parameterized Supabase `.eq()` call; no injection surface. Unchanged by this PR.

**Secret handling** — untouched.

**Injection risks** — none found; no raw SQL, no string-built queries.

**Unsafe defaults** — none introduced. CORS `Access-Control-Allow-Origin: '*'` on this endpoint predates this PR.

**Data exposure — confirmed** (inline comment on `api/functions/artist-page.ts:131-135`): the new `isClaimed` gate only protects `verifiedAt` and the `imageUrl` fallback. `profile.bio`, `profile.customImageUrl`, and `featuredEmbed` are still returned unconditionally whenever a profile row exists, unlike `artist-page-static.ts` where those fields are only ever referenced inside `if (isClaimed)`. The gap is reachable: `claim-artist.ts` sets `artist_profiles.verified_at` and `artists.match_confidence = 'claimed'` in two separate updates and explicitly treats the second one's failure as non-fatal, so `verified_at` can be truthy while `match_confidence` stays non-`'claimed'` — the exact fixture your own "mid-claim" test uses. In that state, a direct call to the public `GET /api/artist-page` endpoint leaks the real bio/custom photo/embed even though the SPA (which only looks at `verifiedAt`) and the edge function both intend to withhold it. See the inline comment for the full trace and a suggested fix (gate the whole `profile` object, and also gate `linkDividers` the same way the edge function does).

## Regression

**Changed assumptions:** `getArtistProfileBySlug`'s JSDoc (`api/functions/db.ts:136-138`, not touched by this diff) still says "Fetch ... for a claimed artist" and "Returns null if the artist doesn't exist or isn't claimed" — both now false. It's outside the diff so I couldn't anchor a line comment on it, but it's exactly the kind of stale doc that could lead someone to reintroduce the `match_confidence === 'claimed'` filter this PR just removed. Worth a one-line fix while this is fresh — needs-manual-verification on whether you want it in this PR or a fast-follow.

**Edge cases / null-empty behavior:** the 404 (no artist row) and 400 (no slug) paths are unchanged and covered by tests — confirmed fine.

**State transitions:** covered above — the claimed/verified two-step update is pre-existing behavior, not introduced by this PR, but this PR is the first code to build a security-relevant gate on top of it, so the non-atomicity now matters in a new way.

**API contract drift:** intentional and is the point of the PR (404 → 200 for unclaimed artists) — matches the edge function's existing contract for the same URLs. No new drift beyond the profile-field gap above.

Everything else (test structure, comments, the rest of the diff) looks solid — the new test file is well targeted at the original bug, it's just missing coverage for the two fields called out above.

(Note: I'd normally use Request Changes given the confirmed data-exposure finding, but GitHub doesn't allow that on your own PR from this account, so this is submitted as a comment instead — please treat the finding above as blocking.)

---

PR #274, opened 2026-06-16
feat(UNS-103): Rewrite ArtistPage as thin shell with three render branches

brandonlucasgreen:
## UNS-103 — `<ArtistPage>` rewrite

**Linear:** [UNS-103](https://linear.app/unstream/issue/UNS-103)
**Spec:** `docs/specs/UNS-103-artist-page-rewrite.md`

### Summary

Rewrites `ArtistPage.tsx` from a 450-line multi-branch component into a 145-line thin shell that fetches `/api/artist-page` and delegates rendering to one of three child components:

| State | Component |
|-------|-----------|
| Loading | `<LoadingProfile>` (spinner) |
| Not found / error | `<NotFoundCard>` (404 + MacAppPromo) |
| Claimed artist (`profile.verifiedAt`) | `<RichArtistProfile>` (with `justClaimed`, save button, post-claim banner) |
| Unclaimed artist | `<UnclaimedQuietCard>` (hero, links, claim nudge, save button) |

### Changes

**Removed from `ArtistPage.tsx`:**
- `useNavigate` / `handleSearch` (search bar gone)
- `SearchBar` / `ResultCard` imports (dead code)
- `isClaimedArtist` derivation
- `isProfileRoute` variable and stale `useLocation` comment
- Pre-generated `/data/artists/{slug}.json` fallback fetch
- `macAppPromo` JSX block (moved to `<NotFoundCard>` via `<MacAppPromo>`)
- `claimBannerDismissed` / `claimShareCopied` state (banner now lives in `<RichArtistProfile>` and `<UnclaimedQuietCard>`)

**New components:**
- `<MacAppPromo>` — extracted from old ArtistPage for reuse in NotFoundCard
- `<NotFoundCard>` — 404 card with slug display, /artists link, MacAppPromo
- `<LoadingProfile>` — spinner + "Loading profile…"
- `<UnclaimedQuietCard>` — hero, platform/social links, claim nudge, save button, Powered by Unstream

**Extended `<RichArtistProfile>`:**
- `justClaimed?: boolean` — shows dismissible green "🎉 You're verified! Welcome to Unstream." banner
- `onSave` / `onUnsave` / `isSaved` / `disabledSave` — auth-aware save button in the hero row
- Save button has same Save/Unsave UI as the old unclaimed card

### Test evidence

```
✓ 16 test files, 302 tests passed
  - ArtistPage.test.tsx: 10 tests (loading, claimed, unclaimed, 404, network error, justClaimed, title update, analytics, back link)
[...]

PR #382, opened 2026-08-01
Make Sentry actually deliver, and report 404s on URLs we publish

brandonlucasgreen:
Brandon: *"I get \*zero\* error reports out of Sentry, despite us having set it up."* He's right, and there were two independent reasons — neither of which would have caught the Funkadelic bug anyway. So this fixes the delivery faults *and* adds the signal that would actually have caught it.

## 1. Client events were blocked by our own CSP

`connect-src` allowed `https://*.ingest.sentry.io`. The DSN host is `o4510896048242688.ingest.**us**.sentry.io`. CSP host wildcards are a suffix match, and `.ingest.sentry.io` is not a suffix of `.ingest.us.sentry.io`.

Verified on production rather than inferred from the spec — same fetch, two hosts:

| host | result |
|---|---|
| `…ingest.us.sentry.io` (real) | **BLOCKED: Failed to fetch** |
| `…ingest.sentry.io` (allowed by CSP) | ALLOWED |

**58 client-side capture calls** have never delivered. No CSP violation showed in the console, which is why it went unseen. This is the second time a telemetry tool has silently died on a CSP hostname here — Umami, 2026-04, same shape — so the reasoning is now a comment in `netlify.toml`.

## 2. `SENTRY_DSN` was never set — ⚠️ needs you

`netlify env:list` has `VITE_SENTRY_DSN`, `SENTRY_ORG`, `SENTRY_PROJECT`, `SENTRY_AUTH_TOKEN`, but **no `SENTRY_DSN`**. `api/lib/sentry.ts` reads only that, returns early, and leaves `Sentry` a no-op stub — so all **19 server-side capture calls** do nothing, including the new ones in this PR.

I could not set it: env writes are blocked in my sandbox. **This PR is inert server-side until you run:**

```
npx netlify env:set SENTRY_DSN "<same value as VITE_SENTRY_DSN>" --context production
```

## 3. The 404 signal

The honest part first: **the Funkadelic 404 would not have alerted even with Sentry fully working.** `ArtistPage.tsx` treats 404 as an expected outcome and `artist-page.ts` returned it as a normal response. Generic error reporting can't catch a deliberate 404. So this reports the two cases that are genuinely ours:

- **A 404 on a slug we publish.** `data/artists-manifest.json` feeds the sitemap and the social posts, so those URLs are ones we hand out — a 404 there is our bug. A 404 on anything else is a fan or crawler guessing and stays quiet, or it buries the signal. Deduped per slug for 24h.
- **A failed lookup now 503s instead of 404ing.** `getArtistProfileBySlug` was collapsing "no such artist", "query errored" and "no service-role client" into one `null`, so a Supabase outage would have rendered as every artist page convincingly reporting the artist doesn't exist. It returns `{ bundle, failed }` now.

## 🚩 What this surfaced — bigger than the bug we just fixed

Building the signal meant checking it would actually fire, so I audited the manifest against the database:

**Only 55 of the 791 published artist slugs have an `artists` row. The other 736 404 right now, in production, after #380.**

All 791 are in `sitemap.xml`. I confirmed live: `funkadelic` returns 200 (it has a row), while `air`, `agalloch`, `al-green`, `aiko` all still 404.

This is a *different* bug from #380 — that one fixed artists who have a row. These have no row at all. `git log -S` dates it to **#274 (2026-06-16)**, when `ArtistPage.tsx` stopped reading `data/artists/{slug}.json` and moved to `/api/artist-page`: the 791 generated static files were orphaned, and nothing reads them at runtime now.

**So my earlier "#380 fixes ~790 pages" was wrong** — it fixed 55. Worth deciding before merge, because the new signal will report those 736 on day one (bounded to one event per slug per day). Options: populate the rows, or stop publishing slugs with no data. Happy to take either as the next PR.

## Verification

- New tests: 10 on the endpoint, 5 on the manifest module. **Teeth proven** — neutralised the manifest gate (the "stays quiet" test fails) and the `failed` branch (both 503 tests fail).
- The manifest-module test deliberately does *not* mock the JSON import, since it silently resolving to nothing is the failure worth guarding.
- Confirmed Netlify's bundler inlines the manifest: esbuilt the module and executed it from the bundle — 791 slugs, correct answers. Adds ~127KB to that function's bundle.
- 533 API + 528 web tests pass. Lint unchanged from baseline (77 pre-existing). `tsc` clean on all three touched source files.

### DIFF
--- a/api/functions/artist-page.ts
+++ b/api/functions/artist-page.ts
@@ -4,7 +4,9 @@
 // This is the data source for the React SPA artist page (UNS-102).
 
 import { getArtistProfileBySlug, getArtistReleases } from './db';
-import { checkRateLimit, getClientIp } from './ratelimit';
+import { checkRateLimit, checkSentryDedup, getClientIp } from './ratelimit';
+import { Sentry } from '../lib/sentry';
+import { isPublishedArtistSlug } from '../shared/published-artist-slugs';
 import { PLATFORMS } from '../shared/platform-registry';
 import { isBandcampFriday } from '../shared/bandcamp-friday';
 import { mainLinkDividerIndexes } from '../shared/link-dividers';
@@ -37,10 +39,39 @@ export async function handler(event: { queryStringParameters?: Record<string, st
   }
 
   try {
-    const bundle = await getArtistProfileBySlug(slug);
+    const { bundle, failed } = await getArtistProfileBySlug(slug);
 
-    // getArtistProfileBySlug returns null only if there's no artist row for this slug
+    // The lookup itself broke. Say so with a 503 instead of claiming the artist doesn't exist —
+    // a Supabase outage would otherwise render as every artist page 404ing convincingly.
+    if (failed) {
+      Sentry.captureMessage('[artist-page] artist lookup failed', {
+        level: 'error',
+        tags: { slug },
+        extra: { context: 'artist-page.lookupFailed' },
+      });
+      return {
+        statusCode: 503,
+        headers: { ...CORS_HEADERS, 'Cache-Control': 'no-store' },
+        body: JSON.stringify({ error: 'Artist lookup temporarily unavailable' }),
+      };
+    }
+
     if (!bundle) {
+      // A 404 for a slug we publish in data/artists-manifest.json is our bug: that manifest feeds
+      // the sitemap and the social posts, so the URL is one we've handed out. A 404 for anything
+      // else is just a fan or a crawler guessing, and reporting those would bury the real signal.
+      //
+      // Deduped per slug for 24h. That bound matters right now: at the time of writing only 55 of
+      // the 791 published slugs have an `artists` row, so this fires for the rest until that gap is
+      // closed (populate the rows, or stop publishing slugs with no data). One event per slug per
+      // day keeps that backlog readable instead of one per request.
+      if (isPublishedArtistSlug(slug) && await checkSentryDedup(`artist-page-404:${slug}`, 86400)) {
+        Sentry.captureMessage('[artist-page] 404 for a published artist slug', {
+          level: 'warning',
+          tags: { slug },
+          extra: { context: 'artist-page.publishedSlug404' },
+        });
+      }
       return {
         statusCode: 404,
         headers: CORS_HEADERS,
--- a/api/functions/__tests__/artist-page.test.ts
+++ b/api/functions/__tests__/artist-page.test.ts
@@ -13,6 +13,9 @@ const mocks = vi.hoisted(() => ({
   getArtistProfileBySlug: vi.fn(),
   getArtistReleases: vi.fn(),
   checkRateLimit: vi.fn(),
+  checkSentryDedup: vi.fn(),
+  captureMessage: vi.fn(),
+  isPublishedArtistSlug: vi.fn(),
 }));
 
 vi.mock('../db', () => ({
@@ -22,7 +25,20 @@ vi.mock('../db', () => ({
 
 vi.mock('../ratelimit', () => ({
   checkRateLimit: mocks.checkRateLimit,
+  checkSentryDedup: mocks.checkSentryDedup,
   getClientIp: () => '203.0.113.1',
+}));
+
+// Mocked so these tests assert on what the endpoint *decides to report*, independent of whether a
+// DSN is configured. Whether the events then leave the process is the deployment's problem —
+// which is exactly how 77 call sites managed to report nothing for weeks without a test noticing.
+vi.mock('../../lib/sentry', () => ({
+  Sentry: { captureMessage: mocks.captureMessage },
+}));
+
+// The real module reads data/artists-manifest.json; published-artist-slugs.test.ts covers that.
+vi.mock('../../shared/published-artist-slugs', () => ({
+  isPublishedArtistSlug: mocks.isPublishedArtistSlug,
 }));
 
 import { handler } from '../artist-page';
@@ -53,10 +69,12 @@ describe('GET /api/artist-page', () => {
     vi.clearAllMocks();
     mocks.checkRateLimit.mockResolvedValue({ limited: false });
     mocks.getArtistReleases.mockResolvedValue({ releases: [], total: 0 });
+    mocks.checkSentryDedup.mockResolvedValue(true);
+    mocks.isPublishedArtistSlug.mockReturnValue(false);
   });
 
   it('returns 200 with links for an unclaimed artist', async () => {
-    mocks.getArtistProfileBySlug.mockResolvedValue({ artist: artistRow(), profile: null, links });
+    mocks.getArtistProfileBySlug.mockResolvedValue({ bundle: { artist: artistRow(), profile: null, links }, failed: false });
 
     const res = await call('funkadelic');
     expect(res.statusCode).toBe(200);
@@ -72,7 +90,7 @@ describe('GET /api/artist-page', () => {
   });
 
   it('reports a verified claimed artist as claimed', async () => {
-    mocks.getArtistProfileBySlug.mockResolvedValue({
+    mocks.getArtistProfileBySlug.mockResolvedValue({ failed: false, bundle: {
       artist: artistRow({ match_confidence: 'claimed' }),
       profile: {
         bio: 'Free your mind.',
@@ -82,7 +100,7 @@ describe('GET /api/artist-page', () => {
         link_dividers: null,
       },
       links,
-    });
+    } });
 
     const body = JSON.parse((await call('funkadelic')).body);
     expect(body.profile.verifiedAt).toBe('2026-06-01T00:00:00.000Z');
@@ -92,7 +110,7 @@ describe('GET /api/artist-page', () => {
   it('does not treat a profile row on an unclaimed artist as claimed', async () => {
     // Mid-claim state: the profile row exists but the artist was never marked claimed. The edge
     // function shows crawlers the quiet card here, so the SPA has to as well.
-    mocks.getArtistProfileBySlug.mockResolvedValue({
+    mocks.getArtistProfileBySlug.mockResolvedValue({ failed: false, bundle: {
       artist: artistRow({ match_confidence: 'verified' }),
       profile: {
         bio: null,
@@ -102,7 +120,7 @@ describe('GET /api/artist-page', () => {
         link_dividers: null,
       },
       links,
-    });
+    } });
 
     const body = JSON.parse((await call('funkadelic')).body);
     expect(body.profile.verifiedAt).toBeNull();
@@ -110,7 +128,7 @@ describe('GET /api/artist-page', () => {
   });
 
   it('404s when there is no artist row for the slug', async () => {
-    mocks.getArtistProfileBySlug.mockResolvedValue(null);
+    mocks.getArtistProfileBySlug.mockResolvedValue({ bundle: null, failed: false });
 
     const res = await call('not-a-real-artist');
     expect(res.statusCode).toBe(404);
@@ -119,5 +137,60 @@ describe('GET /api/artist-page', () => {
   it('400s without a slug', async () => {
     const res = await handler({ httpMethod: 'GET', queryStringParameters: {}, headers: {} });
     expect(res.statusCode).toBe(400);
+  });
+
+  // --- the 404 signal (UNS: Funkadelic follow-up) ---
+
+  describe('reporting', () => {
+    it('reports a 404 on a slug we publish', async () => {
+      mocks.getArtistProfileBySlug.mockResolvedValue({ bundle: null, failed: false });
+      mocks.isPublishedArtistSlug.mockReturnValue(true);
+
+      expect((await call('funkadelic')).statusCode).toBe(404);
+      expect(mocks.captureMessage).toHaveBeenCalledTimes(1);
+      const [message, opts] = mocks.captureMessage.mock.calls[0];
+      expect(message).toContain('published artist slug');
+      expect(opts.tags.slug).toBe('funkadelic');
+    });
+
+    it('stays quiet on a 404 for a slug we never published', async () => {
+      // The whole point of gating on the manifest: crawlers and typos must not drown the signal.
+      mocks.getArtistProfileBySlug.mockResolvedValue({ bundle: null, failed: false });
+      mocks.isPublishedArtistSlug.mockReturnValue(false);
+
+      expect((await call('asdfasdf')).statusCode).toBe(404);
+      expect(mocks.captureMessage).not.toHaveBeenCalled();
+    });
+
+    it('reports a published-slug 404 only once per dedup window', async () => {
+      mocks.getArtistProfileBySlug.mockResolvedValue({ bundle: null, failed: false });
+      mocks.isPublishedArtistSlug.mockReturnValue(true);
+      mocks.checkSentryDedup.mockResolvedValue(false); // already reported this hour
+
+      expect((await call('funkadelic')).statusCode).toBe(404);
+      expect(mocks.captureMessage).not.toHaveBeenCalled();
+    });
+
+    it('503s and reports when the lookup itself fails, rather than 404ing', async () => {
+      // The distinction that matters: a database outage must not render as "artist not found" on
+      // every page. A 404 here would be a monitoring signal that actively lies.
+      mocks.getArtistProfileBySlug.mockResolvedValue({ bundle: null, failed: true });
+
+      const res = await call('funkadelic');
+      expect(res.statusCode).toBe(503);
+      expect(res.headers['Cache-Control']).toBe('no-store');
+      expect(mocks.captureMessage).toHaveBeenCalledTimes(1);
+      expect(mocks.captureMessage.mock.calls[0][1].level).toBe('error');
+    });
+
+    it('does not fire the published-slug 404 signal when the lookup failed', async () => {
+      // failed=true must take the 503 branch even for a published slug, so an outage reports as an
+      // outage rather than as ~790 "we publish a broken URL" warnings.
+      mocks.getArtistProfileBySlug.mockResolvedValue({ bundle: null, failed: true });
+      mocks.isPublishedArtistSlug.mockReturnValue(true);
+
+      expect((await call('funkadelic')).statusCode).toBe(503);
+      expect(mocks.captureMessage.mock.calls[0][0]).toContain('lookup failed');
+    });
   });
 });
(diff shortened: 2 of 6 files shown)


api/functions/artist-page.ts
```diff
@@ -37,10 +39,39 @@ export async function handler(event: { queryStringParameters?: Record<string, st
...
     if (!bundle) {
+      // A 404 for a slug we publish in data/artists-manifest.json is our bug: that manifest feeds
+      // the sitemap and the social posts, so the URL is one we've handed out. A 404 for anything
+      // else is just a fan or a crawler guessing, and reporting those would bury the real signal.
+      //
+      // Deduped per slug for 24h. That bound matters right now: at the time of writing only 55 of
+      // the 791 published slugs have an `artists` row, so this fires for the rest until that gap is
+      // closed (populate the rows, or stop publishing slugs with no data). One event per slug per
+      // day keeps that backlog readable instead of one per request.
+      if (isPublishedArtistSlug(slug) && await checkSentryDedup(`artist-page-404:${slug}`, 86400)) {
```
brandonlucasgreen, line 68: **Regression / edge case — confirmed.** The dedup key is built from the raw `slug` (whatever casing the request used), but `isPublishedArtistSlug` matches case-insensitively via `.toLowerCase()`. So `/artist/Funkadelic` and `/artist/funkadelic` each get their own 24h dedup window and can both fire, even though they're "the same" published slug for the purposes of this check. Given the manifest currently has ~736 slugs that will 404 on merge, any mixed-casing traffic (old links, crawlers normalizing differently) multiplies the alert volume this dedup is meant to bound. Suggest `checkSentryDedup(\`artist-page-404:${slug.toLowerCase()}\`, 86400)` to match the semantics of `isPublishedArtistSlug`.

---

brandonlucasgreen (commented):
@brandonlucasgreen Automated senior-staff-engineer pass on this PR. One inline comment plus the rest below (the other findings sit outside this diff's changed lines, so GitHub won't let me anchor them inline).

## Security

- **Auth/authz** — no change; `artist-page` stays an unauthenticated public GET, consistent with existing pattern. No issue.
- **Input validation / injection** — `slug` flows into a Supabase `.eq('slug', slug)` filter (parameterized, no injection surface) and into a `Set.has()` lookup in `published-artist-slugs.ts`. No issue.
- **Secret handling** — no secrets touched. The `SENTRY_DSN` value in the PR description's `netlify env:set` command is blank/redacted, so nothing to flag there.
- **Data exposure** — the new Sentry events tag/extra only `slug` and a static context string, nothing sensitive. No issue.
- **Unsafe defaults — needs-manual-verification.** `checkSentryDedup` (`api/functions/ratelimit.ts:395-409`, pre-existing, not changed here) fails open — `return true` both when Redis is unconfigured and when a Redis call errors — so the gate this PR relies on to keep the new 404 signal to "once per slug per day" silently stops gating during a Redis outage. That's an existing, documented tradeoff for other call sites, but this PR adds a caller whose whole point is bounding a ~736-slug backlog; worth confirming Sentry's ingest quota/alert routing can tolerate the unbounded case before this ships.
- **Unsafe defaults — plausible.** Same function: the dedup check is `get` then `set`, not atomic (`ratelimit.ts:400-402`). Concurrent requests for the same not-yet-seen slug can both read "not seen" before either write lands, so duplicate events are possible under concurrent traffic. Pre-existing pattern, flagged because the new caller is the one most likely to hit it on day one (see below).

## Regression

1. **Confirmed — the stated fix doesn't reach the user.** The PR's whole justification for the 503 branch is *"a Supabase outage would otherwise render as every artist page 404ing convincingly."* But `apps/web/src/pages/ArtistPage.tsx:42` (untouched by this PR) does:
   `r.ok ? r.json() : r.status === 404 ? null : Promise.reject(...)`
   and the `.catch()` at line 48-52 sets `setNotFound(true)` on *any* rejection — including the new 503. So a real outage still renders `NotFoundCard` to fans; only the client-side `Sentry.captureException` differs from a true 404. Worse, `api/edge/artist-page-static.ts:227-238` (also untouched, SSR path) destructures only `data` from the Supabase call and ignores `error` entirely, so on a query error it silently falls through to the SPA (`context.next()`) exactly as it would for a genuine absence — the edge layer can't tell "outage" from "no artist" either. Net effect: the `{ bundle, failed }` distinction this PR introduces in `db.ts`/`artist-page.ts` is real and well-tested at that layer, but neither downstream consumer surfaces it, so the actual user- and crawler-facing symptom during a Supabase outage is unchanged from before this PR. Worth either fixing the SPA/edge handling in this PR or explicitly scoping the PR to "the API now reports outages to Sentry correctly" rather than implying fans stop seeing false 404s.
2. **Confirmed (minor)** — dedup-key casing mismatch, see inline comment on `artist-page.ts:68`.
3. **API contract drift — contained, verified.** `getArtistProfileBySlug`'s return type changed from `ArtistProfileBundle | null` to `{ bundle, failed }`. Grepped the repo: `artist-page.ts` is the only caller outside tests, and it's updated correctly. No other breakage.
4. **State transitions** — `db.ts`'s `maybeSingle()` vs `single()` swap is correct and is exactly the right fix for collapsing "0 rows" into an error; the three-way return (`found` / `absent` / `failed`) is properly exhaustive and each branch has a test.
5. Test coverage for the new behavior (dedup skip, failed-lookup 503, published-vs-unpublished quiet/loud) is solid and does neutralization-test the guards per CLAUDE.md's "teeth proven" convention.

The Sentry CSP host fix (`*.ingest.sentry.io` → `*.ingest.us.sentry.io`) is correct and narrows rather than widens the CSP — no concern there.

---

api/functions/artist-page.ts
```diff
@@ -37,10 +39,39 @@ export async function handler(event: { queryStringParameters?: Record<string, st
...
+    if (failed) {
+      Sentry.captureMessage('[artist-page] artist lookup failed', {
+        level: 'error',
+        tags: { slug },
+        extra: { context: 'artist-page.lookupFailed' },
+      });
+      return {
+        statusCode: 503,
+        headers: { ...CORS_HEADERS, 'Cache-Control': 'no-store' },
+        body: JSON.stringify({ error: 'Artist lookup temporarily unavailable' }),
```
brandonlucasgreen, line 55: **[plausible] Regression — unthrottled Sentry capture on the `failed` (503) branch**

This branch calls `Sentry.captureMessage` unconditionally on every request where the lookup fails:

```ts
if (failed) {
  Sentry.captureMessage('[artist-page] artist lookup failed', { level: 'error', tags: { slug }, ... });
  return { statusCode: 503, ... };
}
```

Contrast with the published-slug-404 branch a few lines below, which the PR deliberately gates with `checkSentryDedup(...)` specifically to avoid "one event per request" ("keeps that backlog readable instead of one per request"). The `failed` branch has no equivalent dedup key.

During a sustained Supabase outage (the exact scenario this branch exists to catch), every non-rate-limited request to `/api/artist-page` — across all slugs, all IPs — will fire its own `captureMessage`. It's bounded per-IP by `checkRateLimit` above, but not bounded in aggregate, so a real outage could still generate a large burst of Sentry events at the moment Sentry capacity/attention matters most.

Not blocking, but worth a `checkSentryDedup('artist-page-lookup-failed', <short ttl>)` guard (or similar) before merge, or a conscious decision that per-request reporting here is acceptable.

---

brandonlucasgreen (commented):
@brandonlucasgreen Staff-eng security/regression pass on this PR (code-only, no live testing).

**Security**

- **Auth/authorization** — no changes to auth paths; `getArtistProfileBySlug` continues to use the service-role client (RLS bypass by design, unchanged). No findings.
- **Input validation** — `slug` is still taken unvalidated from the query string (pre-existing, not introduced here). No new issue.
- **Secret handling** — confirmed: no secret values appear in the diff or PR body. The `SENTRY_DSN` command in the description has an empty value; you still need to set the real one out-of-band as the PR notes. Fine as-is.
- **Injection risks** — Redis dedup key (`artist-page-404:${slug}`) and Supabase `.eq('slug', slug)` are both parameterized/library-mediated, not string-built SQL/commands. No findings.
- **Unsafe defaults** — none introduced. `checkSentryDedup`'s existing fail-open behavior (`if (!r) return true`) is pre-existing, not new to this PR.
- **Data exposure** — the 503 body ("Artist lookup temporarily unavailable") doesn't leak the underlying Supabase error; Sentry tags (`slug`) are public data. No findings. CSP fix (`*.ingest.sentry.io` → `*.ingest.us.sentry.io`) is a correct narrowing to the real DSN host, not a loosening — verified against the actual CSP line in `netlify.toml`.

**Regression**

- **[plausible] Unthrottled Sentry capture on the `failed`/503 branch** — left as an inline comment on `api/functions/artist-page.ts`. The new published-slug-404 path is deliberately deduped (`checkSentryDedup`, 24h/slug) specifically to avoid one event per request; the new `failed` path a few lines above it has no equivalent guard, so a sustained Supabase outage emits one `captureMessage` per non-rate-limited request rather than one per outage. Worth a dedup key or a conscious call that it's fine.
- **Changed assumptions / API contract** — `getArtistProfileBySlug` now returns `{ bundle, failed }` instead of `ArtistProfileBundle | null`. Checked for other call sites: only `api/functions/artist-page.ts` and its test import it (`grep -r getArtistProfileBySlug`), and both are updated in this diff. No dangling caller on the old signature.
- **Null/empty & state transitions** — `db.ts`'s switch from `.single()` to `.maybeSingle()` correctly separates "0 rows" (now `{ bundle: null, failed: false }`) from a real query error (`{ bundle: null, failed: true }`); the >1-row case still surfaces as an error via `.maybeSingle()`, so behavior there is unchanged. Verified against `slugify()` in `scripts/generate-artist-list.ts` that manifest slugs are always lowercase, so `isPublishedArtistSlug`'s case-insensitive lookup is safe in practice even though the backing `Set` isn't itself lowercased at construction — not a live bug given that invariant, just a bit fragile if that ever changes.
- **[needs-manual-verification] Product/ops call already flagged by the PR itself** — merging this means the new 404 signal starts firing for the ~736 published slugs that currently have no `artists` row (per your own PR description), bounded to 1/slug/day. Confirming you've internalized that before merge, not a code defect.
- Frontend (`ArtistPage.tsx`) already treats any non-200/non-404 response as a caught error → `NotFoundCard`, so the new 503 doesn't crash or dead-end the SPA — confirmed by reading the fetch handler, no fix needed there.

No blocking findings — the dedup gap on the 503 path is the only thing I'd want a decision on before merge.

---

brandonlucasgreen (commented):

**Security** — no issues found across auth, input validation, secret handling, injection, unsafe defaults, or data exposure. Notes:
- *confirmed*: `slug` continues through the existing 400-if-missing guard; the new `Sentry.captureMessage` tags just carry it as metadata (Sentry API, not a shell/SQL string), so no new injection surface.
- *confirmed*: CSP change narrows `connect-src` to the actual DSN host (`*.ingest.us.sentry.io`) rather than broadening it — this is a fix, not a loosening.
- *needs-manual-verification*: `api/lib/sentry.ts` (not touched by this PR) embeds the Sentry DSN in a code comment. DSNs are semi-public ingest keys, not high-severity secrets, but this PR is what starts actually driving traffic through that client from more places — worth a glance since it's now more load-bearing.

**Regression**
- *confirmed-safe (API contract drift)*: `getArtistProfileBySlug` changed from `Bundle | null` to `{ bundle, failed }`. Checked for other callers — `api/functions/artist-page.ts` is the only one, and it's updated in the same commit.
- *confirmed*: `.single()` → `.maybeSingle()` correctly separates "no row" / "query error" / "no client" into three states, and `artist-page.ts` now routes `failed` to a 503 instead of a lying 404. New tests cover all three branches.
- *needs-manual-verification (edge case)*: the new `failed` (503) branch calls `Sentry.captureMessage` on **every** failed request with no dedup — unlike the 404-signal branch two lines below it, which is gated by `checkSentryDedup`. During a real Supabase outage this fires once per request (bounded only by the 30 req/min per-IP rate limit across however many IPs are hitting the endpoint). Worth confirming that volume is intended rather than a missed copy-paste of the dedup pattern.
- *false positive (verified by diff)*: `semantic-revert-check` flagged this PR for deleting lines added by #376/#360/#359/#344/#340/#328/#319 in `db.ts` and one line from #380 in `artist-page.ts`. I diffed this branch directly against current `main`: every exported function in `db.ts` is still present, same order, same signatures except `getArtistProfileBySlug`'s intentional change. The flagged "deletions" are comment/line churn on lines the bot's blame traces to those PRs, not actual reverted logic.

No blocking findings — the outstanding item is confirming `SENTRY_DSN` gets set in Netlify per the PR description, since this whole PR is inert server-side without it.

---

brandonlucasgreen (commented):
@brandonlucasgreen Automated security/regression pass on this PR (`artist-page.ts`, `db.ts`, new `published-artist-slugs.ts`, CSP fix in `netlify.toml`).

**Security**

- Auth/authorization: unchanged — `checkRateLimit(ip, 'standard', ...)` (per-IP) still gates the endpoint before any of the new logic runs. *confirmed no issue*
- Input validation: `slug` still comes from the existing required-param check on `event.queryStringParameters?.slug`; the new code only reads it (into Sentry tags and the manifest lookup), doesn't reinterpret it. *confirmed no issue*
- Secret handling: no new secrets introduced.
- Injection risks: `getArtistProfileBySlug`'s `.maybeSingle()` still goes through supabase-js's parameterized `.eq('slug', slug)` — no change in surface. *confirmed no issue*
- Data exposure / CSP: **confirmed no issue** — the `connect-src` fix (`https://*.ingest.sentry.io` → `https://*.ingest.us.sentry.io`) is a targeted correction to the actual region-scoped Sentry ingest host per the inline comment; it doesn't broaden the policy beyond what was already intended, it fixes a host that silently never matched.
- **Plausible, unsafe defaults** — the new `failed: true` (503) branch in `artist-page.ts` calls `Sentry.captureMessage(...)` unconditionally on *every* request, with no `checkSentryDedup` gate. That's inconsistent with the sibling `isPublishedArtistSlug` 404 branch added in this same PR, which explicitly adds dedup "to keep the backlog readable instead of one per request." `checkRateLimit`'s `standard` tier is per-IP (30/min); during an actual Supabase outage, distributed legitimate traffic across many users/IPs hitting different artist slugs would each independently fire an uncapped Sentry event — exactly the flood pattern `checkSentryDedup` exists to prevent elsewhere in this codebase (`search-sources.ts`, `search-musicbrainz.ts`, `bandcamp-probe.ts`). Worth a short-TTL dedup key here too, so an outage doesn't also burn Sentry's quota at the moment its visibility matters most.

**Regression**

- Changed assumptions: `getArtistProfileBySlug` now returns `{ bundle, failed }` instead of `ArtistProfileBundle | null`. Confirmed by grep that `db.ts`, `artist-page.ts`, and its test are the only call sites — no stale destructuring of the old shape elsewhere in the repo.
- Edge cases: `.single()` → `.maybeSingle()` correctly separates "0 rows" (`{bundle: null, failed: false}`) from a real query error (`{failed: true}`) — this is the "never cache/report uncertainty as a negative" fix the new docblock describes, and it's the right fix.
- **Plausible, low severity** — `published-artist-slugs.ts` builds the `publishedSlugs` `Set` from the manifest *without* lowercasing entries, but `isPublishedArtistSlug()` only lowercases the incoming query slug before `.has()`. All 791 current manifest slugs happen to already be lowercase (verified against the file directly), so this is harmless today, but nothing enforces that invariant going forward — a future mixed-case manifest entry would silently fail to match and its 404s would go unreported, which is the opposite of this PR's goal. Lowercasing at `Set` construction time would close the gap.
- State transitions: the 404 → 503 split is a genuine improvement — a DB outage on `/api/artist-page` no longer masquerades as "artist not found," and `Cache-Control: no-store` on the 503 keeps that failure from being cached anywhere.
- API contract drift: 503 is a new status code for this endpoint (previously only 200/400/404). Worth confirming the SPA's fetch handling for this endpoint treats an unexpected 503 gracefully rather than only branching on 200/404, since that path wasn't reachable before.

One plausible security-adjacent finding (unbounded Sentry volume on the failure branch) and one plausible low-severity correctness edge case flagged above; the 404/503 split itself is a solid fix.

---

brandonlucasgreen (commented):

**Security**
- `netlify.toml` CSP change: `connect-src` moves from `https://*.ingest.sentry.io` to `https://*.ingest.us.sentry.io`. This is a *correction* to match the real DSN host, not a broadening — still scoped to a Sentry-ingest-only wildcard, and the reasoning is captured in the added comment as required. — confirmed: no issue.
- New `Sentry.captureMessage` calls tag `slug` (an artist URL slug) — not PII or a secret, consistent with existing tagging elsewhere. — confirmed: no issue.
- `getArtistProfileBySlug`'s `.single()` → `.maybeSingle()` change plus explicit `artistError` handling is a correctness fix; no new untrusted input reaches the query (the slug already flowed through unchanged validation). — confirmed: no issue.
- No secrets logged, no new outbound fetches, no auth/authorization surface touched.

**Regression**
- ⚠️ `semantic-revert-check` flagged this PR as deleting lines originally added by 7 previously-merged PRs (#376, #360, #359, #344, #340, #328, #319) in `db.ts`, plus lines from #380 in `artist-page.ts` and its test. I traced this by hand: the entire `db.ts` diff here touches only `getArtistProfileBySlug`'s docstring, the `.single()`→`.maybeSingle()` line, and its return statements — nothing about release cataloging, typeahead, admin link removal, or the Bandcamp probe cache (the actual features those PRs shipped) is touched. `getArtistProfileBySlug` has exactly one caller (`api/functions/artist-page.ts`, confirmed via repo-wide grep), and it's updated in this same PR to match the new `{ bundle, failed }` shape. This reads as blame-churn on a heavily-edited function/comment block rather than an actual revert — **needs-manual-verification** given this repo's documented revert-bug history (`docs/retros/UNS-100-bifurcation-retro.md`), but I found no dropped functionality.
- `checkSentryDedup('artist-page-404:' + slug, 86400)` only runs when `isPublishedArtistSlug(slug)` is already `true` (via `&&` short-circuit), so unpublished/typo'd slugs never touch Redis or Sentry — this also closes off an unauthenticated-request path to spam Sentry's event quota with arbitrary slugs. — confirmed: correct, and a nice bound.
- The 503-vs-404 split correctly separates "no client" / "query error" / "exception" (`failed: true`) from "confirmed absent" (`bundle: null, failed: false`) — matches the codebase's "never cache uncertainty" principle, and nothing here caches the outcome, so no stale-negative risk. — confirmed: no issue.
- The existing 200-path response shape for a found artist is unchanged (`ArtistProfileBundle` is just wrapped in `{ bundle, failed: false }`) — confirmed no behavior change for current traffic.

**Non-code note:** the PR description itself flags that `SENTRY_DSN` isn't set in production, so server-side reporting added here is inert until that env var is set manually — not a code defect, just don't expect the new 404 signal to actually fire yet.

No blocking findings; the semantic-revert-check flag is worth a manual glance but I found nothing reverted.

---

PR #384, opened 2026-08-01
Backfill artist rows for the 736 published pages that 404'd

brandonlucasgreen:
**Already run against production — all 791 published artist slugs now return 200.** This PR is the script that did it, kept so the work is reviewable and repeatable.

## The problem

`data/artists-manifest.json` (791 artists) feeds the sitemap and the generated social posts, but `/artist/:slug` renders from Supabase — and rows only land there when someone **searches** an artist, via `persistSearchResults`. Nothing ever seeded the published set.

| | before | after |
|---|---|---|
| published slugs with an `artists` row | 55 | **791** |
| returning 404 | **736** | 0 |

Found while building the 404 signal in #382 — checking the alert would actually fire is what surfaced it. This is a *different* bug from #380: that one fixed artists who have a row but were refused by the claimed filter. These had no row at all. `git log -S` dates it to **#274 (2026-06-16)**, when `ArtistPage` moved onto `/api/artist-page` and orphaned `data/artists/`.

## Three decisions worth reviewing

**The slug comes from the manifest, never derived from the name.** `artistSlug()` in db.ts turns accents into hyphens — `Jónsi` → `j-nsi`, `Hüsker Dü` → `h-sker-d` — and disagrees with the manifest on **34 of 791**. Deriving would have filed those under URLs the sitemap doesn't advertise: 404 unfixed, run looking successful. All 34 verified live afterwards.

**Insert-only.** Existing rows are skipped untouched whatever their state, so claimed profiles can't be affected even in principle — the published set and the claimed set turn out to be disjoint anyway. `source='claimed'` was 124 before and after. `artists.slug` is UNIQUE (`artists_slug_key`, confirmed by probing a duplicate insert), so a re-run fails safely rather than duplicating.

**No cataloguing is triggered.** `persistSearchResults` requests a release catalogue for every artist with a Bandcamp link, and *all* 791 have one — reusing it wholesale would have dumped hundreds of crawls on Bandcamp at once. Releases can be catalogued later, deliberately and in batches.

Link filtering imports `isDirectLink` / `EXCLUDED_PLATFORMS` from db.ts rather than restating them, so a backfilled artist stores exactly what a searched artist would. That's the only change to existing code: two symbols made `export`.

## Verification

- Dry run first (the default), then `--limit=3 --write`, checked those three live, then the rest.
- 736 artists, 4,786 links, zero failures, zero skips.
- All 791 slugs return 200, sampled across the alphabet and across every accent case.
- `808-state` rendered in the browser: pills, payout percentages, AI-policy badges, socials, claim CTA.
- Spot-checked stored Bandcamp URLs — the 303s are legitimate artist→album redirects, not dead links.
- 523 API tests pass; lint unchanged from baseline; `tsc` clean on both touched files.

`match_confidence` is `verified` for all 736. That's earned, not assumed: `generate-artist-data.ts` only ever saved artists with a real, non-empty presence on a qualifying platform. It does roughly triple the `verified` population, which feeds typeahead ranking — worth an eye on suggestions over the next few days.

**Rollback**, if ever needed: these rows are identifiable as `source='auto'` with `created_at` on 2026-08-01 and a slug in the manifest; `artist_links` cascade on delete.

## Follow-on

With this merged, #382's published-slug 404 signal starts life quiet instead of reporting 736 known-broken URLs — which was the reason to do this first.

### DIFF
--- a/api/functions/db.ts
+++ b/api/functions/db.ts
@@ -34,8 +34,9 @@ export function artistSlug(name: string): string {
   return name.toLowerCase().replace(/[^a-z0-9]+/g, '-').replace(/^-|-$/g, '');
 }
 
-// Determine if a platform URL is a direct link (not a search URL)
-function isDirectLink(url: string): boolean {
+// Determine if a platform URL is a direct link (not a search URL).
+// Exported so scripts/backfill-published-artist-rows.ts stores exactly what a search would store.
+export function isDirectLink(url: string): boolean {
   const lower = url.toLowerCase();
   return (
     !lower.includes('duckduckgo.com') &&
@@ -45,8 +46,9 @@ function isDirectLink(url: string): boolean {
   );
 }
 
-// Platforms that are manual search links, not real artist presences
-const EXCLUDED_PLATFORMS = new Set(['buymeacoffee', 'kofi', 'ampwall']);
+// Platforms that are manual search links, not real artist presences.
+// Exported alongside isDirectLink for the same reason — see above.
+export const EXCLUDED_PLATFORMS = new Set(['buymeacoffee', 'kofi', 'ampwall']);
 
 // How long before artist data is considered stale (24 hours)
 const FRESHNESS_TTL_MS = 24 * 60 * 60 * 1000;
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "generate:data": "tsx scripts/generate-artist-data.ts",
     "generate:social": "tsx scripts/generate-social-posts.ts",
     "sync:bandcamp-dates": "npx tsx scripts/sync-bandcamp-dates.ts",
+    "backfill:artist-rows": "npx tsx scripts/backfill-published-artist-rows.ts",
     "ingest:try": "npx tsx scripts/ingest-releases.ts",
     "preview:release": "npx tsx scripts/preview-release-page.ts",
     "catalog:artist": "npx tsx scripts/catalog-artist.ts",
--- a/scripts/backfill-published-artist-rows.ts
+++ b/scripts/backfill-published-artist-rows.ts
@@ -1,0 +1,232 @@
+#!/usr/bin/env npx tsx
+/**
+ * Backfill `artists` + `artist_links` rows for the artist pages we publish but never stored.
+ *
+ * Why this exists: `data/artists-manifest.json` (791 artists) feeds the sitemap and the generated
+ * social posts, but /artist/:slug renders from Supabase — and rows only appear there when someone
+ * *searches* an artist (`persistSearchResults`). Nothing ever seeded the published set, so at the
+ * time of writing only 55 of the 791 had a row and the other 736 returned 404 to fans and to
+ * Google. The static data under `data/artists/` has been orphaned since #274 moved the page onto
+ * /api/artist-page; this script pours it into the database the page actually reads.
+ *
+ * Usage:
+ *   npx tsx scripts/backfill-published-artist-rows.ts              # dry run (default)
+ *   npx tsx scripts/backfill-published-artist-rows.ts --write      # actually insert
+ *   npx tsx scripts/backfill-published-artist-rows.ts --limit=10   # try a small slice first
+ *
+ * Safety properties, in order of how much they matter:
+ *
+ *  - **Insert-only.** Any slug that already has a row is skipped untouched, whatever its state.
+ *    Nothing here updates or deletes an existing artist or link. Claimed profiles therefore can't
+ *    be affected even in principle — which is the point, given the artist_links wipe of 2026-07-29.
+ *  - **The slug comes from the manifest, never derived from the name.** `artistSlug()` in db.ts
+ *    strips accents to hyphens (`Jónsi` → `j-nsi`), and 34 of the 791 manifest slugs disagree with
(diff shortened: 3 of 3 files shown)


brandonlucasgreen (commented):

**Security** — *confirmed*, no issues across auth, input validation, secret handling, injection, and unsafe defaults. It's a local, manually-run script: service-role key pulled via `requireEnv` (never logged), insert-only against Supabase through the SDK (no raw SQL string-building), reads only repo-local JSON, and defaults to a dry run (`write` must be explicitly opted into via `--write`).

**Regression**
- *plausible (input validation / edge case)*: `--limit=<value>` isn't validated — `Number('abc')` is `NaN`, and `candidates.slice(0, NaN)` silently resolves to an empty plan (0 inserts) rather than erroring. Low impact given dry-run-by-default and the "Would insert 0 artists" line is at least visible in the output, but it fails quietly rather than loudly on a malformed flag.
- *needs-manual-verification*: the PR description says this already ran against production before being opened (736 rows inserted, all 791 published slugs now 200). Worth confirming nothing's been added to `data/artists-manifest.json` since that run that would still be missing a row — this script is a one-time snapshot, not something that stays current automatically.
- No `semantic-revert-check` flag on this PR, and the two exports added to `db.ts` (`isDirectLink`, `EXCLUDED_PLATFORMS`) are additive only — confirmed no existing call sites changed behavior.

No blocking findings.

---

brandonlucasgreen (commented):
@brandonlucasgreen Automated security/regression pass on this PR (`scripts/backfill-published-artist-rows.ts` + the `db.ts` export widening).

**Security**

- Auth/authorization: N/A — standalone ops script, defaults to dry-run, requires an explicit `--write` flag to mutate anything. Matches the existing `ingest:try` convention in this repo. *confirmed no issue*
- Input validation: no external/user input — the manifest and `data/artists/*.json` are repo-committed files, not attacker-reachable at runtime.
- Secret handling: `SUPABASE_SERVICE_KEY` read via `requireEnv()`, never logged. *confirmed no issue*
- Injection risks: all DB access goes through supabase-js `.insert()` / `.select().in()` — parameterized, no raw SQL. *confirmed no issue*
- Unsafe defaults: **confirmed no issue** — insert-only (never updates/deletes an existing row), explicit skip of slugs that already have an `artists` row, and no release-cataloguing side effect. All three match the file's stated safety properties.
- Data exposure: none — the data written (artist name, links) is the same class of public data `persistSearchResults` already stores from the same source files.
- Minor/needs-manual-verification, low severity: `buildCandidate` does `resolve(repoRoot, 'data/artists', \`${entry.slug}.json\`)` with no explicit format check on `entry.slug` before the path join. Not exploitable today since the manifest is repo-controlled rather than external input, but flagging for completeness since there's no defensive check.

**Regression**

- Changed assumptions: `isDirectLink` / `EXCLUDED_PLATFORMS` exported from `db.ts` — visibility-only change, confirmed by grep that no other file references them, so no behavior change for existing callers.
- Edge cases: `matchConfidence` defaults to `'unverified'` for anything that isn't exactly `'verified'` in the static file — safe default, can't accidentally produce `'claimed'`.
- **Plausible, worth a look** — state transitions / partial failure: `insertCandidate` inserts the `artists` row first, then `artist_links`, with no rollback if the links insert fails. That leaves a permanently orphaned artist row (a name with zero links). Because `findExistingSlugs` only checks whether an `artists` row exists, re-running the script afterward will silently skip that slug forever — a transient link-insert failure permanently strands that artist rather than being retryable on the next run. Consider deleting the artist row on link-insert failure, or having the existing-slugs check also require ≥1 link.
- Off-by-one: link dedup via `Map` keyed by platform is correct; resulting `display_order` values can have gaps after dedup, but that's cosmetic only.
- API contract drift: none — writes the same `artists`/`artist_links` shape `persistSearchResults` already writes.

One plausible regression flagged (orphaned artist row on partial insert failure); everything else confirmed clean against the file's own stated safety properties.

---

brandonlucasgreen (commented):

**Security**
- No concerns. Insert-only ops script run manually with `SUPABASE_SERVICE_KEY` — the same pattern already used by other scripts in `scripts/`, never runtime/request-exposed. All DB access goes through structured Supabase client calls (`.insert()`, `.select().in()`), not raw SQL. — confirmed: no issue.
- `buildCandidate` reads `data/artists/${entry.slug}.json`, where `entry.slug` comes from the repo-committed `data/artists-manifest.json`, not from any external/user input — so building that path isn't a path-traversal concern here. — confirmed: no issue.
- No secrets logged (`requireEnv` fails loudly on missing vars without printing values).

**Regression**
- The `isDirectLink` / `EXCLUDED_PLATFORMS` change in `api/functions/db.ts` only adds `export` — behavior for existing callers (e.g. `persistSearchResults`) is unchanged. — confirmed: no issue.
- Script is correctly scoped insert-only: any slug that already has an `artists` row is skipped untouched (`findExistingSlugs` / `missing = manifest.filter(...)`), so claimed profiles and existing rows can't be affected — matters given the `artist_links` wipe incident referenced in the script's own comments. — confirmed: matches stated intent.
- Not something I can execute here, but worth a manual `--limit=5` dry run before the real `--write` pass, per the script's own usage note — flagging as **needs-manual-verification**, not a code defect.

No blocking findings.

---

PR #385, opened 2026-08-01
Base sitemap lastmod on the database, and stop listing URLs that 404

brandonlucasgreen:
Follow-on to #384. Those 736 pages came back to life today, but the sitemap still tells Google nothing has changed since March.

## The problem

Every artist URL carried the same date:

```
791 artist URLs, all <lastmod>2026-03-24</lastmod>
```

That's `lastUpdated` from the manifest — the day the **static file** under `data/artists/` was generated. Nothing has rendered those files since #274 moved the page onto `/api/artist-page`, so the date describes an artifact no visitor ever sees. It's not merely stale, it's measuring the wrong thing.

`lastmod` is only a hint, and Google weighs it by how accurate it's historically been — but right now it points actively the wrong way on 736 URLs that just changed.

## The change

**`lastmod` comes from `artists.updated_at`.** The resulting spread:

```
737  2026-08-01   ← the backfill
  6  2026-03-28
  5  2026-06-25
  4  2026-06-13
  …  genuine per-artist dates
```

A real distribution, not a blanket "everything changed today" — which is what keeps the field worth trusting. Future builds keep tracking real per-artist changes.

**A manifest slug with no artist row is omitted rather than advertised.** This is the standing fix for what #384 cleaned up by hand: the sitemap can no longer promise Google a URL that 404s. When it drops one it says so, and points at `npm run backfill:artist-rows`.

## Failure handling

Supabase being unreachable is treated as its own case, not folded into "artist missing" — otherwise one expired key would silently ship a sitemap containing zero artists. On missing credentials or a query error it warns and falls back to the old manifest-date behaviour, listing everything. A genuine fault still fails the build rather than deploying whatever was on disk.

## Verification

All four paths exercised:

| path | result |
|---|---|
| real credentials | 791 URLs, real dates |
| no credentials | ⚠️ warns, 791 URLs, manifest dates |
| rejected API key | ⚠️ warns, 791 URLs, manifest dates |
| manifest slug with no row | omitted + actionable message |

Also ran the exact build invocation (`cd apps/web && npx tsx ../../scripts/generate-sitemap.ts` — the build runs it from `apps/web`, so the relative path and module resolution matter) and `npx tsc -b`. Lint unchanged from baseline.

**One thing to confirm on the deploy preview:** whether `SUPABASE_URL` / `SUPABASE_SERVICE_KEY` are scoped to *builds* and not just functions. If they aren't, this degrades to today's behaviour and the fix is a scope change in Netlify, not code. I'll check the preview's `sitemap.xml` once it's built.

## After merging

Worth hitting **"Validate fix"** on the 404 report in Search Console — that asks Google to re-check in bulk rather than waiting for its own crawl schedule, which is the stronger lever of the two.

### DIFF
--- a/apps/web/public/sitemap.xml
+++ b/apps/web/public/sitemap.xml
@@ -17,4747 +17,4747 @@
   </url>
   <url>
     <loc>https://unstream.stream/artist/12-stones</loc>
-    <lastmod>2026-03-24</lastmod>
+    <lastmod>2026-08-01</lastmod>
     <changefreq>monthly</changefreq>
     <priority>0.7</priority>
   </url>
   <url>
     <loc>https://unstream.stream/artist/808-state</loc>
-    <lastmod>2026-03-24</lastmod>
+    <lastmod>2026-08-01</lastmod>
     <changefreq>monthly</changefreq>
     <priority>0.7</priority>
   </url>
   <url>
     <loc>https://unstream.stream/artist/aaron-carter</loc>
-    <lastmod>2026-03-24</lastmod>
+    <lastmod>2026-08-01</lastmod>
     <changefreq>monthly</changefreq>
     <priority>0.7</priority>
   </url>
   <url>
     <loc>https://unstream.stream/artist/abbey-lincoln</loc>
-    <lastmod>2026-03-24</lastmod>
+    <lastmod>2026-08-01</lastmod>
     <changefreq>monthly</changefreq>
     <priority>0.7</priority>
   </url>
   <url>
     <loc>https://unstream.stream/artist/above-beyond</loc>
-    <lastmod>2026-03-24</lastmod>
+    <lastmod>2026-08-01</lastmod>
     <changefreq>monthly</changefreq>
     <priority>0.7</priority>
   </url>
   <url>
     <loc>https://unstream.stream/artist/absurd</loc>
-    <lastmod>2026-03-24</lastmod>
+    <lastmod>2026-08-01</lastmod>
     <changefreq>monthly</changefreq>
     <priority>0.7</priority>
   </url>
   <url>
     <loc>https://unstream.stream/artist/ace-frehley</loc>
-    <lastmod>2026-03-24</lastmod>
+    <lastmod>2026-08-01</lastmod>
     <changefreq>monthly</changefreq>
     <priority>0.7</priority>
   </url>
   <url>
     <loc>https://unstream.stream/artist/achille-lauro</loc>
-    <lastmod>2026-03-24</lastmod>
+    <lastmod>2026-08-01</lastmod>
     <changefreq>monthly</changefreq>
     <priority>0.7</priority>
   </url>
   <url>
     <loc>https://unstream.stream/artist/adam-jones</loc>
-    <lastmod>2026-03-24</lastmod>
+    <lastmod>2026-08-01</lastmod>
     <changefreq>monthly</changefreq>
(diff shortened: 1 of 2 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Automated security/regression pass on this PR (`scripts/generate-sitemap.ts` is the only substantive change; `sitemap.xml` is generated output).

**Security**

- Auth/authorization: N/A — build-time script only, no new endpoint or access path.
- Input validation: slugs are passed through supabase-js's parameterized `.in('slug', ...)` — no injection surface. *confirmed no issue*
- Secret handling: `SUPABASE_URL`/`SUPABASE_SERVICE_KEY` are read from env and never logged; absence triggers a warning + fallback rather than a crash or a leaked value. *confirmed no issue*
- Injection risks: none — no raw SQL, no string-built queries.
- Unsafe defaults: **needs-manual-verification** — `fetchArtistUpdatedAt` only falls back to the old manifest-only behavior when Supabase returns an explicit `{ error }`. If the client call *throws* instead (a real network exception rather than a resolved error field — DNS failure, abort, etc.), that exception isn't caught locally; it propagates to the new `main().catch(error => { ...; process.exit(1); })` at the bottom, which fails the whole build/deploy instead of the documented "fall back to manifest dates" path. Sitemap generation previously could never fail the build at all, so this is a new failure mode. Worth confirming supabase-js reliably resolves network failures into `{error}` rather than throwing — if it can throw, a transient Supabase blip now blocks every deploy.
- Data exposure: none — sitemap contents are already public.

**Regression**

- Changed assumptions: `lastmod` now sourced from `artists.updated_at` instead of the manifest's `lastUpdated`, and slugs with no artist row are omitted rather than always listed. Matches the PR's stated intent and pairs correctly with #384's backfill.
- Edge cases: null `updated_at` per-row falls back to the manifest date; a null lookup result (Supabase unreachable) falls back to listing everything. Both handled.
- Off-by-one: chunked `.in()` lookups (`i += 200`) look correct.
- State transitions: `main()` becoming async + hard `process.exit(1)` on any uncaught error is the one behavior change worth double-checking (see unsafe-defaults note above) — previously this script could not fail the build.
- API contract drift: sitemap XML schema (`loc`/`lastmod`/`changefreq`/`priority`) is unchanged.

No confirmed security issues. One plausible build-reliability edge case flagged above for verification.

---

brandonlucasgreen (commented):

**Security**
- No concerns. `fetchArtistUpdatedAt` uses the existing `SUPABASE_URL`/`SUPABASE_SERVICE_KEY` pattern from a build-time script (not request-exposed), queries via the Supabase client's structured `.in()` filter (no raw SQL / string interpolation), and doesn't log secret values. — confirmed: no issue.

**Regression**
- Verified the null-vs-empty distinction this PR depends on: `fetchArtistUpdatedAt` returns `null` (not an empty `Map`) on any per-chunk Supabase error, and `main()` only omits a manifest slug when `updatedAt` is a real map that doesn't contain it. So a lookup failure correctly falls back to "list everything with manifest dates" instead of silently gutting the sitemap to zero/near-zero URLs. — confirmed: correct.
- Checked for a subtler variant: could an artist row that *exists* but has a null `updated_at` get misclassified as "no row" and wrongly dropped from the sitemap? `supabase/schema.sql` has `updated_at timestamptz not null default now()` on `artists`, so that value can never be null — this edge case can't occur. — confirmed: no issue.

No blocking findings.

---

PR #387, opened 2026-08-01
Link pre-generated artist pages from search results and a new footer index

brandonlucasgreen:
## Summary
- Search results now link to an artist's pre-generated `/artist/:slug` page ("View artist page") whenever we already have a DB row for them, even when a live platform result wins the same-name merge — previously that slug was silently discarded (`mergeStoredArtistsIntoResults`).
- New "Artists You Know" index at `/known-artists`, linked in the footer next to the existing "Indie Artist Index", listing all unclaimed-but-verified artists (`/api/artist-directory?scope=known`).
- `ArtistDirectoryPage` and the new `KnownArtistsPage` now share one `ArtistIndexPage` component instead of duplicating the fetch/letter-grouping/render logic.

## Test plan
- [x] `npm run lint`
- [x] `npx tsc -b` (root + `apps/web`)
- [x] `npm run test:unit` (529 passed)
- [x] `npm run test:api` (541 passed, incl. 2 new test files: `to-stored-result.test.ts`, `artist-directory.test.ts`)
- [x] Verified in the dev browser: footer link renders and routes via SPA navigation, `/known-artists` and `/artists` both render their loading/empty states correctly (live data needs a deploy preview — the dev API stub doesn't implement `/api/artist-directory` or set `matchConfidence`, a pre-existing gap)


api/functions/search-utils.ts
```diff
@@ -1047,8 +1052,14 @@ export function mergeStoredArtistsIntoResults(
...
 
     if (artist.matchConfidence !== 'claimed') {
-      // A stored non-claimed artist only fills a hole.
-      if (sameNameIdx === -1 && !merged.some(r => normalizeForComparison(r.name) === normalizeForComparison(artist.name))) {
+      // A stored non-claimed artist only fills a hole — but if a live result
+      // already covers this name, its knownSlug is worth keeping so the UI
+      // can still link to the pre-generated /artist/ page.
+      if (sameNameIdx !== -1) {
+        if (artist.knownSlug && !merged[sameNameIdx].knownSlug) {
+          merged[sameNameIdx] = { ...merged[sameNameIdx], knownSlug: artist.knownSlug };
```
brandonlucasgreen, line 1060: **Regression — plausible, needs-manual-verification.** @brandonlucasgreen

`sameNameIdx` (line ~1050) matches on `r.matchConfidence !== 'claimed'`, so it's satisfied by both `'verified'` **and** `'unverified'` live results — there's no exclusion of `'unverified'` here. That matters because this file already documents, at the `unverifiedReason` field (lines 73–80), that an `'unverified'` result with `unverifiedReason: 'conflicting-releases'` specifically means "this may be a different artist with the same name" — release data didn't match its verified sibling. The UI (`ResultCardHeader.tsx`) even surfaces a warning badge for exactly this case.

In the common path this is safe: `filterAndSort`'s tie-break (line ~1096-1097) always sorts the `'verified'` sibling ahead of the `'unverified'` one when names tie, so `findIndex` here picks the verified entry first. But if that verified sibling gets dropped entirely by `filterAndSort`'s search-only-link filter (line ~1088-1090) — e.g. all its surviving platforms are search-only links — only the `'unverified'`/`conflicting-releases` entry remains in `results`, and it would receive a `knownSlug` (and therefore a clickable "View artist page" link) sourced from a DB row that may belong to an entirely different real-world artist who merely shares a name.

Given `unverifiedReason: 'conflicting-releases'` exists precisely to flag "might not be the same artist," I'd suggest explicitly excluding `matchConfidence === 'unverified'` (or at least `unverifiedReason === 'conflicting-releases'`) from `sameNameIdx` eligibility here, the same way `'claimed'` is already excluded. Flagging as plausible rather than confirmed since I traced the mechanism through the code but didn't have a live data case to reproduce the precondition (verified sibling filtered out) end-to-end.

---

brandonlucasgreen (commented):
@brandonlucasgreen Senior-engineer pass on this PR (security + regression). No open PRs besides this one.

## Security

- **Auth/authorization** — no new authenticated actions. `GET /api/artist-directory?scope=known` is public with no auth check, matching the existing default-scope branch in the same handler (unauthenticated, public read of claimed-artist directory data). Not a new exposure surface. **confirmed** (non-issue, pattern predates this PR).
- **Input validation** — the new `scope` query param is only ever compared with strict equality (`=== 'known'`); it never reaches a query string or template. All Supabase calls in the diff use static column/value predicates via the query builder (`.eq('match_confidence', 'verified')`). **confirmed** no injection risk.
- **Secret handling** — untouched by this diff.
- **Injection risks** — none found; no raw SQL, no string-built queries, no `dangerouslySetInnerHTML`. The new `<Link to={`/artist/${result.knownSlug}`}>` uses React Router's `to`, not an href built from unescaped HTML — same pattern as the existing `claimedSlug` link. **confirmed** no issue.
- **Unsafe defaults** — none introduced.
- **Data exposure** — `scope=known` returns `slug, name, image_url` for all `match_confidence='verified'` (unclaimed) artists. Same shape/sensitivity as the pre-existing claimed-artist directory response; no PII, no auth-gated data leaked. **confirmed** no issue.

## Regression

- **Changed assumption** — flagged inline on `search-utils.ts`. `mergeStoredArtistsIntoResults`'s "never displaces a live result" guarantee still holds for identity, but the new `knownSlug`-attachment branch matches on name equality without excluding `matchConfidence === 'unverified'`/`conflicting-releases` results — which the codebase already documents as "may be a different artist with the same name." The main pipeline order (verified-before-unverified on a name tie in `filterAndSort`) protects against this in the common case, but not if the verified sibling gets filtered out first. **plausible**, see inline comment for the trace.
- **Edge cases / same-name collisions** — the `sameNameIdx` heuristic (string equality + article-insensitive match, no MBID/URL cross-check) is pre-existing for the `'claimed'` replace-in-place path, but this PR is the first to attach a user-clickable link to a live result on that heuristic alone. Two distinct real-world artists with the same name could get cross-linked. Same root cause as the point above. **plausible**.
- **Off-by-one / state transitions** — none applicable, no loops or counters changed in a way that affects boundaries.
- **API contract drift** — `knownSlug` added to both `AggregatedResult` (api) and `SearchResult` (web) types, mirroring how `claimedSlug` is already duplicated across both — no new drift pattern. `artist-directory.ts`'s `handler()` → `handler(event)` signature change is safe: Netlify always passes the event regardless of whether the function's signature declared a parameter, and this matches every other handler in `api/functions/`. **confirmed** no issue.

Test coverage for the new behavior (`toStoredResult`, `mergeStoredArtistsIntoResults`) looks solid and the shared-component extraction (`ArtistIndexPage`) is a clean dedup. Only blocking-adjacent item is the inline one — recommend excluding `unverified`/`conflicting-releases` results from `sameNameIdx` eligibility before merge, otherwise non-blocking.

---

PR #388, opened 2026-08-02
Releases steps 10, 5 and 6: Jam.coop, catalog-backed alerts, per-fan feeds

brandonlucasgreen:
Completes the Releases V1 scope. Steps 10, 5 and 6, in that order.

## Step 10 — Jam.coop ingest

**Mirlo was re-checked live rather than trusted from the spec:** `api.mirlo.space/robots.txt` still carries `Disallow: /v1/` under its `# Disallow crawling the API` comment, so it stays paused. Subvert still answers HTTP 429 to everything.

**Jam.coop was built instead, and it is the best-shaped source after Bandcamp:**
- Its `robots.txt` contains a single comment and **no `Disallow` at all**
- It is entirely server-rendered (unlike Bandwagon, which returned nothing)
- **One album-page fetch yields everything** — title, artwork, date, price, currency, format. Faircamp needs two requests and never has a date; Bandcamp needs grid-plus-detail.

Two calls worth reviewing:
- **`£7.00 or more` publishes as `7.00`.** The floor is what a fan can actually pay — the same call `ingestFaircampPurchasePage` makes about Faircamp's `data-min`. A genuine `£0.00 or more` stays `0`, which renders "Name your price".
- **Only `£ € $` are mapped.** Any other symbol yields *no offer*, because `formatMoney` defaults a null currency to USD and would render "¥800" as "$800".

Verified against live markup before the parser was written, and the four parser guards were mutation-tested.

## Step 5 — alerts come from the catalog

`check-releases` reads the catalog and falls back to the live scrape **only when an artist has never been catalogued**. That distinction is load-bearing: `null` means "we have not looked", `[]` means "we looked and there is nothing". Collapsing them would either switch alerts off for every artist nobody has saved yet, or re-scrape Bandcamp for everyone forever.

| # | Spec §5 defect | Status |
|---|---|---|
| 1 | Second release from one artist lost permanently (Mac) | ✅ Fixed — see below |
| 2 | Upcoming releases impossible to alert on | ✅ Fixed. The catalogue holds **2 future-dated releases** that could never have fired before |
| 3 | Only latest release per platform seen | ✅ Fixed on the catalog path |
| 4 | Multi-platform releases collapsed | ✅ Fixed — all platforms, artist-paying first. The hardcoded list is *deliberately* left on the live-scrape path (see code comment) |
| 5 | Closed laptop loses releases | 🟡 Partial — server accepts `sinceDays`; shipped clients don't send it yet |
| 6 | Alert state doesn't sync across devices | ❌ Not built — needs a synced table + both clients. **Decide separately** |
| 7 | Thin notification body | ✅ Fixed |

**Defect 1 is the one that lost data.** The outer dedup in `checkAllArtists` matched on artist *name* against the display list, and ran *after* `checkViaAPI` had already recorded the release as known — so an artist's genuinely-new second record was dropped from both the list and the notification while being marked as seen, making it undetectable forever after. Dedup is now a single pure `selectUnseen`, with a regression test that replays the two-check scenario. Neutralising the fix fails that test.

**Backwards-compatible:** `release` (singular) is still populated for the shipped Mac app (3.3.x) and extension (2.5.x); everything else is additive. Both decode `platform` as a plain string and ignore unknown keys.

## Step 6 — per-fan feeds

`/feed/f/{token}.ics|.xml` (private, primary), plus public `/u/{handle}/releases.*` (gated on the existing sharing opt-in) and `/a/{artist}/releases.*`. Token management in `/settings`.

**The token is stored plaintext, unlike `api_keys` which are hashed** — a hashed token could never be shown to its owner again, and a calendar URL has to be re-readable on a new device. It is treated as a credential everywhere else: owner-only RLS, never logged, `private, no-store`, `noindex`, and **404 rather than 401** on a bad token so it can't be probed.

**Deliberate deviation from spec §3, easy to revert.** §3 says *upcoming only* ("a calendar of past releases is a changelog"). This ships upcoming **plus a 30-day trailing window**. Measured first: 621 releases, 613 dated, but only **2 future-dated** and **9** within thirty-days-plus-future — upcoming-only would leave nearly every calendar empty. One constant (`FEED_TRAILING_DAYS`).

Two routing traps: `/a/{artist}/releases.xml` matches the two-segment `release-page` edge pattern and `/u/{handle}/releases.ics` matches `/u/*`, and **edge functions run before redirects** — hence `excludedPattern` on both. And `event.path` is unreliable behind a `status = 200` rewrite; `rawUrl` is not.

## Verification

- Build ✓ · **631 API tests** ✓ · **528 web unit tests** ✓ · **21 Mac XCTests** ✓
- Every new security/correctness guard was **mutation-tested** — neutralise it, confirm the test fails, restore. That caught two of my own tests passing for the wrong reason, one of which was guarding **unreachable dead code** (now deleted).
- **Rendered a real ICS from production data**, not just mocks: Boy Harsher's *GET MEAN* (2026-09-18) — valid, correctly folded, CRLF, `On Bandcamp / from $9 · ≈$7.20–$7.65 to artist`.
- Settings UI verified in dark, light and mobile.

## Migration

`20260802000000_release-feed-tokens.sql` — new `release_feed_tokens` table with owner-only RLS. Applies automatically on merge.

## Not in this PR

- **Defect 6** (cross-device alert state).
- **Native release UI** for the Mac app and extension — explored separately; needs a `get-release` endpoint first.
- **Jam.coop's payout percentage** — they take 15% (min 20p), so it isn't the `null` the registry currently has. Follow-up.

## Two pre-existing problems noticed, not fixed

- `npm run migrate:dry-run` / `migrate:list` are broken — they pass `--project-ref`, which the Supabase CLI no longer accepts.
- A junk row is in the catalogue: *"NanoPolix Nano Car Cloth: Automotive Paint Maintenance and Surface Care"*, almost certainly a bad Discogs artist match.

### DIFF
--- a/api/edge/release-page.ts
+++ b/api/edge/release-page.ts
@@ -55,17 +55,25 @@ const CSS = `
   .site-header { padding: 16px; border-bottom: 1px solid var(--border); display: flex; align-items: center; justify-content: space-between; gap: 16px; }
   .site-header .brand { display: flex; align-items: center; gap: 8px; font-size: 20px; font-weight: 700; color: var(--text); text-decoration: none; flex-shrink: 0; }
   .site-header .brand:hover { opacity: 0.8; }
-  .site-header .header-search { display: flex; gap: 8px; flex: 1; max-width: 420px; margin: 0 auto; }
-  .site-header .header-search input { flex: 1; min-width: 0; background: var(--bg2); border: 1px solid var(--border); border-radius: 8px; padding: 8px 12px; font-size: 14px; color: var(--text); font-family: inherit; }
+  /* Matches the React HeaderSearch this page can't render: a magnifying glass inside the
+     field and no submit button. Enter submits the form, which is what the icon implies and
+     what every other page in the app does — a separate "Search" button only appeared here
+     because plain SSR made it the obvious way to submit. */
+  .site-header .header-search { position: relative; flex: 1; max-width: 420px; margin: 0 auto; }
+  .site-header .header-search-icon { position: absolute; left: 12px; top: 50%; transform: translateY(-50%); width: 16px; height: 16px; color: var(--muted); pointer-events: none; }
+  .site-header .header-search input { width: 100%; min-width: 0; background: var(--bg2); border: 1px solid var(--border); border-radius: 8px; padding: 8px 12px 8px 36px; font-size: 14px; color: var(--text); font-family: inherit; }
+  .site-header .header-search input::placeholder { color: var(--muted); }
   .site-header .header-search input:focus { outline: none; border-color: var(--accent); }
-  .site-header .header-search button { border: 0; border-radius: 8px; padding: 8px 14px; background: var(--accent); color: #fff; font-size: 14px; font-weight: 500; cursor: pointer; font-family: inherit; }
   @media (max-width: 640px) { .site-header .header-search { display: none; } }
   /* Unlike artist-page-static, this page has no SPA counterpart to hand real browsers off to
      (see the note at the top of this file) — every visitor sees this exact header, so it's the
      only place a signed-out fan following a release alert can reach the login page. */
   .site-header .nav-right { display: flex; align-items: center; flex-shrink: 0; }
-  .site-header .nav-link { color: var(--muted); text-decoration: none; font-size: 14px; }
-  .site-header .nav-link:hover { color: var(--text); }
+  /* A filled accent button, matching the React header's Login. It was a muted text link here,
+     which read as secondary navigation rather than the primary action for a signed-out fan
+     arriving from a release alert. */
+  .site-header .nav-button { padding: 8px 16px; border-radius: 8px; background: var(--accent); color: #fff; text-decoration: none; font-size: 14px; font-weight: 600; box-shadow: 0 1px 2px rgba(0,0,0,0.2); }
+  .site-header .nav-button:hover { opacity: 0.9; }
 
   .release-head { display: flex; gap: 20px; align-items: flex-start; }
   .release-art { width: 160px; height: 160px; flex-shrink: 0; border-radius: 12px; border: 1px solid var(--border); background: var(--bg2); object-fit: cover; }
@@ -318,11 +326,11 @@ export default async function handler(request: Request, context: Context) {
          HeaderSearch. Submitting lands on /?q=… where the SPA renders results. Never point
          this at /search — that URL belongs to the noscript-search edge function. -->
     <form class="header-search" action="/" method="get" role="search">
+      <svg class="header-search-icon" fill="none" stroke="currentColor" viewBox="0 0 24 24" aria-hidden="true"><path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M21 21l-6-6m2-5a7 7 0 11-14 0 7 7 0 0114 0z"/></svg>
       <input type="text" name="q" placeholder="Search artists..." aria-label="Search artists" enterkeyhint="search">
-      <button type="submit">Search</button>
     </form>
     <div class="nav-right">
-      <a href="/login" class="nav-link">Login</a>
+      <a href="/login" class="nav-button">Login</a>
     </div>
   </header>
 
--- a/api/functions/__tests__/check-releases-ssrf.test.ts
+++ b/api/functions/__tests__/check-releases-ssrf.test.ts
@@ -14,6 +14,9 @@ const mocks = vi.hoisted(() => ({
   checkRateLimit: vi.fn(),
   getClientIp: vi.fn(() => '1.2.3.4'),
   isStoredArtistLink: vi.fn(),
+  // Null by default: "this artist has never been catalogued", which is what sends every test
+  // in this file down the live-scrape path the SSRF rules govern.
+  getReleasesForAlerts: vi.fn(async () => null),
   fetch: vi.fn(),
   lookup: vi.fn(),
 }));
@@ -32,6 +35,7 @@ vi.mock('../ratelimit', () => ({
 
 vi.mock('../db', () => ({
   isStoredArtistLink: mocks.isStoredArtistLink,
+  getReleasesForAlerts: mocks.getReleasesForAlerts,
 }));
 
 import { handler } from '../check-releases';
@@ -63,6 +67,10 @@ beforeEach(() => {
   vi.clearAllMocks();
   mocks.checkRateLimit.mockResolvedValue({ limited: false });
   mocks.isStoredArtistLink.mockResolvedValue(false);
+  // No catalog for this artist, so every test here exercises the live-scrape path whose SSRF
+  // behaviour this file exists to pin. Set explicitly rather than relying on clearAllMocks
+  // leaving the hoisted default in place.
+  mocks.getReleasesForAlerts.mockResolvedValue(null);
   // Default: everything resolves to an ordinary public address.
   mocks.lookup.mockResolvedValue([{ address: '93.184.216.34', family: 4 }]);
   vi.stubGlobal('fetch', mocks.fetch);
@@ -485,7 +493,13 @@ describe('rate limiting', () => {
 
 describe('backward compatibility with shipped clients', () => {
   // The Mac app (3.3.x) and extension (2.5.x) are already released against this contract.
-  it('keeps the response shape', async () => {
+  //
+  // This used to assert the exact key set, which would forbid ever adding a field. That is
+  // stricter than the actual constraint: the Mac client decodes into a Swift `Codable` struct
+  // and the extension reads named properties off the parsed object, so both ignore keys they
+  // don't know. What they cannot survive is a *removed* or *retyped* field — so that is what
+  // is pinned here, on both response paths.
+  it('keeps the legacy fields a shipped client decodes', async () => {
     mocks.fetch.mockResolvedValue(res(200, { body: '<html></html>' }));
 
     const r = await post({
@@ -494,8 +508,49 @@ describe('backward compatibility with shipped clients', () => {
     });
 
     const parsed = JSON.parse(r.body);
-    expect(Object.keys(parsed).sort()).toEqual(['artistName', 'release']);
-    expect(parsed.artistName).toBe('Someone');
+    expect(parsed).toHaveProperty('artistName', 'Someone');
+    expect(parsed).toHaveProperty('release');
+    expect(parsed.release === null || typeof parsed.release === 'object').toBe(true);
+  });
+
+  it('keeps the legacy release fields when the catalog answers', async () => {
+    mocks.getReleasesForAlerts.mockResolvedValue({
+      artistSlug: 'someone',
+      releases: [
+        {
+          slug: 'a-record',
+          title: 'A Record',
+          releaseDate: '2026-07-20',
+          datePrecision: 'day',
+          status: 'released',
+          artworkUrl: null,
+          sources: [
+            {
+              platform: 'bandcamp',
+              url: 'https://someone.bandcamp.com/album/a-record',
+              offers: [{ price: 8, currency: 'USD', availability: 'available' }],
+            },
+          ],
+        },
+      ],
+    });
+
+    const parsed = JSON.parse((await post({
+      artistName: 'Someone',
+      platforms: { bandcamp: 'https://someone.bandcamp.com' },
+    })).body);
+
+    // Exactly the four keys ReleaseCheckAPI.swift's APIRelease declares, all non-optional there.
+    expect(parsed.release).toMatchObject({
+      releaseName: 'A Record',
+      releaseDate: '2026-07-20',
+      platform: 'bandcamp',
+    });
+    expect(typeof parsed.release.releaseUrl).toBe('string');
+    // The Swift client parses this with a yyyy-MM-dd formatter and throws otherwise.
+    expect(parsed.release.releaseDate).toMatch(/^\d{4}-\d{2}-\d{2}$/);
+    // Not fetched at all: the catalog answered.
+    expect(mocks.fetch).not.toHaveBeenCalled();
   });
 
   it('still rejects a request with no platform URLs', async () => {
--- a/api/functions/__tests__/feed-format.test.ts
+++ b/api/functions/__tests__/feed-format.test.ts
@@ -1,0 +1,270 @@
+// iCalendar and Atom generation.
+//
+// Worth testing carefully because the failure mode is *silent*: a malformed .ics doesn't error
+// in Apple Calendar, it just shows an empty subscription, and an unescaped comma truncates a
+// release title at the comma without anything looking wrong. None of that is visible from
+// reading the output — only from asserting the rules RFC 5545 actually imposes.
+
+import { describe, it, expect } from 'vitest';
+import {
+  buildAtom,
+  buildIcs,
+  escapeIcsText,
+  escapeXml,
+  foldIcsLine,
+  releasePageUrl,
+  type FeedRelease,
+} from '../../shared/feed-format';
+
+const NOW = new Date('2026-08-01T12:00:00Z');
+
+function release(over: Partial<FeedRelease> = {}): FeedRelease {
+  return {
+    artistName: 'Kid Lightbulbs',
+    artistSlug: 'kid-lightbulbs',
+    title: 'Infinite Normal',
+    releaseSlug: 'infinite-normal',
+    releaseDate: '2026-09-01',
+    offerSummary: 'from $8 · ≈$6.80 to artist',
+    platforms: ['Bandcamp', 'Mirlo'],
+    ...over,
+  };
+}
+
+/** Unfold a document back into logical lines, the way a real parser does. */
+function logicalLines(ics: string): string[] {
+  return ics.split('\r\n').reduce<string[]>((out, line) => {
+    if (line.startsWith(' ') && out.length > 0) out[out.length - 1] += line.slice(1);
+    else if (line) out.push(line);
+    return out;
+  }, []);
+}
+
+describe('escapeIcsText', () => {
+  // A raw comma is a value separator in RFC 5545, so an unescaped one silently truncates the
+  // title in the subscriber's calendar. Album titles have commas in them constantly.
+  it('escapes the four characters that change meaning', () => {
+    expect(escapeIcsText('Hello, World')).toBe('Hello\\, World');
+    expect(escapeIcsText('A; B')).toBe('A\\; B');
+    expect(escapeIcsText('back\\slash')).toBe('back\\\\slash');
+    expect(escapeIcsText('one\ntwo')).toBe('one\\ntwo');
+  });
+
+  // Backslashes must be escaped before the escapes this function adds, or those get re-escaped.
+  it('does not double-escape its own output', () => {
+    expect(escapeIcsText('a\\,b')).toBe('a\\\\\\,b');
+  });
+
+  it('handles CRLF and bare CR as one newline each', () => {
+    expect(escapeIcsText('a\r\nb')).toBe('a\\nb');
+    expect(escapeIcsText('a\rb')).toBe('a\\nb');
+  });
+});
+
+describe('foldIcsLine', () => {
+  it('leaves a short line alone', () => {
+    expect(foldIcsLine('SUMMARY:Short')).toBe('SUMMARY:Short');
+  });
+
+  it('folds a long line into 75-octet pieces with a leading space', () => {
+    const folded = foldIcsLine('SUMMARY:' + 'a'.repeat(200));
+    const lines = folded.split('\r\n');
+
+    expect(lines.length).toBeGreaterThan(1);
+    expect(lines[0]).toHaveLength(75);
+    for (const line of lines.slice(1)) expect(line.startsWith(' ')).toBe(true);
+    // Unfolding must reproduce the original exactly.
+    expect(lines.map((l, i) => (i === 0 ? l : l.slice(1))).join('')).toBe('SUMMARY:' + 'a'.repeat(200));
+  });
+
+  // The reason folding counts octets rather than characters: splitting mid-sequence produces
+  // mojibake or a parse error for a title like "Ágætis byrjun".
+  it('never splits a multi-byte character across the boundary', () => {
+    const folded = foldIcsLine('SUMMARY:' + 'é'.repeat(80)); // é is 2 bytes in UTF-8
+
+    for (const line of folded.split('\r\n')) {
+      expect(new TextEncoder().encode(line).length).toBeLessThanOrEqual(75);
+      // A broken split would leave a replacement character behind.
+      expect(line).not.toContain('�');
+    }
+    expect(folded.split('\r\n').map((l, i) => (i === 0 ? l : l.slice(1))).join(''))
+      .toBe('SUMMARY:' + 'é'.repeat(80));
+  });
+
+  it('keeps every emitted line within 75 octets', () => {
+    const folded = foldIcsLine('DESCRIPTION:' + 'ünstream '.repeat(40));
+    for (const line of folded.split('\r\n')) {
+      expect(new TextEncoder().encode(line).length).toBeLessThanOrEqual(75);
+    }
+  });
+});
+
+describe('buildIcs', () => {
+  it('produces a well-formed calendar with one event per release', () => {
+    const ics = buildIcs([release()], 'Unstream — Upcoming releases', NOW);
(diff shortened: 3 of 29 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Reviewed this as a security + regression pass. Scope covered: `me-feed-token.ts`, `feed-releases.ts`, `api/shared/feed-format.ts`, the `release_feed_tokens` migration, the `db.ts`/`check-releases.ts`/`catalog-artist-background.ts`/`release-ingest.ts` Jam.coop + alerts additions, `netlify.toml` routing, and the Mac `ReleaseAlertManager`/`ReleaseCheckAPI` changes. Overall this is unusually well-instrumented for a security review — the "never cache uncertainty" and null-vs-empty distinctions are applied consistently, redirect-hop SSRF re-validation was already shared infra and is reused correctly for Jam.coop, and ICS/XML escaping has direct injection tests.

## Security

**Auth/authorization** — no findings. `me-feed-token.ts` follows the established `me-*` bearer-against-anon-client pattern; the `release_feed_tokens` migration has owner-only RLS on SELECT/INSERT/UPDATE/DELETE, matching sibling tables. `needs-manual-verification`: I can't confirm from the diff alone whether Supabase's default schema privileges (not this migration) actually grant `authenticated` access to the new table — the repo's convention is to rely on project-level defaults rather than explicit `GRANT`s (consistent with every other migration here), so this is a pre-existing assumption, not a new gap, but worth a quick check in the Supabase dashboard post-deploy.

**Input validation** — no findings. `TOKEN_PATTERN` shape-checks before any DB query (confirmed by test: junk paths never call `getFeedTokenOwner`). `resolveWindowDays` clamps `sinceDays` to `(0, 365]` server-side, closing off the "scrape a whole discography through the alerts endpoint" path called out in the PR description. Jam.coop currency parsing correctly refuses to guess a currency rather than defaulting to USD.

**Secret handling** — no findings. Feed tokens are plaintext-by-design (documented rationale for why, unlike hashed `api_keys`), never logged, `Cache-Control: private, no-store`, 404-not-401 on miss, and `Sentry.captureException` in `ReleaseFeedControls.tsx` only ever passes the JS `Error`, not response bodies containing the token.

**Injection** — no findings. All new queries go through the Supabase query builder (parameterized); no raw SQL. ICS (`escapeIcsText`) and Atom (`escapeXml`) escaping is applied to every user/artist-controlled string field and is covered by dedicated tests (comma-truncation, `&`/`<`/`>` in titles).

**Unsafe defaults** — `plausible`, low severity: `me-feed-token.ts` `GET` (api/functions/me-feed-token.ts:87-95) does read-then-write without a transaction — `getFeedToken` returns null, then `setFeedToken` upserts. Two concurrent first-time `GET`s from the same user (double-click, two open tabs) will each generate a different token and both call `upsert(..., { onConflict: 'user_id' })`; the second write silently invalidates the first token that was already returned and displayed to the user. Not exploitable by another party (requires the victim's own session), so I'm not flagging it as a security bug, but it's a real correctness edge case under "state transitions" below.

**Data exposure** — no findings. `getFeedReleasesForHandle` and the token path both collapse "doesn't exist" and "exists but not authorized" into the same 404, preventing enumeration — and it's asserted by test (`feed-releases.test.ts`: "404s a handle that has not opted into sharing").

## Regression

**Changed assumptions** — `check-releases.ts`'s `isWithinWindow` dropped the old `daysDiff >= 0` floor so future-dated releases pass (intentional, per PR description). Note this also removes any *upper* bound on how far in the future a release can be: a release dated years out would satisfy `daysDiff <= windowDays` forever (since `daysDiff` is negative and unbounded below zero). The catalog path (`getReleasesForAlerts`) has the same shape (`.gte('release_date', since)`, no upper cap). `needs-manual-verification`: given today's Bandcamp/Jam.coop preorder conventions this is probably moot in practice, but it's a shared assumption across both new code paths worth confirming intentional rather than an oversight in the future-date fix.

**Edge cases / null vs empty** — no findings; this is the part of the PR that's most carefully handled (`getReleasesForAlerts` returning `null` vs `[]`, `getFeedTokenOwner` returning `null` for both malformed and unknown tokens, `getFeedReleasesForHandle` collapsing two negative cases into one).

**Off-by-one logic** — no findings in the reviewed date/window math (`nextDay` for exclusive DTEND, UTF-8-safe `foldIcsLine` boundary handling).

**State transitions** — see the token-rotation race above (`plausible`). Also worth a glance: `ReleaseAlertManager.selectUnseen` (apps/mac/Unstream/ReleaseAlertManager.swift:88-122) is now the single dedup point and mutates `state` via `inout` — correct per the regression test description, but confirm the `for result in results` loop order is stable/deterministic if `checkReleases` ever returns results in a different order between calls, since `addKnownRelease` is called incrementally inside the loop.

**API contract drift** — `platform` widened from a closed union (`'bandcamp' | 'faircamp' | 'mirlo'`) to `string` server-side; confirmed backward-compatible since both shipped clients decode it as a plain string and ignore unknown keys (stated in the PR description, consistent with the Swift `Codable` structs using optional fields for everything new).

## Non-blocking observation

`db.ts`, `check-releases.ts`, `catalog-artist-background.ts`, and `release-ingest.ts` together account for most of this PR's ~1,700 new backend lines, and none of them are in `api/tsconfig.json`'s typecheck `include` (only `feed-releases.ts`/`me-feed-token.ts`/`feed-format.ts` were added, correctly, per the existing convention). Not a regression you introduced, but flagging per the repo's own documented gotcha: a type error in this new code fails at runtime, not in CI.

---

brandonlucasgreen (commented):
@brandonlucasgreen Security/regression review (automated, scheduled staff-engineer pass). Reviewed against `origin/main` locally: `api/functions/{db,check-releases,release-ingest,catalog-artist-background,feed-releases,me-feed-token}.ts`, `api/shared/feed-format.ts`, `supabase/migrations/20260802000000_release-feed-tokens.sql`, `netlify.toml`, `apps/mac/Unstream/ReleaseAlertManager.swift`, `apps/mac/Unstream/Services/ReleaseCheckAPI.swift`, `apps/web/src/components/ReleaseFeedControls.tsx`.

**Security**

- **Auth/authorization** — `confirmed`, no issues. `me-feed-token.ts` uses the same bearer-against-anon-client pattern as the other `me-*` endpoints; writes go through the service-role client only after that succeeds. `release_feed_tokens` has RLS with owner-only SELECT/INSERT/UPDATE/DELETE policies, and the serving path (`feed-releases.ts`) correctly uses the service-role client since an anonymous calendar client presents no session — the token *is* the authorization there, by design, and that's the right call for this use case (same "secret address" model Google/Apple Calendar use).
- **Secret handling** — `confirmed`, no issues. The feed token is 256-bit (`randomBytes(32)`, base64url), plaintext storage is deliberate and documented (must be re-readable, unlike hashed `api_keys`), never logged, `Cache-Control: private, no-store` on both the mint endpoint and the private feed response, bad token returns 404 not 401 (avoids inviting probing/enumeration), and rotation replaces in place so a leaked URL can be killed. This is a well-reasoned credential design.
- **Injection risks** — `confirmed`, no issues. `escapeIcsText`/`foldIcsLine` in `feed-format.ts` correctly escape backslash/`;`/`,`/newlines for RFC 5545 (order-sensitive, done right) and `escapeXml` covers the five XML entities for the Atom output. All Supabase access goes through the query builder (no raw SQL string concatenation) — no SQL injection surface in the new `db.ts` functions.
- **Unsafe defaults / SSRF** — `confirmed`, no issues. `catalogJamcoop` checks `isUrlHostnameAllowed(storedUrl)` *before* calling `jamcoopArtistUrl()` and fetching — I checked this specifically because `jamcoopArtistUrl()` itself does not re-validate the hostname, only the URL shape, so if that ordering were ever reversed it would become an SSRF hole for a claimed artist's stored URL. As written today the guard is in place. `ingestJamcoopArtistPage` also filters candidate album links to `url.host === base.host`, so discovered album URLs can't drift off-host either.
- **Data exposure** — `confirmed`, no issues. `getFeedReleasesForHandle` returns `null` (→404) for both "no such handle" and "handle exists but hasn't opted into public sharing," so the public feed can't be used to enumerate handles or bypass the sharing opt-in. Token-shape validation (`TOKEN_PATTERN`) happens before any DB query, so malformed-path scanning costs nothing.

**Regression**

- **Changed assumptions** — `confirmed` intentional and correctly implemented: `isWithinLastMonth` → `isWithinWindow` removes the `daysDiff >= 0` floor so future-dated (announced) releases are no longer filtered out — this is the documented fix for spec §5 defect 2, and matches how `getReleasesForAlerts`'s SQL-side window (`gte release_date, since`, no upper bound) already worked.
- **Null vs. empty-array semantics** — `confirmed`, correctly preserved and tested. `getReleasesForAlerts` returns `null` for "never catalogued" (→ live-scrape fallback) vs. `[]` for "catalogued, nothing in window," and `check-releases.ts`'s handler respects that distinction. This is exactly the bug class the codebase's `CLAUDE.md` calls out as the most-repeated failure mode here, and it's handled correctly in both the new catalog path and the SSRF test file's explicit mock default.
- **Dedup regression (defect 1 fix)** — `confirmed` correct. The old `checkAllArtists` had two dedup layers: per-release inside `checkViaAPI`, then a second by-artist-name check in the caller that ran *after* the release was already marked known — so a second, genuinely-new release from the same artist was dropped from the UI *and* the notification while being permanently marked seen. `selectUnseen` consolidates to a single dedup point keyed on release name across platforms (correct: one record on both Bandcamp and Mirlo should be one alert) but distinct within an artist (correct: two different records should be two). This matches the PR's stated regression test.
- **API contract drift** — `confirmed`, backwards-compatible as claimed. `platform` narrowed from a closed union to `string` server-side, but both shipped clients (Mac 3.3.x via `Codable`, extension 2.5.x via named-property reads) decode it as a plain string and ignore unknown keys, and the singular `release` field is still populated alongside the new `releases` array. `ReleaseCheckAPI.swift` decodes `platforms`/`status`/`offerSummary` as optional with sane fallbacks (`?? [release.platform]`, `?? "released"`, `?? ""`), so an old-shape response (just `release`, no `releases`) still round-trips.
- **Edge cases** — `needs-manual-verification`: `resolveWindowDays`/`getReleasesForAlerts` place no *upper* bound on how far in the future a `release_date` can be and still qualify (only `windowDays` bounds how far *back*). Given the design intent (announced releases should always surface), this looks deliberate rather than a bug, but I'd confirm there's nothing upstream (Jam.coop/Faircamp/Discogs date parsing) that could hand this a garbage far-future date and produce a permanently-dangling "announced" alert. Low severity either way since it's cosmetic, not a data-integrity or security issue.
- **Routing traps** — `confirmed` correct. The two `/a/*`- and `/u/*`-adjacent edge functions get `excludedPattern` entries for the new `releases.ics|xml` paths, and the new redirects are placed above the SPA catch-all — matches the PR description's own callout that edge functions run before redirects, and I verified the regex shapes are correct two-segment excludes.

No blocking findings. This is a large, unusually well-documented PR (SSRF guard ordering, null/empty semantics, and the token credential model in particular show the right instincts) and the deploy preview is green. The one item worth a sanity check before merge is the far-future-date edge case above, but it doesn't block.

---

brandonlucasgreen (commented):
@brandonlucasgreen Senior-staff security/regression pass on this PR (full 4693-line diff read, plus cross-checks against the branch for `middleware.ts`'s SSRF allowlist, `safe-fetch.ts`, and `netlify.toml`).

## Security

**1. [secret-handling] needs-manual-verification** — `api/functions/db.ts` (`getFeedTokenOwner`/`getFeedToken`/`setFeedToken`/`deleteFeedToken`)
The module's own comment says the feed token must never be logged, and the known-error path correctly logs only `error.message`. But each function's outer `catch (error) { console.error('[DB] ... error:', error) }` logs the raw thrown error object, not a sanitized message. If the underlying Supabase client ever throws (network blip, timeout) with an error whose message/stack embeds the request URL — PostgREST filters like `.eq('token', token)` serialize as `?token=eq.<value>` — the raw feed token would land in Netlify function logs despite the explicit design intent. This is the same try/catch idiom used everywhere else in `db.ts`, so it's a pre-existing codebase pattern rather than something newly introduced by this PR, and I couldn't confirm from the diff alone whether supabase-js's thrown (as opposed to returned) errors actually embed the URL — hence "needs-manual-verification."

No SSRF, XSS, or injection issues found: the new Jam.coop ingest path is exact-match allowlisted in `middleware.ts` (not wildcarded) and goes through `safe-fetch.ts`'s DNS-resolution-based SSRF defense same as everything else; the feed XML/ICS builders use `escapeXml`/`escapeIcsText` consistently; and the feed-token endpoints return a uniform 404 for both "unknown token" and "not shared," matching the existing no-oracle convention used elsewhere in the codebase.

## Regression

**2. [changed-assumption] confirmed** — `api/functions/check-releases.ts` (`releases[0]` in the catalog branch) + `api/functions/db.ts`'s `getReleasesForAlerts`
This PR deliberately makes future-dated "announced" releases alert-eligible (a real fix — previously they were filtered out for being in the future and aged past the window before ever qualifying). But `getReleasesForAlerts` orders the windowed read purely by `release_date` descending with no split by `status`, and `check-releases.ts` then does `release: releases[0] ?? null` with the comment "the head is the one a single-release client should show." Verified: `toCatalogResults` passes `status` straight through unfiltered/unsorted. So if an artist has both an announced release dated further out and an already-released one from last week, the announced one sorts first and becomes `release`. The already-shipped Mac app and browser extension read only this single `release` field (no `status`), and render fixed "out now" copy — so a fan is told an unreleased pre-order is out now. `release-alerts.test.ts` doesn't appear to exercise this mixed-date-category case.

**3. [edge-case] confirmed** — `api/functions/db.ts`'s `getReleasesForAlerts`, the catalog-existence check
The "has this artist ever been catalogued" check is `count` on `releases` filtered only by `artist_id` — no `is_hidden`/`needs_review` filter. The actual windowed read that follows filters `is_hidden=false` and `needs_review=false`. If an artist's only catalogued release is flagged `needs_review=true` (a plausible outcome of the Jam.coop persist path's fuzzy-duplicate detector, added in this same PR), `count` returns 1 → treated as "catalogued" → the windowed query legitimately returns `[]` → the endpoint answers `source: 'catalog', release: null` and never falls through to the live scrape. This is the exact "count vs. filtered read disagree" shape the codebase's own comments call out as the most repeated bug class here, just triggered by an identity-uncertainty flag rather than a network failure.

**4. [off-by-one] confirmed** — `api/functions/check-releases.ts`'s `resolveWindowDays`
`if (typeof raw !== 'number' || !Number.isFinite(raw) || raw <= 0) return DEFAULT_WINDOW_DAYS;` only guards `<= 0`, so a fractional value like `0.5` passes through to `Math.floor(0.5)` → `0`, silently narrowing the alert window to "today or later only" instead of falling back to the 31-day default. No currently-shipped client sends `sinceDays`, so this is dormant risk rather than an active path today — but it'll bite the first client update that does.

No API-contract-drift issues beyond what's noted above — the new `releases`/`status` fields are additive alongside the legacy `release` field, so old clients keep parsing what they already expect (aside from finding #2).

---

brandonlucasgreen (commented):
@brandonlucasgreen — automated security/regression pass on this PR (three independent review passes: feed-token security surface, release-ingest/SSRF pipeline, mac/web client regression). One confirmed UI regression, nothing blocking.

## Security

**No confirmed or plausible vulnerabilities** in the new feed-token/RSS surface (`me-feed-token.ts`, `feed-releases.ts`, `feed-format.ts`, the new `release_feed_tokens` migration) or in the release-ingest/catalog pipeline (`release-ingest.ts`, `check-releases.ts`, `catalog-artist-background.ts`, the releases/alerts portions of `db.ts`, `search-parsers.ts`, `release-page.ts`). Verified clean: CSPRNG token generation, per-user DB scoping (no IDOR), fail-closed/enumeration-resistant token lookup, XML/ICS escaping in `feed-format.ts`, rate limiting, CORS posture, SSRF allowlisting on the new Jam.coop scrape path (mirrors the existing Bandcamp/Faircamp pattern — hostname derived from the already-allowlisted stored URL, not attacker input), and the "never cache uncertainty" null-vs-empty distinction in `getReleasesForAlerts`.

- **needs-manual-verification (low severity)** — `supabase/migrations/20260802000000_release-feed-tokens.sql`, `api/functions/db.ts`. `release_feed_tokens.token` is stored in plaintext rather than hashed like `api_keys`. Looks deliberate and defensible — a feed token has to be re-displayable to its owner on re-subscribe, unlike an API key — backed by 256 bits of CSPRNG entropy plus rotation/revocation. Flagging only because a DB/replica leak makes every token immediately usable with no cracking step, a different blast radius than a hashed-credential leak. No action needed unless you want to revisit the trade-off.

- **plausible** — `api/functions/check-releases.ts`, `isWithinWindow` (live-scrape path). The date-window filter dropped its lower bound (`daysDiff >= 0`) as part of the fix that lets *future* announced releases alert — correct intent. But there's now no upper bound on a scraped (not catalog-vetted) release date either. A garbled or placeholder far-future date parsed off a Bandcamp/Faircamp page could pass through and get pushed to a fan as a real upcoming release; the catalog path has more structure around this, the live-scrape path doesn't. Worth confirming `parseDateToISO` already bounds plausible years, or adding a cheap upper-cap (e.g. reject dates >1–2 years out) as defense in depth.

## Regression

- **confirmed** — `apps/mac/Unstream/Views/Shared/SupportListView.swift` (interacting with unchanged `ReleaseAlertManager.newRelease(for:)`). This PR intentionally removed the per-artist dedup so two releases from the same artist in one check window both get recorded as separate unread `NewRelease` entries — correct and tested. But `newRelease(for:)`, which `SupportListView`'s badge relies on, still does `.first { ... }` and only ever surfaces one. Scenario: an artist drops two records in the same check window → both alerts fire and are stored, but the Support List card only badges the first; the second stays invisible there until the first is dismissed (it does show correctly in the iOS Releases tab, which iterates the full list). Not data loss, but a visible gap between the new backend behavior and one of its UI surfaces.

- **needs-manual-verification** — `apps/web/src/components/ReleaseFeedControls.tsx`. On a failed rotate/revoke call, the `catch` block sets an error banner but leaves the displayed `ics`/`atom` URLs unchanged. If the server-side write actually succeeded but the response was lost (e.g. network drop after a successful rotate), the UI would keep showing a stale, no-longer-valid feed URL next to the error, with a working Copy button for a dead link. Couldn't confirm server-side idempotency/ordering from the diff alone.

- **test-coverage gap** — `apps/mac/UnstreamTests/ReleaseAlertTests.swift`. Good coverage of the dedup/notification fix this PR is actually about, but `ReleaseCheckAPI.swift` — the file most exposed to API-contract drift against the server changes here — has no decode/parse tests at that layer (legacy `release` vs. `releases` fallback, dual date-format parsing, skip-on-bad-date behavior are only exercised indirectly). Not a bug, just worth closing given how silently this kind of drift fails in Swift's optional decoding.

---

brandonlucasgreen (commented):
@brandonlucasgreen Security/regression review of this PR (large — reviewed in three sections: Jam.coop ingest, catalog-backed alerts, and the per-fan feed token/calendar). No speculation beyond what's visible in the diff; anything I couldn't confirm from the reviewed line ranges is labeled accordingly.

## Security

**1. ICS `UID:` field skips escaping (injection) — `api/shared/feed-format.ts`, `buildIcs`**
`UID:${release.artistSlug}-${release.releaseSlug}@unstream.stream` is built by raw interpolation, unlike every sibling TEXT field in the same function (`X-WR-CALNAME`, `X-WR-CALDESC`, `SUMMARY`, `DESCRIPTION`), which all go through `escapeIcsText`. RFC 5545 treats `UID` as a TEXT value subject to the same comma/semicolon/backslash/CRLF escaping. `releaseSlug`/`artistSlug` currently appear to come from URL path segments during ingest rather than a strict slugify, so I can't rule out an unusual character reaching this field. The missing escaping is **confirmed** as a code fact; real-world exploitability is **needs-manual-verification** (depends on the slug charset guarantee upstream, which is outside what I reviewed). Cheap fix regardless: route it through `escapeIcsText` like its neighbors, for consistency if nothing else.

**2. SSRF allowlist not re-checked on redirect for Jam.coop candidate-page fetches — `api/functions/catalog-artist-background.ts` (~1589–1628)**
`isUrlHostnameAllowed()` is called once against the stored artist URL. If that fetch redirects, `landedUrl = response.url || artistUrl` becomes the unchecked base for extracting album candidate links, and each candidate URL is passed straight to `safeFetch` with no further allowlist check. This mirrors the pre-existing Bandcamp pattern, so it isn't new to this PR, but it now applies to a second platform. Whether it's exploitable depends on whether `safeFetch` (not in this diff) re-enforces the allowlist on every fetch including redirect targets. **Plausible** — worth a quick check of `safe-fetch.ts`.

**3. `check-releases.ts` catalog branch runs before platform/allowlist validation (~1899–1925)**
When an artist is catalogued, the handler returns from the catalog branch without ever validating `request.platforms` or running the existing SSRF-allowlist check — not exploitable since that branch makes no outbound fetch, but it's a real behavior change: platform validation becomes conditional on whether the artist happens to be catalogued. **Plausible**, worth confirming intentional.

**4. `feed-releases.ts` `notFound()` omits `Cache-Control` (~2686–2692)**
Every success path explicitly sets `private, no-store` or `public, max-age=1800`; the 404 path sets `X-Robots-Tag` but no `Cache-Control`. Given the 256-bit token space this is low severity, but it's a real inconsistency with the "nothing about the private feed should be cacheable" posture the rest of the file enforces. **Plausible**.

**5. Token-validation timing difference (~2793–2801)** — a malformed token is rejected by regex before any DB call; a well-formed-but-unknown token incurs a DB round-trip. Response bodies/status are identical either way, so this is a timing side-channel, not a data leak — and with 256 bits of entropy in the actual token, negligible practical impact. **Plausible**, noted for completeness.

Everything else checked out clean against the specific claims in the PR description — confirmed, not just trusted: RLS is owner-only with no anon policy (`auth.uid() = user_id` on all four operations, migration `20260802000000_release-feed-tokens.sql`); token generation is `crypto.randomBytes(32)` (not a weak RNG); no full token value is ever logged or sent to Sentry; `me-feed-token.ts` derives `userId` solely from the caller's own bearer token with no client-suppliable ID that could target another user's token; unknown-handle and not-shared-handle both collapse to an identical 404 (no enumeration oracle); `netlify.toml` `excludedPattern`s correctly cover all the new feed routes against the pre-existing edge-function patterns; no SQL injection anywhere in the `db.ts` changes (parameterized Supabase builders throughout); no `fetch()` of user-supplied URLs in the feed-generation path.

## Regression

**No findings — the null-vs-`[]` catalog distinction is correctly preserved everywhere.** Traced `getReleasesForAlerts` end-to-end: it returns `null` only when the artist has never been catalogued (zero-count case) or on a genuine lookup/query error, and returns `{ releases: [] }` (truthy, non-null) when the artist is catalogued but nothing matches the window — and `check-releases.ts`'s `if (catalogued)` correctly treats that as "don't fall through to live scrape." The only `|| []` coercions in this diff are for a missing DB join (`release_sources`, `release_offers`), not the never-looked state. Client-side, `ReleaseCheckAPI.swift` only falls back to the legacy singular `release` field when `releases` is `nil`, not when the server sends an explicit empty array. This is the exact bug class the PR claims to fix (#defect 1 in the description), and it holds up.

**No findings — the dedup-ordering fix is effective.** `ReleaseAlertManager.selectUnseen` is now the single dedup point: check-then-mark-then-append happens together per result, in one pass, with no later second filter running after the "known" flag is already set. The regression test replaying the original two-check scenario (`testSecondReleaseSurvivesWhileTheFirstIsStillUnread`) passes and specifically exercises the ordering that broke before.

**No findings — window/boundary logic.** `resolveWindowDays` correctly rejects non-numbers/NaN/≤0 (defaults to 31) and caps at 365; the live-scrape and catalog-SQL paths use consistent, inclusive-at-exactly-N-days boundaries. Line-folding in the ICS output counts UTF-8 octets (not JS string length) and never splits a multi-byte sequence. Date-boundary math (`nextDay()`) is tested across a month rollover and a leap day.

**No findings — API contract widening is backward-compatible.** `platform` moved from a closed union to `string` on both server and Swift client (already decoded as plain `String`); new fields (`platforms`, `status`, `offerSummary`) are optional/defaulted on decode, with an explicit test for a `NewRelease` persisted by an older build.

One item I couldn't resolve within the reviewed scope: whether `getReleasesForAlerts`/persist functions in `db.ts` beyond line 2618 (the feed-token and feed-query functions) enforce ownership purely via the token being unguessable (by design, per the code's own comment) — I traced that this is intentional and consistent with the token's role as bearer authorization, not a gap.

---

PR #390, opened 2026-08-02
Make zero-result searches and post-deploy login failures legible in Sentry

brandonlucasgreen:
Two Sentry blind spots, both about not being able to see what actually happened.

## 1. Zero-result searches didn't show the query

The query was already being sent — as `extra`. That's the problem. Every zero-result event shares one message, so Sentry folds them into a single issue where `extra` belongs to whichever event you happen to be looking at. Seeing *which* searches come back empty meant paging through events one at a time.

Query and mode are now **tags** (`api/functions/search-sources.ts`), which get an aggregated value distribution on the issue page and are searchable (`search_query:radiohead`). The raw form stays in `extra`.

`search_mode` is included because it changes the follow-up: `exact` means a playback-detection client (extension, Mac app) sent a real artist name and we have no coverage; `fuzzy` means a human typed it and it may just be a typo.

Both values are already capped at 200 chars by `validateQuery`, which is also Sentry's tag-value limit. Volume is bounded by the existing 24h-per-query `checkSentryDedup`.

## 2. The "unexpected error" on login after a deploy

**It's a stale build chunk, and it isn't really about auth.** Reproduced end-to-end against a real production build with the chunk removed.

Every page in `main.tsx` is `lazy(() => import(...))`, so chunk filenames carry a content hash. A tab left open across a deploy clicks Login and asks for `LoginPage-<oldHash>.js`, which the new deploy doesn't have. `netlify.toml`'s SPA catch-all (`/*` → `/index.html`, 200) then answers with **`200 text/html` rather than 404** — verified against production:

```
curl -sI https://unstream.stream/assets/does-not-exist.js   # 200, content-type: text/html
```

So the browser rejects it as a module script, `React.lazy` rejects, and the error boundary renders "Unstream hit an unexpected error". Reloading works because it fetches the new `index.html`. The PWA compounds it: `registerType: 'autoUpdate'` + `skipWaiting` + `clientsClaim` lets the new service worker claim the *live* page without a reload, so the page is old while its precache is new.

**It was reaching Sentry all along** — as a bare `TypeError: Failed to fetch dynamically imported module`, with no route, no build, and a different title per browser engine. Delivery was confirmed by wrapping `client.getTransport().send` in the live page; note that `read_network_requests` never shows Sentry envelopes (keepalive fetch), so an empty network log is a false negative there.

Now classified in `beforeSend` with `error_type: stale-build-asset`, the `route` the user was trying to reach, the running release, and a single fingerprint. Classification happens *before* the existing benign-error filter so it can't be mistaken for a transient network blip and dropped.

The Sentry release now comes from Netlify's `COMMIT_REF` instead of the literal string `'unknown'` — "which build was this tab running" is the whole question here, and it was unanswerable. Passed to `sentryVitePlugin` too so uploaded source maps land on the same release the client reports.

## Two nearby gaps found along the way

- **`Auth session ended unexpectedly` has never fired once.** The auth listener is installed in an effect with no deps, so its read of `session` is pinned to the initial `null` forever and the condition is unreachable. ESLint had been flagging the missing dependency the whole time. A ref holds the live value, a deliberate sign-out is no longer reported as an unexpected one, and the Supabase event name is attached as a tag.
- **Supabase returns auth failures instead of throwing them**, so `LoginPage`'s `try/catch` never ran and no failed sign-in has ever been reported. Now reported at the service boundary in `services/auth.ts` — except wrong-password, which is the user's typo and would bury every real failure (rate limits, disabled providers, Supabase 5xx) underneath it.

## User-facing change

The error screen gains a **Reload** button that retries the current URL on the new build, instead of only offering a link back to the homepage. Verified: on the reproduced failure it recovers straight to `/login`.

## Verification

- Reproduced the real failure against a production build served through a Netlify-equivalent SPA catch-all, and confirmed the captured event carries `error_type: stale-build-asset`, `route: /login`, `context: app.errorBoundary`, and the running release.
- `npm run lint` — clean for these changes. Three pre-existing `AuthContext.tsx` findings disappear, including the exact stale-dependency warning fixed here; none added.
- `npm run test:unit` — 540 passing, including 12 new. `npm run test:api` — 541 passing.
- Both new tests were checked to actually fail when the code they cover is neutralised.

## Deliberately not included

The permanent fix — a `lazy()` wrapper that reloads once on import failure — is a behaviour change worth deciding on with data in hand. Tracked separately.

Also spotted but out of scope: `toStoredResult` in `search-sources.ts` reads `p.displayName` off a `PlatformLink` type that has no such field, so platform display names are silently blank on every claimed and known artist card. The narrow `include` in `api/tsconfig.json` means the build never typechecks that file. Tracked separately.

### DIFF
--- a/api/functions/search-sources.ts
+++ b/api/functions/search-sources.ts
@@ -2053,6 +2053,21 @@ export async function handler(event: { queryStringParameters?: Record<string, st
       if (shouldCapture) {
         Sentry.captureMessage('Search returned 0 results', {
           level: 'info',
+          // The query has to be a TAG, not just an extra. Every zero-result event
+          // shares one message, so Sentry folds them into a single issue where
+          // `extra` is only readable one event at a time — you can't see which
+          // searches came back empty without paging through them. Tags get an
+          // aggregated value distribution on the issue page and are searchable
+          // (`search_query:radiohead`), which is the whole point of this signal.
+          // Both values are already length-capped by validateQuery (200 chars),
+          // which is also Sentry's tag-value limit.
+          tags: {
+            search_query: normalizedQuery,
+            // 'exact' means a playback-detection client (extension, Mac app) sent a
+            // real artist name and we have no coverage; 'fuzzy' means a human typed
+            // it and it may just be a typo. Very different follow-ups.
+            search_mode: mode,
+          },
           extra: {
             query,
             normalizedQuery,
--- a/apps/web/src/components/AppFallback.tsx
+++ b/apps/web/src/components/AppFallback.tsx
@@ -23,14 +23,28 @@ export function AppErrorFallback() {
     <div className="min-h-screen bg-bg-primary text-text-primary flex flex-col items-center justify-center gap-4 p-6 text-center">
       <h1 className="font-display text-2xl font-semibold">Something went wrong</h1>
       <p className="text-text-muted max-w-md">
-        Unstream hit an unexpected error. Reloading usually clears it.
+        Unstream hit an unexpected error. Reloading usually clears it — that
+        normally means a new version shipped while this page was open.
       </p>
-      <a
-        href="/"
-        className="px-4 py-2 rounded-lg bg-accent-primary text-white text-sm font-medium hover:bg-accent-primary/90 transition-colors"
-      >
-        Back to Unstream
-      </a>
+      <div className="flex items-center gap-3">
+        {/* Reload beats the "Back to Unstream" link for the common cause: a page
+            left open across a deploy, whose old JS files no longer exist. It
+            re-fetches the current URL on the new build instead of dumping the
+            person back on the homepage. */}
+        <button
+          type="button"
+          onClick={() => window.location.reload()}
+          className="px-4 py-2 rounded-lg bg-accent-primary text-white text-sm font-medium hover:bg-accent-primary/90 transition-colors"
+        >
+          Reload
+        </button>
+        <a
+          href="/"
+          className="px-4 py-2 rounded-lg border border-border text-text-primary text-sm font-medium hover:bg-bg-secondary transition-colors"
+        >
+          Back to Unstream
+        </a>
+      </div>
     </div>
   );
 }
--- a/apps/web/src/contexts/AuthContext.tsx
+++ b/apps/web/src/contexts/AuthContext.tsx
@@ -1,4 +1,4 @@
-import { createContext, useContext, useEffect, useState, useCallback, type ReactNode } from 'react';
+import { createContext, useContext, useEffect, useState, useCallback, useRef, type ReactNode } from 'react';
 import type { Session, User } from '@supabase/supabase-js';
 import * as Sentry from '@sentry/react';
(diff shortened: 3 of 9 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Senior-staff security/regression pass on this PR.

## Security

**1. [data-exposure] plausible** — `api/functions/search-sources.ts` (zero-result capture block)
`normalizedQuery` (raw user search-box text, up to 200 chars per `validateQuery`) is now sent to Sentry as a **tag** (`search_query`), not just `extra`. Tags are aggregated and searchable (`search_query:radiohead`), which is the whole point of the change — but it also means anything a user types into the search box, including accidental PII, is now more discoverable/permanent in a third-party service than it was as buried per-event `extra`. This isn't a new leak vector (the same string already went to Sentry as `extra`), but it is a meaningful increase in exposure surface. Worth a conscious call on whether raw query text belongs in a searchable tag as-is.

**2. [secret-handling] no issue** — `apps/web/src/services/auth.ts`'s `reportAuthFailure` correctly excludes `Invalid login credentials` and only forwards `error.message/code/name/status` — never the email or password. Good practice.

**3. [data-exposure] plausible, low severity** — `apps/web/src/main.tsx`'s `beforeCapture` tags `route: window.location.pathname`. For routes like `/a/:slug`, `/u/:handle` this sends a slug/username to Sentry. These are already public/displayed identifiers, not secrets, so severity is low — flagging only because it's new data flowing to a third party.

No SQL/injection or auth/authz issues found; this PR doesn't touch DB queries or privileged endpoints.

## Regression

**1. [state-transition] needs-manual-verification** — `apps/web/src/contexts/AuthContext.tsx`: `signingOutRef` is set `true` before `await supabase.auth.signOut()` and reset in a `finally` after the await resolves. The "Auth session ended unexpectedly" Sentry capture is gated on `!signingOutRef.current`. If supabase-js emits the `SIGNED_OUT` event *after* the `signOut()` promise resolves (timing depends on the installed supabase-js version's internals, not verifiable from the diff alone), the flag will already be back to `false` and a deliberate sign-out gets misreported as an unexpected one at `warning` level. Worth confirming event-ordering against the actual supabase-js version in use.

**2. [state-transition] plausible** — `signingOutRef` is a per-tab ref, not global. Supabase propagates sign-out across tabs (localStorage/BroadcastChannel). If sign-out is triggered in tab A, tab B's listener sees `newSession === null`, `previous !== null`, `signingOutRef.current === false` (tab B never called signOut), so tab B will report a "session ended unexpectedly" warning for what's actually a normal cross-tab propagation. Confirm this is the intended signal rather than a recurring false positive.

**3. [changed-assumption] plausible, low severity** — This capture path was previously dead in production: the closure bug meant `session` was pinned to `null`, so `newSession === null && session !== null` could never be true. This PR both activates it for the first time *and* raises its level `info` → `warning`. Combined with #1/#2 above, expect a burst of new warning-level events post-deploy — some real, some false positives from the scenarios above. Recommend watching the pattern for the first few days before treating every event as a genuine regression.

No off-by-one or API-contract-drift issues — nothing here changes a response shape consumed by other clients.

---

PR #392, opened 2026-08-02
Reload a tab once when its page chunk belongs to an old deploy

brandonlucasgreen:
Follow-up to #390, which made this failure legible in Sentry but left the fix itself open. This is the fix.

## The problem

A deploy renames every hashed chunk. A tab still running the previous build asks for a chunk the new deploy no longer has; Netlify's SPA catch-all answers `200 text/html` instead of `404`; the browser rejects it as a module script; `React.lazy` rejects; the error boundary shows "Unstream hit an unexpected error."

Every route in `main.tsx` is lazy-loaded, which is why this shows up most often as *"I can't log in after a deploy"* — `/login` is one of those routes.

## The fix

`apps/web/src/utils/lazyWithRetry.ts` wraps all 25 `lazy()` calls. On import failure it classifies the message with `isStaleBuildAssetError` (added in #390):

- **Stale chunk, no attempt recorded** → set a `sessionStorage` flag and `window.location.reload()`. The tab picks up the new `index.html` and the new chunk names. `registerType: 'autoUpdate'` means the service worker refreshes itself, so the reload lands on the new build rather than a cached shell.
- **Flag already set** → rethrow. A chunk that is genuinely broken, rather than merely old, must not reload forever — it falls through to the error boundary exactly as today.
- **Successful load** → clear the flag, so a *later* deploy gets its own attempt.

Two details worth a reviewer's eye:

- The reload path returns a promise that never settles. Resolving or rejecting would flash a page (or the error screen) in the moment before navigation; instead Suspense keeps showing the loading fallback.
- Storage failures (Safari private browsing, quota) resolve toward *not* reloading. An attempt that can't be recorded is a reload loop.

## Tests

`apps/web/tests/unit/lazy-with-retry.test.ts` — passthrough on success, reloads once on a stale-chunk error, does not reload a second time, reloads again after a successful load cleared the flag, rethrows a non-stale error untouched.

Mutation-tested before finalizing: removing the once-per-session guard fails "does not reload a second time"; removing the stale-error classification fails "rethrows a non-stale error untouched".

`npm run lint` (baseline unchanged), `npm run test:unit` (545), `npm run test:api` (542), and `tsc -b` at root + `apps/web` + `typecheck:api` all pass.

### DIFF
--- a/apps/web/src/main.tsx
+++ b/apps/web/src/main.tsx
@@ -1,4 +1,4 @@
-import { StrictMode, Suspense, lazy } from 'react'
+import { StrictMode, Suspense } from 'react'
 import { createRoot } from 'react-dom/client'
 import { BrowserRouter, Routes, Route, Navigate } from 'react-router-dom'
 import * as Sentry from '@sentry/react'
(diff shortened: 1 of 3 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Security/regression review of this PR:

**Security** — no findings. `lazyWithRetry.ts` has no auth, input-validation, secret-handling, injection, or data-exposure surface; it only reads/writes a namespaced `sessionStorage` key and calls `window.location.reload()`. Fail-closed by design: if `sessionStorage` is unavailable (Safari private mode, quota), `hasReloaded()` returns `true` and `markReloaded()` returns `false`, so an unrecorded attempt blocks the reload rather than looping — correct default (confirmed).

**Regression** — no findings. The once-per-session guard, the "clear flag on success" path (so a later deploy gets its own attempt), and the never-settling promise on reload (so Suspense doesn't flash the error boundary mid-navigation) are all covered by `apps/web/tests/unit/lazy-with-retry.test.ts`, and the PR notes mutation-testing on the two load-bearing branches. `isStaleBuildAssetError` itself is unchanged pre-existing code from #390, not part of this diff.

One edge case worth a mental note rather than a fix: if a tab reloads once for a stale chunk and the *new* build still can't serve a (different) chunk in the same session, the flag is already set so it rethrows straight to the error boundary rather than reloading again — that's the documented, deliberate tradeoff (a chunk that's still broken after one reload is broken, not stale), not a bug.

No action needed from my side on this one.

---

PR #389, opened 2026-08-02
Give Jam.coop a sourced artist payout (82-85%)

brandonlucasgreen:
Jam.coop had **no payout percentage** in the registry, so Unstream showed no "to artist" estimate for it — a co-op looked worse than Bandcamp.

**Source, verified live at https://jam.coop/docs/about:**
> "coop a 15% fee (minimum 20p) is taken from each sale."

Set to **`82-85%`**, matching the registry's existing string-range style.

**Why a range and not a flat 85%** — recorded as a comment next to the value in the canonical registry entry: the fee is 15% **with a 20p minimum**, and the floor is what bites on cheap releases. On a £1.00 sale the fee is 20p (20%), not 15p. Plenty of Jam.coop releases sell for £0.75–£3.00, so the effective payout at the low end is genuinely under 85%.

## Eight tables, not four

The obvious four (registry, web `sources.ts`, extension `constants.js`, Mac `PlatformCatalog.swift`) plus four more that a grep for other platforms' payout strings turned up:

- **`api/functions/discord-search-background.ts` — the real defect.** It already carried `jamcoop: '86-95%'`: unsourced, above both Bandcamp and Mirlo, and contradicting the 15% fee. The Discord bot has been quoting that to users. Now corrected and commented so it doesn't recur.
- `apps/mac/UnstreamShareExtension/ShareViewController.swift` — a second hand-maintained Swift copy, separate from `PlatformCatalog.swift`. Was `nil`.
- `apps/web/src/data/faq.ts` — the user-facing payout list. **This one is editorial copy — reword freely.**
- `docs/product-overview.md` — said "Cooperative model".

**Deliberately not changed:** `scripts/generate-social-posts.ts` has a `PAYOUT_PCT` table with `jamcoop: '85%'`. It isn't a registry mirror — its numbers already diverge across the board (`faircamp: '100%'` vs 90-97%, `bandcamp: '82%'` vs 80-85%). 85% is Jam.coop's headline figure and defensible, but it is the ceiling. Your call.

## One visible behaviour change

Search ordering is unaffected — nothing in the search path sorts by payout. But **release** contexts do (`payoutRank` in `release-display.ts`, used by the release page, artist-page badges, alerts and feeds), and at `82` Jam.coop now sorts **above** Bandcamp's `80`. That looks correct, but it is a change worth knowing about.

## Verification

[...]

PR #393, opened 2026-08-02
Return one release's buying guide as JSON, for native clients

brandonlucasgreen:
Item 1 of 3 for the native release UI. The Mac app and the browser extension currently `NSWorkspace.shared.open(url)` / `chrome.tabs.create` a release, handing the payout comparison — the whole point of the product — to a browser tab. They couldn't do better because there was no way to fetch per-format offers for a single release. Now there is.

`GET /api/release/{artist}/{release}` returns the same buying guide the HTML page at `/a/{artist}/{release}` renders.

## Not a second renderer

`api/edge/release-page.ts` stays the only thing that renders `/a/{artist}/{release}`; this is a different URL returning data, so it isn't the UNS-100 bifurcation shape. What the two share is the **answer**: both run `getReleaseDetail` and apply the same payout and ordering rules, so a native client can't describe a release differently from the web page a fan would see for it.

## `payoutPercent` is computed server-side

Payout figures are duplicated by hand across eight files in this repo, and that drift is what let the Discord bot quote an unsourced `86-95%` for Jam.coop to users for months (fixed in #389). A client that reads the number off the response cannot drift.

It also carries the **Bandcamp Friday override** to `~97%`, which the release page has always applied and which no client currently knows exists — plus a per-source `bandcampFriday` flag so a client doesn't have to string-match `'bandcamp'` to render the badge.

## The two filters are deliberately asymmetric

- `is_hidden` **is** filtered, in the query rather than after, so a release an artist suppressed is indistinguishable from one never catalogued.
- `needs_review` is **not** filtered, unlike the feed and alert reads. A tier-3 fuzzy flag means "we aren't sure this release is *distinct*", not "it's wrong" — a good reason to keep it out of someone's calendar, a bad reason to 404 a person following a direct link.

Production currently has zero `needs_review` rows, so `release-detail-query.test.ts` is the only place that asymmetry can actually be checked. It's asserted in both directions.

## Absence versus failure

A failed lookup is an **uncached 503**, never a 404. This response carries `Netlify-CDN-Cache-Control: s-maxage=300`, so a 404 during a Supabase incident would be a convincing lie the CDN then holds for five minutes. Never cache uncertainty.

## Other notes

- **Slugs arrive as query params** via Netlify placeholder substitution, not parsed from the path: `event.path` behind a `status = 200` rewrite can be the rewrite *target*. `feed-releases.ts` hit that and has to rebuild the path from `rawUrl`.
- **CORS is a deliberate wildcard**, same as `check-releases.ts` — neither shipped client sends an Origin the shared allowlist accepts, and the body is data the release page already serves publicly.
- **Rate limit `'lenient'`**, since this fires when a person taps a release.
- **Typechecked.** The function and both test files are in `api/tsconfig.json`'s include list, which caught a real defect: `rl.response` is optional on the limiter's result type, so returning it unchecked would have served a rate-limited request as an empty 200.

## Verification

**Against production Supabase**, not just mocks:

- `boy-harsher/get-mean` — four formats ordered digital \$9 → CD \$15 → cassette \$15 → vinyl \$30, future-dated, `status: announced`.
- `kid-lightbulbs/fruit-is-year-3-honeycrush` — Faircamp (90-97%) ordered **above** Bandcamp (80-85%), name-your-price preserved as `price: 0` rather than rendered free, and the Bandcamp source's `detailCheckedAt: null` distinguishable from "read it, no formats".
- `is_hidden` filtering confirmed against three real hidden rows.
- The slug guard's assumption checked against every row: all 842 `releases.slug` match, and the single `artists.slug` that doesn't is an empty string no URL can address.

**Mutation-tested.** All 39 guards across both files were neutralized one at a time — every one turned the suite red, and both files restored byte-identically (sha256-verified). Two needed fixing as a result: the `!client` branch in `getReleaseDetail` had no coverage at all, and the 429 fallback didn't exist until the typecheck forced it.

`npm run build`, `test:api` (687), `test:unit` (545), `typecheck:api` all pass. Lint is red only on the pre-existing `apps/web` baseline — nothing in the touched files.

## Response shape

```json
{
  "artist": { "slug": "boy-harsher", "name": "Boy Harsher", "imageUrl": null },
  "release": {
    "slug": "get-mean", "title": "GET MEAN", "releaseType": "album",
    "releaseDate": "2026-09-18", "datePrecision": "day", "status": "announced",
    "artworkUrl": "https://f4.bcbits.com/img/a0780870664_2.jpg",
    "pricesCheckedAt": "2026-08-01T09:18:25.833+00:00",
    "sources": [{
      "platform": "bandcamp", "name": "Bandcamp",
      "url": "https://boyharsher.bandcamp.com/album/get-mean",
      "payoutPercent": "80-85%", "bandcampFriday": false,
      "detailCheckedAt": "2026-08-01T09:18:25.966+00:00",
      "offers": [{ "format": "digital", "price": 9, "currency": "USD", "availability": "available", "capturedAt": "2026-08-01T09:18:25.833+00:00" }]
    }]
  },
  "pageUrl": "https://unstream.stream/a/boy-harsher/get-mean",
  "bandcampFriday": false
}
```

Two additions to the suggested shape: `pricesCheckedAt` (the **oldest** capture across all sources — claiming "checked today" because one source was re-read would overstate the rest) and `pageUrl`, so a client can offer "open the full guide" without rebuilding the URL shape.

## Note on the plan

`getReleaseDetail` did not exist — not in `api/functions/db.ts` on `main`, and not on the stale `release-json-endpoint` branch either. I wrote it, modelled on the edge function's query. The one intentional difference from a literal copy: it returns `{ detail, failed }` so absence and failure stay distinguishable, following `getArtistProfileBySlug`.

Item 2 (native Mac/extension UI) can start once this merges.

### DIFF
--- a/api/functions/__tests__/release-detail-query.test.ts
+++ b/api/functions/__tests__/release-detail-query.test.ts
@@ -1,0 +1,223 @@
+// `getReleaseDetail` — the query behind /api/release/{artist}/{release}.
+//
+// Two properties here are worth locking down, and they are the two a well-meaning tidy-up would
+// break in opposite directions:
+//
+//   1. **`is_hidden` is filtered.** An artist suppressing a release must be indistinguishable
+//      from one that was never catalogued. Verified against three real hidden rows in production
+//      too, but a filter that only exists in production is one refactor from vanishing.
+//   2. **`needs_review` is *not* filtered** — unlike `getFeedReleasesForUser` and the alert read,
+//      which both exclude it. A tier-3 fuzzy flag means "we aren't sure this release is
+//      *distinct*", not "this is wrong": a good reason to keep it out of someone's calendar, a
+//      bad reason to 404 a person who followed a direct link. Production currently has zero
+//      `needs_review` rows, so this is the only place that asymmetry can actually be checked.
+//
+// Driven against a recording fake Supabase client rather than the module mock the endpoint test
+// uses, following merge-releases-guard.test.ts, because the point is to exercise the real body.
+
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+
+interface Filter {
+  table: string;
+  column: string;
+  value: unknown;
+}
+
+/** Every `.eq()` the function applied, in order — the surface these assertions read. */
+const filters: Filter[] = [];
+
+let artistRow: Record<string, unknown> | null = { id: 'artist-1', name: 'Boy Harsher', image_url: null };
+let artistError: { message: string } | null = null;
+let releaseRow: Record<string, unknown> | null = null;
+let releaseError: { message: string } | null = null;
+
+function makeClient() {
+  return {
+    from(table: string) {
+      return {
+        select(_cols: string) {
+          // `.eq()` chains, and the terminal call is `.maybeSingle()`. Returning the same
+          // builder from `eq` lets an arbitrary number of filters be recorded.
+          const builder = {
+            eq(column: string, value: unknown) {
+              filters.push({ table, column, value });
+              return builder;
+            },
+            maybeSingle() {
+              return table === 'artists'
+                ? Promise.resolve({ data: artistRow, error: artistError })
+                : Promise.resolve({ data: releaseRow, error: releaseError });
+            },
+          };
+          return builder;
+        },
+      };
+    },
+  };
+}
+
+vi.mock('@supabase/supabase-js', () => ({ createClient: () => makeClient() }));
+
+process.env.SUPABASE_URL = 'https://example.supabase.co';
+process.env.SUPABASE_SERVICE_KEY = 'service-key';
+
+const { getReleaseDetail } = await import('../db');
+
+/** A release row as PostgREST returns it: snake_case, nested sources and offers. */
+function row(over: Record<string, unknown> = {}) {
+  return {
+    slug: 'get-mean',
+    title: 'GET MEAN',
+    release_type: 'album',
+    release_date: '2026-09-18',
+    date_precision: 'day',
+    status: 'announced',
+    artwork_url: 'https://f4.bcbits.com/img/a0780870664_2.jpg',
+    release_sources: [
+      {
+        platform: 'bandcamp',
+        url: 'https://boyharsher.bandcamp.com/album/get-mean',
+        detail_checked_at: '2026-08-01T09:18:25.966+00:00',
+        release_offers: [
+          {
+            format: 'vinyl',
+            price: 30,
+            currency: 'USD',
+            availability: 'available',
+            captured_at: '2026-08-01T09:18:25.833+00:00',
+          },
+        ],
+      },
+    ],
+    ...over,
+  };
+}
+
+function releaseFilters() {
+  return filters.filter(f => f.table === 'releases');
+}
+
+beforeEach(() => {
+  filters.length = 0;
+  artistRow = { id: 'artist-1', name: 'Boy Harsher', image_url: null };
+  artistError = null;
+  releaseRow = row();
+  releaseError = null;
+});
+
+describe('the filters it applies', () => {
+  it('scopes the release to the artist id and slug', async () => {
+    await getReleaseDetail('boy-harsher', 'get-mean');
+
+    expect(filters).toContainEqual({ table: 'artists', column: 'slug', value: 'boy-harsher' });
+    expect(releaseFilters()).toContainEqual({ table: 'releases', column: 'artist_id', value: 'artist-1' });
+    expect(releaseFilters()).toContainEqual({ table: 'releases', column: 'slug', value: 'get-mean' });
+  });
+
+  it('filters hidden releases in the query, not after it', async () => {
+    await getReleaseDetail('boy-harsher', 'get-mean');
+    expect(releaseFilters()).toContainEqual({ table: 'releases', column: 'is_hidden', value: false });
+  });
+
+  // The asymmetry with the feed and alert reads, which is deliberate. If this ever starts
+  // filtering, every fuzzy-flagged release 404s for anyone following a direct link to it.
+  it('does not filter needs_review', async () => {
+    await getReleaseDetail('boy-harsher', 'get-mean');
+    expect(releaseFilters().map(f => f.column)).not.toContain('needs_review');
+  });
+});
+
+describe('absence versus failure', () => {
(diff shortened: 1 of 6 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Automated security/regression pass on this PR (`getReleaseDetail` + `GET /api/release/{artist}/{release}`). No blockers found.

**Security**

- **Auth/authorization** — needs-manual-verification (low): endpoint is intentionally unauthenticated, mirroring `check-releases.ts`. Same query and `is_hidden` filter as the edge-rendered `/a/:artist/:release` page, so no data is exposed here that isn't already public. No artist-level suppression flag exists on the `artists` table for either this or the existing edge function, so no regression in scope.
- **Input validation** — confirmed good: `SLUG_PATTERN` (`^[a-z0-9][a-z0-9-]{0,119}$`) rejects anything outside what the two slug generators ever mint (path traversal, uppercase, over-length) before touching the DB, verified against 842 production release slugs.
- **Injection risks** — confirmed none: all Supabase reads use `.eq()` bound params, no string-built SQL/filter clauses.
- **Secret handling** — confirmed none: no new secrets, `getClient()` reuses the existing service-role singleton and the missing-credentials path correctly returns `failed: true` rather than a false negative.
- **Unsafe defaults** — plausible, low severity: `Access-Control-Allow-Origin: *` is a deliberate wildcard (matches `check-releases.ts` precedent) since neither the Mac app nor the extension send an Origin the shared allowlist accepts, and the response carries only public data via GET with no cookie-based auth, so no CSRF/credentialed-read exposure.
- **Data exposure** — confirmed fine: response shape only includes public release/artist/pricing fields already rendered on the public release page.

One needs-manual-verification note, not treated as a real vuln: the new `netlify.toml` redirect substitutes `:artist`/`:release` path segments directly into the target query string. If Netlify's placeholder substitution doesn't re-encode `&`/`=` in a decoded segment, a crafted path could inject extra query params — but the handler only ever reads `artist`/`release` off `queryStringParameters` and validates both against `SLUG_PATTERN` before any DB call, and the underlying function is reachable directly at the same query-param shape anyway, so there's no privilege/behavior delta. Worth a quick manual check against a deploy preview if you want it fully closed out.

**Regression**

- Changed assumptions: `getReleaseDetail` follows the same absence-vs-failure contract as `getArtistProfileBySlug` (`{detail, failed}`), and the query is asserted (via a recording fake Supabase client) to be byte-for-byte the same shape the edge function runs — including the deliberate `is_hidden` filtered / `needs_review` *not* filtered asymmetry.
- Edge cases: null `release_sources`/`release_offers` embeds (PostgREST returns `null`, not `[]`) are normalized to `[]`; zero-price (name-your-price) offers pass through untouched rather than defaulting to free; a rate-limiter result with `limited: true` but no `response` still 429s instead of silently 200ing.
- Off-by-one: slug length bound (`{0,119}` → max 120 chars) matches the "over-long slug" test case.
- API contract drift: new endpoint only, no existing routes/consumers touched. `pricesCheckedAt` is explicitly the *oldest* capture across sources, avoiding an overstated freshness claim if only one source was re-scraped.

Test coverage (mutation-tested per the PR description, 39/39 guards) is thorough enough that I'd weight the needs-manual-verification items above as low-risk rather than blocking.

---

brandonlucasgreen (commented):
@brandonlucasgreen Automated staff-eng pass (security + regression) on this PR. PR #394 was already merged before I got to it, so this is the only open PR reviewed this cycle.

**Security**
- Auth/authorization — n/a, this is intentionally public data (same as `/a/:artist/:release`). No privileged action.
- Input validation — `SLUG_PATTERN` rejects anything outside `[a-z0-9][a-z0-9-]{0,119}`, checked before any DB call. Verified against real minted slugs (`artistSlug()`/`releaseSlug()` in `release-utils.ts`); path-traversal/uppercase/overlong cases 404 without a query. **confirmed sound.**
- Secret handling — no secrets touched; standard `SUPABASE_URL`/`SUPABASE_SERVICE_KEY` via existing `getClient()`. **confirmed clean.**
- Injection risks — all DB access goes through parameterized `.eq()` calls on the Supabase query builder, no string-built SQL. No outbound fetch of user-controlled URLs (no SSRF surface here — `release_sources.url` is stored data, never fetched by this endpoint). **confirmed clean.**
- Unsafe defaults — `Access-Control-Allow-Origin: *` is a deliberate, precedented choice (matches `check-releases.ts`/`feed-releases.ts` for native clients that send no acceptable Origin), documented in a comment. Not a regression from the shared allowlist since it never claimed to cover this endpoint. **confirmed intentional, not a finding.**
- Data exposure — response only contains data already public on the rendered release page (artist name/photo, release/offer info). No PII, no internal error strings leaked (catch blocks return generic messages). **confirmed clean.**

**Regression**
- Changed assumptions — `getReleaseDetail` is new but models the same query `release-page.ts` already runs; `is_hidden` filtered / `needs_review` intentionally not, both covered by dedicated tests (`release-detail-query.test.ts`) including the asymmetry itself.
- Edge cases / null-empty — null `release_sources`/`release_offers` embeds handled (PostgREST returns `null`, not `[]`, for empty embeds) and covered by tests. Zero-price ("name your price") passed through rather than defaulted — matches `formatOfferPrice`'s established rule in `release-display.ts`.
- Off-by-one — `SLUG_PATTERN`'s `{0,119}` after a mandatory first char caps total length at 120, matching the stated intent; verified consistent with the "over-long slug" test case.
- State transitions / absence vs. failure — `{ detail, failed }` correctly distinguishes "not found" (404, cacheable) from "lookup failed" (503, `Cache-Control: no-store`, no CDN header, Sentry-reported). This is exactly the "never cache uncertainty" pattern this codebase has needed enforced before, and it's tested in both directions.
- API contract drift — new route added to `netlify.toml` with query-param substitution (avoids the `event.path`-behind-rewrite trap `feed-releases.ts` had to work around); doesn't touch or shadow any existing route. `rl.response` optionality bug (would have silently served a 200 on rate-limit) was caught by the PR's own typecheck inclusion and is fixed — confirmed by test coverage.

No confirmed or plausible issues found. Nothing needs manual verification beyond what the tests already assert. This is a well-scoped, well-tested addition that follows the codebase's established patterns (lenient rate-limit tier, wildcard CORS precedent, absence-vs-failure discipline). Nice work.

---

PR #394, opened 2026-08-02
Release page: responsive 2-column layout for where-to-buy cards

brandonlucasgreen:
## Summary

- The release page's "where to buy" source cards (Bandcamp, Subvert, Ampwall, Faircamp, etc.) rendered full width on every screen size, which was hard to scan once a release had more than one or two platforms.
- Wraps the cards in a CSS grid (`.sources-grid`, 2 columns, `gap: 12px`, `align-items: start`) instead of each card carrying its own `margin-bottom`.
- Falls back to a single column under 700px so mobile and narrow browsers are unaffected.
- `align-items: start` keeps each card its natural height rather than grid's default stretch, which would otherwise pad a short card (e.g. one offer) with empty space to match a taller neighbor in the same row.

## Test plan

- [x] Verified locally with `npm run preview:release -- kidlightbulbs`, rendering the real `api/edge/release-page.ts` against Kid Lightbulbs' "Ruined Castle" release (the same release from the original screenshot).
- [x] Since the live Bandcamp fetch only surfaces one real source, injected 3 extra mock `.source` cards via the browser console to force a 4-card grid and confirm the CSS.
- [x] Screenshot at 1280px confirms 2-per-row layout.
- [x] Screenshot at 400px confirms single-column fallback.

### DIFF
--- a/api/edge/release-page.ts
+++ b/api/edge/release-page.ts
@@ -79,7 +79,15 @@ const CSS = `
   .release-art { width: 160px; height: 160px; flex-shrink: 0; border-radius: 12px; border: 1px solid var(--border); background: var(--bg2); object-fit: cover; }
   .release-art-fallback { width: 160px; height: 160px; flex-shrink: 0; border-radius: 12px; border: 1px solid var(--border); background: var(--bg2); display: flex; align-items: center; justify-content: center; font-size: 40px; }
 
-  .source { border: 1px solid var(--border); border-radius: 12px; overflow: hidden; margin-bottom: 12px; }
+  /* Two columns on desktop — full-width cards were hard to scan when there were more than two
+     or three platforms. align-items: start keeps each card its natural height instead of
+     grid's default stretch, which would otherwise pad a short card (e.g. one offer) with empty
+     space to match a taller neighbor. */
+  .sources-grid { display: grid; grid-template-columns: repeat(2, minmax(0, 1fr)); gap: 12px; align-items: start; margin-bottom: 12px; }
+  @media (max-width: 700px) {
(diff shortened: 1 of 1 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Automated staff-eng pass (security + regression) on this PR. Noting this PR was already merged by the time this review ran — nothing actionable, but recording the analysis since the semantic-revert-check bot flagged it.

**Security** — n/a. Pure CSS/template change in `api/edge/release-page.ts`: no new input handling, no auth/data-exposure surface, nothing that touches user input, secrets, or outbound requests.

**Regression — semantic-revert-check flag**

The bot flagged this PR for deleting 2 lines added in #362. Checked against the actual diff:

- `margin-bottom: 12px` was removed from `.source`, but it wasn't dropped — it moved to the new `.sources-grid` container (which also adds `gap: 12px` for the inter-card spacing `.source`'s margin used to provide). Net visual spacing is preserved, just relocated from per-card margins to a grid container, which is also more correct (no doubled margin between adjacent cards).
- `${sourcesHtml || fallback}` became `${sourcesHtml ? grid-wrapped : fallback}` — the falsy/empty-state branch is untouched byte-for-byte; only the truthy branch gained a grid wrapper.

This is a false positive from the bot's line-diff heuristic — it can't see that the deleted line's behavior was relocated rather than removed. No functionality from #362 was actually reverted. Confirmed by the PR's own test plan (screenshots at 1280px/400px) and consistent with the single-column `@media (max-width: 700px)` fallback keeping mobile unaffected, as intended.

No other regression concerns: `align-items: start` is scoped to `.sources-grid` only, and the empty-state message path is unchanged.

---

PR #395, opened 2026-08-02
Let claimed artists arrange their releases by hand

brandonlucasgreen:
Releases have always been listed newest first. An artist may want a different order — lead with the record they're promoting, push a compilation down — the same way they already arrange their platform links, so this reuses that control: ▲▼ on each row in **Manage Releases**, with one explicit save.

## Storage

`releases.display_order` (nullable integer), matching `artist_links.display_order`. Writes go through a transactional `set_release_display_order()` RPC, service-role only, for the same reason `replace_artist_links` is one function: a half-applied order is an arrangement the artist never picked, live on their public page. Ids omitted from the array are cleared, so an empty array is "reset" and a release catalogued between page-load and save isn't handed a position nobody chose.

## Which reads honour it

Every surface that returns a list of an artist's releases now sorts `display_order ASC NULLS LAST, release_date DESC, created_at DESC`:

| Surface | |
|---|---|
| `/api/artist-page` | the SPA artist page |
| `/a/:slug` edge render | identical clause — the two renderers of one URL must not disagree |
| `/api/artist-releases` | the editor, so it shows what fans see |

Sorted **in SQL**, not after the fetch: both page reads are limited (60 and 24), so a JS re-sort would drop a release pinned to the top out of the query on a long catalogue while looking correct on a short one.

## Which deliberately don't

`getReleasesForAlerts` and the ICS/Atom feed reads stay chronological, with a comment saying so at each one. An arrangement answers "what should fans see first"; an alert answers "what came out" — and `releases[0]`, which the shipped Mac app and extension read as *the* new release, has to keep meaning newest. `/api/release/{artist}/{release}` (#393) returns a single release and is untouched.

## Behaviour when new releases arrive

`display_order` is null until an artist arranges their catalogue, so this is inert for everyone else. A release catalogued *after* an arrangement lands at the end of it rather than displacing what the artist chose — the page says so, and "Reset to newest first" undoes the whole thing.

## Verification

- Migration and RPC run against a real Postgres 16: arrangement stored, a later release joining unpositioned, reset via empty array, another artist's id ignored rather than stolen, `NULL` array handled.
- `release-order-query.test.ts` locks the SQL ordering for both page reads and the RPC call; API tests cover reorder validation, ownership of every id in the body, duplicate ids, the length cap and reset; page tests cover move → save payload, discard, reset, and that an unsaved arrangement survives the refetch another action triggers. Each new test was confirmed to fail with the logic broken.
- `npm run build` passes. Lint: touched files clean.

No screenshot — `/artist-edit/:slug/releases` is auth-gated and the dev API has no release endpoints, so the jsdom tests are the real-DOM verification here.

### DIFF
--- a/api/edge/artist-page-static.ts
+++ b/api/edge/artist-page-static.ts
@@ -256,12 +256,16 @@ export default async function handler(request: Request, context: Context) {
 
       links = linksData || [];
 
-      // Releases, newest first. `count: 'exact'` alongside a limit gives the full total in the
-      // same round trip, so the "and N more" line is a real number rather than a guess.
+      // Releases, in the artist's own order where they set one and newest first otherwise.
+      // `count: 'exact'` alongside a limit gives the full total in the same round trip, so the
+      // "and N more" line is a real number rather than a guess.
       //
-      // Ordered to match idx_releases_artist_chrono: many releases have no date yet (grid ingest
-      // gets identity and artwork but no dates), and without the created_at tiebreaker those
-      // undated rows would shuffle between requests.
+      // The ordering must stay identical to `getArtistReleases` in api/functions/db.ts — this is
+      // the crawler's view of the same page the SPA renders, and the two disagreeing on release
+      // order is the two-renderers trap in miniature. display_order leads (NULLS LAST, so it's
+      // inert until a claimed artist arranges their catalogue); the chronological part matches
+      // idx_releases_artist_chrono, where created_at is the tiebreaker that stops undated rows —
+      // grid ingest gets identity and artwork but no dates — shuffling between requests.
       const { data: releasesData, count } = await supabase
         .from('releases')
         .select(
@@ -271,6 +275,7 @@ export default async function handler(request: Request, context: Context) {
         )
         .eq('artist_id', artist.id)
         .eq('is_hidden', false)
+        .order('display_order', { ascending: true, nullsFirst: false })
         .order('release_date', { ascending: false, nullsFirst: false })
         .order('created_at', { ascending: false })
         .limit(MAX_RELEASES_SHOWN)
--- a/api/functions/__tests__/artist-releases.test.ts
+++ b/api/functions/__tests__/artist-releases.test.ts
@@ -19,6 +19,7 @@ const mocks = vi.hoisted(() => ({
   createArtistRelease: vi.fn(),
   dismissReleaseReview: vi.fn(),
   mergeReleases: vi.fn(),
+  setReleaseDisplayOrder: vi.fn(),
   getCatalogState: vi.fn(),
   clearCatalogCooldown: vi.fn(),
   clearReleaseDetailCooldown: vi.fn(),
@@ -37,6 +38,7 @@ vi.mock('../db', () => ({
   createArtistRelease: mocks.createArtistRelease,
   dismissReleaseReview: mocks.dismissReleaseReview,
   mergeReleases: mocks.mergeReleases,
+  setReleaseDisplayOrder: mocks.setReleaseDisplayOrder,
   getCatalogState: mocks.getCatalogState,
   clearCatalogCooldown: mocks.clearCatalogCooldown,
   clearReleaseDetailCooldown: mocks.clearReleaseDetailCooldown,
@@ -99,6 +101,7 @@ beforeEach(() => {
   mocks.createArtistRelease.mockResolvedValue({ ok: true, releaseId: RELEASE_A });
   mocks.dismissReleaseReview.mockResolvedValue(true);
   mocks.mergeReleases.mockResolvedValue({ ok: true });
+  mocks.setReleaseDisplayOrder.mockResolvedValue(true);
   mocks.getCatalogState.mockResolvedValue({ ok: true, state: null });
   mocks.clearCatalogCooldown.mockResolvedValue(undefined);
   mocks.clearReleaseDetailCooldown.mockResolvedValue(undefined);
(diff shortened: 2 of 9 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Automated senior-engineer review of #395 (manual release ordering). Reviewed the full diff (`artist-releases.ts`, `db.ts`, the new migration/RPC, `ArtistReleasesPage.tsx`, `releaseOrder.ts`) plus `resolveOwnedArtist`/`verifyReleaseOwnership` in context. CI is green (all 5 checks passing).

## Security

No confirmed or plausible issues. Specifically checked:

- **Authorization** — confirmed. The new `reorder`/`resetOrder` actions go through the same two-layer check as `merge`/`dismiss`: `resolveOwnedArtist` (session user owns a *verified* profile) then `verifyReleaseOwnership(artistId, releaseIds)` before any write. `resetOrder` correctly skips the ownership call only because it always sends `[]`, which touches nothing but the caller's own rows (`verifyReleaseOwnership` itself returns `false` on an empty list, so this isn't a bypass — it's a legitimate short-circuit since there's nothing to check).
- **Injection** — confirmed none. `set_release_display_order` takes typed `uuid`/`uuid[]` params, no string concatenation; PostgREST/RPC parameter binding, not raw SQL.
- **Unsafe defaults** — confirmed none. `display_order` is nullable and defaults to `NULL` for every existing row, and both order-by clauses use `NULLS LAST`, so this is inert for every artist until they touch it. The RPC's own execute grants are `REVOKE`d from `public`/`anon`/`authenticated` and only granted to `service_role` — correct defense-in-depth given PostgREST auto-publishes every public-schema function as an RPC endpoint.
- **Data exposure** — confirmed none. `display_order` is added to the *owner* read's `select()` (needed for the "Reset" button's visibility) but deliberately not added to the public `getArtistReleases` select — only used there in `.order()`, so it never reaches the public API response.
- **Input validation** — confirmed. UUID-shape check on every id, a length cap (`MAX_ORDERED_RELEASES = 500`) checked before the DB round-trip, and a duplicate-id rejection (a repeated id would otherwise silently drop a release from the arrangement via last-write-wins).
- **Transaction safety** — confirmed. One RPC call for the whole write (matches the existing `replace_artist_links` pattern), so a failure can't leave a half-applied arrangement live on the public page.

## Regression

- **False positive, investigated** (`needs-manual-verification` → resolved): the `semantic-revert-check` bot flagged this PR for deleting lines added in #366/#376/#370. Verified the PR's base sha (`e76b5fc2`) against `main`'s current tip — they match, so this isn't based on stale history. The flagged deletions are all comment/copy rewording (e.g. the `artist-page-static.ts` and `db.ts` jsdoc explaining the new ordering, the "Manage Releases" help text) landing next to the new `display_order` clause, not a reverted behavior. No action needed, but flagging per the repo's own guidance to take that bot seriously rather than reflexively dismiss it.
- **Edge cases** — the "release catalogued between page-load and save" case (new row arrives with `display_order = NULL` mid-session) is explicitly handled: `applyPendingOrder` appends unranked ids at the end using real numbers (`pendingOrder.length + index`), avoiding the `Infinity - Infinity → NaN` comparator trap that would otherwise scramble the whole list. Backed by a direct unit test.
- **State transitions** — the unsaved-arrangement-survives-another-action's-refetch case (e.g. clicking Hide while an unsaved reorder is pending) is tested and correctly implemented: `pendingOrder` is separate state from `releases` and is only cleared on explicit save/discard/reset, not on every `fetchReleases()`.
- **Two-renderers rule** — both `getArtistReleases` (SPA) and the `/a/:slug` edge function apply the identical three-level `ORDER BY`, with a `release-order-query.test.ts` that locks the exact clause via a recording fake client rather than a mock — appropriately paranoid given this file's own comment about the two-renderers bug class.
- **Deliberate non-application** — `getReleasesForAlerts` and the ICS/Atom feed reads correctly stay chronological-only (verified in `db.ts`), preserving `releases[0]` semantics for the shipped Mac app/extension.

Nothing to change. Nice attention to the ownership/transaction patterns already established in this codebase.

---

brandonlucasgreen (commented):
@brandonlucasgreen Automated security + regression pass on this PR (manual release ordering).

## Security

- **auth/authorization** — none found. Two-layer ownership check: `resolveOwnedArtist(slug, user.userId)` gates the whole endpoint (unchanged), then `reorder`/`resetOrder` additionally calls `verifyReleaseOwnership(artistId, releaseIds)` (`api/functions/db.ts:2054`), which does an exact-count match against `releases` scoped to `artist_id` — a caller can't smuggle another artist's release id into an arrangement. The new `set_release_display_order` RPC explicitly revokes EXECUTE from `public`/`anon`/`authenticated` and grants only `service_role` (`supabase/migrations/20260802150000_release-display-order.sql`), closing the usual PostgREST auto-exposure hole. `resetOrder` skips the ownership check by design when `releaseIds.length === 0`, but the SQL function still scopes its `UPDATE` to `artist_id = p_artist_id`, so this can't touch another artist's rows — safe. **confirmed-good, no finding.**
- **input validation** — none found. `releaseIds` is validated as an array of UUID-regex-matching strings, capped at 500, and rejected if it contains a duplicate (`api/functions/artist-releases.ts` reorder branch). Matches the test coverage in `artist-releases.test.ts`.
- **secret handling** — n/a, no secrets touched.
- **injection risks** — none found. The RPC uses a parameterized `uuid[]` param and `unnest(...) WITH ORDINALITY`, no string concatenation into SQL.
- **unsafe defaults** — none found. `display_order` is nullable, defaults to `NULL` on the new column, and every read orders `NULLS LAST` — existing artists/pages are unaffected until an artist explicitly arranges releases.
- **data exposure** — none found. The new `displayOrder` field is only returned from the ownership-gated `getArtistReleasesForOwner` path, not the public artist-page read.

## Regression

- **changed assumptions** — handled correctly, worth noting for reviewers: `getReleasesForAlerts` and the feed queries in `db.ts` deliberately do *not* apply `display_order` (well-commented — an alert/feed is about recency, not curation), and `api/functions/__tests__/release-order-query.test.ts` locks the ordering contract between the edge-function read (`artist-page-static.ts`) and `db.ts`'s `getArtistReleases`, which is exactly the "two renderers must agree" trap this repo has been bitten by before.
- **edge cases** — plausible, low severity: `MAX_ORDERED_RELEASES = 500` is a hard cap with no partial-save path — an artist with >500 releases gets a flat 400 on any reorder attempt. Current max catalogue is 33 per the code comment, so not urgent, but if that ever changes there's no degraded UX, just a wall.
- **off-by-one** — none found. `WITH ORDINALITY` is 1-based; the function stores `ord - 1`, correctly producing 0-based positions matching `ascending: true`.
- **state transitions** — none found. `pendingOrder` (client-side unsaved arrangement) is reconciled against server refetches via `applyPendingOrder`, with test coverage for the case where another action (e.g. hide) triggers a refetch mid-arrangement.
- **API contract drift** — none found. `OwnerReleaseItem`/`ArtistReleasesPage` types and the two release-order-consuming renderers were updated together in the same PR.

No blocking issues found.

---

PR #396, opened 2026-08-02
Show a release's buying guide natively in the Mac app and the extension

brandonlucasgreen:
Both clients answered "where should I buy this?" by opening a browser tab — `NSWorkspace.shared.open(url)` in `SupportListView`, `chrome.tabs.create` in the popup and service worker. The payout comparison *is* the product, and it matters at the moment someone is deciding where to buy, so both now render it themselves from the new `GET /api/release/{artist}/{release}` (#393).

Started from the parked `native-release-ui-prototype` branch — `ReleaseFormatting` survives largely intact; the data layer is new.

## macOS — inside the popover, no second window

Per your call: the guide takes over the existing 320pt menu-bar popover behind a back header, rather than opening a window. Escape and the ⌘1/⌘2 shortcuts back out; it resets on reopen, because NSPopover hides its content view rather than destroying it (reopening the menu bar into the release you last looked at would be a bug).

Plumbing is an `@Environment(\.openReleaseGuide)` action — the alert badge that triggers it sits several views down inside the saved-artists list, and threading a closure through every view in between would have been worse.

## iOS

Pushes onto the Releases tab, or presents as a sheet from the saved-artists list. Release rows now lead with the price line (`from $5 · ≈$4–$4.25 to artist`) instead of the platform name, since that's the reason to tap.

## Extension

The alert row expands in place — no tab. Opening the web page is the escape hatch now, not the main action. Guide rendering is its own module (`lib/release-guide.js`) so `popup.js` doesn't grow a second job and so the panel can be rendered outside the extension to look at.

Nothing scratch was left under `apps/extension/` — the harness lives in the scratchpad and symlinks the real CSS and modules.

## Payout numbers come off the wire

`payoutPercent` is computed per source by the endpoint, so no client can drift from the release page the way the Discord bot's hand-copied Jam.coop rate did — and clients now get the Bandcamp Friday override (~97%), which neither knew existed. Source and offer ordering is the server's for the same reason: re-sorting would mean carrying a copy of the payout registry.

## Live bug fixed while here

`ALLOWED_RELEASE_DOMAINS` never listed `unstream.stream`. Once #388 made a catalogued release's `releaseUrl` our own release page, the extension's **Listen button and its notification click both silently did nothing** — the allowlist check failed and the handler returned. That's every catalog-backed alert since #388.

## Verification

- **34 assertions over the real Swift source files** via `swiftc`, including decoding the live production endpoint. That run caught `freshness()` counting *calendar* days rather than elapsed 24-hour periods — the same price would have read "yesterday" to a fan in one timezone and "today" to another. Now matches `relativeDays` in `api/shared/release-display.ts` exactly.
- **XCTest could not be run**: the runner hangs before establishing a connection while the Mac's screen is locked. `build-for-testing` succeeds, so `UnstreamTests/ReleaseDetailTests.swift` (28 tests) compiles and is wired in — worth a re-run with the screen unlocked before merge.
- **New web unit test** drives the extension's JS copy of the money rules and `api/shared/release-display.ts` over the same inputs and asserts they agree, so the hand-maintained copy can't drift silently. Confirmed it has teeth by seeding two defects (6 failures).
- Rendered in the iOS simulator against live data (phone width + a 320×350 popover-sized box) and in a static harness over the real `popup.css`.
- `npm run build`, `test:api` (697), `test:unit` (582) all pass. Lint unchanged from the dirty baseline. macOS + iOS both build.


brandonlucasgreen (commented):
@brandonlucasgreen Automated security + regression pass on this PR (native release buying guide in the Mac app + extension).

## Security

- **auth/authorization** — none found. No auth paths touched; this only adds consumers of the existing public `GET /api/release/{artist}/{release}` endpoint.
- **input validation** — `apps/mac/Unstream/Services/ReleaseDetailService.swift:35-36` — `fetch(artist:release:)` percent-encodes the artist/release slugs with `.urlPathAllowed`, which per Apple's docs is meant for an entire path and doesn't escape `/` (the JS sibling correctly uses `encodeURIComponent`, which does — `apps/extension/popup/popup.js:710`). I checked the only call site: `slugs(fromReleaseURL:)` (same file, line 27-31) derives both strings via `url.path.split(separator: "/")`, so a `/` can never reach `fetch` today — **not currently exploitable**. Flagging as **needs-manual-verification** only because it's the wrong Foundation API for the job and would silently break/mis-encode if this method is ever called with a slug from a different source. Low priority.
- **secret handling** — none found. `baseURL` is a plain public endpoint, no keys introduced.
- **injection risks** — none found. Extension DOM building uses `createElement`/`textContent`, never `innerHTML`; Swift uses typed `Text` views. No SQL touched.
- **unsafe defaults** — **plausible**: `apps/mac/Unstream/Views/Shared/ReleaseGuideView.swift:351-358` (`buyButton`) builds `URL(string: source.url)` from the API response and opens it directly via `NSWorkspace.shared.open` (macOS) / a Safari sheet (iOS), with **no host/scheme allowlist**. This PR adds exactly that guard on the extension side for the equivalent per-source buy link: `apps/extension/lib/release-guide.js:81` → `onOpenUrl(source.url)` → `openReleaseUrl` (`popup.js:1224-1226`) → `isAllowedReleaseUrl`, checked against `ALLOWED_RELEASE_DOMAINS` (`apps/extension/lib/constants.js:11-18`), whose comment says explicitly "so an alert can never open an arbitrary URL." I verified both sides directly — the extension's per-source buy button is gated, the Mac app's is not. `openOnUnstreamLink` (line 367-374) is lower risk since its URL is always either the hardcoded `webPageURL` (unstream.stream) or server-supplied `detail.pageUrl`. The gap is specifically the multi-platform `source.url` values this PR newly renders as clickable buttons. Exploitability depends on the backend `/api/release` response always producing trustworthy http(s) URLs, which is outside this diff — but since the sibling client in the same PR was given this exact protection, the inconsistency looks unintentional rather than a considered trade-off.
- **data exposure** — none found. Logging only prints artist/release slugs and generic error text.

## Regression

- **changed assumptions** — none found; "never cache uncertainty" is honored correctly in both new clients (404 → `notFound`; any other error/status → `unavailable`, never cached as absence).
- **edge cases / null/empty** — none found; empty `sources`/`offers`, null `price`/`payoutPercent`/`detailCheckedAt` are handled consistently across the JS, TS, and Swift copies, backed by a parity test (`apps/web/tests/unit/extension-release-display.test.ts` vs `apps/mac/UnstreamTests/ReleaseDetailTests.swift`).
- **off-by-one** — none found.
- **state transitions** — none found; `PopoverView`'s `openGuide` state is reset on every exit path (appear, popover close, tab switch, shortcuts, Escape), and the tab control is unreachable while a guide is open, so there's no path to a stale mounted guide view.
- **API contract drift** — none found; only new consumers of an unchanged endpoint.
- **minor, non-security** — `apps/extension/lib/release-display.js`'s exported `formatReleaseDate` is cross-tested against the TS original but never actually called from `release-guide.js`/`popup.js` (the guide panel doesn't render a release date). Not a bug, just dead-export drift worth a look given the repo's "no dead weight" convention.

**Suggested action:** add the same domain/scheme allowlist check (mirroring `isAllowedReleaseUrl`) before `open(url)` in `buyButton`, so the two clients enforce the same policy on the same data.

---

brandonlucasgreen (commented):
@brandonlucasgreen Automated security + regression review of this PR (full diff read, cross-checked against `apps/extension/lib/constants.js`, `apps/extension/popup/popup.js`, `apps/extension/lib/release-guide.js`, `api/shared/platform-registry.ts`, and the relevant Swift views on the PR branch).

## Security

- **Extension URL allowlisting mechanics** — none found. `isAllowedReleaseUrl()` (popup.js) does exact-or-`.suffix` hostname matching against `ALLOWED_RELEASE_DOMAINS` and restricts to `http`/`https` — adding `unstream.stream` doesn't open any substring/open-redirect bypass. All server-derived strings (artist name, release title, source name, price/availability labels) are rendered via `textContent`, never `innerHTML` — no XSS risk in the new extension code.
- **[plausible] macOS has no domain allowlist for release URLs at all.** `ReleaseGuideView.swift`'s `buyButton()` / `openOnUnstreamLink` open `source.url` / `detail.pageUrl` via `NSWorkspace.shared.open(url)` / a Safari sheet with only `URL(string:)` parsing — no equivalent to the extension's `isAllowedReleaseUrl` gate. This matches the existing app-wide pattern (`PlatformBadge.swift`, `ArtistResultView.swift`, `SupportListView.swift` do the same), so it's not new to this PR, but it's a real asymmetry between the two clients trusting the same server-computed URLs. Whether it's exploitable depends on whether the catalog backend (out of this diff) can ever be made to store an attacker-controlled URL — can't be confirmed from this diff alone.
- Auth/authorization, secret handling, data exposure: none found — the endpoint is public catalog data, no tokens attached or logged, diagnostics only log artist/release slugs and error objects.

## Regression

- **[confirmed] The extension's new "Buy" buttons reuse `ALLOWED_RELEASE_DOMAINS`, which doesn't cover most of the catalog — so most Buy buttons silently do nothing.** `release-guide.js`'s `buildGuideSource()` wires every source's Buy button to `onOpenUrl(source.url)` (popup.js:1253/1278 → `openReleaseUrl` → `isAllowedReleaseUrl`). That list is `['unstream.stream', 'bandcamp.com', 'mirlo.space', 'qobuz.com', 'ampwall.com', 'faircamp.eu']`. But per `api/shared/platform-registry.ts`, the release catalog can surface Subvert, Beatport, EVEN, Jam.coop, Discogs, Bandwagon, Patreon, Buy Me a Coffee, Ko-fi, Liberapay, Hoopla, Freegal, and "Official Site" — none of those domains are allowlisted. `openReleaseUrl()` is a bare `if (isAllowedReleaseUrl(url)) { chrome.tabs.create({ url }) }` — no `else`, no log, no fallback — so clicking Buy for any of those sources does nothing, with zero indication to the user or to Sentry that anything failed. This is exactly the failure mode this PR's description calls out fixing for `unstream.stream` (missing from the old alert-only allowlist), except here it's introduced across a much larger set of platforms, because the same narrow list — previously scoped correctly to the alert "Listen" button, which only ever sees Bandcamp/Mirlo/Faircamp URLs from the release checkers — is now reused to gate the guide panel's full-catalog Buy buttons. No test in this PR exercises the click/open path, so it wasn't caught. Worth widening `ALLOWED_RELEASE_DOMAINS` to the full purchasable-platform domain list (or dropping the allowlist gate for the guide panel specifically, since those URLs come from the same server response already trusted to render the panel) before merge.
- **[plausible] Swift `ReleaseDetail.swift` models are almost entirely non-optional `Codable` fields**, so if the API ever omits/renames a single field on a single offer/source, `JSONDecoder` throws for the whole response and `ReleaseGuideView` falls back to a generic "Couldn't load prices" — losing all 10 sources instead of degrading the one that changed. The extension's JS side is duck-typed and degrades per-field instead, so this fragility is Swift-only. Not confirmed against a live endpoint change.
- **[needs-manual-verification]** `PopoverView.swift`'s reset-on-reopen logic for the guide (`.onAppear` / `.onReceive(.popoverDidClose)`) and the conditionally-registered Escape-key "Back" button interacting with `NSPopover`'s native Escape-to-close — looks internally consistent and mirrors the existing `showSignIn` pattern, but whether SwiftUI's shortcut wins the race against the popover's native close needs a real run, not just a read.
- Off-by-one / null-empty handling: none found — freshness day-count boundaries and payout math were checked against both `ReleaseDetailTests.swift` and `extension-release-display.test.ts` and agree (including the 86,399s/86,401s boundary); zero-price, sold-out, and empty-sources states are explicitly handled and tested identically in both clients.

Bottom line: the one thing I'd want addressed before merge is the Buy-button allowlist gap above — as written, most non-Bandcamp/Mirlo/Qobuz/Ampwall/Faircamp purchase links in the new guide panel are dead clicks in the extension.

---

brandonlucasgreen (commented):
@brandonlucasgreen Automated security/regression pass on this PR (29 files, +2733/-129). Full diff read in sequential chunks, no truncation.

## Security

**1. Buy button opens API-supplied URLs with no scheme/host validation (macOS/iOS)** — `confirmed`
`apps/mac/Unstream/Views/Shared/ReleaseGuideView.swift:363-368` (`buyButton`) and `:380-383` (`openOnUnstreamLink`, opens `detail?.pageUrl`).

`buyButton` builds `URL(string: source.url)` from the `GET /api/release/{artist}/{release}` response and hands it straight to `NSWorkspace.shared.open(url)` on macOS (a Safari sheet on iOS) — no protocol check, no hostname allowlist. This is meaningfully weaker than the sibling client shipped in the *same PR*: `apps/extension/popup/popup.js`'s `openReleaseUrl` (used by the extension's own new "Buy" button in `release-guide.js`) routes every open through the existing `isAllowedReleaseUrl()`, which requires `http:`/`https:` and matches `ALLOWED_RELEASE_DOMAINS`. If `source.url`/`pageUrl` were ever attacker-influenced (e.g. a poisoned value scraped from a third-party platform page upstream of #393's endpoint), the Swift app would open it unconditionally — including non-http(s) schemes that `NSWorkspace.shared.open` will hand to other installed apps/protocol handlers — while the extension would refuse the same URL.

Caveat: exploitability depends on how `/api/release` (added in #393, not part of this diff) sources/sanitizes `source.url` and `pageUrl` — couldn't trace that end-to-end from this diff alone. Worth adding the same scheme/host check on the Swift side regardless, for parity with the extension and defense-in-depth.

**2. Auth/authorization, input validation, secret handling, unsafe defaults, data exposure** — no findings. The new Swift/JS code doesn't introduce new privileged operations, doesn't handle secrets, and `apps/extension/lib/release-guide.js` builds the entire guide panel via `createElement`/`textContent`/`replaceChildren` — no `innerHTML`, no injection surface.

## Regression

**1. Artist page slug attached even when the DB write it depends on can fail silently** — `confirmed`
`api/functions/search-sources.ts`, new `attachArtistPageSlugs(finalResults)` call added immediately after the existing `try { await persistSearchResults(...) } catch (err) { console.error(...) }` block.

`persistSearchResults`'s failure is only `console.error`-logged, not surfaced or gated on. `attachArtistPageSlugs` then runs unconditionally, assigning `knownSlug` to every verified result regardless of whether the preceding persist actually succeeded. On a transient Supabase error during that request, the response still names a row that was never written — the exact failure mode the function's own doc comment says must not happen ("skip unverified results... `persistSearchResults` doesn't write them, so they have no row" / "names a row that exists rather than guessing at one"). A client then requesting `/api/artist-page?slug=...` gets a 404 for an artist it just displayed. The new `attach-artist-page-slugs.test.ts` only tests the function in isolation, not this call site's dependence on persist success.

**2. Changed assumptions / off-by-one / API contract** — verified, no findings:
- `freshness()` fix: Swift `ReleaseFormatting.freshness` (`floor(now.timeIntervalSince(captured)/86_400)`) matches `relativeDays` in `api/shared/release-display.ts` (`Math.floor((now - captured)/86_400_000)`) — both count elapsed 24h periods, not calendar days. Correct.
- `ALLOWED_RELEASE_DOMAINS` fix: only `unstream.stream` added; the existing suffix-match logic (`hostname === domain || hostname.endsWith('.' + domain)`) is unchanged and correct — no widening for other domains, and lookalike hosts (`unstream.stream.evil.example`) are rejected by both the JS and Swift slug parsers, which require exact-hostname match and are unit-tested for it.
- Popover state reset: `PopoverView` resets `route = []` on `.onAppear`/`.onReceive(.popoverDidClose)`, addressing the reuse concern from NSPopover hiding rather than destroying its content view — no stale-release flash found in the static code.

**3. Needs-manual-verification (per the PR description itself):** XCTest suite (`ReleaseDetailTests.swift`, 28 tests) compiled but couldn't run because the runner hangs while the screen is locked — re-run with the screen unlocked before merge, as noted in the PR description.

---

brandonlucasgreen (commented):
@brandonlucasgreen Automated security/regression review. Full diff read (3,423 lines, all 29 changed files, 100% coverage) plus targeted checks of pre-existing code the new hunks call into.

## Security

**1. Mac app's Buy button opens server-supplied URLs with no domain/scheme allowlist — plausible, low-to-medium severity**
`apps/mac/Unstream/Views/Shared/ReleaseGuideView.swift` (`buyButton(_:)`) does `URL(string: source.url)` → `NSWorkspace.shared.open(url)` with no host/scheme validation, where `source.url` comes from the new `GET /api/release/{artist}/{release}` response. Contrast with the extension, where this same PR gives `openReleaseUrl` an explicit `ALLOWED_RELEASE_DOMAINS` check ("so an alert can never open an arbitrary URL", `apps/extension/lib/constants.js`) before `chrome.tabs.create`. This pattern is pre-existing on the Mac app (`NewReleaseBadge.openInBrowser`, `LinkActions.swift` already open URLs unconditionally) — not a new hole — but this PR adds a new UI surface (one Buy button per marketplace source, up from one URL per alert) that inherits it rather than adopting the allowlist approach shipped for the extension in the same PR. Worth hardening for consistency between the two clients this PR ships together.

**2. `ALLOWED_RELEASE_DOMAINS` addition (`unstream.stream`) — confirmed correctly scoped, no issue.** Both consumers (`popup.js`, `service-worker.js`) use exact-match-or-proper-subdomain logic, not substring matching, so a lookalike host (`evilunstream.stream`) is correctly rejected.

**3. No SSRF, injection, or secret-handling issues — confirmed.** No new server-side outbound fetches. No `innerHTML` in the new extension code (`lib/release-guide.js`, `popup.js`) — built via `createElement`/`.textContent`. Artist/release slugs are percent-encoded before URL interpolation on both extension (`encodeURIComponent`) and Swift (`.addingPercentEncoding`, context-appropriate allowed-character sets) sides.

**4. No auth/authorization changes needed or made — confirmed.** The relevant endpoints are public/unauthenticated/read-only both before and after; new fields (`offerSummary`/`platforms`) are pricing data already publicly visible on the release page.

## Regression

**5. `attachArtistPageSlugs` can hand out a slug for an artist that was never actually persisted — confirmed design gap, low-to-medium severity.** `api/functions/search-sources.ts`'s new `attachArtistPageSlugs` assigns `knownSlug` assuming `persistSearchResults` wrote a matching row, but `persistSearchResults` (`api/functions/db.ts:2803`) silently skips artists with zero valid platform links and swallows per-artist upsert failures in a `try/catch` (`console.error` only, no signal back). A `verified` result can get a `knownSlug` that later 404s on the native client's "Releases & prices" button. Impact is soft — `ArtistReleasesView.swift` catches `.notFound` and shows "We don't have a page for this artist yet." rather than crashing — but it's the exact class of "don't hand out an address that 404s" bug the PR's own comments say they're avoiding, and the existing test (`attach-artist-page-slugs.test.ts`) only covers slug-assignment logic in isolation, not this persistence-success gap.

**6. `freshness()` elapsed-days fix — confirmed correct.** Uses `floor(timeIntervalSince / 86_400)`, timezone-independent elapsed-24h-period math, not calendar-day counting. Matches the accompanying XCTest assertions.

**7. NSPopover state reset (drill-down `route` stack) — confirmed correct.** Reset to `[]` in both `.onAppear` and `.popoverDidClose`, mirroring the existing `showSignIn` reset pattern. No stale-state leak found between opens.

**8. Feature-parity gap — needs-manual-verification.** The extension's release guide (`lib/release-display.js`, `lib/release-guide.js`) never renders the freshness/"prices checked" line that the Mac app's `ReleaseGuideView.footnote(_:)` shows. Likely a deliberate scope reduction for the smaller popup panel, but worth confirming it's intentional since the two clients ship together in this PR.

**9. API contract drift — mostly tolerated correctly.** New optional fields (`offerSummary`, `platforms`, `knownSlug`) are correctly declared `Optional` in Swift to tolerate an older server deploy. One brittleness note (needs-manual-verification if it matters to you): `ReleaseDetail.bandcampFriday: Bool` fields are non-optional, so a server response ever omitting them would fail the whole decode — though `ReleaseDetailService.fetch`'s `do/catch` turns that into a graceful "Couldn't load prices just now" + Try Again rather than a crash, so risk is low.

## Coverage note

`api/functions/release.ts` (the endpoint referenced throughout the PR description as landing in #393) is **not** touched by this diff — the only server-side change touching release data is the additive mapping in `api/functions/artist-page.ts`. `apps/extension/background/service-worker.js` is also not touched in this diff (only `constants.js`'s domain list, `popup.js`, and the two new `lib/` files are) — worth double-checking that's expected.

---

apps/mac/Unstream/Views/Shared/ReleaseGuideView.swift
```diff
@@ -0,0 +1,468 @@
...
+    /// Only used before a response arrives — once one has, `detail.pageUrl` is authoritative so
+    /// the app and the site can't disagree about where a release lives.
+    private var webPageURL: String {
+        "https://unstream.stream/a/\(target.artistSlug)/\(target.releaseSlug)"
+    }
+
+    private func open(_ url: URL?) {
+        guard let url else { return }
+        #if os(macOS)
+        NSWorkspace.shared.open(url)
```
brandonlucasgreen, line 412: **[confirmed] security — unsafe default: no scheme/host allowlist before opening server-supplied URLs.**

`open(_:)` (line 409-415) hands `source.url` (buy button, line 378/388) and `detail?.pageUrl` (openOnUnstreamLink, line 393-400) straight to `NSWorkspace.shared.open(url)` with only `URL(string:)` parsing — no scheme check, no hostname allowlist. Same pattern repeats in `ArtistReleasesView.swift`'s `open(_ url: URL?)` (line 287-291) for every release row's `linkActions`/drag/context-menu URL, all built from `source.url`/`release.slug` off the same `GET /api/release` and `GET /api/artist-page` responses.

This is the asymmetric half of a control this PR *does* ship: the extension's equivalent per-source Buy button (`apps/extension/lib/release-guide.js:81` → `openReleaseUrl` → `isAllowedReleaseUrl`) requires `http`/`https` and a domain in `ALLOWED_RELEASE_DOMAINS` before opening anything — comment there says explicitly "so an alert can never open an arbitrary URL." The Mac app gets no equivalent gate for the same server-computed data, and `NSWorkspace.shared.open` will hand a non-http(s) scheme to whatever app/handler claims it, which is a materially different risk profile than a browser tab.

Exploitability turns on whether `/api/release`'s `source.url`/`pageUrl` (added in #393, not part of this diff) can ever contain an untrusted value — can't be fully traced from this diff alone — but the two clients shipped in the same PR should enforce the same policy on the same data. Recommend adding a Swift equivalent of `isAllowedReleaseUrl` before every `open(...)` call in `ReleaseGuideView.swift` and `ArtistReleasesView.swift`.

---

apps/extension/lib/constants.js
```diff
@@ -2,7 +2,15 @@
...
 // Run `npm run sync:bandcamp-dates` after updating Bandcamp Friday dates in the web app.
 
+// Where a release alert is allowed to send someone. Guards both the popup's link and the
+// notification click, so an alert can never open an arbitrary URL.
+//
+// `unstream.stream` is first because it is now the *usual* destination, not an exception: a
+// catalogued release's `releaseUrl` is our own release page, so a fan lands on the payout
+// comparison rather than one shop. Leaving it out meant every catalog-backed alert had a Listen
+// button and a notification that silently did nothing when clicked.
 export const ALLOWED_RELEASE_DOMAINS = [
```
brandonlucasgreen, line 12: **[confirmed] regression — most "Buy" buttons in the new guide panel are dead clicks.**

`ALLOWED_RELEASE_DOMAINS` only lists `unstream.stream`, `bandcamp.com`, `mirlo.space`, `qobuz.com`, `ampwall.com`, `faircamp.eu`. This list was correctly scoped when it only gated the release-alert "Listen" link (which only ever sees Bandcamp/Mirlo/Faircamp/Qobuz URLs). This PR reuses the same check for the new guide panel's per-source Buy buttons (`release-guide.js:81` → `openReleaseUrl`, `popup.js:1237-1240`), but `SOURCE_CONFIG` in this same file (lines 20+) lists many more purchasable platforms the catalog can surface — `jamcoop`, `officialsite`, `discogs`, `bandwagon`, `nina`, `patreon`, `kofi`, `buymeacoffee`, `hoopla`, `freegal` (plus `beatport`, per `PAYOUT_PERCENTAGES`) — none of whose domains are in this allowlist.

`openReleaseUrl` (`popup.js:1237-1240`) is a bare `if (isAllowedReleaseUrl(url)) { chrome.tabs.create({ url }) }` with no `else`, no console warning, no Sentry signal. So clicking "Buy" for any of those sources silently does nothing — no error, no fallback, no indication anything failed. Given the release catalog is expected to include exactly these platforms (per `api/shared/platform-registry.ts`), this will affect a meaningful fraction of real Buy clicks. No test in this PR exercises the click/open path, so it wasn't caught.

Recommend widening `ALLOWED_RELEASE_DOMAINS` to the full purchasable-platform domain list, or dropping the allowlist gate specifically for the guide panel's Buy buttons since those URLs come from the same server response already trusted to render the panel's prices.

---

api/functions/search-sources.ts
```diff
@@ -2044,6 +2073,21 @@ export async function handler(event: { queryStringParameters?: Record<string, st
...
+    // search returned none. Measured on production before the fix — of six real searches, only
+    // the one *claimed* artist came back with a slug, while all six had rows carrying 16-21
+    // catalogued releases apiece. The data was there; the address for it wasn't.
+    //
+    // Deriving it client-side was the alternative and is rejected on purpose: `artistSlug` would
+    // then have to be reimplemented in Swift and in the extension's JavaScript, which is the same
+    // hand-copied-rule drift that the release endpoint's server-side `payoutPercent` exists to
+    // prevent. One definition, on the server.
+    //
+    attachArtistPageSlugs(finalResults);
```
brandonlucasgreen, line 2089: **[confirmed] regression — `attachArtistPageSlugs` can hand out a slug for a row that was never persisted.**

`attachArtistPageSlugs(finalResults)` runs unconditionally right after the `persistSearchResults(...)` call above (lines 2068-2074), whose failure is only `console.error`-logged, not surfaced or gated on. `persistSearchResults` (`api/functions/db.ts:2803`) also silently skips any artist with zero valid direct-link platforms and swallows per-artist upsert errors in its own try/catch. On a transient Supabase error (or an artist that gets filtered out of the persist batch), `attachArtistPageSlugs` still assigns `knownSlug = artistSlug(result.name)` to that same `verified` result, because it has no way to know the write didn't happen.

This is the exact failure mode the function's own doc comment (lines 1946-1949, "Skip unverified results… `persistSearchResults` doesn't write them, so they have no row" / "names a row that exists rather than guessing at one") says it's avoiding — it only accounts for the *unverified* case, not the *persist-failed* case. A client then requests `/api/artist-page?slug=...` / `/api/release/{artist}/{release}` for a page that 404s. Impact is soft (native clients render "We don't have a page for this artist yet" on 404), but it's a real address-that-404s handed out by design, and `attach-artist-page-slugs.test.ts` only tests the slug-assignment logic in isolation, not this call site's dependency on persist success.

---

apps/mac/Unstream/Services/ReleaseDetailService.swift
```diff
@@ -0,0 +1,92 @@
...
+            return try JSONDecoder().decode(ArtistPage.self, from: data)
+        } catch {
+            print("[ReleaseDetailService] Artist page decode failed for \(slug): \(error)")
+            throw ReleaseDetailError.unavailable
+        }
+    }
+
+    func fetch(artist: String, release: String) async throws -> ReleaseDetail {
+        guard let encodedArtist = artist.addingPercentEncoding(withAllowedCharacters: .urlPathAllowed),
+              let encodedRelease = release.addingPercentEncoding(withAllowedCharacters: .urlPathAllowed),
```
brandonlucasgreen, line 64: **[needs-manual-verification] input validation — wrong Foundation API for path-segment encoding.**

`artist`/`release` are percent-encoded with `.urlPathAllowed`, which is meant for an entire path and does not escape `/` — the JS sibling correctly uses `encodeURIComponent` per segment (e.g. `apps/extension/popup/popup.js:1330-1332`'s `fetchReleaseDetail`). The only call site today, `slugs(fromReleaseURL:)` (lines 27-33), derives both strings by splitting `url.path` on `/`, so a `/` can never reach `fetch` currently — not exploitable as written. Flagging because it's the wrong API for the job and would silently mis-encode/break path segmentation if this method is ever called with a slug from a different source (e.g. a future artist-search entry point). Low priority; worth swapping to `.urlQueryAllowed` or manual per-segment encoding for correctness regardless of current exploitability.

---

apps/mac/Unstream/Models/ReleaseDetail.swift
```diff
@@ -0,0 +1,220 @@
...
+struct ReleaseDetailSource: Codable, Hashable, Identifiable {
+    let platform: String
+    /// The platform's proper display name, from the server's registry — "Jam.coop", not "Jamcoop".
+    let name: String
+    let url: String
+    /// A range like "80-85%", or nil for a platform with no published rate (Discogs listings are
+    /// secondhand, so the artist genuinely receives nothing — worth saying, not worth hiding).
+    let payoutPercent: String?
+    /// True while Bandcamp is waiving its revenue share, which makes `payoutPercent` ~97%.
+    let bandcampFriday: Bool
```
brandonlucasgreen, line 42: **[plausible] regression — non-optional `Codable` fields make the whole response fragile to a single field change, unlike the JS sibling.**

`bandcampFriday: Bool` here (and other non-optional fields across `ReleaseDetailSource`/`ReleaseDetailRelease`/`ReleaseDetail`) mean that if the API ever omits or renames a single field on a single offer/source, `JSONDecoder.decode` throws for the *entire* response, and `ReleaseGuideView` falls back to a generic "Couldn't load prices" — losing all sources instead of degrading just the one that changed. `apps/extension/lib/release-guide.js` reads the equivalent JSON in a duck-typed way (`detail.bandcampFriday`, `source.payoutPercent ?`, etc.) and degrades per-field instead, so this all-or-nothing fragility is Swift-only. Impact is contained by `ReleaseDetailService.fetch`'s do/catch turning the decode failure into a graceful error state rather than a crash, so this is a UX/resilience note rather than a crash risk — flagging for awareness given the two clients are meant to stay in parity.

---

brandonlucasgreen (commented):
@brandonlucasgreen Automated security + regression review at the current head (`5bdc579`). Four earlier automated reviews exist on this PR but none match this head commit — three more commits landed after the last one (`0c3058b`), including a substantial new extension feature (artist releases panel) and a Mac-app UX pass (keyboard nav, drag, right-click) that expand the URL-opening surface — so I re-verified everything end to end rather than skip.

## Findings (5 line comments)

**Security — 1 confirmed, 1 needs-manual-verification:**
- **[confirmed]** `ReleaseGuideView.swift` / `ArtistReleasesView.swift` (macOS/iOS): every `open(url)` call — the buy button, "Open on Unstream", and every release row's drag/context-menu action — opens server-supplied URLs via `NSWorkspace.shared.open` with no scheme or hostname allowlist. The extension shipped in this same PR gives its equivalent Buy button exactly this protection (`isAllowedReleaseUrl` / `ALLOWED_RELEASE_DOMAINS`), so the two clients enforce different policies on the same server data. Exploitability depends on whether `/api/release` (added in #393, outside this diff) can ever return an untrusted URL — not traceable from this diff alone — but the inconsistency between sibling clients in the same PR looks unintentional.
- **[needs-manual-verification]** `ReleaseDetailService.swift`: `fetch(artist:release:)` uses `.urlPathAllowed` (doesn't escape `/`) instead of per-segment encoding like the JS sibling. Not currently exploitable — the only call site can't produce a `/` in either string — but the wrong API for the job.

**Regression — 2 confirmed, 1 plausible:**
- **[confirmed]** `apps/extension/lib/constants.js`: `ALLOWED_RELEASE_DOMAINS` (narrowly scoped for the old alert-only Listen link) is now reused to gate every source's Buy button in the new guide panel, but the catalog's `SOURCE_CONFIG` covers many more platforms (Jam.coop, Patreon, Ko-fi, Discogs, Bandwagon, Nina, Buy Me a Coffee, Hoopla, Freegal, Beatport, Official Site) whose domains aren't allowlisted. `openReleaseUrl` has no `else`/log/fallback, so most non-Bandcamp/Mirlo/Qobuz/Ampwall/Faircamp Buy clicks silently do nothing.
- **[confirmed]** `api/functions/search-sources.ts`: `attachArtistPageSlugs` runs regardless of whether the preceding `persistSearchResults` actually succeeded (its failure is only `console.error`'d), so a transient write failure can hand a client a slug for a page that 404s — the exact "don't name a row that doesn't exist" bug the function's own comments say to avoid, just for a case the comments don't cover.
- **[plausible]** `ReleaseDetail.swift`: non-optional `Codable` fields make the whole decode fail on any single missing/renamed field, unlike the extension's duck-typed JS, which degrades per-field. Contained by a graceful error fallback, so UX risk not crash risk.

## Scope / coverage note

Full diff read (35 files, +3298/-141) across all seven commits, including the two commits added after the last automated review (`678125c` Mac-assed keyboard/drag/right-click pass, `ae79bf1c` extension artist-releases panel + two unrelated bugfixes). Verified previously-reported items are still present or resolved at this head rather than assuming prior findings still apply (e.g. the earlier "dead `formatReleaseDate` export" note is now stale — it's used by the new `artist-releases.js`). Did not independently trace `/api/release`'s upstream data sourcing (#393, outside this diff), so URL-trust exploitability for finding #1 is capped at "plausible/confirmed-as-inconsistency" rather than confirmed end-to-end.

**Overall risk:** Medium — no auth/injection/secret issues found; the confirmed items are a real UX regression (dead Buy buttons) and a soft data-integrity gap (404-prone slugs), plus a genuine client asymmetry on URL-opening safety worth closing before merge.

---

brandonlucasgreen (commented):
@brandonlucasgreen Staff-eng security/regression pass on this one (full 4,119-line diff read, plus the already-merged endpoint/DB code it builds on).

**Security**

- **SSRF allowlist change — confirmed correct.** `apps/extension/lib/constants.js` `ALLOWED_RELEASE_DOMAINS` had exactly the literal string `'unstream.stream'` added; the matching logic in `popup/popup.js` (`isAllowedReleaseUrl`) is unchanged and not substring-based (`hostname === domain || hostname.endsWith('.' + domain)`), so there's no bypass risk (e.g. `evil-unstream.stream` is still correctly rejected). Minimal, matches the described bug fix.
- **XSS/injection — confirmed, no issue.** `release-guide.js`, `artist-releases.js`, and the new release-row code in `popup.js` all set server-supplied strings (title, price, payout text, platform name) via `textContent`, never `innerHTML`. No injection path for untrusted Bandcamp/Discogs text.
- **New endpoint `GET /api/release/{artist}/{release}` — confirmed, no issue.** Already merged via #393 and unmodified by this PR, but load-bearing for it: public/unauthenticated by design (same data the HTML release page already serves), strict slug regex before any DB call, parameterized Supabase queries, and correctly returns 503 (not cached) on a lookup failure vs. 404 on genuine absence — follows the repo's "never cache uncertainty" rule.
- **macOS has no equivalent domain allowlist — needs-manual-verification.** `ReleaseGuideView.swift`'s Buy button/link actions call `NSWorkspace.shared.open(url)` directly on server-supplied `source.url` with no scheme/domain gate, unlike the extension. This matches pre-existing platform-badge behavior elsewhere in the app so it's not new risk from this PR, but the two clients now diverge in how defensively they treat the same server response — worth confirming that's an accepted, deliberate asymmetry rather than an oversight now that a lot more surface calls it.

**Regression**

- **Extension "Buy" button silently no-ops for several live platforms — confirmed, most important finding.** `ALLOWED_RELEASE_DOMAINS` still only lists `unstream.stream`, `bandcamp.com`, `mirlo.space`, `qobuz.com`, `ampwall.com`, `faircamp.eu`. The new guide's Buy button (`release-guide.js:buildGuideSource` → `onOpenUrl` → `popup.js:openReleaseUrl` → `isAllowedReleaseUrl`) routes *every* platform's source URL through that same list. `api/shared/platform-registry.ts` includes `discogs.com`, `jam.coop`, `bandwagon.fm`, `patreon.com`, `buymeacoffee.com`, and `ko-fi.com` — none allowlisted. `release-guide.js`'s own comment even calls out "Discogs is the live case." So clicking Buy on a Discogs/Jam.coop/Patreon/Ko-fi/Buy Me a Coffee/Bandwagon offer does nothing — no error, `chrome.tabs.create` just never fires — which is the exact same silent-no-op bug class this PR's description says it fixed for `unstream.stream`. Recommend extending the allowlist to cover every `homepageUrl` domain in the platform registry, or deriving it from that registry so it can't drift again.
- **`freshness()`/`relativeDays` fix — confirmed correct.** Swift `ReleaseFormatting.freshness` (new) now matches `api/shared/release-display.ts::relativeDays` (unchanged) — both use elapsed-24h-period arithmetic, not calendar-day subtraction — and `ReleaseDetailTests.swift` pins the 23:59-vs-24:01-elapsed boundary. Matches the PR description's claim.
- **Feature-parity gap, minor — confirmed.** The extension's `release-guide.js` never renders `pricesCheckedAt`/freshness at all (Swift's `ReleaseGuideView.footnote` does). Not a bug, but worth a follow-up since "no client re-derives this, they all show what the server sent" is the PR's own stated design goal and the extension currently just omits the field rather than showing it.
- **macOS popover state reset — plausible, unverified by execution.** `PopoverView.swift`'s `route: [PopoverRoute]` resets in `onAppear` (fires every reopen, since `NSPopover` hides rather than destroys its view), on `.popoverDidClose`, and on the search-tab notification; Escape/⌘[/back button all pop the stack. Traced by code read only — the PR description itself notes XCTest couldn't run locked-screen, so this is a good candidate for the manual open→drill-in→close→reopen smoke test it already flags as outstanding.
- **Server-driven payout ordering — confirmed, no issue.** Both clients render `sources`/`offers` in server order with explicit "do not re-sort" comments, and neither hardcodes a payout percentage. `attach-artist-page-slugs.test.ts`/`artist-page.test.ts` assert Discogs's lower price doesn't out-rank Bandcamp's better payout server-side, so the clients displaying server order is doing the right thing.
- **API contract — confirmed, no issue.** Swift `CodingKeys` and the JS field access both match `release-detail.ts`/`artist-page.ts` field names exactly (`offerSummary`, `platforms`, `payoutPercent`, `bandcampFriday`, `detailCheckedAt`, `pricesCheckedAt`, `pageUrl`); Swift's `Codable` optionals mean an older-deploy response missing newer fields degrades rather than crashes.
- **`attachArtistPageSlugs` — confirmed, no issue.** Never overwrites an existing claimed/known slug and skips unverified results, covered by dedicated tests — avoids linking a release to an artist page that might 404.

Net: one confirmed regression worth fixing before merge (the extension allowlist gap above), everything else checks out or is flagged for a manual pass you were already planning.

---

PR #397, opened 2026-08-02
Handle service worker registration failures gracefully

brandonlucasgreen:
## Summary

Fixes unhandled promise rejections from service worker registration failures being reported as errors in Sentry. Service worker registration can be declined for legitimate reasons (Google's crawlers, incognito/private browsing, enterprise policy) that don't indicate a broken page. These are now recorded as breadcrumbs instead of errors.

## Changes

- **New service**: `apps/web/src/services/registerServiceWorker.ts` — registers the PWA service worker with proper error handling. Catches registration failures and logs them as Sentry breadcrumbs rather than unhandled rejections.
- **Updated entry point**: `apps/web/src/main.tsx` — calls `registerServiceWorker()` on production builds instead of relying on vite-plugin-pwa's auto-injected `registerSW.js`.
- **Vite config**: `apps/web/vite.config.ts` — disabled `injectRegister` in VitePWA plugin to prevent the unguarded registration call.
- **Test coverage**: `apps/web/tests/unit/register-service-worker.test.ts` — comprehensive unit tests covering successful registration, declined registration (breadcrumb recording), and browsers without service worker support.

## Implementation Details

The original issue: vite-plugin-pwa injects a single unguarded line (`navigator.serviceWorker.register('/sw.js', { scope: '/' })`) with no `.catch()` handler. When registration is declined (e.g., by Google's Web Rendering Service for Read-Aloud, or in private browsing), the unhandled rejection is caught by Sentry's global handler and reported as an error-level event, polluting error metrics.

The solution defers registration to the window `load` event and wraps it with a `.catch()` that records the failure as an info-level breadcrumb with the rejection reason. This preserves the information for debugging while correctly classifying declined registrations as non-errors.

The implementation is guarded behind `import.meta.env.PROD` since vite-plugin-pwa only emits `/sw.js` during production builds.

https://claude.ai/code/session_01UJ5mKPVKjJRzY5MpQSkkaW

### DIFF
--- a/apps/web/src/main.tsx
+++ b/apps/web/src/main.tsx
@@ -3,6 +3,7 @@ import { createRoot } from 'react-dom/client'
 import { BrowserRouter, Routes, Route, Navigate } from 'react-router-dom'
 import * as Sentry from '@sentry/react'
 import { initSentry } from './services/sentry'
+import { registerServiceWorker } from './services/registerServiceWorker'
 import { AuthProvider } from './contexts/AuthContext'
 import './index.css'
 import App from './App.tsx'
@@ -12,6 +13,11 @@ import { ScrollToTop } from './components/ScrollToTop'
 import { lazyWithRetry } from './utils/lazyWithRetry'
 
 initSentry()
+
+// Ours rather than vite-plugin-pwa's injected registerSW.js, which reports a
+// declined registration as an unhandled rejection. Guarded on PROD because the
+// plugin only emits /sw.js for a build, exactly as the injected script was.
+if (import.meta.env.PROD) registerServiceWorker()
 
 // Lazy-load non-critical pages to reduce initial bundle size.
 // lazyWithRetry adds a one-shot reload when a chunk belongs to a superseded deploy.
--- a/apps/web/src/services/registerServiceWorker.ts
+++ b/apps/web/src/services/registerServiceWorker.ts
@@ -1,0 +1,41 @@
+import * as Sentry from '@sentry/react'
+
+/**
+ * Register the PWA service worker, and treat a refusal as a refusal rather than
+ * a crash.
+ *
+ * vite-plugin-pwa injects this for us by default, but the `registerSW.js` it
+ * generates is one unguarded line:
+ *
+ *   navigator.serviceWorker.register('/sw.js', { scope: '/' })
+ *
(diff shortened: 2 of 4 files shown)


brandonlucasgreen (commented):

**Security**

- Auth/authorization: none found — no auth-relevant code touched.
- Input validation: none found — no external/user input is handled by this change.
- Secret handling: none found.
- Injection risks: none found — the Sentry breadcrumb `data.reason` is a plain string from `error.message` (or `String(error)`), never interpolated into HTML/DOM/SQL.
- Unsafe defaults: none found — the change *tightens* the previous default (vite-plugin-pwa's auto-injected `registerSW.js` had no `.catch()` at all; this replaces it with an explicit, caught registration).
- Data exposure: none found. `data.reason` will typically be a generic browser string (e.g. `"Rejected"`, a `DOMException` name) — not PII, not a secret. [needs-manual-verification] if you want extra certainty: worth a spot-check in the Sentry dashboard after a few real declines land, just to confirm no browser ever stuffs anything sensitive into that rejection message — but nothing in the code itself points to a problem.

**Regression**

- Changed assumptions [confirmed, verified by reading the rest of the codebase]: registration moves from vite-plugin-pwa's auto-injected script to this explicit call, gated by `import.meta.env.PROD` and deferred to `window.addEventListener('load', ...)`. I checked for any other code relying on vite-plugin-pwa's `virtual:pwa-register` / `useRegisterSW` / periodic-update-check hook (which some vite-plugin-pwa setups use for auto-update polling) — there is none in this repo, so nothing is silently lost by turning `injectRegister` off.
- Edge case [needs-manual-verification]: registration is deferred via `window.addEventListener('load', ...)`. If `load` has already fired by the time `registerServiceWorker()` runs, the listener would never fire and the SW would silently never register. In practice this call sits at the top of `main.tsx`, which executes as a deferred ES module before `window.load`, so this should be safe — but it's worth a quick real-browser check (Network tab / Application tab) post-merge rather than trusting timing assumptions alone.
- Off-by-one / state transitions / API contract drift: none found — no state machine or API surface is touched.

Overall: small, well-scoped, well-tested change (`register-service-worker.test.ts` covers success, decline, and no-SW-support paths). No confirmed security or regression issues.

---

PR #399, opened 2026-08-02
Stop reporting an in-app browser's own script as our error

brandonlucasgreen:
Sentry has been collecting "undefined is not an object (evaluating
'window.webkit.messageHandlers')", with frames sendDataToNative and
sendPageHideMessage, all attributed to https://unstream.stream/:1.

None of that is ours. Neither function name appears anywhere in this repo,
and our document has no inline JavaScript at all — index.html loads the
bundle and GoatCounter as external files, and the og-metadata edge function
that renders the homepage adds no scripts — so line 1 of the document can
never be our code. It's an iOS in-app browser injecting its own script into
the page it opened: that script reports back to its host app over
window.webkit.messageHandlers and throws when the bridge is gone, which the
pagehide frame explains, as the webview tears the bridge down while the
listener still fires. Injected scripts have no source URL of their own, so
WebKit blames the document and the event lands in our project.

Drop it in beforeSend. We can't cause it, fix it, or see which app the user
was in. If Unstream ever loads itself in a WKWebView, the filter has to go —
it would be hiding a real bug — and the comment says so.

Also read the message from the event as well as originalException. Errors
that arrive through the global handler often carry no originalException,
which is exactly the shape this one has, and a filter that can only read one
of the two would have missed it.

Claude-Session: https://claude.ai/code/session_01NwS2rvTzx2dfXd9n6vzjvy

### DIFF
--- a/apps/web/src/services/sentry.ts
+++ b/apps/web/src/services/sentry.ts
@@ -38,6 +38,29 @@ export function isStaleBuildAssetError(message: string): boolean {
   )
 }
 
+/**
+ * True when an error came from a WebKit-to-native bridge script that something
+ * outside our page injected into it.
+ *
+ * Sentry attributes these to us because the stack points at
+ * `https://unstream.stream/:1` — but our document has no inline JavaScript
+ * (`apps/web/index.html` loads the bundle and GoatCounter as external files, and
+ * the og-metadata edge function adds no scripts), so line 1 of the document is
+ * never our code. It's an in-app browser: iOS apps that open links in a WKWebView
+ * inject their own script, which reports home over `window.webkit.messageHandlers`
+ * and throws when that bridge isn't there — usually from a `pagehide` listener
+ * firing after the webview has already torn it down. The frame names in those
+ * traces (`sendDataToNative`, `sendPageHideMessage`) exist nowhere in this repo.
+ *
+ * Nothing we ship causes it and nothing we ship can fix it, so it's noise. If
+ * Unstream ever loads itself in a WKWebView of its own (the Apple app is native
+ * SwiftUI today, with no webview anywhere), delete this filter — at that point it
+ * would be hiding a real bug.
+ */
+export function isInjectedNativeBridgeError(message: string): boolean {
+  return message.includes('webkit.messageHandlers')
+}
+
 export function initSentry(): void {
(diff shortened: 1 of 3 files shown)


brandonlucasgreen (commented):

**Scope reviewed:** full diff (`apps/web/src/services/sentry.ts`, `apps/web/tests/unit/sentry-injected-bridge.test.ts`) — 2 files, +75/-1.

## Security
- **Auth/authorization:** N/A — no auth surface touched.
- **Input validation:** N/A — no external/user input parsed. `isInjectedNativeBridgeError` only substring-matches text already captured by Sentry's own instrumentation.
- **Secret handling:** N/A — no secrets introduced.
- **Injection risks:** none — `confirmed`. The new code only does `.includes()` string checks; nothing is eval'd, interpolated into HTML, or used to build a URL/query.
- **Unsafe defaults:** none — `confirmed`. Filter is narrowly scoped (`webkit.messageHandlers` substring) and only suppresses Sentry reporting, not app behavior.
- **Data exposure:** none — `confirmed`. Grepped the repo (`apps/web`) for `webkit`; only other hits are unrelated `AppleWebKit` UA-string checks in `apps/web/server/search/*`, so the new filter can't accidentally swallow real errors that happen to share the string.

## Regression
- **Changed assumptions:** `errorMessage` now joins `hint?.originalException?.toString()` **and** `event.exception?.values?.[0]?.value` (previously only the former). This widens matching for the pre-existing `isStaleBuildAssetError`, `Network Error`, and `AbortError` filters too, not just the new bridge filter — intentional per the PR description, but worth knowing it's a behavior change for those existing filters, not just additive. Label: `plausible` (widening is correct in the cases shown, but there's no test asserting the pre-existing filters still behave correctly against the new joined-string shape).
- **Edge cases:** when both `originalException` and `event.exception.values[0].value` are empty/undefined, `errorMessage` becomes a single space `' '` (from `.join(' ')`) rather than `''`. All downstream checks are `.includes()`, so this is harmless — `confirmed` no impact.
- **Off-by-one/state/API contract:** none applicable — no stateful logic or API surface here.

No blocking findings. Small, well-scoped, well-tested (4 unit tests covering the new predicate, including a negative case for a real app error message).

---

PR #402, opened 2026-08-02
Fix saving an artist while logged in, and make a failed save visible

brandonlucasgreen:
Saving an artist while signed in was broken. Two independent bugs, both ending in a 400 from `/api/saved-artists` — and because the client rolls a failed save back, the heart filled in and immediately emptied again with nothing to explain it.

## What was wrong

**1. The artist page sent the artists-table UUID where a slug belongs.**

Saving is keyed by slug end to end: `handleSave` validates the artist-slug format, stores `artist_slug`, and returns slugs, so `savedArtistIds` holds slugs too. `ArtistPage.tsx` read `payload.artist.id` — the `artists.id` UUID — and passed it to both `saveArtist()` and `isArtistSaved()`.

So every save from `/a/:slug` and `/artist/:slug` failed validation. Two knock-ons: even a save that *had* worked would render as unsaved, because a UUID can never match a set of slugs; and unsave silently no-op'd, because `handleRemove` didn't validate, matched no rows, and returned `200 success`.

`artist.id` is still correct for the admin "Catalog releases now" button — that endpoint validates a UUID — so the two identifiers are now named apart instead of one value being quietly wrong for one of its two uses.

**2. The slug format check rejected real artist slugs.**

`SLUG_REGEX` was `/^[a-z0-9](?:[a-z0-9-]{1,18}[a-z0-9])$/` — 3 to 20 characters, copied from migration 021's `username_format`. That constraint describes *usernames*, which people choose; artist slugs come out of `slugify()` on an artist's name. It rejected **24 of the 791 published slugs**, including `explosions-in-the-sky`, `queens-of-the-stone-age`, `bt`, `iq` and `x`, so saving those artists from search results failed too — as did any `nameonly-…` result key long enough to trip it. Now 1–60 characters, against a longest published slug of 38.

Worth flagging: widening the bound removes the *only* thing that was rejecting a UUID, since lowercase hex and hyphens satisfy the character class. That would have turned bug 1 silent instead of loud — a row stored under the UUID, named by titleizing it, and a `/a/{uuid}` link on the fan's public share page that 404s. So the UUID shape is now refused explicitly with a message saying what was wrong. The character class, which is what actually stops stored XSS reaching the edge SSR, is unchanged.

## Making it visible next time

The bugs being invisible for weeks was the more expensive problem, so this closes that on both sides of the request.

**The endpoint reported nothing at all.** Every failure path was a `console.error` — a Lambda log nobody reads — behind a generic 4xx/5xx. The browser extension fires its save and never looks at the response, so a rejected save there was invisible at both ends. Each refusal and each failed write now reaches Sentry: a missing `artistId`, a malformed slug, an id sent where a slug belongs, the upsert failing, the tombstone write failing, the list and bulk-check queries failing, the support toggles failing, a thrown error, and Supabase not being configured at all. The database's own message and code go on the event — "permission denied for table saved_artists" is a different morning than "deadlock detected".

Deduped on the reason alone with a 5-minute window, not on the rejected value: `reason` is a fixed set, so a client looping on a rejection can't flood the project by varying the slug it sends. The first event in each window carries the value. No user id in the dedup key, and the offending value is truncated before it reaches a tag.

**The client said something failed, but never what.** "Save artist failed (rolled back)" had been firing all along with no status code and no server message, so a 400 from a validation bug we shipped looked exactly like a flaky network. Status code is now a tag, so the issue page separates a client bug from an expired session from an outage at a glance, and the server's explanation rides along in `extra`. Same for remove and for the saved-artists load.

**Remove now validates its identifier the way save does**, so it can't answer 200 for a removal that did nothing. Every write has always gone through the format check, so no row stored under a UUID can exist to be locked out by this.

Deliberately *not* reported: a 401, which is just an expired token and would bury everything real; and a remove that matches no rows, which happens benignly when two devices unsave the same artist.

## Why nothing caught it

`npm run dev` serves no `/api/saved-artists` route at all, so this path can't be exercised locally — and `ArtistPage.test.tsx` mocked `saveArtist` with a fixture id of `'artist-1'`, a short placeholder that happened to pass the old regex.

The new tests drive the real handler and the real `AuthProvider` rather than re-simulating their logic. Checked against the pre-fix code: 10 of the 20 new function tests fail on the old regex, so they would have caught the original bugs.

I also corrected two test fossils that asserted a UUID contract for this endpoint under comments claiming to mirror `saved-artists.ts`. They didn't — and that stale mental model is plausibly how the artist page came to send an id in the first place.

## Verification

`npm run build` passes end to end — 588 web unit tests, 745 API tests, both typechecks. Lint is unchanged at its pre-existing 67 errors, all in files this branch doesn't touch. No migration, no config change, no new environment variable.

Untested by me: the deployed behaviour against the real Supabase, since the save path can't run locally. Worth a click on an artist page on the deploy preview — including one of the long-slug artists like `explosions-in-the-sky`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JWqKm9xnycdezrNzH3zpVZ)_


api/functions/saved-artists.ts
```diff
@@ -13,14 +13,84 @@
...
+  // Defense in depth: reject slugs that don't match the canonical artist slug format. This
+  // prevents stored XSS via malicious slugs reaching the edge function SSR — the character class
+  // is what does that work, by admitting nothing that can break out of HTML.
+  //
+  // The bound used to be 3–20 characters, copied from migration 021's `username_format`. That
+  // constraint is about *usernames*, which people choose; artist slugs are generated by slugify()
+  // from an artist's name and are neither that short nor that bounded. It rejected 24 of the 791
+  // published artist slugs — 'explosions-in-the-sky', 'queens-of-the-stone-age', 'bt', 'iq', 'x'
+  // — so saving those artists 400'd. 60 is comfortably above the longest one we publish (38).
+  const SLUG_REGEX = /^[a-z0-9](?:[a-z0-9-]{0,58}[a-z0-9])?$/;
```
brandonlucasgreen, line 71: **Regression — edge case: the new 60-char cap can still reject a real save path.** `confirmed` (reproduced below)

The comment says 60 is "comfortably above the longest [slug] we publish (38)," which is true for `data/artists/*` slugs, but `artistId` here isn't only ever a published slug — for verified-but-unclaimed search results, `ResultCard.tsx` falls back to the synthetic key `nameonly-${normalizedName}` (and `search-sources.ts:1414`/`1434` append `-${Date.now()}` when disambiguating same-name artists), and that's exactly what gets sent to `saveArtist`/`removeSavedArtist` for those results.

`"nameonly-" (9) + normalizedName + "-" + 13-digit epoch ms (14)` leaves only ~37 chars for the artist name before hitting the 60-char ceiling. Any band with a longer name than that fails the format check on save, e.g.:

```
nameonly-and-you-will-know-us-by-the-trail-of-dead-1785689203616   (64 chars, > 60)
```

This is the *exact* silent-400/heart-fills-and-empties bug this PR sets out to fix, still reachable through the disambiguated-search-result save path the PR didn't touch. Worth either excluding the `nameonly-...` id space from this slug check (it's never looked up against `artists.slug` anyway — `findExistingArtist` will just miss) or raising the bound / stripping the suffix before validating.

---

api/functions/saved-artists.ts
```diff
@@ -199,15 +279,20 @@ async function handleSave(user: { userId: string; email: string }, body: Record<
...
-  // This prevents stored XSS via malicious slugs reaching the edge function SSR.
-  const SLUG_REGEX = /^[a-z0-9](?:[a-z0-9-]{1,18}[a-z0-9])$/;
-  if (!SLUG_REGEX.test(artistSlug)) {
-    return { statusCode: 400, headers: CORS_HEADERS, body: JSON.stringify({ error: 'Invalid artist slug format' }) };
+  const invalid = invalidArtistSlugReason(artistSlug);
+  if (invalid) {
+    await reportSavedArtistFailure(`save-${invalid.reason}`, `Save rejected — ${invalid.error}`, {
+      artistSlug: forTag(artistSlug),
+      artistSlugLength: artistSlug.length,
+      artistName,
```
brandonlucasgreen, line 293: **Security: data exposure / unbounded input to Sentry.** `plausible`

`artistSlug` is truncated with `forTag()` before being attached to the report, but `artistName` (`body.name`, fully client-controlled, no length or content validation anywhere in this handler) is passed straight through unbounded. A caller can put an arbitrarily large or arbitrary string in `name` and have it land in Sentry's `extra` on every `save-*` validation failure. Low severity (authenticated endpoint, rate-limited, Sentry payload limits will eventually clip it), but worth the same `forTag()` treatment for consistency and to avoid surprising Sentry ingestion costs/truncation-in-the-wrong-place.

---

api/functions/saved-artists.ts
```diff
@@ -13,14 +13,84 @@
...
+ * fixed set, so a client looping on a rejection can't flood the project by varying the slug it
+ * sends. The first event in each window carries the value, which is all that's needed to diagnose.
+ * No user id in the key — dedup keys shouldn't hold PII.
+ */
+async function reportSavedArtistFailure(
+  reason: string,
+  message: string,
+  detail: Record<string, unknown> = {},
+): Promise<void> {
+  if (!await checkSentryDedup(`saved-artists:${reason}`, 300)) return;
```
brandonlucasgreen, line 45: **Security: unsafe default — new Sentry-capture path fails open when Redis is down.** `needs-manual-verification`

`checkSentryDedup` (unchanged, in `ratelimit.ts`) returns `true` — "go ahead and capture" — whenever `getRedis()` is unavailable, by design ("Redis unavailable → don't gate Sentry on it"). That's a reasonable default on its own, but this PR is what wires `reportSavedArtistFailure` into every validation-failure branch of `/api/saved-artists` (previously these paths only did `console.error`, which called nothing external). So a Redis outage now turns every malformed-`artistId` request into an un-deduped `Sentry.captureMessage` call on an authenticated-but-otherwise-unbounded-in-count endpoint (IP rate limiting via `checkRateLimit('standard')` still applies, but that's per-IP, not global). Not exploitable without an independent Redis outage, so flagging for awareness/verification rather than as a standalone bug — worth confirming Sentry's own rate limiting is enough of a backstop, or adding a coarse in-process cap for this specific fail-open case.

---

apps/web/tests/unit/AuthContext-save-reporting.test.tsx
```diff
@@ -0,0 +1,201 @@
...
+describe('AuthContext — reporting a failed save', () => {
+  beforeEach(() => {
+    cleanup();
+    vi.clearAllMocks();
+    localStorage.clear();
+    held.current = null;
+    mocks.onAuthStateChange.mockReturnValue({ data: { subscription: { unsubscribe: vi.fn() } } });
+  });
+
+  it('puts the status code and the server\'s explanation on a rejected save', async () => {
```
brandonlucasgreen, line 89: **Regression — flaky test (state transitions / timing).** `confirmed` (reproduced)

`puts the status code and the server's explanation on a rejected save` is intermittently failing (`captureMessage` called 0 times, so `Sentry.captureMessage` never fired) when this file runs as part of the full suite alongside the other files this PR touches. Reproduced locally: passed 8/9 runs of

```
npx vitest run api/functions/__tests__/saved-artists-reporting.test.ts api/functions/__tests__/xss-defense.test.ts \
  apps/web/tests/unit/ArtistPage.test.tsx apps/web/tests/unit/AuthContext-save-reporting.test.tsx \
  apps/web/tests/unit/AuthContext-saved-artists.test.ts apps/web/tests/unit/saved-artists-api.test.ts
```
but failed once with exactly this test as the only failure; running the file alone or filtered with `-t` never reproduced it. Looks like a race between the mocked `fetch` resolving and React's `act()` batching under the thread pool (the test does `await act(async () => { await api().saveArtist(...) })` off a value captured in a `useEffect`-refreshed ref). Since this runs in the build-blocking `npm run test:unit` pipeline, this is a real (if low-probability) source of spurious CI failures on unrelated PRs, not just this one.

---

brandonlucasgreen (commented):

**Summary:** The core fix (ArtistPage was saving `artist.id` — a UUID — instead of `artist.slug`, and the slug-format regex was rejecting 24 real published slugs) is correct, well-tested, and the failure-visibility work (Sentry on every save/remove failure path) is a solid observability improvement. No auth/authorization or injection issues found — auth checks are unchanged, the shared slug/UUID validation is applied consistently to save and remove, and the character class still blocks anything that could break out of HTML in the edge SSR.

**Findings: 4 total** — 0 `confirmed`-security, 2 `plausible`/`needs-manual-verification`-security, 2 regression (1 confirmed edge case, 1 confirmed flaky test).

- Security: `plausible` — `artistName` sent unbounded to Sentry `extra` (unlike `artistSlug`, which is truncated via `forTag()`).
- Security: `needs-manual-verification` — the new Sentry-report wiring fails open (uncapped, un-deduped) if Redis is down, a new-to-this-PR attack surface on validation-failure paths that previously called nothing external.
- Regression (confirmed): the raised 60-char slug bound still isn't enough for the `nameonly-<name>-<timestamp>` synthetic ids used to save unclaimed search results with longer band names — reproduced a 64-char id that fails the exact validation this PR just relaxed.
- Regression (confirmed): `AuthContext-save-reporting.test.tsx`'s status-code test is intermittently flaky in the full suite (reproduced: 1 failure in 9 runs), a timing race unrelated to the change under test.

Overall risk: low. Nothing here blocks merge; the two security notes are worth a look before or shortly after shipping, and the `nameonly-` edge case is a real (if narrower) recurrence of the bug this PR fixes.

---

PR #403, opened 2026-08-02
Move artist removal from the dashboard to a danger zone

brandonlucasgreen:
Removing a claimed artist sat one click away from Edit and View on the
dashboard, which is too easy for an action that hands the page back. The
button is gone from there; removal now lives at the bottom of the edit
page in a danger-zone container behind typing the artist's name.

The dashboard button was a stub that only showed a "coming soon" toast,
so this adds the real thing: DELETE /api/artist-profile deletes the
artist_profiles row (bio, photo, featured release, dividers) and reverts
the artists row to auto-discovered. The artist row itself stays — it is
public search data that existed before the claim — as do the platform
links, and the artist can claim the page again later.

Left flagged as claimed, the artist row would never expire and enrichment
would keep skipping it, so the page would stay frozen on the last edit of
an artist who just asked to be removed. A failed revert can't be fixed
automatically, so it reports to Sentry.

Also adds artist-profile.ts to the API typecheck include (it passes
strict cleanly) and pulls the shared cache purge out of the PUT path.

Claude-Session: https://claude.ai/code/session_01WVFKxLzSw5nvK2K72bc1aM

### DIFF
--- a/api/functions/__tests__/artist-profile-remove.test.ts
+++ b/api/functions/__tests__/artist-profile-remove.test.ts
@@ -1,0 +1,161 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+
+// DELETE /api/artist-profile removes an artist's claim. It is the only destructive action an
+// artist can take on their own profile, so these pin down that it can't be reached without
+// ownership, that it doesn't delete another user's row, and that a page whose claim is gone
+// doesn't stay flagged as claimed (which would freeze it — see db.ts getArtistBySlug).
+
+const mocks = vi.hoisted(() => ({
+  mockFrom: vi.fn(),
+  mockResolveOwnedArtist: vi.fn(),
+  mockCreateClient: vi.fn(),
+  mockCheckRateLimit: vi.fn(() => Promise.resolve({ limited: false })),
+  mockGetClientIp: vi.fn(() => '127.0.0.1'),
+  mockCacheDeleteByArtist: vi.fn(() => Promise.resolve()),
+  mockCaptureException: vi.fn(),
+}));
+
+vi.mock('../db', () => ({
+  getClient: () => ({ from: mocks.mockFrom }),
+  resolveOwnedArtist: mocks.mockResolveOwnedArtist,
+}));
+vi.mock('@supabase/supabase-js', () => ({
+  createClient: mocks.mockCreateClient,
+}));
+vi.mock('../ratelimit', () => ({
+  checkRateLimit: mocks.mockCheckRateLimit,
+  getClientIp: mocks.mockGetClientIp,
+}));
+vi.mock('../cache', () => ({
+  cacheDeleteByArtist: mocks.mockCacheDeleteByArtist,
+}));
+vi.mock('../../lib/sentry', () => ({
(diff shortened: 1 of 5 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Staff-eng security/regression pass on this one (full diff read, all 5 changed files).

**Security**

- **Auth/authorization — confirmed OK.** `DELETE` reuses the same `authenticateRequest` bearer check as `PUT`, then `resolveOwnedArtist(slug, userId)` requires both `profile.user_id === userId` *and* `verified_at` set before anything happens. The delete itself is additionally scoped with `.eq('artist_id', ...).eq('user_id', ...)` even though ownership was just proven — good defense in depth, and it's covered by a dedicated 403 test ("someone else's profile").
- **Input validation — confirmed OK.** Missing `slug` → 400 before touching the DB. No raw SQL/string interpolation anywhere; all Supabase calls are parameterized.
- **Secret handling — confirmed OK.** No new secret handling; `NETLIFY_API_TOKEN`/`NETLIFY_SITE_ID` usage is just refactored into `purgeArtistCaches`, not changed.
- **Injection — confirmed OK.** Nothing new here.
- **Unsafe defaults — needs-manual-verification (pre-existing, not introduced by this PR).** `CORS_HEADERS` on this file is `Access-Control-Allow-Origin: '*'` and now also covers the new destructive `DELETE`. Since auth is bearer-token (not cookie-based) this isn't classic CSRF, but it does mean any origin's JS can call this endpoint if it holds a valid token. Worth a follow-up glance at whether this file should match the tighter CORS pattern used elsewhere (`buildCorsHeaders`/`buildPublicCorsHeaders` in `middleware.ts`) — not blocking, since `PUT` already had the same wildcard.
- **Data exposure — confirmed OK.** Response body is just `{ success: true }` / an error string; no row data leaked.

**Regression**

- **Changed assumptions — confirmed OK.** Removing a claim un-verifies (`match_confidence: 'unverified', source: 'auto'`) rather than deleting the `artists` row, so search/enrichment keep working for the now-unclaimed page. Reasoning is documented inline and matches `getArtistBySlug`'s claimed-row skip logic described in `db.ts`.
- **Edge cases — confirmed OK.** Race/no-op case (`count: 0`, e.g. double-submit or already-removed) returns 409 and explicitly skips the `artists` revert (`updateFn` not called) — tested. Delete-error path also correctly skips the revert and reports to Sentry — tested.
- **Partial-failure state transition — confirmed OK, by design.** If the profile delete succeeds but the `artists` revert fails, the endpoint still returns 200 (the claim *is* gone either way) but fires `Sentry.captureException` so the now-frozen `artists` row doesn't fail silently — tested explicitly.
- **Dead-code cleanup — confirmed OK.** The `DashboardPage` toast `'info'` variant and its icon branch were fully removed together with the only caller (`handleRemoveClaimedProfile` stub); grepped the file, no leftover `'info'` references.
- **API contract — confirmed OK.** New `DELETE` takes `slug` via query string (consistent with other delete endpoints per the inline comment), frontend (`ArtistEditPage.handleRemoveArtist`) matches: `?slug=...` + `Authorization` header, no body.

No confirmed or plausible bugs found. One low-severity, pre-existing CORS note above for awareness, not a blocker.

---

PR #404, opened 2026-08-02
Refresh saved artists' release catalogues on a schedule

brandonlucasgreen:
## The problem

Every release-catalog trigger was demand-driven — a fan saving an artist, a search resolving one, an artist's own button, the admin command. And `check-releases` only *reads* the catalogue; it never refreshes it, and only falls back to a live scrape for an artist the catalogue has **never** seen.

So: a fan saves an artist → the catalogue is built once → the weekly alert check reads that same catalogue forever. If nobody ever searches that artist again, **their new releases are never discovered**. Popular artists stayed fresh incidentally because people search them; the long tail of saved-but-not-searched artists silently stopped producing alerts. That's worse than the pre-catalogue behaviour, where the client re-scraped weekly, and it undermines pillar 3 of the V1 Releases scope.

## What this adds

`api/functions/recatalog-sweep.ts`, run daily by `.github/workflows/recatalog-sweep.yml` (GitHub Actions cron — there are no scheduled Netlify functions in this repo).

It selects the stalest catalogues **among artists somebody has saved**, ordered:

1. Never attempted first — saved, but the save-time request never got through (usually the hourly cap). Nothing else retries those.
2. Then oldest `last_attempted_at`.
3. Then most savers, **as a tiebreak only**. Popularity-first would starve exactly the long tail the sweep exists for.

Then it asks `requestArtistCatalog` for up to 25 of them. **No second rate limiter** — the 7-day cooldown and the hourly cap in `claimArtistForCatalog` still decide, which is also what makes a duplicate invocation a no-op. The cooldown is read up front only to pick who to ask about, so a bounded batch isn't spent on artists that'll be refused a moment later.

## Cadence, and what it costs Bandcamp

**Daily, 05:00 UTC** — off both US and European peak.

One sweep is at most 25 artists through the existing background function, which paces itself: a second between artists, at most 100 release-page fetches per run on top of one grid page each. On the order of 125 sequential requests spread over minutes — a trickle, less than a handful of real visitors searching.

Daily rather than weekly because the **7-day cooldown, not the cron, is what bounds per-artist crawling**: an artist can't be re-crawled more than once a week however often this runs. Daily just spreads the same work across seven small runs instead of one lump, and it shortens how long a newly-saved artist waits for a first catalogue if their save-time request was dropped by the cap.

## Why a new `scheduled` trigger

- **Budget.** `CATALOG_HOURLY_CAP` is keyed by trigger and counted across *all* attempts in the last hour. Give the sweep the smallest budget and an ordinary hour of search traffic leaves it nothing — it runs once a day at a fixed time and can't wait for a quieter moment the way a search can be dropped and re-triggered by the next visitor. Give it the largest and a machine outranks a person. So: 120, between `searched` (60) and `saved` (240).
- **Observability.** `last_trigger = 'scheduled'` is how you tell from the table alone whether the sweep is still running.

## Observability

A sweep that silently stops working recreates the exact bug it fixes, so:

- **Every refusal is a non-2xx that fails the workflow, plus a Sentry event** — disabled deploy, unreadable database, handshake that never left. It returns a real summary rather than 202 for this reason (a `-background` function would 202 whatever happened).
- `200` with `requested: 0` is a legitimate quiet day. The counts in the body (`savedArtists`, `inCooldown`, `neverAttempted`, `stalestAttemptedAt`) are what let a reader tell that from broken selection.
- **`recordCatalogOutcome` now reports the "found 0 where it previously found 20" transition to Sentry.** That case is recorded as a *success*, so nothing else about it looks wrong. Fires on every trigger, not just the sweep, and only on the transition — an artist who has always had 0 isn't news.
- `requestArtistCatalog` now returns whether the handshake completed. Not "the crawl was accepted" — the dispatcher 202s anything — but a failed handshake proves it didn't, and the sweep has no next visitor to ask again. User-facing callers ignore it and still never see an error.

## Also in here

- The background function's private copy of the internal-secret check moved into `middleware.ts` as `isInternalRequest()`, so both endpoints authenticate identically instead of via a third copy.
- Fixed the trigger parse (`=== 'saved' ? 'saved' : 'searched'`) that would have silently spent the *search* budget on a scheduled sweep.
- `recatalog-sweep.ts` and `catalog-artist-background.ts` added to `api/tsconfig.json`'s narrow include, so they're actually typechecked.
- CLAUDE.md: a "Keeping catalogues fresh" section, and the workflow list — which was missing `upstash-keepalive.yml` too.

## Verification

- `npm run build`, `npm run test:api` (791 passing), `npm run test:unit` (625 passing), `npm run typecheck:api` — all clean. `dispatch.xml` / `sitemap.xml` reverted after the build.
- **The migration was validated against a real `postgres:16`.** The risk was the auto-generated constraint name: if `DROP CONSTRAINT IF EXISTS` had named it wrong it would silently no-op, `ADD CONSTRAINT` would succeed, and *both* constraints would remain — the old one still rejecting `'scheduled'` while the migration reported success. Confirmed: exactly one constraint afterwards, `'scheduled'` accepted, `'saved'`/`'searched'` still accepted, garbage still rejected, re-running is a no-op.
- **Every test was checked by mutation** — dropping the `deleted = false` filter, inverting never-attempted ordering, removing the cooldown filter, reverting the trigger allowlist, removing the per-artist `try/catch`, and turning a failed read into an empty result each fail the test that claims to protect it.

## Deploy note

Needs a GitHub repo secret: **`INTERNAL_FUNCTION_SECRET`**, matching the Netlify value. Without it the workflow fails loudly on the first run rather than reporting a fake success.

### DIFF
--- a/.github/workflows/recatalog-sweep.yml
+++ b/.github/workflows/recatalog-sweep.yml
@@ -1,0 +1,66 @@
+name: Release catalogue sweep
+
+# Keeps saved artists' release catalogues from going stale.
+#
+# Every other catalog trigger is demand-driven: a save, a search, a button. So an artist who is
+# saved but never searched gets catalogued once and never again, and their release alerts
+# quietly stop — the long tail this feature exists for. See api/functions/recatalog-sweep.ts.
+#
+# A GitHub Actions cron rather than a scheduled Netlify function: there are no scheduled Netlify
+# functions in this repo, and this is the established precedent (schedule-social-posts.yml,
+# upstash-keepalive.yml).
+#
+# ## Cadence: daily, 05:00 UTC
+#
+# What this actually costs the sites we crawl is one batch of at most 25 artists, run through
+# the existing background function, which paces itself: one second between artists, at most 100
+# Bandcamp release-page fetches per run on top of one grid page each. So a sweep is on the order
+# of 125 requests to Bandcamp, sequential, spread over minutes — a trickle, and less than a
+# handful of real visitors searching. 05:00 UTC keeps it off both US and European peak.
+#
+# Daily rather than weekly because the batch is small and the 7-day cooldown is what actually
+# bounds per-artist crawling: an artist can't be re-crawled more than once a week however often
+# this runs. Daily just means the 25 slots are spread across seven small runs instead of arriving
+# in one lump, and it caps how long a newly-saved artist waits for a first catalogue if the
+# save-time request was dropped by the hourly cap.
+#
+# Safe to over-run. A duplicate invocation finds the same artists claimed and does nothing, so
+# re-running by hand is free.
+
+on:
+  schedule:
+    - cron: '0 5 * * *' # Daily, 05:00 UTC
+  workflow_dispatch: # Manual trigger from the GitHub UI
+
+jobs:
+  sweep:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Request a re-catalogue of the stalest saved artists
+        env:
+          INTERNAL_FUNCTION_SECRET: ${{ secrets.INTERNAL_FUNCTION_SECRET }}
+        run: |
+          if [ -z "$INTERNAL_FUNCTION_SECRET" ]; then
+            echo "::error::INTERNAL_FUNCTION_SECRET secret is not set"
+            exit 1
+          fi
+
+          response=$(curl -sS -X POST -w '\n%{http_code}' \
+            -H "Authorization: Bearer $INTERNAL_FUNCTION_SECRET" \
+            https://unstream.stream/.netlify/functions/recatalog-sweep)
+
+          status=$(echo "$response" | tail -n1)
+          body=$(echo "$response" | sed '$d')
+
+          # Any non-200 fails the run on purpose. A sweep that reports success while refusing to
+          # do anything — cataloging disabled, database unreadable — is the exact silent failure
+          # this workflow exists to prevent.
+          if [ "$status" != "200" ]; then
+            echo "::error::Sweep failed with HTTP $status: $body"
+            exit 1
+          fi
+
+          # Worth reading rather than skipping past: `requested` falling to 0 every day while
+          # `savedArtists` is large means everything is inside its cooldown (fine) or that
+          # selection has broken (not fine, and it looks identical from here).
+          echo "Sweep OK: $body"
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -208,9 +208,25 @@ page from Bandcamp on demand. Only the two database reads are stubbed; there is 
 connection, so nothing can be written. One Bandcamp request per release page you open.
 
 Once ingest is live, `release_catalog_state` is the observability surface:
-`last_attempted_at`, `releases_found`, `last_error`, `consecutive_failures`. A run that suddenly
-finds 0 releases where it previously found 20 is a parser break or a bot challenge, not an
-artist deleting their catalog.
+`last_attempted_at`, `releases_found`, `last_error`, `consecutive_failures`, `last_trigger`. A run
+that suddenly finds 0 releases where it previously found 20 is a parser break or a bot challenge,
+not an artist deleting their catalog — `recordCatalogOutcome` reports exactly that transition to
+Sentry, since it is otherwise recorded as a perfectly ordinary success.
+
+### Keeping catalogues fresh
+
+Every other catalog trigger is demand-driven — a save, a search, an artist's own button, the admin
+command — and `check-releases` only *reads* the catalogue. So without a scheduled refresh an artist
+who is saved but never searched gets catalogued once and their release alerts quietly stop. That
+is what `api/functions/recatalog-sweep.ts` fixes, run daily by `.github/workflows/recatalog-sweep.yml`.
+
+The sweep selects the stalest catalogues **among artists somebody has saved** (`getStaleSavedArtistCatalogs`
+in `db.ts`: never-attempted first, then oldest `last_attempted_at`, savers only as a tiebreak) and
+asks `requestArtistCatalog` for up to 25 of them under the `scheduled` trigger. It adds no rate
+limiting of its own — the 7-day cooldown and the hourly cap in `claimArtistForCatalog` still decide
+— so running it twice is a no-op. It returns a real summary rather than 202, and any refusal is a
+non-2xx that fails the workflow: a scheduled job that reports success while doing nothing is the
+same silent failure it was built to fix.
 
 ### Platform registry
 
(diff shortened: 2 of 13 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Staff-eng security/regression pass on this PR (recatalog sweep). Scoped to the diff only, no speculation beyond what the code shows.

## Security

**Auth/authorization** — needs-manual-verification (operationally sound, unverifiable from code)
`recatalog-sweep.ts` and the refactored `catalog-artist-background.ts` both gate on the new `isInternalRequest()` in `middleware.ts`, a straight lift of the existing `timingSafeEqual` Bearer-token check (length-checked before compare, fails closed when `INTERNAL_FUNCTION_SECRET` is unset). Consolidating the duplicated copy is a net security improvement — one auth path to audit instead of two near-identical ones. The one thing I can't verify from the diff is that the `INTERNAL_FUNCTION_SECRET` GitHub Actions secret is actually set to match the Netlify value, as the PR description calls out — if it isn't, the workflow fails loudly on first run rather than silently, which is the right failure mode either way.

**Secret handling** — confirmed clean
The workflow reads `INTERNAL_FUNCTION_SECRET` from `secrets.*` into an env var and never echoes it; only the JSON response body (which we control) is echoed. No secret leakage into logs.

**Injection risks** — confirmed clean
Bash in the new workflow quotes `$INTERNAL_FUNCTION_SECRET` properly and never interpolates response content into a command. `getStaleSavedArtistCatalogs` uses Supabase client query builders throughout (no raw SQL), and the new migration's `DROP CONSTRAINT IF EXISTS` / `ADD CONSTRAINT` is static SQL with no interpolated input.

**Unsafe defaults** — confirmed clean
`isCatalogEnabled()` fails closed (must be explicitly `'true'`), `isInternalRequest()` fails closed on a missing secret, and the new `scheduled` trigger is added to an allowlist (`=== 'saved' || === 'scheduled' ? ... : 'searched'`) rather than a cast — an unrecognized trigger can't slip past `CATALOG_HOURLY_CAP` as an undefined key.

**Data exposure / SSRF** — confirmed clean
The only outbound fetch this PR touches (`requestArtistCatalog` → `catalog-artist-background`) targets `process.env.URL`, the deploy's own site — not a user-controlled or externally-sourced URL, so the SSRF allowlist in `middleware.ts` doesn't apply here and doesn't need to. `recatalog-sweep.ts` reads only `artist_id`/timestamps/counts out of Supabase and returns aggregate counts in its response body, nothing PII-bearing.

## Regression

**Changed assumptions** — confirmed, and it's a real fix, not a leftover bug
`catalog-artist-background.ts`'s trigger parse changed from `=== 'saved' ? 'saved' : 'searched'` to an explicit allowlist including `'scheduled'`. Without this the sweep's requests would have silently been counted against the `searched` hourly budget (60/hr) instead of the new `scheduled` budget (120/hr) — exactly the kind of budget bug that's hard to notice because it fails as "some sweeps get dropped," not as an error.

**Edge cases / null handling** — confirmed clean
`getStaleSavedArtistCatalogs`'s sort comparator explicitly special-cases `lastAttemptedAt === null` (never-attempted-first) before falling into numeric comparison, and compares attempt timestamps as parsed instants rather than strings, called out for PostgREST's variable offset notation — a string comparison there would have been a subtle, hard-to-spot ordering bug. `savers.size === 0` short-circuits to an empty result rather than falling through to an empty `.in()` filter.

**API contract drift** — confirmed clean
`requestArtistCatalog`'s return type changed `Promise<void>` → `Promise<boolean>`. Checked both other call sites (`db.ts:3075`, `saved-artists.ts:352`) — both just `await` it and discard the value, so this is a non-breaking widening. The one test asserting the old `resolves.toBeUndefined()` was updated to `resolves.toBe(false)` in the same diff.

**State transitions** — confirmed clean
The sweep's own cooldown pre-filter (`getStaleSavedArtistCatalogs`) and `claimArtistForCatalog`'s authoritative cooldown check both key off `last_catalogued_at` against the same `RECATALOG_COOLDOWN_HOURS` constant (now exported instead of duplicated) — no drift between the two checks that could either double-count a cooldown or let it slip.

**Off-by-one / bounds** — confirmed clean
`SWEEP_BATCH_SIZE = 25` in `recatalog-sweep.ts` matches `MAX_ARTISTS_PER_RUN = 25` in `catalog-artist-background.ts` (verified against the actual constant, not just the comment claiming it), so one sweep is exactly one background invocation with no silently-truncated remainder.

## Summary

No confirmed or plausible bugs found. This is a tightly-scoped, well-tested change (mutation-checked tests per the PR description, migration validated against a real Postgres instance) that fixes a real latent bug (the trigger-parse budget leak) as part of adding the feature. Only open item is the operational one above — confirming the `INTERNAL_FUNCTION_SECRET` repo secret is set — which fails safely either way.

---

brandonlucasgreen (commented):

**Security**
- Authz: `isInternalRequest()` (`middleware.ts`) fails closed when `INTERNAL_FUNCTION_SECRET` is unset, and compares with `timingSafeEqual` after a length check (needed because `timingSafeEqual` throws on length mismatch — length itself isn't secret, so this isn't a timing leak). Applied identically to both `recatalog-sweep` and `catalog-artist-background` now that the duplicate inline check was consolidated. **confirmed** correct, matches the existing pattern for `resolve-url`/`search-sources`.
- `recatalog-sweep`'s new endpoint takes no attacker-controlled input beyond the bearer token — it only reads artist ids already in the DB and dispatches to the (also internally-authenticated) `catalog-artist-background`. **confirmed** no new attack surface.
- The GitHub Actions workflow reads the secret from `secrets.INTERNAL_FUNCTION_SECRET`, sends it only as a bearer header over HTTPS to `unstream.stream`, and doesn't echo it into logs. **confirmed**.

**Regression / edge cases**
- Cooldown consistency: `getStaleSavedArtistCatalogs`'s pre-filter (`last_catalogued_at > now - 7d`) and `claimArtistForCatalog`'s own cooldown check use the same `RECATALOG_COOLDOWN_HOURS` constant and the same comparison direction — I traced both and they agree, so the sweep can't hand `claimArtistForCatalog` a batch that gets mass-rejected a moment later. **confirmed** correct.
- Trigger handling: extending the allowlist from `'saved' | 'searched'` to include `'scheduled'` is a straightforward, safe addition — the prior code already defaulted any non-`'saved'` value to `'searched'` rather than casting a raw string, so this isn't closing a pre-existing gap, just adding a third recognized value the same way. **confirmed** no regression.
- Plausible, minor: `getStaleSavedArtistCatalogs`'s `saved_artists` read uses `.limit(10_000)` with no `ORDER BY`. Postgres doesn't guarantee which rows a `LIMIT` without an order returns once the table exceeds that, so at scale the sweep could non-deterministically miss artists near the boundary between runs. The code comments call this "a guard, not a limit," which is fine today — flagging so it isn't forgotten if the saved-artist population grows past 10k.
- Test coverage is genuinely thorough, not just claimed: I read through `recatalog-sweep-selection.test.ts` and `recatalog-sweep.test.ts` and confirmed cases for tombstoned saves, the exactly-outside-cooldown boundary, chunked `in()` lookups, savers-as-tiebreak-only (never overriding staleness), a failed read reported rather than treated as "nothing to do," and one artist throwing mid-batch not taking down the rest. **confirmed**.

Nothing here blocks merging — the `.limit(10_000)`/no-`ORDER BY` item is worth a follow-up ticket rather than a change to this PR.

---

PR #405, opened 2026-08-02
Add Mirlo as a release source via the Canimus federation catalog

brandonlucasgreen:
Adds Mirlo — the last source in the locked V1 Releases scope — through the **Fairplayer/Canimus federation catalog** rather than Mirlo's REST API.

**Merged only once Mirlo confirms the robots.txt question below.** The code is gated off by default, so merging it changes nothing until a flag is set.

## What the federation angle actually changes

The hope was that Canimus would federate across Fairplayer-family platforms and widen coverage. It doesn't, and that's the headline finding:

- `/v1/sm/canimus.json` is **per-host**. `{host}` is `mirlo.space`. Canimus is a *protocol* other platforms could adopt (Mirlo's docs name Faircamp as a hope), not a cross-platform index that exists today.
- It carries **only artists who individually opted in** to federated syndication — so it is **narrower** than the REST API, not wider.

What it does change is the cost shape, and that's a genuine win:

| | REST `/v1/artists/{slug}` | Canimus `/v1/sm/canimus.json` |
|---|---|---|
| Requests | one **per artist, per run** | one **per six hours, for everyone** |
| Coverage | every public Mirlo artist | opted-in artists only |
| Price / currency / pre-order | yes | **no** |
| Artist's declared off-platform links | no | yes |
| Opt-out signal | no | yes (`deleted[]`) |
| Sanctioned for aggregators | not stated | **documented for exactly this** |

So the trade is: give up prices and some coverage, get a per-artist cost of zero and an endpoint Mirlo published specifically so aggregators could use it.

**The missing price matters for the product.** A Mirlo release lands with a source and no offer, so the payout comparison — the thesis — can't render for it. No zero is invented: a fabricated "name your price" would misstate an artist's own terms. The honest fix is Mirlo's REST API under an agreed key.

## Design

- **One shared fetch.** Memoized per invocation, cached in Redis for 6h. The Mirlo pass is the only source with no budget parameter, because an extra artist costs Mirlo nothing.
- **Never caches uncertainty.** A refused fetch, a non-200, a non-JSON body, or a 200 that isn't a Canimus root all return `null` and are never cached. `ingestCanimusCatalog` returns `null` rather than an empty catalog for a non-root document, so a challenge page can't read as "nobody opted in".
- **Every URL host-pinned.** A federation format exists to name other hosts, and `release_sources.url` is a link fans click. Off-host artist and release URLs are dropped, not trusted.
- **Dedup unchanged, under-merge preserved.** Matches on `match_key` alone (Canimus types everything `"album"`, so the stored type is title-inferred and too weak to partition identity by). Exact key merges; `isFuzzyReleaseMatch` queues for review; nothing new auto-merges.
- **Live data guarded.** Mirlo carries a release dated **2925-11-02**; `parseReleaseDate` bounds it out. Empty-title drafts are dropped rather than stored as untitled.

## Typecheck coverage

`catalog-artist-background.ts` joins `api/tsconfig.json`'s include, bringing `release-ingest.ts`, `release-utils.ts` and the catalog write path under strict mode for the first time. It was already clean — verified the include has teeth by breaking a signature and watching six errors appear.

## The overlap measurement (§9's open question)

`scripts/measure-mirlo-overlap.ts` answers *"real Mirlo↔Bandcamp overlap rate — every dedup rule depends on it and I'm guessing."* It compares the catalog against releases **already in our database**, so it costs one request to Mirlo and none to Bandcamp, and it measures the production path rather than an abstract resemblance. Gated behind the same flag.

**Sample size is much better than the spec assumed.** The "Mirlo on 0 of 791 artists" figure came from the static SEO set. Live `artist_links` has **66 Mirlo links, 64 of which are on artists who also have a Bandcamp link** — a real sample, not a handful.

## Verification

- `npm run build` ✅ · `npm run test:api` ✅ 783 passing (20 new) · `npm run test:unit` ✅ · typecheck ✅
- Lint unchanged from baseline (71 problems / 64 errors both with and without this diff)
- Host-pinning tests proven to have teeth: disabling the pin fails exactly the two tests that assert it

**Not verified against a live response.** No request has been made to Mirlo. Fixtures are built from the documented shape, so these tests pin *our* decisions, not Mirlo's field names.

### DIFF
--- a/api/functions/db.ts
+++ b/api/functions/db.ts
@@ -994,6 +994,7 @@ export interface ArtistForCatalog {
   discogsUrl: string | null;
   faircampUrl: string | null;
   jamcoopUrl: string | null;
+  mirloUrl: string | null;
   officialSiteUrl: string | null;
 }
 
@@ -1015,7 +1016,7 @@ export async function getArtistForCatalog(artistId: string): Promise<ArtistForCa
         .from('artist_links')
         .select('platform, url')
         .eq('artist_id', artistId)
-        .in('platform', ['bandcamp', 'discogs', 'faircamp', 'jamcoop', 'officialsite']),
+        .in('platform', ['bandcamp', 'discogs', 'faircamp', 'jamcoop', 'mirlo', 'officialsite']),
     ]);
 
     if (artistError || !artistRow) {
@@ -1031,6 +1032,7 @@ export async function getArtistForCatalog(artistId: string): Promise<ArtistForCa
       discogsUrl: links.find(l => l.platform === 'discogs')?.url ?? null,
       faircampUrl: links.find(l => l.platform === 'faircamp')?.url ?? null,
       jamcoopUrl: links.find(l => l.platform === 'jamcoop')?.url ?? null,
+      mirloUrl: links.find(l => l.platform === 'mirlo')?.url ?? null,
       officialSiteUrl: links.find(l => l.platform === 'officialsite')?.url ?? null,
     };
   } catch (error) {
@@ -1657,6 +1659,182 @@ export async function persistJamcoopReleases(
     return written;
   } catch (error) {
     console.error('[DB] persistJamcoopReleases error:', error);
+    return [];
+  }
+}
+
+/**
+ * Write a Mirlo catalog pass, from the Canimus federation document.
+ *
+ * Matches on `match_key` alone, for the same reason `persistJamcoopReleases` and
+ * `persistFaircampReleases` do: Canimus types every release `"album"`, so the type we store is
+ * inferred from the title and is far too weak to partition identity by. Partitioning on it
+ * would mint a duplicate row every time Bandcamp had already typed the same record correctly —
+ * which, on the artists most likely to be on both, is most of them.
+ *
+ * Dates behave like the Jam.coop writer: Canimus publishes one, so this fills
+ * `release_date`/`date_precision` on a row that lacks them, under the same never-overwrite
+ * rules — a stored date wins over a new one, and a date a verified artist curated is never
+ * touched at all.
+ *
+ * No offers are written. The Canimus shape carries no price (see the Mirlo section of
+ * release-ingest), and a source row with no offer is an honest "we know it's here, we don't
+ * know what it costs" — whereas inventing a zero would read as "name your price" and misstate
+ * an artist's own terms.
+ */
+interface MirloReleaseToPersistRow {
+  title: string;
+  slug: string;
+  matchKey: string;
+  releaseType: string;
+  releaseDate: string | null;
+  datePrecision: string;
+  status: string;
+  artworkUrl: string | null;
+  externalUrl: string;
+}
+
+export async function persistMirloReleases(
+  artistId: string,
+  releases: MirloReleaseToPersistRow[]
+): Promise<PersistedRelease[]> {
+  const client = getClient();
+  if (!client || releases.length === 0) return [];
+
+  try {
+    const { data: existingRows, error: readError } = await client
+      .from('releases')
+      .select('id, slug, match_key, release_type, release_date, artwork_url, curated_fields')
+      .eq('artist_id', artistId);
+
+    if (readError) {
+      console.error('[DB] persistMirloReleases read failed:', readError.message);
+      return [];
+    }
+
+    type ExistingRow = {
+      id: string;
+      slug: string;
+      match_key: string;
+      release_type: string;
+      release_date: string | null;
+      artwork_url: string | null;
+      curated_fields: string[] | null;
+    };
+    const existing = (existingRows as ExistingRow[]) || [];
+    const byMatchKey = new Map(existing.map(r => [r.match_key, r]));
+    const takenSlugs = new Set(existing.map(r => r.slug));
+    const claimedKeys = await getClaimedSourceKeys(client, existing.map(r => r.id));
+
+    const written: PersistedRelease[] = [];
+
+    for (const release of releases) {
+      const prior = byMatchKey.get(release.matchKey);
+
+      let releaseId: string;
+      let curatedFields: string[];
+
+      if (prior) {
+        const curated = new Set(prior.curated_fields ?? []);
+        curatedFields = [...curated];
+        const patch: Record<string, unknown> = {};
+
+        if (!curated.has('artwork_url') && release.artworkUrl && !prior.artwork_url) {
+          patch.artwork_url = release.artworkUrl;
+        }
+        if (!curated.has('release_date') && release.releaseDate && !prior.release_date) {
+          patch.release_date = release.releaseDate;
+          patch.date_precision = release.datePrecision;
+          // Status moves with the date, but only when this pass is the one supplying it.
+          patch.status = release.status;
+        }
+
+        if (Object.keys(patch).length > 0) {
+          const { error } = await client.from('releases').update(patch).eq('id', prior.id);
+          if (error) {
+            console.error('[DB] persistMirloReleases update failed:', error.message);
+            continue;
+          }
+        }
+        releaseId = prior.id;
+      } else {
+        const fuzzy = existing.find(c => isFuzzyReleaseMatch(c.match_key, release.matchKey));
+
+        let slug = release.slug;
+        if (takenSlugs.has(slug)) slug = `${slug}-${release.matchKey.slice(0, 6)}`;
+        takenSlugs.add(slug);
+
+        const { data: inserted, error } = await client
+          .from('releases')
+          .insert({
+            artist_id: artistId,
+            title: release.title,
+            slug,
+            match_key: release.matchKey,
+            release_type: release.releaseType,
+            release_date: release.releaseDate,
+            date_precision: release.datePrecision,
+            status: release.status,
+            artwork_url: release.artworkUrl,
+            source: 'auto',
+            ...(fuzzy && { needs_review: true, flagged_against_release_id: fuzzy.id }),
+          })
+          .select('id')
+          .single();
+
+        if (error || !inserted) {
+          console.error('[DB] persistMirloReleases insert failed:', error?.message);
+          continue;
+        }
+        releaseId = (inserted as { id: string }).id;
+        curatedFields = [];
+
+        const createdRow: ExistingRow = {
+          id: releaseId,
+          slug,
+          match_key: release.matchKey,
+          release_type: release.releaseType,
+          release_date: release.releaseDate,
+          artwork_url: release.artworkUrl,
+          curated_fields: [],
+        };
+        byMatchKey.set(release.matchKey, createdRow);
+        existing.push(createdRow);
+
+        if (fuzzy) {
+          const { error: flagError } = await client
+            .from('releases')
+            .update({ needs_review: true, flagged_against_release_id: releaseId })
+            .eq('id', fuzzy.id);
+          if (flagError) console.error('[DB] persistMirloReleases fuzzy-flag failed:', flagError.message);
+        }
+      }
+
+      const source = await upsertReleaseSource(
+        client,
+        releaseId,
+        'mirlo',
+        release.externalUrl,
+        release.externalUrl,
+        claimedKeys
+      );
+      if (!source) {
+        console.error('[DB] persistMirloReleases source write failed for release', releaseId);
+        continue;
+      }
+
+      written.push({
+        releaseId,
+        sourceId: source.id,
+        url: source.url,
+        detailCheckedAt: source.detail_checked_at,
+        curatedFields,
+      });
+    }
+
+    return written;
+  } catch (error) {
+    console.error('[DB] persistMirloReleases error:', error);
     return [];
   }
 }
--- a/api/functions/__tests__/mirlo-canimus.test.ts
+++ b/api/functions/__tests__/mirlo-canimus.test.ts
@@ -1,0 +1,293 @@
+// Parsing Mirlo's Fairplayer/Canimus federation catalog into release rows.
+//
+// The fixtures here are built from the documented response shape in Mirlo's API docs, not
+// captured from a live response — Mirlo's robots.txt disallows `/v1/`, which is where the
+// federation endpoint lives, so no request has been made pending Mirlo's own answer on that.
+// What that means for these tests: they pin *our* decisions (host pinning, empty-title drafts,
+// the 2925 date, opted-out artists) rather than certifying Mirlo's field names. The first live
+// response is what confirms the shape, and `ingestCanimusCatalog` returning null on anything
+// that isn't a Canimus root is what makes a shape mismatch loud instead of silent.
+
+import { describe, it, expect } from 'vitest';
+import {
+  ingestCanimusCatalog,
+  buildMirloReleases,
+  mirloArtistSlug,
+  type CanimusArtist,
+} from '../release-ingest';
+
+const CATALOG_URL = 'https://mirlo.space/v1/sm/canimus.json';
+const NOW = new Date('2026-08-02T00:00:00Z');
+
+function album(overrides: Record<string, unknown> = {}) {
+  return {
+    type: 'album',
+    name: 'Garden Invicta',
+    url: 'https://mirlo.space/m-corp/release/garden-invicta',
+    release_date: '2024-01-15T00:00:00.000Z',
+    license: 'CC BY-SA 4.0',
+    artist: 'M Corp',
+    images: { cover: { src: 'https://mirlo.space/img/cover.jpg', alt: null, width: 600, height: 600 } },
+    description: 'About this album',
+    children: [],
+    ...overrides,
+  };
+}
+
+function root(children: unknown[], deleted: unknown[] = []) {
+  return { type: 'root', url: 'https://mirlo.space', children, deleted };
+}
+
+function artistNode(overrides: Record<string, unknown> = {}) {
+  return {
+    type: 'artist',
+    name: 'M Corp',
+    url: 'https://mirlo.space/m-corp',
+    images: { cover: { src: 'https://mirlo.space/img/artist.jpg' } },
+    summary: 'Short description',
+    description: 'Full bio',
+    links: [{ name: 'Bandcamp', href: 'https://mcorp.bandcamp.com', type: 'bandcamp' }],
+    children: [album()],
+    ...overrides,
+  };
+}
+
+describe('ingestCanimusCatalog', () => {
+  it('parses artists and their releases from a root document', () => {
+    const catalog = ingestCanimusCatalog(root([artistNode()]), CATALOG_URL, NOW);
+
+    expect(catalog).not.toBeNull();
+    expect(catalog!.artists).toHaveLength(1);
+
+    const artist = catalog!.artists[0];
+    expect(artist.name).toBe('M Corp');
+    expect(artist.slug).toBe('m-corp');
+    expect(artist.releases).toHaveLength(1);
+    expect(artist.releases[0]).toMatchObject({
+      title: 'Garden Invicta',
+      url: 'https://mirlo.space/m-corp/release/garden-invicta',
+      releaseDate: '2024-01-15',
+      datePrecision: 'day',
+      status: 'released',
+      artworkUrl: 'https://mirlo.space/img/cover.jpg',
+    });
+  });
+
+  it('returns null — not an empty catalog — for a body that is not a Canimus root', () => {
+    // A login wall, an error envelope or a challenge page must be distinguishable from
+    // "nobody has opted in", or a bad response silently wipes every Mirlo artist.
+    expect(ingestCanimusCatalog({ error: 'unauthorized' }, CATALOG_URL, NOW)).toBeNull();
+    expect(ingestCanimusCatalog('<html>Just a moment…</html>', CATALOG_URL, NOW)).toBeNull();
+    expect(ingestCanimusCatalog(null, CATALOG_URL, NOW)).toBeNull();
+    expect(ingestCanimusCatalog(root([]), CATALOG_URL, NOW)).toEqual({ artists: [], deletedSlugs: [] });
+  });
+
+  it('drops a release whose url points at another host', () => {
+    // A federation document exists to name other hosts, and release_sources.url is rendered as
+    // a link a fan clicks. An off-host release URL is never attributed to the artist.
+    const catalog = ingestCanimusCatalog(
+      root([artistNode({ children: [album({ url: 'https://evil.example/m-corp/release/x' })] })]),
+      CATALOG_URL,
+      NOW
+    );
+
+    expect(catalog!.artists[0].releases).toHaveLength(0);
+  });
+
+  it('drops an artist whose url points at another host', () => {
+    const catalog = ingestCanimusCatalog(
+      root([artistNode({ url: 'https://evil.example/m-corp' })]),
+      CATALOG_URL,
+      NOW
+    );
+
+    expect(catalog!.artists).toHaveLength(0);
+  });
+
+  it('drops empty-title drafts rather than storing them as untitled', () => {
+    const catalog = ingestCanimusCatalog(
+      root([artistNode({ children: [album({ name: '   ' }), album({ name: 'Real Record' })] })]),
+      CATALOG_URL,
+      NOW
+    );
+
+    expect(catalog!.artists[0].releases.map(r => r.title)).toEqual(['Real Record']);
+  });
+
+  it('rejects the live 2925 date rather than sorting it to the top of every chronology', () => {
+    // Mirlo genuinely carries a release dated 2925-11-02. Unbounded it lands in every
+    // calendar subscriber's feed forever.
+    const catalog = ingestCanimusCatalog(
+      root([artistNode({ children: [album({ release_date: '2925-11-02T00:00:00.000Z' })] })]),
+      CATALOG_URL,
+      NOW
+    );
+
+    const release = catalog!.artists[0].releases[0];
+    expect(release.releaseDate).toBeNull();
+    expect(release.datePrecision).toBe('unknown');
+    // Undated reads as released: promising a fan something is "coming" when we don't know is
+    // the worse error.
+    expect(release.status).toBe('released');
+  });
+
+  it('marks a genuine future date as announced', () => {
+    const catalog = ingestCanimusCatalog(
+      root([artistNode({ children: [album({ release_date: '2027-09-07T00:00:00.000Z' })] })]),
+      CATALOG_URL,
+      NOW
+    );
+
+    expect(catalog!.artists[0].releases[0]).toMatchObject({
+      releaseDate: '2027-09-07',
+      status: 'announced',
+    });
+  });
+
+  it('refuses a non-http artwork src', () => {
+    const catalog = ingestCanimusCatalog(
+      root([artistNode({ children: [album({ images: { cover: { src: 'javascript:alert(1)' } } })] })]),
+      CATALOG_URL,
+      NOW
+    );
+
+    expect(catalog!.artists[0].releases[0].artworkUrl).toBeNull();
+  });
+
+  it('reads opted-out artists off the deleted array', () => {
+    const catalog = ingestCanimusCatalog(
+      root([artistNode()], [{ type: 'artist', name: 'Gone Artist', url: 'https://mirlo.space/gone-artist' }]),
+      CATALOG_URL,
+      NOW
+    );
+
+    expect(catalog!.deletedSlugs).toEqual(['gone-artist']);
+  });
+
+  it('keeps artist-declared off-host links, which are compared but never fetched', () => {
+    const catalog = ingestCanimusCatalog(root([artistNode()]), CATALOG_URL, NOW);
+
+    expect(catalog!.artists[0].links).toEqual([
+      { type: 'bandcamp', href: 'https://mcorp.bandcamp.com' },
+    ]);
+  });
+
+  it('ignores nodes that are not artists, and releases that are not albums', () => {
+    const catalog = ingestCanimusCatalog(
+      root([
+        { type: 'playlist', name: 'Staff picks', url: 'https://mirlo.space/playlists/1' },
+        artistNode({ children: [album(), { type: 'track', name: 'A Track', url: 'https://mirlo.space/x' }] }),
+      ]),
+      CATALOG_URL,
+      NOW
+    );
+
+    expect(catalog!.artists).toHaveLength(1);
+    expect(catalog!.artists[0].releases).toHaveLength(1);
+  });
+});
+
+describe('buildMirloReleases', () => {
+  const base: CanimusArtist = {
+    name: 'M Corp',
+    url: 'https://mirlo.space/m-corp',
+    slug: 'm-corp',
+    links: [],
+    releases: [],
+  };
+
+  it('maps a release to a persistable row', () => {
+    const [row] = buildMirloReleases({
+      ...base,
+      releases: [
+        {
+          title: 'Garden Invicta',
+          url: 'https://mirlo.space/m-corp/release/garden-invicta',
+          releaseDate: '2024-01-15',
+          datePrecision: 'day',
+          status: 'released',
+          artworkUrl: 'https://mirlo.space/img/cover.jpg',
+        },
+      ],
+    });
+
+    expect(row).toMatchObject({
+      title: 'Garden Invicta',
+      slug: 'garden-invicta',
+      matchKey: 'gardeninvicta',
+      releaseDate: '2024-01-15',
+      externalUrl: 'https://mirlo.space/m-corp/release/garden-invicta',
+    });
+  });
+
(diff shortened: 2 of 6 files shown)


brandonlucasgreen (commented):

**Security**
- SSRF: `fetchCanimusCatalog` (`catalog-artist-background.ts`) routes every request through `isUrlHostnameAllowed()` + `safeFetch()`, which re-validates every redirect hop and resolves DNS to reject private/metadata addresses. `mirlo.space` / `*.mirlo.space` were already on `ALLOWED_OUTBOUND_HOSTNAMES` on `main`. **confirmed** — correctly wired, no gap.
- Host pinning on ingested data: every URL pulled out of the Canimus document (artist URL, release URL, artwork `src`) is checked against the fetch host via `sameHostUrl`/`canimusArtistSlug` before being trusted; off-host artist/release entries are dropped rather than stored, and artwork `src` is restricted to `http(s)://` (blocks `javascript:`/`data:` into the ``). Covered directly by tests (`drops a release whose url points at another host`, `refuses a non-http artwork src`). **confirmed**.
- `mirloArtistSlug` (applied to artist-stored URLs, which a claimed profile controls) rejects non-Mirlo hosts and reserved path segments (`v1`, `api`, `login`, …), so a malicious or malformed stored link can't smuggle a foreign host or the `/v1/` API path itself into the join key. **confirmed**.
- No new externally-reachable endpoint: the Mirlo pass runs inside the existing internally-authenticated `catalog-artist-background` handler, so it adds no new authz surface. **confirmed**.
- Never-cache-uncertainty: `cacheGetOrFetch`'s `shouldCache` predicate refuses to cache `null` (fetch/parse failure), and `ingestCanimusCatalog` returns `null` — not an empty catalog — for anything that isn't a genuine Canimus root (login wall, challenge page, wrong shape), verified directly by test. **confirmed**.

**Regression / edge cases**
- Plausible, minor: `fetchCanimusCatalog`'s pagination loop stops after `CANIMUS_MAX_PAGES` (10 × 200 = 2000 artists) with no log line distinguishing "that was genuinely the last page" from "we hit the cap with more data behind it." Today's opt-in population is far under that, but if it ever grows past 2000, artists beyond the cutoff would silently drop out of every subsequent pass with nothing in the logs to explain why — worth a `console.warn` on the truncation case specifically, per the "no silent caps" principle in CLAUDE.md. Non-blocking.
- No behavior change to existing sources: `catalogArtist`'s no-links-stored guard is only widened (not altered) when `MIRLO_CANIMUS_ENABLED` is unset, which is the default, and the whole Mirlo pass is then a no-op. **confirmed** no regression to shipped Bandcamp/Discogs/Faircamp/Jam.coop behavior while the flag is off.

Nice work overall — host-pinning discipline, "return null not empty" cache semantics, and the 2925-date test are the kind of detail that actually prevents the bug class this codebase keeps hitting. Only the pagination-truncation logging gap above is worth a follow-up; nothing here should block merging.

---

brandonlucasgreen (commented):

- **Auth/authorization** — reuses the existing internal shared-secret check; no new auth surface. *confirmed*
- **Input validation** — `ingestCanimusCatalog`/`parseCanimusArtist`/`parseCanimusRelease` type-guard every field (`isRecord`, `type` checks) and drop empty-title drafts rather than storing "Untitled". *confirmed*
- **Secret handling** — no new secrets; `measure-mirlo-overlap.ts` reads existing `.env` vars and doesn't log them. *confirmed*
- **Injection** — all writes go through the Supabase client with typed columns, no string-built SQL. *confirmed*
- **SSRF** — every outbound request is checked with `isUrlHostnameAllowed()` before `safeFetch`. I verified `mirlo.space` / `*.mirlo.space` are already on `ALLOWED_OUTBOUND_HOSTNAMES` (middleware.ts:330-331), so this isn't a feature that silently no-ops for lack of an allowlist entry. *confirmed*
- **XSS via artwork** — `canimusCoverUrl` rejects anything not `http(s)://` before it's stored as `artwork_url` (blocks `javascript:` etc.), covered by a test. *confirmed*
- **Unsafe defaults** — the whole path is gated behind `MIRLO_CANIMUS_ENABLED`, default off, and both `ingestCanimusCatalog` and `fetchCanimusCatalog` fail closed (return `null`) instead of caching an empty/uncertain result as a real negative. *confirmed*
- **Cross-host attribution** — artist and release URLs are host-pinned to the fetched document's own host (`sameHostUrl`), with tests for off-host URLs being dropped rather than trusted. *confirmed*

**Regression**
- *plausible, needs-manual-verification* — `catalogArtist`'s "nothing to crawl" check now includes `mirloUrl` when the flag is on, but PR #406's new sweep pool (`CATALOGUEABLE_PLATFORMS = ['bandcamp','discogs','faircamp','jamcoop']`) doesn't include `mirlo`. Harmless today since the flag defaults off, but if both land and the flag is later flipped on, mirlo-only artists won't be picked up by the scheduled sweep even though `catalogArtist` would happily crawl them on demand. Worth a one-line addition to that list whenever the flag actually goes live.
- No other changed-assumption or edge-case regressions found — `persistMirloReleases` follows the same never-overwrite-curated-fields / fuzzy-match-flagging pattern as the existing Jam.coop and Faircamp writers, and slug/match-key collision handling is covered by tests.

Deploy preview is green. Nice discipline keeping this fully gated until Mirlo confirms the endpoint is fair game.

---

brandonlucasgreen (commented):

**Security**
- SSRF: every Canimus fetch is gated by `isUrlHostnameAllowed()` before `safeFetch`, and `mirlo.space` / `*.mirlo.space` are already on `ALLOWED_OUTBOUND_HOSTNAMES`. **confirmed**
- Data trust: artist and release URLs parsed out of the federation document are host-pinned via `sameHostUrl()` before being stored as `release_sources.url` (a link fans click) — an entry pointing at another host is dropped, not trusted. Artwork `src` is restricted to `http(s)://` only, blocking `javascript:`/`data:` injection into the rendered ``. **confirmed**
- Unsafe defaults: the whole feature is off by default (`MIRLO_CANIMUS_ENABLED === 'true'`), and the standalone `scripts/measure-mirlo-overlap.ts` independently refuses to run without the same flag. **confirmed**
- **needs-manual-verification:** the fetched endpoint, `/v1/sm/canimus.json`, sits under the `/v1/` path prefix that this PR's own comments say Mirlo's robots.txt disallows (`# Disallow crawling the API`). The reasoning that this specific document is a deliberate syndication export rather than API crawling is plausible, and gating all requests behind `MIRLO_CANIMUS_ENABLED` pending Mirlo's own confirmation is the right shape — but nothing in code enforces that the confirmation actually happened before someone flips the flag in production. Worth confirming there's a paper trail (issue/email from Mirlo) before that env var is ever set to `true` anywhere — this is the "respect other people's servers" principle from CLAUDE.md, and this codebase has had real outages from scraping disallowed paths before.
- Never-cache-uncertainty is respected: `cacheGetOrFetch`'s `shouldCache` predicate refuses to cache a `null` (failed/unparseable) catalog, and the per-invocation memo is explicitly reset (`__resetCanimusMemo`) so a warm Lambda can't serve a stale snapshot past the 6h Redis TTL. **confirmed**
- No secrets or auth changes in this PR; all Supabase writes are parameterized and scoped by `artist_id` like the existing Jam.coop/Faircamp writers. **confirmed**

**Regression**
- **plausible:** `CATALOGUEABLE_PLATFORMS` in `db.ts` (the sweep's pool query, `getStaleCatalogCandidates`, added in #407) was not updated to include `'mirlo'`, while this PR's `catalogArtist` eligibility check now treats a stored `mirloUrl` as sufficient reason to catalog. CLAUDE.md calls out, verbatim, that this list must stay identical to `catalogArtist`'s "nothing to fetch" check for exactly this reason. Net effect once `MIRLO_CANIMUS_ENABLED=true`: an artist whose *only* stored link is `mirlo` will never be selected by the 6-hourly sweep (silent under-coverage — the opposite direction from the failure-poisoning case the doc warns about, but still a gap) — their Mirlo releases would only ever get pulled in via a demand-driven trigger. Worth a one-line addition to `CATALOGUEABLE_PLATFORMS` before or alongside enabling the flag.
- Pagination is bounded (`CANIMUS_MAX_PAGES = 10`) with short/empty-page termination, so a host that ignores `skip` can't cause an unbounded loop. **confirmed**
- `persistMirloReleases` follows the existing Jam.coop/Faircamp writer pattern (match on `match_key`, curated-field guard, fuzzy-match flagging for review) — no divergence spotted. **confirmed**

No blocking findings; the robots.txt item above is the one worth a deliberate answer before flipping the flag, and the `CATALOGUEABLE_PLATFORMS` gap is a small follow-up.

---

brandonlucasgreen (commented):

**Security**
- auth/authorization: N/A, `catalog-artist-background.ts`'s `isAuthorized` internal-secret check is untouched.
- input validation / SSRF: confirmed solid. `fetchCanimusCatalog` checks `isUrlHostnameAllowed()` before every fetch (and `mirlo.space`/`*.mirlo.space` are already on `ALLOWED_OUTBOUND_HOSTNAMES`), then goes through `safeFetch`. Every URL pulled out of the response — artist URL, release URL, cover image — is re-validated: `sameHostUrl` pins to the exact host the catalog was fetched from and rejects non-http(s) schemes, and `canimusCoverUrl` separately rejects anything not starting `http(s)://` (blocks `javascript:`/`data:` in what becomes an ``). `mirloArtistSlug` (used to join a stored `mirlo` artist_links row back to the catalog) independently validates host + scheme + a reserved-segment blocklist, so a stored link on some other domain can't spoof a join. Confirmed by the new tests (`refuses a non-Mirlo host`, `drops a release whose url points at another host`, `refuses a non-http artwork src`).
- unsafe defaults: `isMirloCanimusEnabled()` defaults closed (`process.env.MIRLO_CANIMUS_ENABLED === 'true'`) — fails safe.
- secret handling: none introduced.
- injection: none — all Supabase calls are parameterized client calls, no raw SQL.
- data exposure: none new; Canimus is a public, opt-in catalog.

No security findings.

**Regression — one confirmed, in scope of the diff**

`api/functions/db.ts` around line 661:

```ts
const CATALOGUEABLE_PLATFORMS = ['bandcamp', 'discogs', 'faircamp', 'jamcoop'] as const;
```

used by `getStaleCatalogCandidates` (the pool for the 6-hourly recatalog-sweep), is **not** updated to include `'mirlo'`. But this PR *does* update the parallel list in `getArtistForCatalog` (adds `'mirlo'`) and `catalogArtist`'s "is there anything to fetch" check (adds `mirloUrl` to the OR, gated on the same flag) to treat a Mirlo link as catalogueable.

The comment sitting directly above that constant is explicit about why this has to stay in sync:

> "Keep this list identical to that condition or the sweep will spend its batch on artists with nothing to fetch and turn the failure counters into noise."

This PR breaks that invariant in the other direction: an artist whose *only* catalogueable link is `mirlo` will never appear in `getStaleCatalogCandidates`'s pool, so once `MIRLO_CANIMUS_ENABLED=true`, those artists will never be picked up by the periodic sweep — only by demand-driven triggers (search/save). That's the exact "saved but never searched → catalogue quietly goes stale" failure mode the "Keeping catalogues fresh" doc section was written to fix, reintroduced for one platform. It's dormant today only because the flag defaults off.

**Confirmed** (verified directly in the code, not speculative) — recommend adding `'mirlo'` to `CATALOGUEABLE_PLATFORMS` (or making it flag-conditional the same way `catalogArtist` is) before or alongside flipping `MIRLO_CANIMUS_ENABLED`.

Nothing else stood out — dedup/date-bounding/pagination edge cases (2925 date, empty-title drafts, non-root response, exact-page-size pagination) all have direct test coverage and looked correct on read-through.

---

brandonlucasgreen (commented):

Note: this branch is 5 commits behind current `main` — it predates the recatalog-sweep feature (#404/#406/#407/#408/#409). That gap is the source of the one high-severity regression finding below.

## Security

No confirmed issues. Specifically checked and clean:
- **Flag gating**: `isMirloCanimusEnabled()` uses strict `=== 'true'`, not a truthy check — no falsy-string bug.
- **Auth on the background function**: requires a valid `Bearer` token (`timingSafeEqual` against `INTERNAL_FUNCTION_SECRET`, fails closed if unset) *and* `RELEASE_CATALOG_ENABLED`, both ahead of the per-artist loop — Mirlo cataloging can't run unauthenticated or with cataloging off.
- **SSRF**: `mirlo.space`/`*.mirlo.space` already allowlisted on `main`; the Canimus fetch goes through `isUrlHostnameAllowed()` + `safeFetch()`.
- **Host pinning**: real, not just asserted — `sameHostUrl` rejects any artist/release/deleted-slug URL whose host doesn't match the fetched catalog's own host, and `canimusCoverUrl` additionally requires `^https?:\/\//i` on artwork src (rejects `javascript:`/`data:`). Both covered by tests.
- **Date bound**: `parseReleaseDate` bounds to `[1900, now+3y]` and correctly rejects the live 2925-11-02 Mirlo release while still accepting a genuine 2027 pre-order.
- **Never-cache-uncertainty**: correctly implemented — `shouldCache: catalog => catalog !== null`, and `fetchCanimusCatalog` returns `null` (not an empty catalog) on refused fetch, non-OK status, JSON parse failure, or "parsed but not a Canimus root" (the last case also logged).
- **Injection/XSS**: no raw SQL; release fields flow through the existing `escapeHtml`-based edge renderer, unmodified by this PR.

Two lower-severity items:
- **plausible** — Canimus's `deleted[]` array is parsed into `CanimusCatalog.deletedSlugs` but nothing ever reads it. An artist who opts into Mirlo federation, gets synced, then revokes consent simply disappears from `children` on the next poll — `catalogMirlo` no-ops rather than removing their earlier rows, so previously-synced Mirlo releases/links stay live indefinitely after an opt-out. Sits awkwardly against the PR's own "consent is per-artist and explicit" rationale.
- **confirmed** (minor/cosmetic) — `catalog-artist-background.ts:364-370`: with the flag off and an artist whose only stored link is Mirlo, the "nothing to catalog" error reads "no bandcamp, discogs, faircamp, jam.coop, or mirlo link stored" even though the mirlo link *is* stored — it's the flag that's off, not the link that's missing. Misleading for debugging.

## Regression

1. **confirmed** — `api/functions/db.ts`: this branch doesn't yet have `CATALOGUEABLE_PLATFORMS`/`getStaleCatalogCandidates` (added to `main` after this branch diverged), so the PR has no corresponding change adding `mirlo` to that list, and there's no `catalog-artist-links.test.ts` on this branch to guard the sync. Per this repo's own documented pattern, that platform list must stay identical to `catalogArtist`'s "no link stored" check, which this PR *does* extend with `mirlo`. On merge/rebase onto current `main`, if `mirlo` isn't manually added to `CATALOGUEABLE_PLATFORMS`, any artist whose only catalog-eligible link is Mirlo will catalog fine on save/search but never appear in the 6-hourly sweep's pool — reproducing the exact "saved-but-never-searched goes stale" bug the sweep exists to fix, scoped to Mirlo. This needs to be handled explicitly during the rebase, not assumed to carry over.

No findings in API contract drift (new fields are additive/nullable), partial-write atomicity (matches the existing per-source pattern, `recordCatalogOutcome` fed the real written count), or dedup/auto-merge correctness (exact `match_key` only, `isFuzzyReleaseMatch` flags for review but never auto-merges, same as every other source).

**needs-manual-verification**: no integration-level test exercises `catalog-artist-background.ts`'s orchestration itself (the flag gate's effect on `catalogArtist`, or `loadCanimusCatalog`'s memoization/never-cache-on-failure behavior) — only the pure parsing functions in `release-ingest.ts` are tested. Worth a look given how central "never cache a failed fetch" is to this PR's design.

---

brandonlucasgreen (commented):

## Regression

**1. `catalogMirlo` can't distinguish a real Mirlo outage from a legitimate "not opted in"** — `confirmed`
`api/functions/catalog-artist-background.ts`, new `catalogMirlo` (~L207-232) and `loadCanimusCatalog` (~L121-135): `const catalog = await loadCanimusCatalog(); if (!catalog) return 0;` treats a genuine fetch failure (timeout, non-200, bad body) identically to "this artist isn't in the federation catalog" — both return `0`, neither throws. `catalogArtist`'s outcome logic only flags a failure when `bandcampError && totalFound === 0` (unchanged), so Mirlo failures never bump `consecutive_failures` or set `last_error` — they read as an ordinary zero-release success. This is exactly the "never cache uncertainty" failure mode from CLAUDE.md, one layer below the Redis cache (which itself is handled correctly — see below). `catalogBandcamp`'s docstring explicitly calls out this distinction as load-bearing; `catalogMirlo` doesn't follow it. Note: Discogs/Faircamp/Jamcoop already share this gap, so this extends a pre-existing convention rather than introducing a new one — but worth closing here since Mirlo may become some artists' only catalogable source.

**2. `CATALOGUEABLE_PLATFORMS` in `db.ts` not updated for Mirlo** — `confirmed`
`api/functions/db.ts:685` (`['bandcamp', 'discogs', 'faircamp', 'jamcoop']`) vs. `catalog-artist-background.ts:~83`, which now also checks `mirloUrl` before deciding there's something to catalog. CLAUDE.md is explicit these two lists must stay identical. Confirmed by grep: `recatalog-sweep.ts` and its tests have no mention of `mirlo`. Once `MIRLO_CANIMUS_ENABLED` flips on, an artist whose only link is Mirlo will never enter `getStaleCatalogCandidates`' sweep pool — their catalog only refreshes if someone searches/saves them again, the exact silent-staleness gap the sweep exists to close. Currently inert behind the flag, but ships latent.

**3. `deletedSlugs` (Mirlo federation opt-outs) parsed but never consumed** — `confirmed`, low severity while flag is off
`release-ingest.ts` builds `deletedSlugs`; `catalog-artist-background.ts` merges them into the catalog (~L188) but nothing reads them afterward. If an artist opts out of federation after having Mirlo releases ingested, their stale `release_sources` rows for `platform: 'mirlo'` are never cleaned up or flagged. The type/parsing exists specifically for this, so it looks like an intended step that didn't get wired up — worth closing before the flag flips on.

## Security

**4. No `failureTtlSeconds` on the Canimus Redis cache** — `plausible` (respect-third-party-server risk)
`loadCanimusCatalog`: `cacheGetOrFetch('canimus:mirlo:v1', fetchCanimusCatalog, CANIMUS_CACHE_TTL_SECONDS, catalog => catalog !== null)` omits the 5th arg. Per `cache.ts`'s contract, that means a failed fetch isn't cached at all — every invocation during a Mirlo outage re-attempts the request, and this function runs on ordinary demand-driven triggers (search/save), not just the sweep. Compare `search-sources.ts:384-390` and `search-musicbrainz.ts:439-446`, which both pass a short failure TTL for this exact reason. Needs manual verification of real production invocation volume to size the actual impact, but the missing argument is clear from the diff.

**5. Robots.txt-gated launch relies solely on an env var, no enforced check** — `needs-manual-verification`
PR description says this merges pending Mirlo's robots.txt confirmation, gated by `MIRLO_CANIMUS_ENABLED`. That's reasonable, but nothing in code stops the flag from being flipped in Netlify before that confirmation lands — it's a boolean plus a comment, not a technical safeguard. Not a code bug, but worth a belt-and-suspenders check given this repo's history of self-inflicted scraping outages.

**Checked, no finding:**
- `mirlo.space` / `*.mirlo.space` already in `ALLOWED_OUTBOUND_HOSTNAMES` (middleware.ts:330-331), pre-existing and unmodified — no SSRF allowlist gap.
- All Canimus-returned URLs are host-pinned (`sameHostUrl`) before use/storage; artwork `src` restricted to `http(s)` only — no `javascript:`/`data:` injection surface.
- No secrets introduced; Canimus catalog is unauthenticated.
- No changes to `/api/search/sources` response shape or `search-utils.ts` — no contract drift for web/extension/Apple clients.
- No dev-shim drift: this feature is entirely in `api/functions/` (the production path), nothing needed in `apps/web/server/`.

---

api/functions/db.ts
```diff
@@ -1015,7 +1016,7 @@ export async function getArtistForCatalog(artistId: string): Promise<ArtistForCa
         .from('artist_links')
         .select('platform, url')
         .eq('artist_id', artistId)
-        .in('platform', ['bandcamp', 'discogs', 'faircamp', 'jamcoop', 'officialsite']),
+        .in('platform', ['bandcamp', 'discogs', 'faircamp', 'jamcoop', 'mirlo', 'officialsite']),
```
brandonlucasgreen, line 1019: **Regression — confirmed by reading the code, not speculation.**

This adds `mirlo` here (and `catalogArtist` in `catalog-artist-background.ts` correspondingly gates its "does this artist have anything to fetch" check on `isMirloCanimusEnabled() ? artist.mirloUrl : null`), but `CATALOGUEABLE_PLATFORMS` in this same file (line 685: `['bandcamp', 'discogs', 'faircamp', 'jamcoop']`) is *not* updated to include `mirlo`.

That constant's own docstring (lines 678-683) says: *"Keep this list identical to that condition or the sweep will spend its batch on artists with nothing to fetch and turn the failure counters into noise."* `getStaleCatalogCandidates` builds its entire recatalog-sweep pool from `CATALOGUEABLE_PLATFORMS`.

Net effect once `MIRLO_CANIMUS_ENABLED=true` is flipped: `catalogArtist` will correctly treat a mirlo-only artist as having something to fetch, but `getStaleCatalogCandidates` will never include that artist in its sweep pool — so `recatalog-sweep` silently never refreshes mirlo-only artists (or picks up a first Mirlo catalog for an artist whose only link is Mirlo), even though the two call sites are documented as needing to agree. It's the same "these two lists must match" hazard the comment already warns about, just in the direction of under-inclusion rather than over-inclusion, so it's silent rather than failure-counter noise.

Harmless today since the flag defaults off, but easy to forget by the time it's flipped on — nothing exercises this path yet. Worth adding `mirlo` to `CATALOGUEABLE_PLATFORMS` now (unconditionally is fine — `isCatalogueableLink` already returns `true` for anything but `bandcamp` search-link placeholders — or gated the same way `catalogArtist` gates it) so the two lists can't drift apart.

---

brandonlucasgreen (commented):
@brandonlucasgreen Automated senior-engineer review.

**Security**
- Auth/authorization: no new externally-reachable surface. `catalog-artist-background.ts` keeps its existing internal-secret `isAuthorized` gate, untouched by this PR. confirmed.
- Input validation / SSRF: every outbound Canimus request goes through `isUrlHostnameAllowed()` before `safeFetch`, matching the codebase's mandatory SSRF pattern (`mirlo.space` / `*.mirlo.space` was already allowlisted in `middleware.ts`, pre-dating this PR). Beyond that, `ingestCanimusCatalog` host-pins **every URL in the response body itself** (`sameHostUrl`) before trusting it as a release or artist URL — including artwork `src`, which is further restricted to `http(s)://` only before it can land in an ``. This is good, deliberate defense-in-depth against a federation format whose whole point is naming other hosts. confirmed, no findings.
- Injection risks: all Supabase writes (`persistMirloReleases`) go through the client's parameter binding, matching the existing `persistJamcoopReleases`/`persistFaircampReleases` pattern. confirmed, no findings.
- Unsafe defaults: `MIRLO_CANIMUS_ENABLED` defaults to off (`process.env.MIRLO_CANIMUS_ENABLED === 'true'`), and the standalone measurement script (`scripts/measure-mirlo-overlap.ts`) refuses to run without the same flag, with a clear message about *why* (robots.txt `/v1/` disallow, pending Mirlo's answer). This is exactly the right shape for a feature that can't be enabled until an external party confirms it's okay. confirmed, no findings.
- Data exposure: no new PII paths; Mirlo-declared artist links are read and compared but never fetched or written as a source (correctly commented as such).

**Regression**
- Left a line comment on `api/functions/db.ts` (`getArtistForCatalog`'s `.in('platform', [...])`, ~line 1019): `CATALOGUEABLE_PLATFORMS` (line 685), which the recatalog sweep's pool query (`getStaleCatalogCandidates`) is built from, was not updated to include `mirlo` even though `catalogArtist`'s "anything to fetch" check now considers `mirloUrl`. The constant's own docstring says these two must stay in sync. Confirmed by reading both call sites — not speculative. Harmless while the flag is off, but will silently under-populate the sweep pool for mirlo-only artists the moment `MIRLO_CANIMUS_ENABLED` flips on, and nothing in this PR's tests would catch it since the flag is off in test/dev too.
- Minor, lower-confidence (plausible, not confirmed as a live bug): the Canimus pagination loop in `catalog-artist-background.ts` (`fetchCanimusCatalog`, `for (let page = 0; page < CANIMUS_MAX_PAGES; page++)`, `CANIMUS_MAX_PAGES = 10` × `CANIMUS_PAGE_SIZE = 200` = 2,000-artist ceiling) only stops early on a short/empty final page. If the catalog ever reaches exactly 10 full pages, the loop exits at the `MAX_PAGES` bound with no error/warning — unlike `readAllPages` in the same codebase, which explicitly `console.error`s and refuses a truncated read when it hits its own backstop. Given the opt-in nature of Canimus this is unlikely to bite soon, but it's an inconsistency with the "never silently truncate" principle this file otherwise follows carefully (e.g. the `ingestCanimusCatalog` null-vs-empty distinction is exemplary). Consider a log line if `page === CANIMUS_MAX_PAGES - 1` and the last page was full.
- State transitions / never-overwrite rules in `persistMirloReleases`: checked against `persistJamcoopReleases` for consistency (curated-field guards, `needs_review`/`flagged_against_release_id` fuzzy-match flow, exact-match-only on `match_key`) — matches the established pattern. confirmed, no findings.
- Edge cases: empty-title Canimus drafts dropped, future-dated releases (a real 2925-11-02 row in Mirlo's live data) bounded by `parseReleaseDate`, `type !== 'root'` and non-JSON bodies both correctly return `null` rather than an empty catalog (so a challenge page can't be cached as "no artists opted in"). confirmed, no findings — this is the one thing the PR description flags as its biggest risk area, and it's handled correctly.
- API contract drift: none — this only adds a new, flag-gated code path; nothing existing changes shape.

One item to resolve before flipping the flag: the `CATALOGUEABLE_PLATFORMS` gap above.

---

brandonlucasgreen (commented):

**Security**
- SSRF: every outbound Canimus request goes through `isUrlHostnameAllowed()` + `safeFetch` (`mirlo.space`/`*.mirlo.space` already on the allowlist), and every URL parsed *out of* the response — artist URL, release URL — is host-pinned via `sameHostUrl()`/`canimusArtistSlug()` against the host the document was actually fetched from, not trusted from the payload. *confirmed*
- The one deliberately un-pinned field, `CanimusArtist.links` (artist-declared off-platform links), is never fetched and never reaches `persistMirloReleases` — only `buildMirloReleases(entry)` output does, which doesn't carry `links`. Matches the comment's claim that it's compare-only. *confirmed*
- Artwork URL is scheme-checked (`^https?:\/\//i`) before being stored/rendered in an ``, blocking `javascript:`/`data:` injection. *confirmed*
- Feature is off by default (`MIRLO_CANIMUS_ENABLED`), including the measurement script, which additionally refuses to run without the same flag — no request reaches Mirlo until the robots.txt question is resolved. *confirmed*
- No secrets involved; no auth changes to the function (still gated by the existing shared-secret header, untouched here). *confirmed*

**Regression**
- *Plausible* — `catalogArtist`'s per-artist bail-out ("nothing to catalog") now treats a Mirlo-only link as catalogueable (`!mirloUrl` added to the guard), but `CATALOGUEABLE_PLATFORMS` in `db.ts` (used by `getStaleCatalogCandidates` to build the recatalog-sweep pool) was **not** updated to include `'mirlo'`. CLAUDE.md calls this out explicitly as a class of bug to avoid ("Keep that platform list identical to `catalogArtist`'s ... check"). As shipped, once the flag is on, an artist with *only* a Mirlo link will never be picked up by the 6-hour sweep — they'd only get catalogued via a demand-driven trigger (search/save/admin button). That may be an acceptable/intentional scope limit for v1, but it's the exact mismatch the existing comment warns about, so worth a one-line decision either way (add `mirlo` to the sweep pool, or note why it's excluded).
- Date/status handling (2925 date rejection, empty-title drafts dropped, dedup on `match_key` only since Canimus types everything `"album"`) is well covered by the new tests and matches the existing Jam.coop/Faircamp patterns for "never overwrite a curated field." *confirmed*
- PR is currently in a **dirty** (conflicting) mergeable state against `main` — will need a rebase before merge regardless of the above. *needs-manual-verification*

Nothing here should block on security grounds; the Mirlo-in-sweep-pool question is the one item worth a decision before merge.

---

brandonlucasgreen (commented):
@brandonlucasgreen Security/regression review of this PR (automated).

**Security**

- **SSRF — needs-manual-verification, but looks correct.** `fetchCanimusCatalog` checks `isUrlHostnameAllowed(url)` before every page fetch and routes through `safeFetch`. `mirlo.space` / `*.mirlo.space` are already present in `ALLOWED_OUTBOUND_HOSTNAMES` (`api/functions/middleware.ts`), so no allowlist change was needed or missed.
- **Untrusted document parsing — no_change_needed / confirmed-safe.** `ingestCanimusCatalog` pins every artist and release URL to the fetch host (`sameHostUrl`) before trusting it, exactly matching the existing `resolveReleaseUrl`/`resolveSameHost` pattern for fetched-markup hrefs. Artwork `src` is restricted to `http(s)://` (`canimusCoverUrl`), which correctly refuses a `javascript:` URI (covered by a test). Good adherence to "respect other people's servers" / SSRF-mandatory principles in CLAUDE.md.
- **Artist-declared `links[]` — plausible, low severity.** `parseCanimusArtist` accepts any `href` string for artist-declared links (no host pin, no scheme check), with a comment stating they're "only ever compared, never fetched and never written as a source." Confirmed true for this PR's code: `MirloReleaseToPersist`/`persistMirloReleases` never touch `.links`, and the only consumer is the read-only `measure-mirlo-overlap.ts` (console output only). Flagging as needs-manual-verification only because `CanimusArtist.links` is exported and nothing at the type level stops a future caller from rendering `href` directly as an `<a href>` — a `javascript:` value would sail through today's parser.
- **Unsafe defaults — none.** `MIRLO_CANIMUS_ENABLED` correctly defaults to off (`=== 'true'`) so no outbound Mirlo request happens until the flag is explicitly set, consistent with the "not ours to assume away" stance on Mirlo's `Disallow: /v1/`.
- **Secrets/auth/injection — none found.** No new endpoints, no raw SQL, Supabase writes go through the existing parameterized client methods.

**Regression**

- **Confirmed: sweep/catalogArtist eligibility drift.** `catalogArtist` (`api/functions/catalog-artist-background.ts`) now treats an artist as catalogueable when `mirloUrl` is set (gated by the flag):
  ```
  const mirloUrl = isMirloCanimusEnabled() ? artist.mirloUrl : null;
  if (!artist.bandcampUrl && !artist.discogsUrl && !artist.faircampUrl && !artist.jamcoopUrl && !mirloUrl) { ... error ... }
  ```
  But `CATALOGUEABLE_PLATFORMS` in `api/functions/db.ts` (line 694) is still `['bandcamp', 'discogs', 'faircamp', 'jamcoop']` — **not updated to include `mirlo`**. That constant's own doc comment (lines 684–692, 707–709) says it must stay identical to `catalogArtist`'s "is there anything to fetch" check specifically so `getStaleCatalogCandidates` (the recatalog-sweep pool) and `catalogArtist` "cannot disagree about who is worth a run." They now disagree: once `MIRLO_CANIMUS_ENABLED=true`, an artist with *only* a `mirlo` link is legitimately catalogueable per `catalogArtist`, but is invisible to `getStaleCatalogCandidates`'s pool query (which filters `artist_links` by `CATALOGUEABLE_PLATFORMS` before anything else runs). Net effect: such an artist gets catalogued once via a demand-driven trigger (search/save) and then never refreshed by the 6-hourly sweep — precisely the staleness failure mode `recatalog-sweep.ts` was built to close (see CLAUDE.md's "Keeping catalogues fresh"). Currently dormant (flag is off), but it'll surface silently the moment the flag flips unless `CATALOGUEABLE_PLATFORMS` is updated alongside it.
- **Plausible, low severity: silent pagination truncation.** `fetchCanimusCatalog` caps at `CANIMUS_MAX_PAGES = 10` (2,000 artists). The loop's only two exits are "short page → done" and "hit the page cap" — the latter isn't distinguished in the logs, so a catalog that happens to be an exact multiple of 2,000 (or larger) would be silently truncated with a log line indistinguishable from a complete fetch. Given Mirlo's catalogue is described as small today this is unlikely to bite soon, but worth a one-line log (`page === CANIMUS_MAX_PAGES` case) so a future silent truncation isn't mistaken for a complete sync.
- **Edge case, informational:** `MIRLO_RESERVED_SEGMENTS` is a hand-maintained list (`v1`, `api`, `auth`, `login`, ... `admin`) rather than derived from Mirlo's actual routes. Missing a reserved path (e.g. a future `/search` or `/terms`) would only cause a benign no-match against the Canimus catalog (host + protocol are still validated), not a security issue — just a data-quality miss worth revisiting if Mirlo adds routes.

Nothing above blocks merge given the feature flag defaults off; the `CATALOGUEABLE_PLATFORMS` gap is the one item I'd fix (or explicitly accept) before flipping `MIRLO_CANIMUS_ENABLED` to `true` anywhere.

---

PR #406, opened 2026-08-02
Sweep every artist with something to crawl, not just saved ones

brandonlucasgreen:
Follow-up to #404. You were right that saved-only was too narrow — the numbers are more lopsided than either of us assumed.

## The measurement

| | count |
|---|---|
| artists total (all created by search) | 3,421 |
| **artists with a catalogue-able link** | **2,564** |
| artists ever catalogued | 125 |
| **distinct saved artists** | **9** |

Nine. The sweep's entire eligible universe fit inside one 25-artist batch, and #404's first run consumed five of them. It would have sat idle almost every run. The cadence justification in that PR — "175/week, comfortably above the saved population" — was true and beside the point.

Alerts aren't the only consumer either. `/a/:slug` renders a release list for any catalogued artist, and those pages exist because somebody *searched* — so an unrefreshed catalogue is a visibly out-of-date artist page, not just a missed alert.

## What changed

**Pool**: every artist with a `bandcamp`, `discogs`, `faircamp` or `jamcoop` link.

Not *every searched artist*, though — 2,724 of the 3,421 have no crawlable link, and `catalogArtist` records a linkless artist as an **error**, bumping `consecutive_failures` and writing `last_error`. Sweeping them would spend the batch on artists with nothing to fetch and turn the failure counters into noise. `CATALOGUEABLE_PLATFORMS` is deliberately identical to that check, and excludes `officialsite` (only followed to *discover* other platforms).

**Ordering** gains a tier on top of the existing ones:

1. **Saved artists** — an alert is a promise to a person, and with single figures of them it costs the rest of the pool almost nothing to let them jump the backfill.
2. Never catalogued.
3. Stalest `last_attempted_at`.
4. Savers, pure tiebreak.

**Cadence**: daily → every 6 hours. Batch size can't grow (25 is one background invocation; more is silently dropped), so frequency is the only throughput lever. 4 × 25 = 100/day:

- saved artists come round as soon as their 7-day cooldown expires
- the tail rotates roughly monthly — right for artist-page freshness
- the ~2,400 artists with no catalogue drain in about three weeks
- against Bandcamp: 400–500 sequential requests/day, under one a minute averaged, robots-permitted paths

## A silent truncation fixed on the way

**PostgREST caps every response at 1,000 rows regardless of the `.limit()` you ask for, and truncates with no error.** A single `.select()` over `artist_links` returns 1,000 of ~3,900 rows and looks completely successful — my own first measurement in this session reported 697 catalogue-able artists for exactly that reason.

#404's `.limit(10_000)` was harmless only because `saved_artists` has 31 rows. A reviewer flagged the missing `ORDER BY` there; the real problem was a tenth of the size they thought. All three reads now page through `.range()` until a short page, and **refuse rather than return a truncated pool** if they exceed the backstop — a partial pool that reads as complete is the same silent-failure class as everything else in this feature.

## Verification

- `npm run build`, `test:api` (801 passing), `test:unit` (625), `typecheck:api` — clean. `dispatch.xml`/`sitemap.xml` reverted.
- **Mutation-checked**: collapsing the pager to a single read, admitting `officialsite` to the pool, dropping the saved-first tier, and counting tombstoned saves each fail the test that claims to protect them.
- New paging tests drive a fake client that reproduces the 1,000-row cap, so they'd fail against the old `.limit()` shape.

No migration, no new secrets — this is selection logic and a cron line.

### DIFF
--- a/.github/workflows/recatalog-sweep.yml
+++ b/.github/workflows/recatalog-sweep.yml
@@ -1,35 +1,38 @@
 name: Release catalogue sweep
 
-# Keeps saved artists' release catalogues from going stale.
+# Builds and refreshes release catalogues in the background.
 #
 # Every other catalog trigger is demand-driven: a save, a search, a button. So an artist who is
-# saved but never searched gets catalogued once and never again, and their release alerts
-# quietly stop — the long tail this feature exists for. See api/functions/recatalog-sweep.ts.
+# catalogued once and not searched again quietly stops producing release alerts, and their
+# artist page keeps showing the releases they had that day. See api/functions/recatalog-sweep.ts.
 #
 # A GitHub Actions cron rather than a scheduled Netlify function: there are no scheduled Netlify
 # functions in this repo, and this is the established precedent (schedule-social-posts.yml,
 # upstash-keepalive.yml).
 #
-# ## Cadence: daily, 05:00 UTC
+# ## Cadence: every 6 hours
 #
-# What this actually costs the sites we crawl is one batch of at most 25 artists, run through
-# the existing background function, which paces itself: one second between artists, at most 100
-# Bandcamp release-page fetches per run on top of one grid page each. So a sweep is on the order
-# of 125 requests to Bandcamp, sequential, spread over minutes — a trickle, and less than a
-# handful of real visitors searching. 05:00 UTC keeps it off both US and European peak.
+# One run asks for at most 25 artists — one background invocation's worth — so four runs a day
+# is 100 artists daily. Measured against the pool on 2026-08-02 (2,564 artists with a
+# catalogue-able link, 9 of them saved by anybody) that means:
 #
-# Daily rather than weekly because the batch is small and the 7-day cooldown is what actually
-# bounds per-artist crawling: an artist can't be re-crawled more than once a week however often
-# this runs. Daily just means the 25 slots are spread across seven small runs instead of arriving
-# in one lump, and it caps how long a newly-saved artist waits for a first catalogue if the
-# save-time request was dropped by the hourly cap.
+#   - Saved artists come round as soon as their 7-day cooldown expires. They sort first, and
+#     there are single figures of them, so an alert never waits behind the backfill.
+#   - The rest rotate roughly monthly, which is the right order for artist-page freshness.
+#   - The ~2,400 artists with no catalogue at all drain in about three weeks.
+#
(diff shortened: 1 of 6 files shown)


brandonlucasgreen (commented):

- **Auth/authorization** — unchanged; same `isInternalRequest` shared-secret check on the function, same GitHub secret in the workflow. *confirmed*
- **Input validation / injection** — purely internal reads via the Supabase client (`.eq`, `.in`, `.range`), no user input reaches a query. *confirmed*
- **Secret handling** — no new secrets. *confirmed*
- **Unsafe defaults / data exposure** — N/A, no new external surface introduced.

**Regression**
- **Confirmed — `stalestAttemptedAt` no longer means what its name/doc-comment claims.** The new sort puts `saved` artists ahead of everyone else regardless of staleness (`db.ts`: `if (a.saved !== b.saved) return a.saved ? -1 : 1;`), which your own test demonstrates — a saved artist last attempted 200 hours ago sorts ahead of an unsaved artist last attempted 5,000 hours ago (`"puts saved artists ahead of everyone, however fresh they are"`). But `recatalog-sweep.ts` still reports `stalestAttemptedAt: candidates[0]?.lastAttemptedAt`, with the JSDoc still saying "How stale the stalest artist was." Post-merge, a saved-heavy batch can report a *recent* `stalestAttemptedAt` while a genuinely stale unsaved artist sits later in that same batch — the exact "looks like a quiet, healthy run but isn't" failure mode this file's own comments warn about elsewhere. Low severity (it's a workflow-log field, not something blocking a fix), but the name is now misleading. Consider renaming it (e.g. `firstCandidateAttemptedAt`) or computing an actual min over the batch.
- **Needs-manual-verification — outbound Bandcamp volume roughly 4x's.** Pool goes from 9 saved artists to ~2,564 catalogueable ones, and cadence goes from daily to every 6 hours (1 run/day → 4 runs/day), which by the PR's own estimate takes Bandcamp request volume from ~125/day to ~400-500/day. The reasoning in the workflow comment reads as sound, but given this repo's documented history of self-inflicted scraping outages, it's worth someone consciously signing off on the 4x jump rather than it landing purely as a side effect of widening the pool.
- Paging correctness (`readAllPages` against the documented PostgREST 1,000-row silent-truncation bug), the saved/pool interaction, and the rest of the sort tiebreak all check out against their tests.

Deploy preview is green.

---

PR #407, opened 2026-08-03
Stop cataloguing Bandcamp search placeholders that can only 404

brandonlucasgreen:
## What was wrong

`release_catalog_state` had exactly one error message in it: `bandcamp responded 404`. All 18 rows carrying it turned out to be the same thing.

The stored `bandcamp` link was `https://bandcamp.com/search?q=<name>` — the "go search Bandcamp yourself" placeholder `attachAmpwallAndSearchLinks` writes when nothing resolved a real artist page. `bandcampMusicUrl()` reduces any URL to its origin plus `/music`, so every one of them derives the same URL:

```
https://bandcamp.com/search?q=Mount%20Eerie  →  https://bandcamp.com/music  →  404
```

One failure reproduced 18 times, not 18 failures. Checked against the three candidate causes: **wrong slug 0, stale-once-valid 0, URL shape 18/18.**

## Why now

Measured 2026-08-03: **189 placeholder rows, 10.4% of all 1,822 stored Bandcamp links**, every one of them in the sweep pool. #406 had just widened the scheduled sweep to all 2,564 catalogue-able artists every 6 hours, so the ~171 not yet reached were queued to fail identically — each burning one of 25 batch slots and climbing an exponential backoff it could never escape.

- **73** were in the pool *only* because of the placeholder — no other catalogue-able link at all.
- **116** also had a real Discogs/Faircamp/jam.coop link. Those passes still ran, but `if (bandcampError && totalFound === 0)` stamped the Bandcamp 404 as `last_error` — which is why it was the only message in the table. It was masking what the other passes actually did.

## The probe is not at fault

The placeholder is the probe's *negative* answer being written to disk as a link, not a false-accept. Spot-checking all 18 against `bandcamp_slug_probes` and independently curling the candidate slugs:

| Probe verdict | n | Verified |
|---|---|---|
| `absent` | 7 | ✅ both slug candidates 404 |
| `rejected_empty` (parked) | 3 | ✅ Viagra Boys / Mount Eerie / ChatGPT — 200 with 0 grid items |
| `accepted`, placeholder stored anyway | 1 | Vienna Vienna |
| no probe row | 5 | 3 have live populated pages |

The probe's design premise held: it correctly refused three parked accounts a naive slug check would have accepted.

## The fix

`isBandcampSearchLink()` already existed in `search-utils.ts` with a comment drawing exactly this distinction; it just wasn't consulted at the catalogue boundary. `isCatalogueableLink` in `db.ts` now applies it in both places — `getStaleCatalogCandidates`'s pool and `getArtistForCatalog`'s result — sharing one predicate so "who is worth a run" and "is there anything to fetch" cannot drift, as CLAUDE.md requires.

The migration clears the backoff already recorded, scoped to rows that never catalogued successfully so a genuine dead link keeps backing off. Validated against a throwaway Postgres 16 including idempotency; a 404 from an artist who *had* succeeded, a bot-challenge row and a healthy row were all left untouched.

## Testing

- New `catalog-artist-links.test.ts` — `getArtistForCatalog` was mocked everywhere and never tested directly. Driven against a recording fake client so the real query body runs.
- Extended `recatalog-sweep-selection.test.ts`; its `link()` fixture now carries a URL, since the pool judges shape as well as platform.
- Verified the 4 new assertions fail without the predicate.
- `npm run build`, `test:api` (812 passed), `typecheck:api` all green. Lint unchanged from the pre-existing `apps/web` baseline — nothing here is in lint scope.

## Deliberately not included

Two things I found and left alone, happy to take either as follow-ups:

1. **A coverage bug**, separate from this one: 5 of the 18 were never probed at all, and **Warren Harrison (16 releases), melondruie (16) and Nixienoise (6)** have live catalogues sitting at the obvious `<name>.bandcamp.com`. Different fix, different blast radius.
2. All 189 placeholder rows carry **`is_direct: true`**, which is a lie about what the row is — but changing it touches search display.

### DIFF
--- a/api/functions/__tests__/catalog-artist-links.test.ts
+++ b/api/functions/__tests__/catalog-artist-links.test.ts
@@ -1,0 +1,147 @@
+// Which stored links getArtistForCatalog hands the crawler.
+//
+// This is the other half of the pool decision made in recatalog-sweep-selection.test.ts, and the
+// two must agree: getStaleCatalogCandidates decides who is worth a run, catalogArtist decides
+// what to fetch for them. If one counts a link the other refuses, the sweep spends its batch on
+// artists it then records as failures — which is precisely the loop this pair of checks closes.
+//
+// Driven against a recording fake Supabase client rather than a module mock, following
+// recatalog-sweep-selection.test.ts, so the real query body runs.
+
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+
+const tables: Record<string, Record<string, unknown>[]> = {
+  artists: [],
+  artist_links: [],
+};
+
+function makeClient() {
+  return {
+    from(table: string) {
+      return {
+        select(_columns: string) {
+          const filters: { column: string; value: unknown }[] = [];
+          let wantedPlatforms: string[] | null = null;
+
+          const rows = () => {
+            let out = tables[table] ?? [];
+            for (const f of filters) out = out.filter(r => r[f.column] === f.value);
+            if (wantedPlatforms) {
+              const wanted = new Set(wantedPlatforms);
+              out = out.filter(r => wanted.has(r.platform as string));
+            }
+            return out;
+          };
+
+          const builder = {
+            eq(column: string, value: unknown) {
+              filters.push({ column, value });
+              return builder;
+            },
+            in(_column: string, value: string[]) {
+              wantedPlatforms = value;
+              return builder;
+            },
+            maybeSingle() {
+              return Promise.resolve({ data: rows()[0] ?? null, error: null });
+            },
+            then(resolve: (r: unknown) => unknown, reject: (e: unknown) => unknown) {
+              return Promise.resolve({ data: rows(), error: null }).then(resolve, reject);
+            },
+          };
+          return builder;
+        },
+      };
+    },
+  };
+}
+
+vi.mock('@supabase/supabase-js', () => ({ createClient: () => makeClient() }));
+
+process.env.SUPABASE_URL = 'https://example.supabase.co';
+process.env.SUPABASE_SERVICE_KEY = 'service-key';
+
+const { getArtistForCatalog } = await import('../db');
+
+const ARTIST_ID = 'artist-1';
+
+function link(platform: string, url: string) {
+  return { artist_id: ARTIST_ID, platform, url };
+}
+
+beforeEach(() => {
+  tables.artists = [{ id: ARTIST_ID, name: 'Warren Harrison' }];
+  tables.artist_links = [];
+});
+
+describe('getArtistForCatalog', () => {
+  it('returns a real Bandcamp artist page', async () => {
+    tables.artist_links = [link('bandcamp', 'https://warrenharrison.bandcamp.com')];
+
+    const artist = await getArtistForCatalog(ARTIST_ID);
+
+    expect(artist?.bandcampUrl).toBe('https://warrenharrison.bandcamp.com');
+  });
+
+  it('refuses a bandcamp.com/search placeholder', async () => {
+    // The placeholder search-utils writes when nothing resolved a real page. bandcampMusicUrl
+    // reduces it to https://bandcamp.com/music, which 404s every single time — so handing it to
+    // the crawler can only ever produce a failure and a longer backoff.
+    tables.artist_links = [link('bandcamp', 'https://bandcamp.com/search?q=Warren%20Harrison')];
+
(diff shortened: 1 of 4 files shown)


brandonlucasgreen (commented):

- **Auth/authorization** — no new endpoint or auth surface; this only changes which stored links `getStaleCatalogCandidates`/`getArtistForCatalog` treat as crawlable. *confirmed*
- **Input validation** — `isCatalogueableLink` delegates to the existing `isBandcampSearchLink()` (`url.includes('bandcamp.com/search')`), already unit-tested and used elsewhere in `search-sources.ts`; no new parsing logic introduced. *confirmed*
- **Secret handling / injection** — no secrets touched. The migration is a static `UPDATE ... WHERE last_error = 'bandcamp responded 404' AND last_catalogued_at IS NULL` with no interpolation. I traced the string back to its source: `catalogBandcamp` throws `` `bandcamp responded ${response.status}` `` (catalog-artist-background.ts:388), caught and stored verbatim via `recordCatalogOutcome`'s `outcome.error.slice(0, 500)` (db.ts:522) — for a 404 this is exactly `bandcamp responded 404` with no artist-specific suffix, so the WHERE clause is an exact match, not a guess. *confirmed*
- **Unsafe defaults** — none; this narrows what gets crawled, doesn't widen any allowlist or auth check. *confirmed*
- **Data exposure** — none; no new data returned to a client, only internal catalog-sweep filtering. *confirmed*

**Regression**
- **Changed assumption, verified consistent** — `getStaleCatalogCandidates`'s pool-building and `getArtistForCatalog`'s per-artist fetch now share one predicate (`isCatalogueableLink`), which is exactly the thing CLAUDE.md calls out ("the two must agree... otherwise the sweep spends its batch on artists it then records as failures"). Confirmed both call sites use the same function, so they can't drift independently.
- **Edge cases** — well covered: placeholder-only artist (falls through to the pre-existing "nothing stored" branch, not a fetch), placeholder + real link on another platform (kept, other platforms untouched), deep Bandcamp paths and custom Pro domains (correctly *not* treated as placeholders). All four asserted directly and confirmed to fail without the predicate per the PR description.
- **Migration idempotency** — confirmed by inspection: the WHERE clause matches on `last_error`, which the UPDATE itself sets to `NULL`, so a second run matches zero rows. Scoping to `last_catalogued_at IS NULL` correctly excludes artists who once had a real Bandcamp presence and later 404'd for an unrelated reason (a genuine dead link), so their backoff is left intact.
- **Semantic-revert-check false positive** — the bot flagged deletions in `db.ts` and the sweep-selection test tracing back to #406/#404/#388/#376/#344. Checked the actual diff: every flagged deletion is a single-line loop body or field list being replaced by a strictly-more-filtered version (e.g. `for (...) if (row.artist_id) pool.add(...)` → same loop plus an `isCatalogueableLink` guard), not a rollback of prior functionality. No real revert.

Tight, well-scoped fix — root-caused to one string, verified against real probe data, and the shared-predicate design is exactly the "can't disagree" property the codebase has been burned by not having before. Nothing here should block merging.

---

brandonlucasgreen (commented):

**Security**
- No new external input surface and no injection risk — all DB access goes through the existing Supabase client with parameterized filters. **confirmed**
- New migration `20260803000000_clear-bandcamp-placeholder-404-backoff.sql` is a plain `UPDATE` on an existing server-only table (no new table/column, so no RLS policy needed), scoped by an exact `last_error` match **and** `last_catalogued_at IS NULL`, and is idempotent (re-running matches nothing once cleared). **confirmed**
- `WHERE last_error = 'bandcamp responded 404'` is an exact string match against a message produced elsewhere in the codebase. If that literal ever drifts before this migration runs, the backfill silently matches fewer/zero rows instead of failing loud. Low impact (one-time backfill, not a recurring path) but worth a quick eyeball that the string still matches production's `release_catalog_state.last_error` at merge time. **plausible**

**Regression**
- The core fix — `isCatalogueableLink` — is shared between `getStaleCatalogCandidates` (the sweep's pool) and `getArtistForCatalog` (what the crawler actually fetches). That's exactly the invariant this codebase has been burned by before (pool and crawl-check disagreeing → false failures poisoning `consecutive_failures`), enforced here as one function instead of two copies that could drift. **confirmed**
- Edge cases look well covered: placeholder-only artist, placeholder + real Discogs/Faircamp/jam.coop link, deep Bandcamp paths (`/album/...`, `/music`), and the null-`artist_id` skip. No gaps found. **confirmed**

No blocking findings.

---

PR #408, opened 2026-08-03
Fix migrate:dry-run and migrate:list, which had stopped checking anything

brandonlucasgreen:
## The problem

Both scripts passed `--project-ref`, which `supabase db push` and `supabase migration list` no longer accept (CLI 2.111.0).

The failure mode is the bad kind. The CLI didn't error — it printed a help blob and exited, so a dry run **appeared to complete without validating a single migration**:

```
$ npm run migrate:dry-run
{"_tag":"Help","doc":{"description":"Push new migrations to the remote database.", ... }}
{"_tag":"Errors","errors":[{"code":"UnrecognizedOption","message":"Unrecognized flag: --project-ref ..."}]}
```

Found while trying to validate the migration in #407. CLAUDE.md documents `npm run migrate:dry-run` as the local safety net before a migration auto-deploys on merge to `main`, so it has been quietly not being one.

## The fix

`--linked` is what both commands take now, and what `.github/workflows/supabase-migrate.yml` already uses — so the local check and the deploy path agree again.

That requires a one-time link, so `migrate:link` is added alongside. It writes only to the gitignored `supabase/.temp/`.

```bash
npm run migrate:link      # once per checkout
npm run migrate:dry-run
```

## Verification

Both commands now parse their flags and reach the real check, failing with an actionable error rather than a help dump:

```
Cannot find project ref. Have you run supabase link?
```

I stopped there rather than authenticating as you — `supabase projects list` hung on an interactive login prompt, so there's no access token in my shell. Completing the link and confirming a green dry run is one command on your side.

CLAUDE.md updated with the link step, and with a note that the dry run needs credentials so it isn't a substitute for validating a migration's SQL against a throwaway Postgres.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -116,6 +116,7 @@ npm run generate:social   # Generate social media posts via Buffer
 npm run sync:bandcamp-dates  # Sync Bandcamp release dates
 npm run ingest:try -- <artist>   # Dry-run release ingest against a real Bandcamp page
 npm run sentry:sourcemaps    # Upload source maps to Sentry
+npm run migrate:link      # One-time: link this checkout to the Supabase project
 npm run migrate:dry-run   # supabase db push --dry-run against the linked project
 npm run migrate:list      # List applied vs pending migrations
 ```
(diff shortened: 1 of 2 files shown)


brandonlucasgreen (commented):

**Security:** No findings. This PR only touches `package.json` scripts and `CLAUDE.md` docs — swaps `--project-ref <ref>` for `--linked` in `migrate:dry-run`/`migrate:list` and adds a one-time `migrate:link` step. No new code path, no external input, no secrets, no auth/CORS/DB changes touched. **confirmed**

**Regression:** No functional risk — these are local dev-only Supabase CLI invocations, not anything that runs in production or CI. `--project-ref` was apparently dropped by the CLI for these subcommands (silently printing a help blob and exiting 0), which is the bug being fixed here, not introduced. **confirmed**

Nothing further to flag.

---

PR #409, opened 2026-08-03
Stop synthesising Bandcamp search links, and delete the stored ones

brandonlucasgreen:
Your call: *"that's not a critical feature, I don't think it's worth showing anymore."*

## What this removes

`https://bandcamp.com/search?q=<name>` was the fallback written whenever nothing resolved a real Bandcamp artist page. It claimed a presence we hadn't found, and because it was stored as an ordinary `bandcamp` link it was indistinguishable from a real artist page downstream — the release crawler derived `/music` from each one and got a 404 every time. #407 taught the crawler to skip them; this removes the source.

Three live producers, all now silent when there's no real URL:

| File | Path |
|---|---|
| `attachAmpwallAndSearchLinks` | `api/functions/search-utils.ts` — the main path |
| MusicBrainz-only result card | `api/functions/search-sources.ts` |
| Client-side MusicBrainz fallback card | `apps/web/src/services/sources.ts` |

The dev API (`apps/web/server/`) is updated in step, so local search matches production rather than drifting further.

## The gate, not just the label

`isDirectLink` now rejects the pattern too. That isn't cosmetic: `persistSearchResults` filters on it —

```ts
p => !EXCLUDED_PLATFORMS.has(p.sourceId) && isDirectLink(p.url)
```

— so it decides whether a link gets a row **at all**. Had it listed `bandcamp.com/search`, none of the 189 rows would ever have been written and #407 wouldn't have needed a guard at the catalogue boundary. This is the upstream fix; #407 was the downstream one.

## Data cleanup

The migration deletes the 189 stored rows. Measured immediately beforehand: **all** are `source = 'search'` (none artist-curated, so nothing here is somebody's own edit), and every affected artist keeps at least one other link — **0 would be left with none**. #407's migration already cleared their catalogue backoff, so no state rows need touching.

Validated against a throwaway Postgres 16: both placeholder shapes deleted including mixed case, while `https://warrenharrison.bandcamp.com`, `.../album/x`, a Bandcamp Pro custom domain, and Discogs' and Subvert's own search links all survived. Second run is a no-op.

## This is invisible to users — deliberately checked

I ran the change and `origin/main` side by side in the browser against a search for `Darkitecture` (an artist with genuinely no Bandcamp presence). **The rendered card is identical.** The client already filtered the placeholder out of the display via its own `isDirectLink` check in `ResultCardUtils.ts`, so the link was in the data and the API response but never on screen.

So the value here is entirely in what stops reaching the database. Worth knowing if you were expecting a visible difference — there isn't one.

Separately, and pre-existing: that `Darkitecture` card renders with **no platform pills at all**, before and after. A result with nothing actionable on it may be worth a look on its own, but it isn't this change.

## Testing

- New `api/functions/__tests__/direct-link-gate.test.ts` — `isDirectLink` had no test at all despite gating every stored link.
- New cases in `disambiguation.test.ts`, including one asserting a result with no Bandcamp link still survives `filterAndSort`. The placeholder was never counted as search-only, so it used to prop results up; Subvert still does that job, but it's pinned now rather than assumed.
- Two tests in `sources-service.test.ts` updated — they asserted the placeholder was present and first. They encoded the old behaviour.
- `npm run build`, `test:api` (825), `test:unit` (629), `typecheck:api` all green. Lint unchanged from the `apps/web` baseline.

## Also confirmed harmless

`splitSuspiciousPlatforms` treats Bandcamp as a `RELIABLE_RELEASE_PLATFORMS` member, so I checked whether removing the placeholder changes disambiguation. It doesn't: a placeholder carries no `allReleaseTitles`, so it always landed in the "no release data, keep with verified" branch and could never trigger a split.

### DIFF
--- a/api/functions/__tests__/direct-link-gate.test.ts
+++ b/api/functions/__tests__/direct-link-gate.test.ts
@@ -1,0 +1,45 @@
+// What `isDirectLink` lets into artist_links.
+//
+// It reads like a labelling helper and is easy to mistake for one, but `persistSearchResults`
+// filters on it (`db.ts`: `!EXCLUDED_PLATFORMS.has(p.sourceId) && isDirectLink(p.url)`), so it
+// decides whether a link gets a row at all. That makes it the last check standing between a
+// synthesised search URL and every downstream consumer that assumes a stored link is a real
+// destination — including the release crawler, which derives `/music` from it and fetches.
+//
+// The gap it had: `bandcamp.com/search` was not on the list, so 189 placeholder rows were
+// written and later crawled as artist pages, 404ing every time (#407).
+
+import { describe, it, expect } from 'vitest';
+import { isDirectLink } from '../db';
+
+describe('isDirectLink — what reaches artist_links', () => {
+  it.each([
+    ['a Bandcamp search placeholder', 'https://bandcamp.com/search?q=Mount%20Eerie'],
+    ['the same with different casing', 'https://BandCamp.com/Search?q=Mount%20Eerie'],
+    ['a Ko-fi DuckDuckGo fallback', 'https://duckduckgo.com/?q=site:ko-fi.com+Artist'],
+    ['a Google search', 'https://google.com/search?q=Artist'],
+    ['the Ampwall explore fallback', 'https://ampwall.com/explore?searchStyle=search&query=Artist'],
+    ['the BuyMeACoffee explore page', 'https://buymeacoffee.com/explore-creators'],
+  ])('rejects %s', (_label, url) => {
+    expect(isDirectLink(url)).toBe(false);
+  });
+
+  it.each([
+    ['a bare artist subdomain', 'https://melondruie.bandcamp.com'],
+    ['an album page', 'https://warrenharrison.bandcamp.com/album/some-record'],
+    ['a /music page', 'https://nixienoise.bandcamp.com/music'],
+    ['a Bandcamp Pro custom domain', 'https://music.sufjan.com'],
+    ['a Mirlo artist page', 'https://mirlo.space/warren-harrison'],
+    ['a Discogs artist page', 'https://www.discogs.com/artist/4861285'],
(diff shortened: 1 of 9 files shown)


brandonlucasgreen (commented):

**Security**
- auth/authorization: N/A, no auth surface touched.
- input validation: `isDirectLink`'s new `bandcamp.com/search` check is a plain lowercase substring test, consistent with the existing checks on the same line — needs-manual-verification only in the trivial sense that substring checks can theoretically over/under-match, but the added test (`direct-link-gate.test.ts`) pins the one case that matters (an artist subdomain containing "search" still passes). No finding.
- secret handling / injection: N/A. The migration's `DELETE ... WHERE url ILIKE '%bandcamp.com/search%'` is a static literal, no interpolation.
- unsafe defaults: N/A.
- data exposure: N/A — this only stops writing/deletes rows that were never real artist links.

No security findings.

**Regression**
- Changed assumption: "a `bandcamp` platform entry always exists on a result" no longer holds. Checked the three other producers this touches in tandem (`search-utils.ts`, `search-sources.ts`, `apps/web/src/services/sources.ts`) plus the dev API mirror (`apps/web/server/search/index.ts`) — all four are updated consistently, and I didn't find a fifth producer: grepped `apps/mac` and `apps/extension` for `bandcamp.com/search`, no hits. (`apps/web/server/search/bandcamp.ts` still builds a `bandcamp.com/search` URL, but that's the live-scrape fetch target, not a stored placeholder — out of scope and pre-existing.)
- Null/empty behavior: new tests cover a result with zero Bandcamp platforms surviving `filterAndSort` (previously the placeholder silently counted as "has Bandcamp"), and a `sourceId` array with no bandcamp entry rather than a placeholder. Looks correctly pinned, not just assumed.
- The `isDirectLink` gate change and the migration's `ILIKE` pattern are the same shape as the app-level check, so the two can't drift from each other.

No regression findings — this one's clean.

---

PR #410, opened 2026-08-03
Fix duplicate artist profiles: fold accents in slugs, and merge the duplicates that exist

brandonlucasgreen:
Artists have been writing in about duplicate profiles. There are **27 duplicate artist row pairs** on production and **5 shadow a claimed profile** — Kid Lightbulbs, Snap Infraction, Shannon Curtis, Loki P-A-W.net and Stan Stewart each have a second, near-empty page at a different URL.

Separately, fans reported being unable to find artists whose name has an accent. That turned out to be the *same* problem: `Björk` and `Bjork` are two rows, and a search for "bjork" lands on the one with 1 link instead of the one with 10.

## What causes it

Three mechanisms, measured:

| Cause | Pairs | Status |
|---|---|---|
| Machine-key names — `kidlightbulbs` beside `Kid Lightbulbs` | 9 | Dead since #338, which added `displayName` to every name-only source |
| Sources spelling one artist differently — `Big Thief` / `Bigthief` | 13 | Still live; not addressed here (see below) |
| Accent mangling — `Björk` → slug `bj-rk` | 5 | Generator fixed in this PR |

## Commit 1 — fold accents in slugs

`artistSlug` ran the raw name through `/[^a-z0-9]+/ -> '-'`, so `Sébastien Tellier` became `s-bastien-tellier` and `Łukasz` lost its first letter entirely. Unicode NFD alone doesn't fix it: a letter carrying its stroke in the codepoint (ł, ø, đ, æ) has no combining mark to strip. `foldToAscii` adds a short, explicit transliteration map for those — deliberately not a library, so every entry is readable.

Deliberately **not** wired into `normalizeForComparison`, which keys the Bandcamp probe cache and every match map — widening that would shift existing cache keys and silently change which names count as the same artist.

Changing `artistSlug` changes what an existing row's slug computes to, so this adds **`artist_slug_aliases`**. Three rules matter:

- Resolved only *after* `artists.slug` misses, so a live slug always wins.
- Via a separate `resolveArtistSlugAlias`, not a fallback inside `getArtistBySlug` — that runs at the front of every search and misses on nearly all of them, so it must not pay a round trip.
- The edge renderer **301s** rather than serving one page at two URLs; it only ever sees crawlers, and humans fall through to the SPA.

Migration validated against a throwaway `postgres:16`: applies clean, idempotent, PK prevents double-claiming an alias, CHECK constrains `reason`, cascade works, RLS on with no policies (server-only, matching `bandcamp_slug_probes`).

## Commit 2 — merge the duplicates that exist

Evidence-gated, because **name similarity is not evidence**: `Tiger Cub` and `Tigercub` are different bands (zero shared release titles), and so are `Honeycrush` and `Honey Crush`, which an earlier fix deliberately separated.

| Evidence | Meaning |
|---|---|
| `provenance` | loser's name **is** `normalizeForComparison(winner's name)` — only the old bug produced that |
| `release-overlap` | the rows share a release title |
| `accent-fold` | identical once folded. Uses `foldToAscii`, so `Błoto` → `Bloto` ≠ `Boto` and that pair stays unmerged |
| `name-only` | refused without an explicit override |

Blockers refuse a merge even with evidence: both rows claimed, both holding an `artist_profiles` row (UNIQUE on `artist_id`, so one artist's edits would be discarded), or both having releases with none shared.

The merge covers all seven tables carrying `artist_id`. Two are subtle:

- `artist_analytics` is counted fresh, not read off the pair snapshot — the admin list can be hours old and a stale zero would silently drop an artist's history.
- `saved_artists` has **no foreign key**, is keyed by `(user_id, artist_slug)`, and nulls `artist_id` on delete. The **slug** has to be rewritten or a fan's save survives pointing at a dead page. A test fails if that's reduced to reassigning `artist_id`.

The loser's slug is aliased **before** the row is deleted, so the URL never has a gap.

Re-slugging is included and only moves rows whose stored slug equals what the *old* `artistSlug` produced. **24 rows fail that test** because an artist picked their slug by hand — `muz4now`, `boezi`, `courstellation`, `cbertine` — 11 of them claimed. Re-slugging those would break URLs artists share.

## How to run it

One module behind both surfaces so they can't drift:

```bash
npx tsx scripts/merge-duplicate-artists.ts list
```

…or the **Duplicate artist rows** section now on `/admin/verify`. Dry run is the default everywhere, Apply only appears after a Preview showing every write, and the server re-derives the evidence per request so a stale page can't talk it into a merge.

## Result

**13 of 27 pairs mergeable** (9 provenance, 3 accent-fold, 1 release-overlap), 14 held for review, 17 slugs to fix. **Nothing is merged or re-slugged by this PR** — that's a separate, reviewed action in the admin UI.

Held back and needing a human: `Stan Stewart` (both rows claimed), `Choan Gálvez` (both hold a profile), `Tigercub`/`Tiger Cub` (probably wants a merge *override* rather than a merge, since search currently conflates two real bands).

## Not in scope

The 13 casing pairs are still arriving, and the mechanism is harder than it looks — it needs choosing a canonical name across sources that disagree, and the same machinery would wrongly merge Tiger Cub with Tigercub. Worth its own change.

## Verification

- 877 API tests pass; 32 new for the merge, 19 for accent folding
- Two bugs were caught by tests written to fail first (the stale analytics count, the `saved_artists` slug rewrite)
- Two more were caught only by running the CLI over production read-only: the winner rule was electing rows named `snapinfractionfeatmadeline`, and `Björk`/`Bjork` was landing in "needs a human"
- `typecheck:api` clean, with all three new files added to the narrow `include` and coverage proven by a deliberate type error
- Full build green; lint unchanged from the baseline on `main` (71 problems before and after)
- An existing test pinned `sigur-r-s`; updated, and strengthened to assert agreement with `artistSlug` rather than a hand-maintained string

### DIFF
--- a/api/edge/artist-page-static.ts
+++ b/api/edge/artist-page-static.ts
@@ -232,11 +232,37 @@ export default async function handler(request: Request, context: Context) {
         .single()
         .abortSignal(controller.signal);
 
-      if (!artistData) {
-        clearTimeout(timeoutId);
-        return context.next();
-      }
       artist = artistData;
+
+      // A retired slug still resolves: the artist was merged into another row, or re-slugged when
+      // accent folding was fixed (`bj-rk` -> `bjork`). Checked only *after* the live slug misses, so
+      // a real artist who later takes that slug always wins.
+      //
+      // A 301 rather than rendering the page here — this branch only ever serves crawlers, and
+      // rendering identical content at two URLs is duplicate content plus a second page competing
+      // with the canonical one. Humans never reach this code (see the crawler check above): they
+      // fall through to the SPA, whose data endpoint resolves the alias itself.
+      //
+      // Duplicates resolveArtistSlugAlias in api/functions/db.ts — edge functions run on Deno and
+      // cannot import from api/functions.
+      if (!artist) {
+        const { data: alias } = await supabase
+          .from('artist_slug_aliases')
+          .select('artists!inner(slug)')
+          .eq('alias', slug.toLowerCase())
+          .maybeSingle()
+          .abortSignal(controller.signal);
+
+        const canonical = (alias as { artists?: { slug?: string } } | null)?.artists?.slug;
+        clearTimeout(timeoutId);
+        if (!canonical) return context.next();
+
+        const prefix = url.pathname.startsWith('/artist/') ? '/artist/' : '/a/';
+        return new Response(null, {
+          status: 301,
+          headers: { Location: `${prefix}${canonical}`, 'Cache-Control': 'public, max-age=3600' },
+        });
+      }
 
       const { data: profileData } = await supabase
         .from('artist_profiles')
--- a/api/functions/__tests__/artist-merge.test.ts
+++ b/api/functions/__tests__/artist-merge.test.ts
@@ -1,0 +1,453 @@
+// Merging duplicate artist rows: what counts as evidence, and what the merge actually writes.
+//
+// This is the most destructive code in the repo — it deletes artist rows and rewrites six tables —
+// so the tests assert the *sequence* of writes, not just the outcome. Two rules carry the most
+// weight, both learned from real data on 2026-08-03:
+//
+//   - Name similarity is NOT evidence. `Tiger Cub` and `Tigercub` are different bands (zero shared
+//     release titles), as are `Honeycrush` (Brooklyn) and `Honey Crush` (Orlando). Merging on names
+//     alone would recombine what an earlier fix deliberately separated.
+//   - `saved_artists` has no foreign key and is keyed by (user_id, artist_slug), with artist_id set
+//     to NULL on delete. So a merge that only reassigns artist_id leaves fans' saves pointing at a
+//     dead slug — silently, because the row survives by design.
+
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+
+interface Op { table: string; kind: 'select' | 'update' | 'delete' | 'upsert'; payload?: unknown }
+
+const ops: Op[] = [];
+const tables: Record<string, Record<string, unknown>[]> = {};
+
+function makeClient() {
+  const match = (rows: Record<string, unknown>[], filters: [string, unknown][], ins: [string, unknown[]][]) =>
+    rows.filter(r => filters.every(([c, v]) => r[c] === v) && ins.every(([c, vs]) => vs.includes(r[c])));
+
+  return {
+    from(table: string) {
+      const filters: [string, unknown][] = [];
+      const ins: [string, unknown[]][] = [];
+      const rowsOf = () => (tables[table] ??= []);
+
+      const builder: Record<string, unknown> = {
+        select() {
+          ops.push({ table, kind: 'select' });
+          return builder;
+        },
+        eq(c: string, v: unknown) { filters.push([c, v]); return builder; },
+        in(c: string, v: unknown[]) { ins.push([c, v]); return builder; },
+        order() { return builder; },
+        range(from: number, to: number) {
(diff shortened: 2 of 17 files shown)


brandonlucasgreen (commented):

## Security

No confirmed issues. Specifically checked and clean:
- **Auth**: `admin-artist-duplicates.ts` gates every read/write behind `authenticateAdmin` (the fresh server round-trip, not the fast local-JWT path).
- **CLI trust boundary**: `scripts/merge-duplicate-artists.ts` uses the service-role key directly and is a local operator tool, not HTTP-exposed — not "bypassing" the API's auth, a different intentional trust boundary.
- **Input validation**: `winnerId`/`loserId`/`artistId` validated against a UUID regex; the server re-derives the duplicate pair and evidence server-side rather than trusting client-supplied values, so a tampered payload can't force an unevidenced merge.
- **Unsafe defaults**: dry-run is the true default in both the API and core function (only an explicit `dryRun: false` writes); `force` requires exact `=== true` and the web UI never sends it.
- **Data exposure**: no PII in the duplicates list response.
- **Injection / SSRF**: no raw SQL or new external fetches in this diff; everything goes through the Supabase query builder.
- **Migration**: idempotent, RLS enabled with no policies + explicit "server-only, deliberate" comment, matching the `bandcamp_slug_probes` precedent.

## Regression

1. **confirmed** — `api/functions/artist-merge.ts` / `supabase/migrations/20260803180000_artist-slug-aliases.sql:26`: `artist_slug_aliases.artist_id` is `ON DELETE CASCADE`. `mergeArtistPair` upserts the new `loser.slug → winner.id` alias but never repoints *pre-existing* alias rows that already point at `loser.id`. When the loser is deleted, Postgres cascades those older aliases away with it. Concretely: R1 gets accent-reslugged (`bj-rk → R1`), then R1 is later found to be a duplicate and merged as the loser into R2 — `bjork → R2` is created, but the earlier `bj-rk → R1` alias is silently destroyed, so `/a/bj-rk` 404s instead of redirecting. This is exactly the population most likely to hit it (reslugged rows are disproportionately likely to also be spelling duplicates), and neither new test file covers the chained scenario.
2. **confirmed** — `api/tsconfig.json`: `artist-lookup.ts` and `artist-lookup-v1.ts` were both modified (new `resolveArtistSlugAlias` fallback) but aren't in the typecheck `include` and aren't reachable from any included root. A type error in either won't fail `npm run build` per this repo's own documented gotcha — it'll only surface at runtime in production.
3. **plausible** — `apps/web/src/components/AdminDuplicateArtists.tsx`: the "Fix slug" (reslug) action fires immediately with `dryRun: false`, no preview/confirm step, unlike the merge flow's explicit Preview→Apply gate one section above it.
4. **plausible** — same file/flow: no Sentry breadcrumb on a successful reslug write (merge gets one), so reslug — also an irreversible production write — has a thinner audit trail.
5. **plausible** — `api/edge/artist-page-static.ts`: its inline alias lookup (necessarily duplicated for Deno) skips the `/^[A-Za-z0-9-]+$/` format guard that `db.ts`'s `resolveArtistSlugAlias` applies. Not currently exploitable — `.eq()` is parameterized and the redirect's `canonical` value is read from `artists.slug`, never reflected from user input, so this is not an open redirect — just an inconsistency between the two otherwise-parallel implementations worth tidying.

**needs-manual-verification**: whether a crash mid-merge (after evidence rows move to the winner but before the pair fully completes) could cause a re-run of `findDuplicateArtistPairs` to reclassify the same pair's evidence differently — plausible from the logic, not verified against a live/staged DB.

---

PR #411, opened 2026-08-03
Add merge and ignore buttons to the duplicate review queue

brandonlucasgreen:
The review list on `/admin/verify` was read-only. The 14 pairs that need a human decision were a list you could look at and not act on.

## Two buttons, because there are two right answers

| Action | For | How it's guarded |
|---|---|---|
| **Merge anyway** | pairs that really are one artist | Passes `force` — by definition these have no automatic evidence, so the admin *is* the evidence. Preview-first: the button only appears after a dry run has shown every write. |
| **Not duplicates** | pairs that are two different acts | Records the decision in a new `artist_duplicate_dismissals` table with an optional note and the admin's email. |

Ignoring matters as much as merging. `Tigercub`/`Tiger Cub`, `Honeycrush`/`Honey Crush` and `Boto`/`Błoto` are provably different artists — without a way to say so they sit at the top of the queue forever, and a queue that always shows the same un-actionable rows is one nobody reads. That's how a real duplicate gets missed.

## Three design points

**Keyed on the pair, not the shared name.** Keying on the name would also hide a genuinely *new* third artist who normalises the same way. The table enforces `artist_id_a < artist_id_b` so a pair has exactly one representation, and `dismissalKey` sorts to match.

**Dismissed pairs are still returned**, flagged, and shown under "Marked as different artists" with a restore button. A one-way hide would mean a mis-click silently loses a real duplicate.

**A dismissed pair refuses to merge even with `force`.** `force` overrides the automatic checks, not a recorded human decision — restoring is the explicit way back. Otherwise the queue's own record would be untrustworthy. A test fails if that guard weakens to `!opts.force`.

## Failure behaviour

A failed read of the dismissals table refuses the whole listing rather than reporting every pair as active — showing settled pairs again would invite a merge the admin already rejected. Verified against production, where the table doesn't exist yet:

```
[merge] failed reading artist_duplicate_dismissals: Could not find the table ...
Could not read artist_duplicate_dismissals
```

Which is the point, and also means **there's a deploy-ordering window**: the Netlify function deploy and the Supabase migration workflow both fire on merge, and if the function wins the race the duplicates panel 503s until the migration lands. Failing loudly for a couple of minutes beats risking a merge over a dismissal, so I've left it that way — worth a glance at `/admin/verify` a few minutes after merging.

Deliberately **not** consulted by the duplicate prevention in `persistSearchResults`: that path matches an incoming search result against a stored artist by shared platform URL, and the incoming result has no artist row yet, so there is no pair to look up.

## Verification

- Migration validated against a throwaway `postgres:16`: applies clean, idempotent, reversed id order rejected by the CHECK, same pair twice rejected by the PK, cascade drops the dismissal with either artist, RLS on with no policies.
- 898 API tests pass (10 new). Teeth proven on the force guard by weakening it to `!opts.force` and watching the test fail.
- The fake client in `artist-merge.test.ts` gained failure injection so the "couldn't read dismissals" case is actually covered rather than asserted against an undefined variable.
- `typecheck:api` clean, build green, lint unchanged from baseline (71 problems before and after).
- CLI parity: `ignore <slugA> <slugB> [--note "..."]` and `unignore`.

### DIFF
--- a/api/functions/__tests__/artist-merge.test.ts
+++ b/api/functions/__tests__/artist-merge.test.ts
@@ -18,6 +18,9 @@ interface Op { table: string; kind: 'select' | 'update' | 'delete' | 'upsert'; p
 const ops: Op[] = [];
 const tables: Record<string, Record<string, unknown>[]> = {};
 
+/** Set to a table name to make that table's paged read fail, as a dropped connection would. */
+let failingTable: string | null = null;
+
 function makeClient() {
   const match = (rows: Record<string, unknown>[], filters: [string, unknown][], ins: [string, unknown[]][]) =>
     rows.filter(r => filters.every(([c, v]) => r[c] === v) && ins.every(([c, vs]) => vs.includes(r[c])));
@@ -37,6 +40,9 @@ function makeClient() {
         in(c: string, v: unknown[]) { ins.push([c, v]); return builder; },
         order() { return builder; },
         range(from: number, to: number) {
+          if (failingTable === table) {
+            return Promise.resolve({ data: null, error: { message: 'connection reset' } });
+          }
           const rows = match(rowsOf(), filters, ins).slice(from, to + 1);
           return Promise.resolve({ data: rows, error: null });
         },
@@ -62,7 +68,12 @@ function makeClient() {
             return Promise.resolve({ data: null, error: null });
           };
           return { eq: (c: string, v: unknown) => { filters.push([c, v]); return apply(); },
-                   in: (c: string, v: unknown[]) => { ins.push([c, v]); return apply(); } };
+                   in: (c: string, v: unknown[]) => { ins.push([c, v]); return apply(); },
+                   // Composite-key delete, as restoreArtistDuplicatePair uses.
+                   match: (obj: Record<string, unknown>) => {
+                     for (const [c, v] of Object.entries(obj)) filters.push([c, v]);
+                     return apply();
+                   } };
         },
         upsert(row: Record<string, unknown>) {
           ops.push({ table, kind: 'upsert', payload: row });
@@ -82,8 +93,10 @@ vi.mock('@supabase/supabase-js', () => ({ createClient: () => makeClient() }));
 process.env.SUPABASE_URL = 'https://example.supabase.co';
 process.env.SUPABASE_SERVICE_KEY = 'k';
 
-const { findDuplicateArtistPairs, mergeArtistPair, findReslugCandidates, reslugArtist } =
-  await import('../artist-merge');
+const {
+  findDuplicateArtistPairs, mergeArtistPair, findReslugCandidates, reslugArtist,
+  dismissArtistDuplicatePair, restoreArtistDuplicatePair, dismissalKey,
+} = await import('../artist-merge');
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 const client = makeClient() as any;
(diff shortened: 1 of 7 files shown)


brandonlucasgreen (commented):

## Security

**1. Admin auth enforced server-side** — `confirmed` (pass, no action needed)
`api/functions/admin-artist-duplicates.ts:34-40` calls `authenticateAdmin()` before the body is parsed, and the new `dismiss`/`restore` branch (~L75-93) runs after that gate using the validated `admin.email`. `authenticateAdmin` (middleware.ts:184-195) fails closed on an invalid/missing bearer token or an email mismatch, returning 401 before any DB write. A non-admin cannot reach `dismissArtistDuplicatePair`/`restoreArtistDuplicatePair` directly.

**2. `note` field isn't runtime-type-checked before `.slice()`** — `plausible`
`admin-artist-duplicates.ts`, dismiss branch: `note: body.note?.slice(0, 500) ?? null`. The TS type (`note?: string`) is compile-time only — `JSON.parse(event.body)` can hand back `note: 123` or `note: {}` for any request with a valid admin token but a malformed payload, and `.slice` on a non-string throws an unhandled exception (500) instead of this file's usual clean `400`. Compare to `dryRun`/`force` in the same handler, which use `!== false` / `=== true` and are safe under any input type. Suggest a `typeof body.note === 'string'` guard.

**3. Merge vs. dismiss race window** — `plausible / needs-manual-verification`
`mergeArtistPair` re-derives the pair via a fresh `findDuplicateArtistPairs(client)` call at the start of the merge action, so an already-dismissed pair is correctly rejected up front. But the merge writes (`artist_links`, `releases`, `artist_analytics`, …) are sequential and non-transactional with no re-check of `dismissed` partway through. Two admins racing a dismiss against an in-flight merge on the same pair could still let the merge complete after the dismissal lands. This extends a pre-existing (non-transactional) pattern rather than introducing it, but it undercuts the PR's own claim that a dismissed pair can't be merged even with `force`. Would need a live-concurrency test against real Postgres to confirm exploitability.

**4. UUID case ordering in `dismissalKey`** — `needs-manual-verification`, low severity
`UUID_REGEX` accepts mixed-case UUIDs, but `dismissalKey` orders pairs via plain JS string comparison (`idA < idB`), while the DB's `CHECK (artist_id_a < artist_id_b)` compares canonical `uuid` values. All UUIDs currently produced in this codebase are lowercase, so this is unlikely to trigger — noting for completeness only.

**5. RLS/migration** — `confirmed` correct
`supabase/migrations/20260803210000_artist-duplicate-dismissals.sql` enables RLS with no policies (service-role only, matching the existing `bandcamp_slug_probes` pattern) and documents that choice. FK `on delete cascade`, composite PK, and the ordering `CHECK` are all present.

**6. No SQL injection / XSS surface** — `confirmed`. All DB access is parameterized via the Supabase client; `note` renders as plain JSX text in `AdminDuplicateArtists.tsx` (no `dangerouslySetInnerHTML`), so React escapes it.

## Regression

**7. API contract** — `confirmed` no drift. New `dismissed`/`dismissal` fields are additive on `DuplicatePair`; existing `merge`/`reslug` response shapes are unchanged, and the CLI script (`scripts/merge-duplicate-artists.ts`) was updated in the same diff.

**8. Edge cases** — `plausible`, see finding 3 for the concurrent state-transition case. Recommend a follow-up to wrap the dismiss/merge check-then-act in a single transaction or add a re-check immediately before the first write.

---

PR #412, opened 2026-08-04
Page the artist-directory reads so the index can't silently truncate

brandonlucasgreen:
## Why

`/api/artist-directory?scope=known` — the "Artists You Know" set behind [/known-artists](https://unstream.stream/known-artists) — read the whole table with a bare `.select()`:

```ts
const { data: artistRows, error } = await supabase
  .from('artists')
  .select('slug, name, image_url')
  .eq('match_confidence', 'verified');
```

PostgREST caps every response at 1,000 rows and truncates **silently** — no error, no flag, the rows simply aren't in `data`. There are **958 verified artists** today, so this endpoint was ~42 artists away from quietly dropping people off the page with nothing to notice it by. This is the same class of bug as the one that gave the recatalog sweep a pool of 697 instead of 2,564 (#404), and CLAUDE.md's "Paging, not `.limit()`" section exists because of it.

Nothing alerts on this. The only defence is reading every `.select()` and asking how big the table gets.

## What changed

**Both reads now page.** `readAllPages` in `db.ts` — already doing this for the sweep — is exported for the purpose, and takes the caller's own `.range(from, to)`, so `artist-directory.ts` keeps its own client.

- `scope=known` (958 rows): the live fix.
- The claimed path's `artist_profiles` read (128 rows): headroom rather than a live fix, but the failure mode is identical and just as invisible. Commented as such so the number doesn't read as urgent.

**The `.in('id', artistIds)` lookup is chunked at 200.** Two hazards in one call, not one: the row cap, *and* the id list travelling in the query string, where a few thousand UUIDs is a 414 rather than a slow query. Because `id` is unique, a 200-id chunk can match at most 200 rows and can never reach the cap — so the chunks need no `.range()` paging on top of the chunking, which the comment says explicitly so it doesn't read as a half-done fix.

## One behaviour change worth a look

A **failed** `artist_profiles` read used to return `200 {artists: []}` *with* `Cache-Control: public, max-age=300` — a transient error pinned as "there are no claimed artists" for five minutes after it had cleared. That's caching uncertainty on an error path. It now returns 500, matching what the `scope=known` branch already did: uncached, and `ArtistIndexPage` already retries 5xx once before giving up. Either way the user sees an empty list in the worst case; the difference is whether the CDN keeps serving it.

## Test approach

The new tests mock what PostgREST actually does rather than just offering a `.range()` method. A mock whose chain *only* has `.range()` makes unpaged code fail with `range is not a function` — a shape error, not the bug. So `cappedQuery` gives the query object both a `.range(from, to)` **and** a `then` that resolves to `rows.slice(0, 1000)` with `error: null`, exactly as production would.

The unpaged handler therefore still gets a successful-looking response and fails on the row count. Confirmed by reverting the fix and running the suite:

```
× scope=known pages past the 1,000-row PostgREST cap instead of truncating
  AssertionError: expected [ …(1000) ] to have a length of 1042 but got 1000
× default scope pages the profile read and chunks the id lookup
  AssertionError: expected [ …(1000) ] to have a length of 1150 but got 1000
```

4 tests → 7, covering: paging past the cap with the tail intact and the exact ranges requested (`[[0, 999], [1000, 1999]]`), chunked `.in()` with every id asked about exactly once and no chunk over 200, and the uncacheable 500 on a failed profile read.

`artist-directory.ts` and its test are added to `api/tsconfig.json`'s typecheck include, per CLAUDE.md's note about that list's narrow coverage. I confirmed the include actually bites by planting a deliberate type error and watching it fail.

## Verification

- `npm run test:api` — 901 passed (49 files)
- `npm run test:unit` — 636 passed (47 files)
- `npx tsc -b` and `npm run typecheck:api` — clean
- `npm run lint` — 71 problems, unchanged from the pre-existing baseline on `main` (it only covers `apps/web`; nothing here is in scope)
- No generated files dirtied (`sitemap.xml` / `dispatch.xml` untouched)

No migration, no schema change, no new env var. `/known-artists` and `/artists` are the only consumers, plus `scripts/generate-social-posts.ts`, which reads the default scope.

### DIFF
--- a/api/functions/__tests__/artist-directory.test.ts
+++ b/api/functions/__tests__/artist-directory.test.ts
@@ -11,6 +11,36 @@ vi.mock('@supabase/supabase-js', () => ({
 
 import { handler } from '../artist-directory';
 
+/** PostgREST's `max-rows` on this project. Reads above it are truncated silently. */
+const CAP = 1000;
+
+/**
+ * A filtered query that behaves the way PostgREST really does.
+ *
+ * Awaiting it directly — the bare `.select().eq()` this endpoint used to do — resolves to the
+ * *first `CAP` rows only*, with `error: null`, exactly as production would. `.range(from, to)`
+ * returns that window instead. So a handler that forgets to page still gets a successful-looking
+ * response here and fails on the row count, which is the failure that matters.
+ */
+function cappedQuery<T>(rows: T[]) {
+  const respond = (data: T[]) => Promise.resolve({ data, error: null });
+  return {
+    range: vi.fn((from: number, to: number) => respond(rows.slice(from, to + 1))),
+    then: (
+      onFulfilled: (value: { data: T[]; error: null }) => unknown,
+      onRejected?: (reason: unknown) => unknown
+    ) => respond(rows.slice(0, CAP)).then(onFulfilled, onRejected),
+  };
+}
+
+function verifiedArtists(count: number) {
+  return Array.from({ length: count }, (_, i) => ({
+    slug: `artist-${String(i).padStart(4, '0')}`,
+    name: `Artist ${String(i).padStart(4, '0')}`,
+    image_url: null,
+  }));
+}
+
 describe('artist-directory handler', () => {
   beforeEach(() => {
     vi.resetAllMocks();
@@ -25,13 +55,10 @@ describe('artist-directory handler', () => {
         eq: vi.fn((column: string, value: string) => {
           expect(column).toBe('match_confidence');
           expect(value).toBe('verified');
-          return Promise.resolve({
-            data: [
-              { slug: 'zzz-artist', name: 'ZZZ Artist', image_url: null },
-              { slug: 'patrick-hardy', name: 'Patrick Hardy', image_url: 'https://img/p.jpg' },
-            ],
-            error: null,
-          });
+          return cappedQuery([
+            { slug: 'zzz-artist', name: 'ZZZ Artist', image_url: null },
+            { slug: 'patrick-hardy', name: 'Patrick Hardy', image_url: 'https://img/p.jpg' },
+          ]);
         }),
       })),
     });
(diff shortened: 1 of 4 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Automated senior-engineer review.

**Security**
- Auth/authorization: no change — endpoint's auth posture is unchanged from `main` (public, service-role client server-side). needs-manual-verification only in the sense that this PR doesn't touch it either way.
- Input validation: `scope` param is compared with `===`, never interpolated into a query. `.in('id', ids)` uses the Supabase client's parameter binding, not string concatenation — no injection risk. confirmed.
- Injection risks: none found. confirmed.
- Unsafe defaults: none introduced.
- Data exposure: response shape unchanged (slug/name/imageUrl only).

**Regression**
- Changed assumption verified against source: `readAllPages` (exported, unchanged logic) pages correctly with `.range(from, to)` and refuses (doesn't truncate) past its 50-page backstop, matching the existing sweep helper's behavior. confirmed.
- Edge case: `profiles.rows.length === 0` (genuinely empty) is now correctly distinguished from a read error — previously both were folded into the same `200 {artists: []}` path with a 5-minute `Cache-Control`, which pinned a *transient failure* as "no claimed artists" for 5 minutes. Now errors return an uncached 500. This is a deliberate, well-flagged behavior change (called out in the PR description) and looks correct — confirmed via test `default scope returns an uncacheable 500...`.
- ID-chunking (`ID_CHUNK = 200`) correctly avoids both the 1,000-row PostgREST cap and a query-string-length 414, and chunk failure short-circuits to 500 rather than silently returning a partial list. confirmed.
- API contract: response shape and status codes for the two scopes are unchanged from callers' perspective (`/known-artists`, `/artists`, `scripts/generate-social-posts.ts`).

No findings. Test coverage for the exact regression class (958-row truncation) is good — the mock (`cappedQuery`) actually mimics PostgREST's silent-truncation behavior rather than just adding a `.range()` stub, so an un-paged handler would genuinely fail these tests.

---

brandonlucasgreen (commented):

**Security**
- Auth/authz: no change — endpoint stays public-read, no new surface. *confirmed*
- Input validation / injection: all reads go through the Supabase client (`.eq`, `.range`, `.in`) with no string-built SQL; `scope` handling is unchanged. *confirmed*
- Secret handling: none touched. *confirmed*
- Unsafe defaults: the claimed-profile failure path changed from `200 {artists:[]}` cached 5 min to an uncached `500`. That's a deliberate, documented tightening (a transient error can no longer be pinned as "no claimed artists" for 5 minutes), not a new risk. *confirmed*
- Data exposure: none — same fields returned as before. *confirmed*

**Regression**
- Paging logic (`readAllPages`, reused from `db.ts`): `range(from, from+PAGE-1)` with a `< PAGE` short-page stop is correct, including the exact-1000 boundary (one extra empty round trip, no data loss). *confirmed*
- `.in('id', artistIds)` chunking at 200: since `id` is a primary key, a 200-id chunk can never itself hit the 1,000-row PostgREST cap, so skipping `.range()` on the chunk loop is sound. *confirmed*
- API contract: `artists: [...]` shape is unchanged; the profile-read failure now returns 5xx instead of 2xx, and the PR notes `ArtistIndexPage` already retries 5xx once — worth a quick manual smoke-check that the retry actually fires from this exact response shape, since that's the one behavior a consumer could be surprised by. *needs-manual-verification*
- **Cross-PR conflict**: this PR and #413 both branch from the same base commit (`3462159`) and both rewrite the same `scope === 'known'` block in `api/functions/artist-directory.ts` — this PR adds paging via `readAllPages`, #413 adds a `died_on`/buyable-platform filter with its own separate paging loop. Whichever merges second will need a manual rebase, and if #413's version isn't reconciled with this one, the claimed-scope paging + `.in()` chunking fix here would silently disappear again. Recommend merging this one first and rebasing #413 on top so it reuses `readAllPages` for the known-scope read too rather than duplicating the loop. *needs-manual-verification*

No blocking issues. Nice test approach on `cappedQuery` — it actually exercises the truncation bug instead of just checking the range() call shape.

---

PR #413, opened 2026-08-04
Clean up the pre-generated artist database

brandonlucasgreen:
Removes the artist records that should never have existed, and stops the index and social posts making claims the data does not support. Started from `/artist/chatgpt`; the audit that followed covered all 961 verified rows.

## What was removed

Five rows, deleted from production and denylisted so a search cannot recreate them:

| Row | Why |
|---|---|
| ChatGPT | OpenAI's product — the MusicBrainz entry's official site is chatgpt.com |
| Saturday Night Live | NBC show, stored with the broadcaster's own socials |
| LANDR | Mastering SaaS, stored with instagram.com/landr.music |
| Coca-Cola | Beverage brand, no direct-purchase presence anywhere |
| Absurd | German NSBM band; the stored `horde_absurd` Instagram confirms the identity |

Two lists, kept separate on purpose: `api/lib/non-artist-names.ts` makes a factual claim ("this is not a musical act"), `api/lib/excluded-artists.ts` is an editorial one. `scripts/prune-non-artist-rows.ts` applies both, dry-run by default.

**Deliberately kept.** American Express, masterclass, Now Playing and Seoul Metro are live Bandcamp pages with real releases by independent artists who happen to share a name with a company — each verified by fetching the page. Blocking those would remove the artists Unstream exists to surface, and the test pins that distinction in both directions.

## Deceased artists

107 of the set are dead. Their `/artist/` pages stay — an estate really does sell Elliott Smith, MF DOOM and John Prine on Bandcamp. What was wrong is the copy around them: the index says they "have music available for direct purchase" and the weekly posts say "support them directly". Five had already been published: Sara Tavares, Dusty Hill, Brook Benton, Lhasa de Sela, Lex Barker.

`artists.died_on` records the date, the index filters on it, and `generate-social-posts.ts` now excludes a death date the way it already excluded comedians and actors.

Only dates resolved through the artist's **own MusicBrainz ID** are written. Name matching paired "Sebastian Bach" with Johann Sebastian Bach and "Jack White" with a footballer during research, and a wrong date silently delists a living artist. `NULL` means "no date recorded", never "alive". The 56 provable dates are seeded in the migration so the filter works the moment it applies; ~50 name-only candidates are left out pending confirmation.

That generator had a matching bug: every Wikidata batch warned and continued on error, so an outage produced an empty exclusion set — indistinguishable from "nobody to exclude", when what it excludes is a post telling people to support someone who died. It now tracks completeness and drops the pool for the run.

## Nothing to buy

14 rows claimed direct purchase with no such link. Qobuz, Beatport and Discogs count, so 11 are fine. The index now derives the qualifying set from the platform registry's own categories, which leaves out the 3 that had only Instagram and a search placeholder — and means a new platform is covered by editing the registry.

## Subvert search placeholders

`subvert.fm` is `searchOnly` and its `searchUrlTemplate` is a `/discover` URL, which `isDirectLink` did not reject — so **321 of 349** stored Subvert links are the search box. The #407 Bandcamp defect on another platform. Scoped to the `/discover` path: the other 28 are real artist pages artists added themselves.

## Also

The known-scope read is now paged. PostgREST truncates at 1,000 rows silently and the set is at 958, so it was weeks away from dropping artists with no error to show for it. The claimed-scope read has the same ceiling at 128 rows and is left alone with a comment.

## Verification

- 921 API tests, 636 web unit tests, `tsc -b`, `typecheck:api`, lint at its 71-problem baseline.
- Every new test confirmed to fail with its behaviour removed — the gate, the paging, both index filters.
- Both migrations applied to `postgres:16` in Docker: idempotent, the seed skips a row whose `died_on` was corrected by hand, and the Subvert delete leaves artist-curated links intact.

## Note before merging

`api/functions/artist-directory.ts` reads `died_on` with a fallback select that drops the column, because a push to main deploys the functions and runs `supabase-migrate` concurrently — they are not ordered. The window degrades to an unfiltered index rather than a 500. That fallback can be deleted once the column has shipped.


brandonlucasgreen (commented):

**Security**
- Auth/authz: the denylist checks in `persistSearchResults` sit before any write for that result (`api/functions/db.ts:3266-3282`), so a denylisted name never reaches the DB even transiently. *confirmed*
- Input validation / injection: no user input reaches SQL directly; the migrations use static `VALUES`/`ILIKE` literals, and the two new lib files (`non-artist-names.ts`, `excluded-artists.ts`) are hardcoded sets keyed on `artistSlug()` output. *confirmed*
- Secret handling: the two new scripts (`backfill-artist-death-dates.ts`, `prune-non-artist-rows.ts`) require `SUPABASE_SERVICE_KEY` from env/`.env`, consistent with existing scripts; nothing logged. *confirmed*
- `prune-non-artist-rows.ts` refuses to delete a row with a claimed profile or any saves, which is the right conservative default for a destructive script. *confirmed*
- RLS: `died_on` inherits the existing `artists` table policies (public SELECT, service-role write); migration comment states this explicitly rather than leaving it implicit. *confirmed*

**Regression**
- *Plausible* — `readVerifiedArtists`'s two-select fallback (`api/functions/artist-directory.ts`) catches **any** error on the `died_on`-inclusive query, not specifically "column does not exist." If the first select fails for an unrelated transient reason (timeout, connection blip) partway through paging, it silently retries from page 0 with the column-less select and returns `200` with an **unfiltered** list — i.e. the exact "dead artists shown as available to support" bug this PR exists to fix, on any DB hiccup, not just the intended migration-ordering race. Consider checking `error.message` for the missing-column signature specifically, or at minimum logging when the fallback is taken so it's visible when it fires outside the deploy window.
- *Plausible* — the new partial index `idx_artists_living ON artists (match_confidence) WHERE died_on IS NULL` won't actually be used by the known-scope query: that query only does `.eq('match_confidence', 'verified')` in SQL and applies `.filter(a => !a.died_on)` client-side afterward. Postgres can't select a partial index unless the query's WHERE clause includes a matching predicate, so as written this index adds write overhead with no read benefit. Either add `.is('died_on', null)` to the query (and drop the buyable-platform filter interaction accordingly) or drop the index until the filter moves into SQL.
- **Cross-PR conflict with #412** (see comment there): both PRs rewrite the same `scope === 'known'` block in `artist-directory.ts` from the same base commit. If this merges without incorporating #412's fix, the claimed-profile-scope paging and `.in()` id-chunking protection is lost again. *needs-manual-verification*
- The `died_on` fallback comment says it's temporary ("delete once the column has shipped") — worth a follow-up ticket so it doesn't linger silently masking the broader error case above.

Everything else — the denylist test coverage (collision cases in both directions), the death-date evidence rule (MusicBrainz ID only, never name match), and the Subvert `/discover` scoping (URL-shape not host, preserving the 28 real artist links) — looks careful and well-tested.

---

PR #414, opened 2026-08-05
Use one "View artist page" link treatment for all artist results

brandonlucasgreen:
## What

Search results used two different treatments for the same action:

- **Claimed artists** → accent-coloured pill, checkmark icon, "View profile"
- **Pre-generated artist results** → muted link, chevron icon, "View artist page"

Both now use the muted "View artist page" treatment. The destination still differs — `/a/:slug` for a claimed profile, `/artist/:slug` for a pre-generated page — only the visual treatment and label are unified.

## Why

The claimed/unclaimed split is an internal distinction. To a fan scanning results, both links make the same promise ("here's more about this artist"), so they shouldn't look like two different affordances.

Side effect worth noting: the accent pill was the only visual marker that a profile was claimed. Claimed results now look the same as any other artist result in the card header.

## Verification

- Rendered the real component with both fixture results in a throwaway Vite harness: identical label, icon and classes; hrefs correctly `/a/kid-lightbulbs` and `/artist/nine-inch-nails`.
- `npx tsc -b apps/web` clean, no new lint errors, `npm run test:unit` 628 passing (the one failing file — an `api/search/enrichment.ts` import error — fails identically on a clean `main`).

### DIFF
--- a/apps/web/src/components/ResultCardHeader.tsx
+++ b/apps/web/src/components/ResultCardHeader.tsx
@@ -14,6 +14,13 @@ interface ResultCardHeaderProps {
   onSave?: (e: React.MouseEvent) => void;
 }
 
+/** Where this result's artist page lives, or null if it doesn't have one. */
+function artistPagePath(result: SearchResult): string | null {
+  if (result.matchConfidence === 'claimed') return `/a/${result.claimedSlug || result.id}`;
+  if (result.knownSlug) return `/artist/${result.knownSlug}`;
+  return null;
+}
+
 export function ResultCardHeader({
   result,
   isAdmin,
@@ -25,6 +32,7 @@ export function ResultCardHeader({
   onSave,
 }: ResultCardHeaderProps) {
   const [imageError, setImageError] = useState(false);
+  const artistPath = artistPagePath(result);
 
   return (
     <div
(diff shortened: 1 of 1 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Security/regression review of this PR (automated).

**Security**
- No findings. The diff is a pure client-side refactor (`ResultCardHeader.tsx`): the two existing `<Link>` branches are collapsed into one `artistPagePath()` helper. No new user input is introduced, nothing is rendered via `dangerouslySetInnerHTML`, and destinations still go through React Router's `<Link to=...>` rather than a raw URL/window navigation.

**Regression**
- Logic equivalence check: old ternary was `claimed → /a/:slug` else `knownSlug → /artist/:slug` else nothing. New `artistPagePath()` implements the identical precedence and falls back to `null`, and the render guard changed from `... : null` to `{artistPath && (...)}`, which is behaviorally identical for a `string | null`. **no_change_needed**.
- The load-bearing comment explaining why this must stay a React Router `<Link>` (Safari 26 back-button / UNS-99 regression, #269) was preserved, just moved — good, since that context is exactly the kind CLAUDE.md flags as "load-bearing."

No blocking issues found.

---

PR #415, opened 2026-08-05
Add Mirlo as a release source via its REST API

brandonlucasgreen:
Closes step 10 of the locked V1 Releases scope, and closes the Mirlo question that #405 was paused on. **Supersedes and replaces [#405](https://github.com/brandonlucasgreen/unstream/pull/405)**, now closed — Mirlo granted Unstream permission and issued an API key on 2026-08-05, which made the REST API available, and it has the prices Canimus doesn't.

Two commits, reviewable independently:

1. `f253664` — the Mirlo REST ingest
2. `e8e6e3c` — the "Digital" release-type label (touches the artist page, release page and admin queue, so worth reading on its own)

## One request per artist, discography *and* prices

`GET /v1/artists/{slug}` returns everything in one response, so **there is no detail pass at all**. Bandcamp needs a grid plus one fetch per release; Jam.coop needs an artist page plus one per album. Mirlo needs one. It is now the cheapest and richest release source in the codebase.

## Verified against live data, which is what #405 couldn't do

#405's fixtures came from pasted documentation, so they pinned *our assumptions* rather than Mirlo's field names. Every fixture here was captured from real responses — **209 releases across 31 artists** — and three findings changed the implementation:

| Finding | Why it matters |
|---|---|
| `minPrice` is **integer cents**, and `null` is not `0` | 56 of 209 releases have no price configured. Publishing those as `0` renders "Name your price" and advertises terms the artist never set. They get **no offer**. |
| **Currency casing is inconsistent** — `GBP` (40) *and* `gbp` (8), `usd` *and* `USD` | Stored raw, one currency becomes two. |
| **Drafts bypass every visibility flag** | Both real drafts report `isPublic: true`, `hideFromSearch: false`, `isGettable: true`. The filter keys on the empty title and the `mi-temp-slug-` prefix instead. |

Also: `releaseDate` was non-null on all 209 (contradicting an earlier note), genuine future dates exist, and `isPreorder: true` appears — so status gets a real signal rather than an inference from the date.

## `platformPercent` is deliberately ignored

It looks like a per-release payout share, which would beat the registry's static range outright. It isn't trustworthy as one: values run 7, 10, 15, 20, 25, 50, 100; it **contradicts the artist's own `defaultPlatformFee`** (one artist has `defaultPlatformFee: 7` with every release at `50`); and it reached `100` on a free release, which is meaningless as a fee. None of the outliers correlate with `fundraiser` or `isAllOrNothing`.

Payout stays on the platform registry until Mirlo tells us what the field means. A wrong number about an artist's money is worse than a coarse one. **Open question to raise with Mirlo** — if it means what it might, it's a real upgrade later.

## The API key gates nothing

Verified the same day: this endpoint returns **byte-identical responses** with `mirlo-api-key`, with `Authorization: Bearer`, and with **no auth at all**. It's a public read endpoint.

The key is still sent — it identifies us and future-proofs against them gating or rate-limiting by key — but an absent key must not disable the pass, and nothing here assumes an authenticated response differs. **What unblocked this work was the permission, not the key.** `Disallow: /v1/` is still in Mirlo's robots.txt; we poll it on direct consent rather than by ignoring it.

## No feature flag, and what that means on merge

`RELEASE_CATALOG_ENABLED` (production-only) is already the master gate, and #405's `MIRLO_CANIMUS_ENABLED` existed solely because the robots question was open. A second flag would be speculative config.

**So merging starts cataloguing Mirlo-linked artists in production** — ~66 artists had a Mirlo link as of 2026-08-02 — and adds `mirlo` to the recatalog sweep's pool. That's the intent, but it isn't dormant.

## The "Digital" label (second commit)

Mirlo, Faircamp and Jam.coop expose no type field, so their releases store `release_type: 'other'`. All five render sites already omitted the label entirely, so those rows showed no type at all. Where **every** platform selling a release is download-only, they now read "Digital".

Done in the display layer, not the data: extending the `release_type` CHECK constraint would put a *format* into a column that answers "what kind of work is this", and would make `'digital'` reachable for a Bandcamp release that only exists on vinyl. `digitalOnly` lives on the platform registry with per-platform evidence; absent means "not established", not "sells physical". One unflagged platform drops the label — a release on Mirlo *and* Bandcamp may well be a vinyl pressing.

## Verification

- `npm run build` ✅ · `test:api` **958 passing** (+5) · `test:unit` **640 passing** (+4) · `typecheck:api` ✅
- Lint unchanged from the repo's 71-problem / 64-error baseline
- **Eleven mutations proven to fail exactly their intended test** across both commits — including cents-not-divided, currency-not-normalized, uncertainty-collapsed-to-`[]`, null-minPrice-coerced-to-zero, and the vacuous `[].every()` that would have claimed "Digital" for a release with no sources
- Two teeth-probes initially revealed *weak tests rather than working ones*: the draft filter was only being exercised by the title check, and the sweep's `it.each` proved inclusion but never exclusion. Both were rewritten.
- End-to-end dry run against real artists via the new `npm run ingest:mirlo`: `timerival` 32 of 33 written (draft skipped, 400 cents → `4 USD`), `clive-murray` `type:'lp'` → `album` with GBP preserved, `dreamofomni` 666 cents → `6.66 CAD` with two `isGettable:false` CD comps correctly yielding no offer

## Before merging

**Set `MIRLO_API_KEY` in Netlify, Functions scope.** Not blocking — the pass works without it, it just isn't attributed to us.

Still untouched, per your earlier call: the three pre-existing unauthenticated `/v1/` calls in search, `enrichment.ts` and `check-releases.ts`.

### DIFF
--- a/api/functions/catalog-artist-background.ts
+++ b/api/functions/catalog-artist-background.ts
@@ -16,6 +16,7 @@ import {
   persistDiscogsReleases,
   persistFaircampReleases,
   persistJamcoopReleases,
+  persistMirloReleases,
   persistMusicBrainzEnrichment,
   persistReleaseDetail,
   persistReleases,
@@ -40,8 +41,10 @@ import {
   ingestFaircampReleasePage,
   ingestJamcoopAlbumPage,
   ingestJamcoopArtistPage,
+  ingestMirloArtist,
   ingestMusicBrainzReleaseGroups,
   jamcoopArtistUrl,
+  mirloArtistSlug,
   type DiscogsArtistReleaseEntry,
   type DiscogsReleaseDetailRaw,
   type IngestedOffer,
@@ -186,6 +189,42 @@ interface JamcoopBudget {
   deadline: number;
 }
 
+// --- Mirlo budgets -------------------------------------------------------------
+//
+// The cheapest source in the codebase: `/v1/artists/{slug}` returns the whole discography **and
+// its prices** in one request, so an artist costs exactly one fetch and there is no detail pass.
+// Compare Bandcamp (grid + one page per release) and Jam.coop (artist page + one per album).
+//
+// Mirlo granted Unstream permission to use this endpoint and issued an API key (2026-08-05), so
+// the `Disallow: /v1/` in their robots.txt is superseded by direct consent rather than ignored.
+// The pacing below is still deliberately gentle: a small co-op's own servers, and the recatalog
+// sweep runs four times a day.
+
+const DELAY_BETWEEN_MIRLO_FETCHES_MS = 1_000;
+
+/**
+ * Invocation-wide. One request per artist, so this is effectively "how many Mirlo artists can one
+ * batch cover" — set to the batch ceiling (`MAX_ARTISTS_PER_RUN`) since no artist needs a second.
+ */
+const MAX_MIRLO_FETCHES_PER_RUN = 25;
+
+interface MirloBudget {
+  fetchesLeft: number;
+  deadline: number;
+}
+
+/**
+ * Sent as `mirlo-api-key` when set.
+ *
+ * Verified live 2026-08-05: this endpoint returns byte-identical responses with the key, with a
+ * bearer token, and with no auth at all — so the key is **not** what grants access, and an absent
+ * key must not disable the pass. It is sent because Mirlo issued it for this use and it lets them
+ * attribute our traffic; if they later gate or rate-limit by key, we are already correct.
+ */
+const MIRLO_API_KEY = process.env.MIRLO_API_KEY;
+
+const MIRLO_USER_AGENT = 'Unstream/1.0 (https://unstream.stream - ethical music finder)';
+
 export async function handler(event: {
   httpMethod?: string;
   headers?: Record<string, string | undefined>;
@@ -244,6 +283,10 @@ export async function handler(event: {
     fetchesLeft: MAX_JAMCOOP_FETCHES_PER_RUN,
     deadline: Date.now() + ENRICHMENT_CUTOFF_MS,
   };
+  const mirloBudget: MirloBudget = {
+    fetchesLeft: MAX_MIRLO_FETCHES_PER_RUN,
+    deadline: Date.now() + ENRICHMENT_CUTOFF_MS,
+  };
   const enrichmentDeadline = Date.now() + ENRICHMENT_CUTOFF_MS;
 
   for (const [index, artistId] of artistIds.entries()) {
@@ -263,6 +306,7 @@ export async function handler(event: {
         discogsBudget,
         faircampBudget,
         jamcoopBudget,
+        mirloBudget,
         enrichmentDeadline
       );
       if (found === null) {
@@ -297,6 +341,7 @@ async function catalogArtist(
   discogsBudget: DiscogsBudget,
   faircampBudget: FaircampBudget,
   jamcoopBudget: JamcoopBudget,
+  mirloBudget: MirloBudget,
   enrichmentDeadline: number
 ): Promise<number | null> {
   const artist = await getArtistForCatalog(artistId);
@@ -304,8 +349,18 @@ async function catalogArtist(
     await recordCatalogOutcome(artistId, { error: 'artist not found' });
     return null;
   }
-  if (!artist.bandcampUrl && !artist.discogsUrl && !artist.faircampUrl && !artist.jamcoopUrl) {
-    await recordCatalogOutcome(artistId, { error: 'no bandcamp, discogs, faircamp, or jam.coop link stored' });
+  // Keep this list identical to CATALOGUEABLE_PLATFORMS in db.ts — the sweep picks its pool from
+  // that constant, and an artist it sweeps but this rejects is recorded as a failure.
+  if (
+    !artist.bandcampUrl &&
+    !artist.discogsUrl &&
+    !artist.faircampUrl &&
+    !artist.jamcoopUrl &&
+    !artist.mirloUrl
+  ) {
+    await recordCatalogOutcome(artistId, {
+      error: 'no bandcamp, discogs, faircamp, jam.coop, or mirlo link stored',
+    });
     return null;
   }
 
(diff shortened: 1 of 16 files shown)


api/functions/catalog-artist-background.ts
brandonlucasgreen, line 1: **needs-manual-verification — bypasses the `isUrlHostnameAllowed()` allowlist convention**

`catalogMirlo` fetches with `globalThis.fetch` instead of `safeFetch`, same as the existing Discogs/MusicBrainz passes in this file. CLAUDE.md states "SSRF protection is mandatory for any code that fetches an external URL... don't add a raw `fetch(userUrl)`." The `apiUrl` here isn't literally `userUrl` — it's a hardcoded `https://api.mirlo.space` host with only a URL-encoded path segment coming from `mirloArtistSlug(storedUrl)`, and that function already rejects any `storedUrl` whose host isn't `mirlo.space`/`*.mirlo.space` before a slug is ever derived — so I don't see an actual SSRF path here (the fetch target's host is never attacker/artist-controlled). Flagging as needs-manual-verification only because it's a second precedent-following deviation from the stated "always route through the allowlist" rule rather than an addition to `ALLOWED_OUTBOUND_HOSTNAMES`, worth a conscious call rather than silent precedent-copying.

Related, lower severity: this fetch (like Discogs/MusicBrainz) has no per-request timeout/AbortController. `budget.deadline` is only checked *before* starting a fetch, not enforced during it, so one hanging `api.mirlo.space` response could run past `enrichmentDeadline` for the whole batch. Pre-existing pattern, not introduced by this PR, but worth a shared fix if it's ever going to be addressed.

@brandonlucasgreen

---

brandonlucasgreen (commented):
@brandonlucasgreen Reviewed #415 (Mirlo REST ingest, both commits). Scope: auth/authz, input validation, secret handling, injection, unsafe defaults, data exposure, plus regression risk (changed assumptions, edge cases, off-by-one, state transitions, API contract drift). This was checked against the actual branch (`origin/mirlo-rest-releases`), not just the diff.

## Security

- **Auth/authorization — confirmed, unchanged.** `catalog-artist-background.ts` still gates on the existing shared-secret `isInternalRequest` check; the Mirlo path adds no new entry point, it's invoked from inside the already-authenticated handler.
- **SSRF — confirmed safe, one item flagged inline as needs-manual-verification.** `catalogMirlo` fetches a hardcoded `https://api.mirlo.space/v1/artists/{slug}` host via `globalThis.fetch` rather than `safeFetch`/`isUrlHostnameAllowed()`. The only variable is a path segment from `mirloArtistSlug(storedUrl)`, which itself rejects any URL whose host isn't `mirlo.space`/`*.mirlo.space` before deriving a slug, and the segment is `encodeURIComponent`-encoded before interpolation. No attacker/artist-controlled host reaches `fetch`. It matches the existing Discogs/MusicBrainz precedent in the same file. See inline comment for the "worth a conscious call" nuance re: CLAUDE.md's allowlist-first rule.
- **Secret handling — confirmed clean.** `MIRLO_API_KEY` is read from env, sent only as the `mirlo-api-key` header, never logged (including in `scripts/ingest-mirlo.ts`, which only logs *whether* it's set). Absence is a documented no-op, not a failure — the PR verified live that the endpoint is unauthenticated regardless.
- **Injection — confirmed none.** All new Supabase access (`persistMirloReleases`, `getArtistForCatalog`) goes through the parameterized query builder (`.eq()`, `.insert()`, `.update()`); no raw SQL, no string-built queries.
- **Unsafe defaults — confirmed correct, this is the PR's main hazard and it's handled well.** `minPrice: null` (no price set) is kept distinct from `minPrice: 0` (name-your-price) rather than being coerced to a `0`/free offer — verified by dedicated tests. Currency casing (`gbp`/`GBP`) is normalized; a missing/absent currency is left `null` rather than defaulted to `USD` (which would have mislabeled a price in another currency). `platformPercent` is deliberately never used for payout math, with the outlier data documented in a comment. Drafts (`mi-temp-slug-…`, empty title) are excluded even though Mirlo's own visibility flags (`isPublic`, `hideFromSearch`, `isGettable`) don't mark them — this is the kind of "trust upstream flags" bug that would otherwise slip through, and it's covered by both an observed-shape test and a constructed one for the untested branch.
- **Data exposure — confirmed none.** No new PII surface; artwork/title continue through the existing `escapeHtml()` path in the edge renderers, unchanged by this diff.

## Regression

- **Changed assumptions — confirmed handled correctly, and specifically tested for.** `CATALOGUEABLE_PLATFORMS` (db.ts) and `catalogArtist`'s "nothing to fetch" condition (catalog-artist-background.ts) are two independently-maintained lists that must agree, per the file's own comment — this PR keeps them in sync and adds the negative-case test (`treats a platform outside the catalogueable list as not crawlable`) that would have caught a future drift between them.
- **Edge cases / null-empty behavior — confirmed correct.** `releaseTypeLabel([])` (no platforms) correctly returns `''` rather than falling through to "Digital" — "no evidence" is kept distinct from "evidence of digital," mirroring the ingest layer's own null-vs-zero discipline. Mixed-platform releases (e.g. Mirlo + Bandcamp) correctly suppress the "Digital" label since Bandcamp isn't `digitalOnly`.
- **State transitions — confirmed correct.** `persistReleaseDetail` is called with `status: null` from the Mirlo pass specifically because status/date were already written in the same call from the same response; `persistReleaseDetail`'s existing `detail.status !== null` guard means this correctly skips re-patching status/date instead of clobbering a `curated`/already-correct value with `null`.
- **API contract drift — needs-manual-verification, not blocking.** The whole PR rests on `/v1/artists/{slug}` continuing to return the `{ result: { urlSlug, trackGroups } }` shape observed 2026-08-05; `ingestMirloArtist` correctly returns `null` (not `[]`) for any unrecognized shape so a future contract change fails loud via Sentry rather than silently recording "no releases," which is the right posture — just noting it's inherently unverifiable from static review alone.

## Not blocking, from the PR description

Two items the author already flagged and I'd just second: `MIRLO_API_KEY` needs to be set in Netlify (Functions scope) before/at merge, and merging starts live cataloging for ~66 Mirlo-linked artists immediately (no feature flag) — both are called out already, not new findings.

No confirmed security issues. One inline item worth a conscious decision rather than silent precedent-following; everything else checks out.

---

PR #416, opened 2026-08-05
Fix the intermittent failure in AuthContext-save-reporting.test.tsx

brandonlucasgreen:
## What was wrong

`apps/web/tests/unit/AuthContext-save-reporting.test.tsx` failed intermittently — seen once in a `npm run test:unit` run, then unreproducible in isolation. Because the build gate blocks Netlify deploys, a flake here makes a real regression and a spurious failure indistinguishable, and the usual response ("re-run it") is exactly what would let a real one through.

## The race

`mountSignedIn` waited for the DOM to read `signed-in`, but every test then drives `held.current` — which is written by a **passive effect**. Those are two different events:

- the text lands when React commits;
- the effect flushes on a later macrotask;
- `waitFor` can resolve on the commit's MutationObserver microtask, i.e. before the effect.

`api()` then handed back the pre-session context snapshot, whose `saveArtist` hits `if (!session) return` ([AuthContext.tsx:185](https://github.com/brandonlucasgreen/unstream/blob/main/apps/web/src/contexts/AuthContext.tsx#L185)). The call under test silently did nothing, so nothing was reported — surfacing either as `captureMessage` "Number of calls: 0" or as a `TypeError` on `mock.calls[0][1]`.

No product code is at fault; the assertions were correct, the setup wasn't.

## Reproduction

Repetition doesn't surface it — **250 unloaded iterations and 10 clean full-suite runs all passed.** CPU contention does: 10 background busy-loops plus 120 runs of the single file produced **4 failures, across three different tests** (not just the one originally reported).

An instrumented probe pinned the mechanism: of 60 iterations, the single one where `held.current.session` was still `null` after the wait was exactly the one where nothing was reported (`{"iterations":60,"staleHeld":1,"noReport":1}`).

## The fix

- `mountSignedIn` waits on `held.current?.session` — the value the test is about to use — instead of a rendered proxy for it. `Probe` returns `null` since nothing reads its output any more. This costs ~50ms per test (interval polling rather than a DOM mutation) and is tolerant of any session-arrival timing.
- `reportOptions(message)` looks the report up by message rather than reading `mock.calls[0][1]`. Two of the three reproduced failures were opaque `TypeError: Cannot read properties of undefined` from that index; a missing report now names itself and lists what *was* reported.

## Verification

- **0 failures in 250 runs** under the identical load that produced 4/120 before the fix.
- `test:unit` 636 passed · `test:api` 923 passed · root and `apps/web` `tsc -b` clean · lint unchanged at the known 71-problem baseline.

## For reviewers

Two things I found but deliberately left alone, to keep this diff to the flake:

- **One test was passing for the wrong reason.** "rolls the optimistic save back" only asserts `isArtistSaved` is `false` — which a no-op save also satisfies, so it went green under the race while the code under test never ran. The fix protects it, but it still has no teeth of its own against a regression in the rollback path.
- **The fetch spy is never restored.** `vi.clearAllMocks()` clears calls but not spies, and each `mockResolvedValue` hands out a *single* `Response`, whose body can only be read once. Nothing triggers a second fetch per test today, so it isn't firing — but it's the next flake in the same family if one ever does. Happy to harden it here or in a follow-up.

### DIFF
--- a/apps/web/tests/unit/AuthContext-save-reporting.test.tsx
+++ b/apps/web/tests/unit/AuthContext-save-reporting.test.tsx
@@ -10,7 +10,7 @@
 
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 import { useEffect } from 'react';
-import { render, screen, waitFor, cleanup, act } from '@testing-library/react';
+import { render, waitFor, cleanup, act } from '@testing-library/react';
 import { AuthProvider, useAuth } from 'src/contexts/AuthContext';
 
 const mocks = vi.hoisted(() => ({
@@ -54,7 +54,7 @@ const held: { current: ReturnType<typeof useAuth> | null } = { current: null };
 function Probe() {
   const auth = useAuth();
   useEffect(() => { held.current = auth; });
-  return <div data-testid="probe">{auth.session ? 'signed-in' : 'signed-out'}</div>;
+  return null;
 }
 
 function api(): ReturnType<typeof useAuth> {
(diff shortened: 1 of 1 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Automated security + regression review.

**Scope:** test-only change (`apps/web/tests/unit/AuthContext-save-reporting.test.tsx`), no production code touched.

### Security
No findings. No auth/authz, input-validation, secret-handling, injection, unsafe-default, or data-exposure surface in this diff — it's confined to a test file.

### Regression
No findings — **needs-manual-verification cleared**: the `semantic-revert-check` bot flagged this PR for deleting 15 lines added in #402. I fetched both branches and diffed the actual test bodies: the "deleted" lines are the old, race-prone versions of the same five assertions #402 added (status_code / serverError checks for the 400, 401, 502, remove-artist, and dashboard-load cases). All five are still present and unweakened — they're now looked up by `reportOptions(message)` instead of a positional `reportOptions()` (`mock.calls[0]`), which is a strict improvement, not a coverage loss:
- `beforeEach` calls `vi.clearAllMocks()`, so for the current single-report-per-test cases the old index-0 lookup and the new keyed lookup return identical data — behavior is unchanged.
- The keyed lookup only diverges (for the better) if a test ever triggers more than one `captureMessage` call: previously it would silently grab call 0 regardless of which report the test claimed to check; now it throws a clear error naming what was actually reported if the expected message is missing.
- The root-cause fix itself (`api()`/`mountSignedIn` waiting on `held.current?.session` — written by the mounted `Probe`'s passive effect — instead of `waitFor` on rendered DOM text) is a legitimate fix for a real race: React can commit the "signed-in" text before the effect that populates `held.current` has flushed, so the old test could read a stale pre-session snapshot whose `saveArtist` short-circuits on `if (!session) return`.

This is a false-positive on the revert-check bot, not a regression — verdict: **confirmed false positive**, safe to ignore per-PR.

No other regression concerns (no edge-case, state-transition, or API-contract changes — the diff doesn't touch `AuthContext` itself, only the test harness around it).

---

PR #418, opened 2026-08-07
Resolve retired artist slugs in the release JSON and artist feed

brandonlucasgreen:
Two small, unrelated fixes.

## 1. `release-detail.ts` didn't resolve slug aliases

`GET /api/release/{artist}/{release}` — the JSON the Mac app's `ReleaseGuideView` and the extension's `lib/release-guide.js` render the buying guide from — looked the artist up with a plain `.eq('slug', …)`. `artist_slug_aliases` holds 44 rows in production, most minted by the accent-folding reslug in #410 (`beyonc` → `beyonce`), so a release URL built before that 404'd in the native "Where to buy" panel while `/api/artist` resolved the same alias happily.

It now reads the alias table on a miss, the way [artist-lookup.ts](api/functions/artist-lookup.ts:31) does:

- **Only on a miss**, so a live slug always wins and the hot path pays nothing.
- **Never after a failed lookup.** A Supabase outage is not evidence that a slug was retired; retrying through the alias table would spend a second query to reach the same 503, and the 503 is the thing that has to survive.
- `pageUrl` and the CDN cache tag are built from the **canonical** slug, so a client's "open the full guide" lands on a URL that renders and one purge clears both entries.

### The two neighbours I was asked to check

**`getReleasesForAlerts` — no change, and not for symmetry's sake.** It looks up by `artistSlug(artistName)`, derived from a *name* rather than read off a URL, and `artistSlug` folds accents ([db.ts:60](api/functions/db.ts:60)). It re-derives the current canonical form on every call, so it never sees a retired slug. Changing it would add a query to the alert path for a case that can't arise.

**`/a/{artist}/releases.{ics,xml}` — fixed.** Same gap, worse failure mode: a calendar client keeps fetching whatever URL it was handed, forever, and never reports a 404 to anybody. An artist re-slugged after somebody subscribed silently lost them. Self-link and feed id are canonical too, so a reslug doesn't split subscribers across two feed identities.

### One correction to the premise

The brief said an aliased release URL "opens fine in a browser." It doesn't. `api/edge/release-page.ts` doesn't resolve aliases either — on an unknown artist slug it `context.next()`s to `artist-page-static`, which hands non-crawlers straight to the SPA, and the SPA has **no route** for a two-segment `/a/` path ([main.tsx:79](apps/web/src/main.tsx:79) is `/a/:slug` only). So `/a/{retired-slug}/{release}` renders nothing in a browser today.

That doesn't change the fix — resolving here is correct and matches every other slug-serving endpoint — but the JSON endpoint is now *ahead* of the HTML page rather than catching up to it. Recorded in the file header. Fixing the page needs the Deno-duplicated alias query plus a decision about 301-vs-render, which is more than this PR should carry.

## 2. CLAUDE.md's catalogue-platform list was stale

The "Keeping catalogues fresh" section described the sweep pool as bandcamp/discogs/faircamp/jamcoop in both sentences — including the one warning not to let that list drift from `catalogArtist`'s check. #415 added `'mirlo'` to `CATALOGUEABLE_PLATFORMS` and to `catalogArtist` together, so only the documentation drifted. While there, sanity-checked the rest of the section against the code:

- "run daily" → **every six hours** (`cron: '0 1,7,13,19 * * *'`, 25 artists a run, 100 a day). The Deployment section already said six hours; the two disagreed.
- "No rate limiting of its own" → `getStaleCatalogCandidates` *does* drop cooldown artists up front, reading the same `RECATALOG_COOLDOWN_HOURS`; `claimArtistForCatalog` remains the authority. Said so.
- The population figures are now explicitly dated measurements (2026-08-02: 2,564 catalogue-able / 9 saved; 2026-08-07: 5,977 releases, 803 catalogued artists, 36 live `saved_artists` rows) with a note to re-measure rather than reason from them.
- Fixed a "see … below" pointing at a section that is above.

Ordering, batch size and the `getStaleCatalogCandidates` priority description all match the code as written.

## Testing

- `api/functions/__tests__/release-detail.test.ts` — 7 new cases: alias resolution, canonical `pageUrl`/`Cache-Tag`, no alias read on the hit path, 404 when neither live nor aliased, 404 when the alias resolves but the release is gone, no alias read after a failed lookup, and no second query when an alias points at the slug already tried.
- `api/functions/__tests__/feed-releases.test.ts` — 3 new cases for the artist feed.
- **Teeth proven:** with the alias branch removed from both files, 4 of the new tests fail (the rest are guardrails that should pass either way).
- `npm run test:api` 968 passed · `npm run typecheck:api` clean (both files are in the include list) · `npm run lint` at the known 71-problem baseline, unchanged.

⚠️ `npm run test:unit` has **one pre-existing failure** unrelated to this branch: `RichArtistProfile.test.tsx > renders the payout percent on each pill` expects `80-85%` and today (2026-08-07) is a Bandcamp Friday, so the component renders `~97%`. Verified failing identically on `main`. It will pass again tomorrow, and it will fail again on the next Bandcamp Friday — worth a separate fix.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -219,24 +219,34 @@ Sentry, since it is otherwise recorded as a perfectly ordinary success.
 Every other catalog trigger is demand-driven — a save, a search, an artist's own button, the admin
 command — and `check-releases` only *reads* the catalogue. So without a scheduled refresh an artist
 who is saved but never searched gets catalogued once and their release alerts quietly stop. That
-is what `api/functions/recatalog-sweep.ts` fixes, run daily by `.github/workflows/recatalog-sweep.yml`.
+is what `api/functions/recatalog-sweep.ts` fixes, run every six hours by
+`.github/workflows/recatalog-sweep.yml` (25 artists per run, so 100 a day).
 
-The pool is **every artist with a bandcamp, discogs, faircamp or jamcoop link** — measured
-2026-08-02: 2,564 of them, against 9 artists saved by anybody, so a saved-only sweep sat idle
-almost every run. Alerts aren't the only consumer either: `/a/:slug` renders a release list for
-any catalogued artist, and those pages exist because somebody *searched*. Keep that platform list
-identical to `catalogArtist`'s "no bandcamp, discogs, faircamp, or jam.coop link stored" check —
-an artist with only an official site is recorded as a *failure*, so sweeping them poisons
-`consecutive_failures`.
+The pool is **every artist with a bandcamp, discogs, faircamp, jam.coop or mirlo link** —
+`CATALOGUEABLE_PLATFORMS` in `db.ts` is the list. Keep it identical to `catalogArtist`'s "no
+bandcamp, discogs, faircamp, jam.coop, or mirlo link stored" check — an artist with only an
+official site is recorded as a *failure*, so sweeping them poisons `consecutive_failures`. The two
+have to change together: Mirlo was added to both in the same PR (#415) for exactly that reason.
+
+Why the pool isn't saved-only, which is how it shipped: measured 2026-08-02 there were 2,564
+artists with a catalogue-able link against 9 saved by anybody, so the sweep's whole universe fit
+in one batch and it sat idle almost every run. Alerts aren't the only consumer either — `/a/:slug`
+renders a release list for any catalogued artist, and those pages exist because somebody
+*searched*. The ratio is what matters and it hasn't moved: as of 2026-08-07 there are 5,977
+releases across 803 catalogued artists, and 36 live `saved_artists` rows in total. Treat every
+number in this paragraph as a dated measurement, not a current count — re-measure before
+reasoning from one.
 
 `getStaleCatalogCandidates` in `db.ts` orders them: saved first (an alert is a promise to a
 person, and there are few enough that they never starve behind the backfill), then never
 catalogued, then oldest `last_attempted_at`, then savers as a pure tiebreak. It asks
-`requestArtistCatalog` for up to 25 under the `scheduled` trigger. No rate limiting of its own —
-the 7-day cooldown and the hourly cap in `claimArtistForCatalog` still decide — so running it
-twice is a no-op. It returns a real summary rather than 202, and any refusal is a non-2xx that
-fails the workflow: a scheduled job that reports success while doing nothing is the same silent
-failure it was built to fix.
+`requestArtistCatalog` for up to 25 under the `scheduled` trigger. It adds no limit of its own: it
+drops artists inside the 7-day re-catalogue cooldown up front, reading the same
+`RECATALOG_COOLDOWN_HOURS` constant so a bounded batch isn't spent on artists that would be
+refused a moment later, but `claimArtistForCatalog` — that cooldown plus the per-trigger hourly cap
+— stays the authority. So running the sweep twice is a no-op. It returns a real summary rather than
+202, and any refusal is a non-2xx that fails the workflow: a scheduled job that reports success
+while doing nothing is the same silent failure it was built to fix.
 
 **Paging, not `.limit()`.** PostgREST caps every response at 1,000 rows whatever limit you ask
 for, and truncates *silently*. A single `.select()` over `artist_links` returns 1,000 of ~3,900
@@ -369,7 +379,7 @@ GitHub Actions (`.github/workflows/`):
 - `schedule-social-posts.yml` — weekly (Mondays) social post generation, committed back to the repo.
 - `semantic-revert-check.yml` — runs `scripts/semantic-revert-check.py` on every PR to flag changes that quietly undo earlier fixes. If it flags your PR, take it seriously: the bug loops it was built for are described in `docs/retros/UNS-100-bifurcation-retro.md`.
 - `upstash-keepalive.yml` — twice weekly, writes one Redis key so the free-tier database isn't reaped for inactivity (which silently killed caching and rate limiting once).
-- `recatalog-sweep.yml` — every 6 hours, POSTs `/.netlify/functions/recatalog-sweep` so release catalogues get built and stay fresh. See "Keeping catalogues fresh" below.
+- `recatalog-sweep.yml` — every 6 hours, POSTs `/.netlify/functions/recatalog-sweep` so release catalogues get built and stay fresh. See "Keeping catalogues fresh" above.
 
 ## Engineering principles
 
--- a/api/functions/__tests__/feed-releases.test.ts
+++ b/api/functions/__tests__/feed-releases.test.ts
@@ -14,6 +14,7 @@ const mocks = vi.hoisted(() => ({
   getFeedReleasesForUser: vi.fn(),
   getFeedReleasesForArtist: vi.fn(),
   getFeedReleasesForHandle: vi.fn(),
+  resolveArtistSlugAlias: vi.fn(),
 }));
 
 vi.mock('../ratelimit', () => ({
@@ -25,6 +26,7 @@ vi.mock('../db', () => ({
   getFeedReleasesForUser: mocks.getFeedReleasesForUser,
   getFeedReleasesForArtist: mocks.getFeedReleasesForArtist,
   getFeedReleasesForHandle: mocks.getFeedReleasesForHandle,
+  resolveArtistSlugAlias: mocks.resolveArtistSlugAlias,
 }));
 
 import { handler, parsePath, requestPath } from '../feed-releases';
@@ -70,6 +72,7 @@ beforeEach(() => {
   mocks.getFeedReleasesForUser.mockResolvedValue([]);
   mocks.getFeedReleasesForArtist.mockResolvedValue(null);
   mocks.getFeedReleasesForHandle.mockResolvedValue(null);
+  mocks.resolveArtistSlugAlias.mockResolvedValue(null);
 });
 
 describe('requestPath', () => {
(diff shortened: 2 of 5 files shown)


brandonlucasgreen (commented):

**Security**
- Auth/authorization: unchanged; both endpoints keep their existing rate limiting (`checkRateLimit`, 'lenient') and CORS posture.
- Input validation / injection: `resolveArtistSlugAlias` validates the slug against `/^[A-Za-z0-9-]+$/` before it's interpolated into the PostgREST filter, mirroring `getArtistBySlug`'s existing guard — same pattern, no gap. `release-detail.ts` still checks `SLUG_PATTERN` before the alias lookup even runs. Supabase query builder is used throughout (`.eq(...)`), no raw SQL. No injection risk found.
- Secret handling: n/a, nothing touched here.
- Unsafe defaults / data exposure: the alias read is correctly gated to run only on a miss and only when `!failed` (not after a lookup failure) — so a Supabase outage still returns a 503 rather than spending a second query and risking a false 404 that the CDN would then cache. This matches the file's own "never cache uncertainty" comment and is exercised by the test suite (`does not consult the alias table when the lookup itself failed`).

**Regression**
- Edge cases look covered: alias resolves but release is gone (still one 404, not a third query), alias points at the already-tried slug (no redundant query), live slug always wins over an alias. All backed by explicit tests.
- `pageUrl`/`Cache-Tag` correctly built from the canonical slug rather than the requested one, avoiding a client being handed a link that 404s on the (unmodified) HTML page — the PR is explicit that `release-page.ts` doesn't resolve aliases yet, which is accurately reflected in the code comments rather than papered over.
- CLAUDE.md changes are documentation-only; the corrected figures/schedule match the code as described.

---

brandonlucasgreen (commented):

**Security:**
- No SQL injection, cache-tag injection, or open-redirect risk: `artist`/`release` path params are validated against `SLUG_PATTERN` before any DB call, `resolveArtistSlugAlias` independently re-validates, all Supabase reads use `.eq()` (parameterized by PostgREST), and `pageUrl` is hardcoded to `https://unstream.stream/a/...` with only regex-validated slug segments interpolated. — *confirmed*
- No new authorization/data-exposure issue: both endpoints were already fully public before this PR, and `resolveArtistSlugAlias`'s `artist_id` join means an alias can only ever resolve to *that same artist's* current live slug — it can't leak or shadow a different artist's data. — *confirmed*
- **[low-medium, confirmed]** `resolveArtistSlugAlias` (`api/functions/db.ts:3041-3066`) collapses "the alias query itself failed" (transient DB error) into the same `return null` as "no alias exists for this slug." Every other lookup this PR touches (`getReleaseDetail`) explicitly separates a real error from a legitimate miss and reports failures to Sentry; this helper doesn't — a Supabase hiccup during alias resolution silently becomes a plain 404 with no Sentry signal. This is the same principle as CLAUDE.md's "never cache uncertainty" rule, applied to a 404 instead of a cached negative. Mitigating factor: neither new 404 path sets `Netlify-CDN-Cache-Control`, so this is a one-request silent miss, not a cached one — downgraded from "cached uncertainty" to "silent unreported failure" accordingly. The function predates this PR, but the PR adds two new production call sites that inherit the gap.

**Regression:**
- **[low, confirmed]** The PR's stated reason for *not* touching `getReleasesForAlerts` — "`artistSlug(name)` re-derives the current canonical form on every call" — holds for the accent-folding reslug this PR targets, but not universally: `artist-profile.ts`'s self-service `PUT` handler lets a claimed artist set an arbitrary `newSlug` unrelated to `artistSlug(name)` (confirmed via `artist-merge.ts`'s `legacyArtistSlug` logic and named examples — `muz4now`, `boezi`, `courstellation`, `cbertine`). For such an artist, `getReleasesForAlerts` computes a slug that no longer matches the stored row, and the self-service rename path never writes an `artist_slug_aliases` row either (only `reslugArtist()` does) — so aliasing wouldn't fix this case even if applied there. Bounded impact: `check-releases.ts` falls through to a live scrape when this returns null, so the effect is a degraded alert (extra scrape, no upcoming/multi-platform detection) rather than a hard outage, and this is pre-existing behavior unrelated to this diff's actual changes.
- Verified safe by construction: alias resolution always joins through to the artist row's *current* slug (never chains through another alias), so a stale/chained alias pointing at a dead intermediate can't occur, even after a double reslug. — *confirmed*
- Verified correct: `release-detail.ts`'s `if (!detail && !failed)` correctly skips the alias branch on a real lookup failure (503 path), and a failure on the *retry* after alias resolution still correctly surfaces as 503 rather than a false 404. — *confirmed*
- **[info, confirmed]** `feed-releases.ts` passes `route.slug` straight to `getFeedReleasesForArtist`/`resolveArtistSlugAlias` without the cheap `SLUG_PATTERN` pre-check that `release-detail.ts` has — not a security issue (the helpers guard themselves), just a missed cost optimization, and pre-existing rather than introduced by this diff.

CLAUDE.md doc fix (catalogue-platform list, cron cadence, cooldown description) checked against the actual code in `db.ts` and matches — no factual errors introduced.

---

brandonlucasgreen (commented):

**Security**

No issues found. Both Supabase calls use parameterized `.eq()`, not string interpolation. `resolveArtistSlugAlias` regex-guards its input (`^[A-Za-z0-9-]+$`) before touching the alias table, and `release-detail.ts`'s handler already validates both path segments against `SLUG_PATTERN` before any of this code runs. Confirmed via the `artist_slug_aliases` migration that an alias row maps directly to the artist's current `slug` (not a chain), so a second reslug just adds another alias row and still resolves in one hop — no multi-hop gap. `artists` has no hidden/unpublished flag and alias rows FK-cascade-delete with the artist, so nothing is exposed that wasn't already public via the live slug. Building `pageUrl`/Cache-Tag from the canonical (DB-sourced) slug isn't a poisoning vector — it's not client-echoed input.

**Regression**

`api/functions/feed-releases.ts:221-233`, backed by `getFeedReleasesForArtist` in `api/functions/db.ts` (~4334-4372), `confirmed` — unlike `getReleaseDetail`/`release-detail.ts` (which correctly track a `failed` flag and gate `if (!detail && !failed)`), `getFeedReleasesForArtist` conflates "artist not found" and "Supabase query errored" into the same plain `null` return, so `feed-releases.ts`'s new alias fallback can't distinguish them and runs on both. During a Supabase outage, the first lookup errors and returns `null` exactly as it would for a genuinely unknown artist, so the code proceeds to a second failing alias query, then returns a plain 404 to a calendar client instead of any signal that this was transient — the exact "failure indistinguishable from doesn't-exist" pattern CLAUDE.md's "never cache uncertainty" principle calls out (not cached here, but silently misrepresented), and it doubles DB round-trips per request during the outage itself. The PR's stated invariant (b) — "never after a failed lookup" — isn't actually achievable for this file because the data layer it calls doesn't expose the signal needed to implement it, even though the PR correctly implements it in `release-detail.ts`. No test in the diff exercises this path for the feed endpoint; `release-detail.test.ts` has "does not consult the alias table when the lookup itself failed" but `feed-releases.test.ts` has no equivalent.

Verified with no regression found: live-slug-first precedence is correct in both files; the canonical-slug echo in `pageUrl` doesn't break either named consumer (`ReleaseGuideView.swift`, `apps/extension/lib/release-guide.js` both treat it as opaque, never compare against the requested slug); and feed-subscriber identity isn't actually at risk — a URL built on a retired slug 404'd outright before this PR, so there was no working subscription to break, only broken ones being fixed.

---

brandonlucasgreen (commented):
@brandonlucasgreen

Reviewed the diff (`CLAUDE.md`, `api/functions/release-detail.ts`, `api/functions/feed-releases.ts`, and their tests) against the full surrounding files, plus the existing `resolveArtistSlugAlias`/`getArtistBySlug`/`getReleaseDetail` implementations in `db.ts` and the `artist-lookup.ts` precedent this PR mirrors.

## Security

- `resolveArtistSlugAlias` (unmodified, pre-existing) validates the slug against `/^[A-Za-z0-9-]+$/` before it's ever interpolated into a PostgREST filter, and both call sites (`release-detail.ts`, `feed-releases.ts`) only invoke it with a slug that already passed the caller's own gate (`SLUG_PATTERN` in `release-detail.ts`) or is bound by parameter, not string concatenation (Supabase `.eq()` in `feed-releases.ts`'s `getFeedReleasesForArtist`). — confirmed, no injection surface.
- No new table/column is introduced (confirmed via `git diff --stat`; no `supabase/migrations/` file in this PR), so RLS is not applicable here — the backing `artist_slug_aliases` table was created and reviewed in an earlier PR.
- Spoofing check: alias resolution always maps a fixed `alias -> artist_id`, then the live slug re-query is scoped by that same `artist_id`/current slug (`getReleaseDetail`, `getFeedReleasesForArtist`). The live slug is always tried first and always wins (`if (!detail && !failed)` / `if (!artist)` gates), so a retired alias can never shadow a real artist who later claims that slug, and one artist's data can never be served under another still-active artist's slug. — confirmed correct, matches the file's own stated invariant and is covered by the "never re-queries when the live slug answers" / "alias resolves but release is gone" tests.
- No secrets, tokens, or PII touched; Sentry tags in the `failed` branch carry only slugs, consistent with existing usage.

## Regression

- **needs-manual-verification**: `release-detail.ts`'s `Sentry.captureMessage` in the `failed` branch tags `{ artist, release }` using the *original requested* slug, not `artistSlug` (the post-resolution value). In practice this only diverges when the *second* `getReleaseDetail` call (after a successful alias resolution) itself fails — a narrow window — but in that case the Sentry event reports the retired slug rather than the canonical one actually queried when the failure occurred. Low impact (observability only, not user-facing), worth a quick look but not blocking. `api/functions/release-detail.ts` (the `if (failed)` block).
- **plausible**: Both endpoints now do up to 2 DB round trips on any release/artist "not found" response (live-slug lookup + alias lookup) instead of 1, including for a valid live artist with a mistyped release slug (`release-detail.ts`) or any nonexistent artist slug (`feed-releases.ts`). This is an accepted, deliberate trade-off per the PR's own comments ("only after the live slug has had its go") and is mitigated by the existing 'lenient' rate limiter, but it does mildly amplify DB load for scanning/miss-heavy traffic on both endpoints. Not a functional bug.
- **confirmed (non-issue, verified)**: `feed-releases.ts`'s `getFeedReleasesForArtist` collapses a DB error and "artist not found" into the same `null` (no `failed` signal), so on a Supabase outage this PR's new alias-lookup branch also fires and spends an extra query before returning 404 — unlike `release-detail.ts`, which explicitly skips alias resolution on `failed`. Verified this is **not a regression**: `getFeedReleasesForArtist` had no `failed` signal before this PR either (same as `getArtistBySlug`, which `artist-lookup.ts` already treats the same way), so behavior on outage (404 either way) is unchanged — only cost, not correctness, is affected.
- No API response shape/contract changes beyond `pageUrl`/`Cache-Tag`/`selfUrl`/`feedId` now reflecting the canonical slug instead of the requested one when an alias was resolved — this is the intended fix, verified against `release-page.ts` (confirmed it does a plain `.eq('slug', …)` with no alias resolution, matching the PR's own header comment claiming this) so the canonical `pageUrl` this endpoint returns will actually render.
- Test coverage for the new branches is thorough on both files: live-slug hit (no alias query), alias hit + re-query, alias miss (404), alias resolves but release/artist still missing (one extra query, still 404), and the outage case explicitly skipping alias resolution in `release-detail.ts`. `feed-releases.ts`'s tests don't add an explicit outage-path case, but per the point above that's consistent with pre-existing behavior, not a gap introduced here.
- `CLAUDE.md` doc changes (recatalog-sweep cadence/pool/cooldown wording) spot-checked against `api/functions/db.ts` (`RECATALOG_COOLDOWN_HOURS`, `CATALOGUEABLE_PLATFORMS`, `getStaleCatalogCandidates`) — accurate, no drift introduced.

---

PR #419, opened 2026-08-07
Fix the browser extension's release-alert layer

brandonlucasgreen:
`apps/extension/background/service-worker.js` was last modified by a *search* PR that predates the entire Releases feature, and #396 updated only `popup/popup.js`. So the alert layer still spoke the pre-catalog `check-releases` contract while the popup had been rewritten against the new one.

## What was broken

**1. Only one release per artist could ever surface.** The worker read the singular `release`, took the newest, and marked only that one known. By the next check the newest was already known and nothing looked past it — so an artist's *second* record in the window was unreachable forever. Same permanent-loss shape that was fixed on the Mac by `ReleaseAlertManager.selectUnseen`; `selectUnseenReleases` mirrors that rule exactly.

**2. `offerSummary` was a field the popup read and nothing ever wrote.** Because the read had a `|| \`on \${platformDisplayName(...)}\`` fallback, the dead field rendered as plausible content instead of `undefined`, which is why nobody noticed.

**3. `platforms[]` was dropped.** A record on both Bandcamp and Mirlo showed as one platform — the payout comparison, which is the product's whole thesis, invisible at peak purchase intent.

**4. `status` was dropped.** So an alert said `"X" is out now on Bandcamp!` about a record dated next month. That doesn't merely fail, it states something false.

## What changed

- **Every release, deduped once.** `selectUnseenReleases` iterates everything the server sent, dedups by release name *within* an artist (so one record on two platforms is one alert, but a second record is always new), and marks each one known as it accepts it.
- **`platforms`, `status` and `offerSummary` persisted.** The popup row and the notification now read `out now on Bandcamp and Mirlo · from $8 · ≈$6.80 to artist`, and an announced release reads `announced for 1 September on Bandcamp` — word for word what the Mac says.
- **`sinceDays` is sent**, computed from `lastCheckDate` with 7 days of padding and clamped to the server's own 31/365 bounds. A browser closed for six weeks used to come back, take the 31-day default, find nothing, record a fresh check date, and lose the rest of the gap permanently.
- **The three-platform gate is gone.** `check-releases` reads the catalogue first and ignores `platforms` entirely on that path; only the live-scrape fallback needs URLs. Production holds 1,953 Discogs artist links against 1,810 Bandcamp, so skipping anyone without bandcamp/faircamp/mirlo skipped the largest link population over a question one DB read answers. An artist with no scrapeable link now sends an empty object and gets a real answer if they're catalogued (the endpoint 400s only when the catalogue *also* misses).

## One thing found along the way

**Firefox was running no background code at all.** `manifest-firefox.json` declared the background as a classic script, but `service-worker.js` has used ES `import`s since before this change — so the first import was a syntax error and release alerts, the badge and auth were all dead in Firefox. Fixed with `"type": "module"`, and the both-manifests requirement is now noted at the top of the file so the next import doesn't repeat it. It fails silently unless you open the extension console, which is presumably how it survived.

## Testing

Rules live in `lib/release-alerts.js` as pure functions so they're testable without `chrome.*`.

- `extension-release-alerts.test.ts` pins the rules against the same cases `apps/mac/UnstreamTests/ReleaseAlertTests.swift` pins — Swift can't be driven from vitest, so agreeing suites are what stop the clients drifting.
- `extension-release-check.test.ts` drives the **real service worker** end to end under a stubbed `chrome.*`, triggered the way the popup triggers it. All four defects were wiring, not logic, so testing the helpers alone would not have caught any of them.

Each fix was reverted individually to confirm exactly the intended test fails — including replaying the original bug (mark known, then skip the artist if they have any known release), which fails the second-record test and nothing else.

`npm run lint` holds at the known 71-problem baseline. `npm run test:api` is clean (958). `npm run test:unit` has one failure, `RichArtistProfile > renders the payout percent on each pill`, which **fails identically on a clean checkout of `main`** — it's today, 7 August, being a Bandcamp Friday, so the component renders the 93-97% override instead of the hardcoded 80-85%. Unrelated to this PR, and it will break the build one day a month until it's fixed.

I could not load the extension in a live Chrome or Firefox — launching a browser was denied in this session. What is verified instead: both manifests parse and point at files that exist, and the service worker's whole module graph loads and executes its real check path under the stubbed `chrome.*` in the integration test.

Extension bumped to 2.7.0 in both manifests.

### DIFF
--- a/apps/extension/background/service-worker.js
+++ b/apps/extension/background/service-worker.js
@@ -1,9 +1,22 @@
 // Unstream Chrome Extension - Service Worker
 // Handles API calls, caching, badge updates, release alerts, and auth
+//
+// This file uses ES module imports, so **both** manifests must declare it as a module —
+// `"type": "module"` alongside `service_worker` in manifest.json and alongside `scripts` in
+// manifest-firefox.json. Firefox's manifest was missing it, which meant the first `import` was a
+// syntax error and none of this ran there at all: no release alerts, no badge, no auth. It fails
+// silently unless you open the extension's console, so check both files when adding an import.
 
 import { ALLOWED_RELEASE_DOMAINS } from '../lib/constants.js';
 import { getStoredSession, handleMagicLinkCallback, getAccessToken, signOut } from '../lib/supabase.js';
 import { reconcileCustomSites } from '../lib/custom-sites.js';
+import {
+  releasesFromResponse,
+  releaseNotificationBody,
+  releaseNotificationTitle,
+  selectUnseenReleases,
+  sinceDaysForCheck,
+} from '../lib/release-alerts.js';
 
 const API_BASE = 'https://unstream.stream/api';
 const CACHE_TTL = 5 * 60 * 1000; // 5 minutes
@@ -546,62 +559,41 @@ async function checkForNewReleases() {
 
   // Get current state
   const { releaseCheckState = { knownReleases: {}, newReleases: [] } } = await chrome.storage.local.get('releaseCheckState');
+  if (!releaseCheckState.knownReleases) releaseCheckState.knownReleases = {};
+
+  // How far back to ask. Left unsent, the server looks back 31 days, so a browser that was closed
+  // for six weeks came back, found nothing older than that, and recorded a fresh check date —
+  // losing the gap permanently.
+  const sinceDays = sinceDaysForCheck(releaseCheckState.lastCheckDate);
 
   const foundNewReleases = [];
 
   for (const artistName of artistNames) {
     const artistData = savedArtistsData[artistName];
-    if (!artistData || !artistData.platforms) continue;
+    if (!artistData) continue;
 
-    // Build platforms dict
+    // Every link we hold for this artist, keyed by source.
+    //
+    // No filtering, and no skipping an artist who has none: the endpoint reads the catalogue
+    // first and ignores these entirely on that path — only its live-scrape fallback uses them,
+    // and only the sources it can scrape. Skipping anyone without a Bandcamp, Faircamp or Mirlo
+    // link meant never asking about the largest link population we have (there are more Discogs
+    // artist links in production than Bandcamp ones) over a question one database read answers.
     const platforms = {};
-    for (const p of artistData.platforms) {
-      if (['bandcamp', 'faircamp', 'mirlo'].includes(p.sourceId)) {
-        platforms[p.sourceId] = p.url;
-      }
+    for (const p of artistData.platforms || []) {
+      if (p && p.sourceId && p.url) platforms[p.sourceId] = p.url;
     }
-
-    if (Object.keys(platforms).length === 0) continue;
 
     // Call release check API
     try {
-      const result = await checkReleaseAPI(artistName, platforms);
+      const result = await checkReleaseAPI(artistName, platforms, sinceDays);
 
-      if (result && result.release) {
-        const release = result.release;
-        const key = artistName.toLowerCase();
-
-        // Check if already known (by release name across all platforms)
-        const knownReleases = releaseCheckState.knownReleases[key] || [];
-        const alreadyKnown = knownReleases.some(
-          kr => kr.releaseName.toLowerCase() === release.releaseName.toLowerCase()
-        );
-
-        if (!alreadyKnown) {
-          // New release found!
-          const newRelease = {
-            id: crypto.randomUUID(),
-            artistName: artistName,
-            releaseName: release.releaseName,
-            releaseDate: release.releaseDate,
-            releaseUrl: release.releaseUrl,
-            platform: release.platform,
-            detectedAt: new Date().toISOString(),
-          };
-
-          foundNewReleases.push(newRelease);
-
-          // Add to known releases
-          if (!releaseCheckState.knownReleases[key]) {
-            releaseCheckState.knownReleases[key] = [];
-          }
-          releaseCheckState.knownReleases[key].push({
-            releaseName: release.releaseName,
-            releaseDate: release.releaseDate,
-            platform: release.platform,
-          });
-        }
-      }
+      // Every release in the window, not just the newest — an artist who put out two records
+      // since the last check produces two alerts. selectUnseenReleases marks each one known as
+      // it accepts it, and is the only dedup point.
+      foundNewReleases.push(
+        ...selectUnseenReleases(releasesFromResponse(result), artistName, releaseCheckState.knownReleases)
+      );
     } catch (error) {
       console.error(`Release check failed for ${artistName}:`, error);
     }
@@ -631,13 +623,13 @@ async function checkForNewReleases() {
 }
 
 // Call the release check API
-async function checkReleaseAPI(artistName, platforms) {
+async function checkReleaseAPI(artistName, platforms, sinceDays) {
   const url = `${API_BASE}/check-releases`;
 
   const response = await fetch(url, {
     method: 'POST',
     headers: { 'Content-Type': 'application/json' },
-    body: JSON.stringify({ artistName, platforms }),
+    body: JSON.stringify({ artistName, platforms, sinceDays }),
   });
 
   if (!response.ok) {
@@ -676,14 +668,16 @@ async function dismissRelease(releaseId) {
   return { success: true };
 }
 
-// Send notification for new release
+// Send notification for new release.
+// Title and body come from release-alerts.js so this says the same thing the Mac app does — and,
+// for an announced release, stops claiming a record dated next month is already out.
 async function sendReleaseNotification(release) {
   try {
     await chrome.notifications.create(`release-${release.id}`, {
       type: 'basic',
       iconUrl: chrome.runtime.getURL('icons/icon128.png'),
-      title: `New Release from ${release.artistName}`,
-      message: `"${release.releaseName}" is out now on ${release.platform.charAt(0).toUpperCase() + release.platform.slice(1)}!`,
+      title: releaseNotificationTitle(release),
+      message: releaseNotificationBody(release),
       priority: 2,
     });
   } catch (error) {
--- a/apps/extension/lib/release-alerts.js
+++ b/apps/extension/lib/release-alerts.js
@@ -1,0 +1,213 @@
+// Release alerts: which of the server's results are new, and what an alert actually says.
+//
+// Pure functions, split out of the service worker so they can be tested without chrome.* — and
+// because every rule here already exists on the Mac side. `selectUnseenReleases` mirrors
+// `ReleaseAlertManager.selectUnseen`, `releaseNotificationBody` mirrors its `notificationBody(for:)`,
+// and `formatPlatformList` its `formatPlatformList`. apps/web/tests/unit/extension-release-alerts.test.ts
+// pins the same cases apps/mac/UnstreamTests/ReleaseAlertTests.swift pins, so the two clients
+// can't drift into saying different things about the same release.
+//
+// The server contract these read is documented at the top of api/functions/check-releases.ts:
+// `releases[]` carries every release in the window with every platform it's on, its status, and a
+// price summary. The extension used to read only the singular `release` and keep four of its
+// fields, which is what all of this fixes.
+
+import { SOURCE_CONFIG } from './constants.js';
+
+/** Never ask for a narrower window than the server's own default. */
+export const MIN_SINCE_DAYS = 31;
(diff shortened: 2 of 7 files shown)


apps/extension/background/service-worker.js
```diff
@@ -546,62 +559,41 @@ async function checkForNewReleases() {
 
   // Get current state
   const { releaseCheckState = { knownReleases: {}, newReleases: [] } } = await chrome.storage.local.get('releaseCheckState');
+  if (!releaseCheckState.knownReleases) releaseCheckState.knownReleases = {};
+
+  // How far back to ask. Left unsent, the server looks back 31 days, so a browser that was closed
+  // for six weeks came back, found nothing older than that, and recorded a fresh check date —
+  // losing the gap permanently.
+  const sinceDays = sinceDaysForCheck(releaseCheckState.lastCheckDate);
```
brandonlucasgreen, line 567: **Regression — confirmed.** `sinceDays` here is computed from `releaseCheckState.lastCheckDate`, but further down (`releaseCheckState.lastCheckDate = Date.now();`, just before `scheduleNextReleaseCheck()`) that checkpoint is advanced unconditionally — even when every `checkReleaseAPI(...)` call in the loop above threw.

The per-artist `try/catch` around `checkReleaseAPI` swallows the error and moves to the next artist, so a run where the server or network is unreachable for *every* artist still advances the checkpoint to "now." The next run's `sinceDaysForCheck` then measures elapsed time from that false checkpoint rather than from the last time data was actually fetched. Concretely: if `unstream.stream` is down for several consecutive weekly alarm cycles while the browser stays open, `lastCheckDate` keeps sliding forward every 7 days regardless, so the requested window never grows past `MIN_SINCE_DAYS` (31) — once the true outage exceeds that plus padding, releases from the gap are silently and permanently missed. This is the same "never cache uncertainty" failure class `CLAUDE.md` calls out as the most-repeated bug here, just applied to `lastCheckDate` instead of a probe cache.

Suggest only advancing `lastCheckDate` when at least one `checkReleaseAPI` call in the run actually succeeded (e.g. track a `anySucceeded` flag through the loop), rather than unconditionally after it.

@brandonlucasgreen

---

brandonlucasgreen (commented):

**Security**
- Auth/authorization: n/a (unauthenticated extension → public `check-releases` endpoint, unchanged).
- Input validation / injection: `platforms` now forwards *every* stored `sourceId`/`url` pair instead of filtering to `bandcamp`/`faircamp`/`mirlo`. Checked the receiving endpoint (`api/functions/check-releases.ts`, not touched by this PR) — `mayFetch()` already re-validates every URL server-side via `isSafePublicHostname` + `isUrlHostnameAllowed` + `isStoredArtistLink` (which itself confirms the URL is actually stored for that artist before it's fetched), so widening what the client sends doesn't widen what the server will act on. **No issue** — needs-manual-verification only in the sense that this depends on `check-releases.ts` staying unchanged, which it is here.
- Secret handling: no tokens/keys touched. Notification/popup strings use `textContent`, not `innerHTML` — no new injection surface from artist/release names.
- Unsafe defaults / data exposure: none spotted.

**Regression**
- Confirmed finding left as an inline comment on `service-worker.js` line 567/616: `lastCheckDate` advances unconditionally even when every `checkReleaseAPI` call in the run failed, so a sustained server/network outage while the browser stays open can silently lose releases published during the gap — the same "never cache uncertainty" bug class this PR is otherwise fixing.
- Verified the dedup/window/wording logic (`selectUnseenReleases`, `sinceDaysForCheck`, `releaseSummaryLine`) matches its test coverage and the stated Mac-side parity; no off-by-one or null-handling issues found there.
- Firefox manifest fix (`"type": "module"`) matches the stated root cause and is a plausible, narrowly-scoped fix.

No other findings.

---

brandonlucasgreen (commented):

**Security:** No high/medium issues.
- Release strings (`releaseName`, `artistName`, `releaseSummaryLine()` output — all ultimately scrape-derived) are assigned via `element.textContent` everywhere, never `innerHTML`; `chrome.notifications.create()` also only accepts plain text. No XSS vector despite the data's untrusted origin. — *confirmed*
- `check-releases.ts` remains unauthenticated by design (IP rate-limited only, deliberate wildcard CORS, documented as intentional). Removing the client-side "three-platform gate" doesn't expand the trust boundary — the catalog path ignores `platforms` entirely, and the live-scrape fallback still runs every URL through `mayFetch()`/`isUrlHostnameAllowed()` regardless of what the client sends. — *confirmed*
- `resolveWindowDays` on the server only clamps the *ceiling* (365), not a floor — a caller-supplied `sinceDays` of e.g. 5 is honored as-is. The extension's own `sinceDaysForCheck` always returns ≥31 so this isn't exploitable through the shipped client, but the code comment describing a server-enforced floor overstates what the server actually does. — *confirmed, info*

**Regression:**

1. **[medium, confirmed]** `service-worker.js`'s `checkForNewReleases()` swallows every per-artist `checkReleaseAPI` failure in a local try/catch and continues, then writes `lastCheckDate = Date.now()` unconditionally after the loop — with no check that any request actually succeeded. This is the same "recorded a fresh check date and lost the gap" shape the PR calls out for the *old* code. It's largely self-healing in the common case because `MIN_SINCE_DAYS = 31` in `release-alerts.js` floors the window well above the 7-day check cadence, but the case this doesn't cover is a long-dormancy wake-up (browser closed >24 days) whose *first* catch-up check itself fails — that stamps "now" and discards the real gap permanently.

2. **[low, plausible]** `lastCheckDate` is one global watermark, not per-artist. If one artist's request fails mid-batch while the rest succeed, that artist's effective coverage narrows on the next cycle along with everyone else's, with no per-artist bookkeeping to compensate. Mitigated by the same `MIN_SINCE_DAYS` floor as #1.

3. **[low, needs-manual-verification]** Removing the platform gate means every saved artist now issues a `check-releases` request per cycle instead of just bandcamp/faircamp/mirlo-linked ones. For a user with a large saved-artist list this could approach the shared 5000/day per-IP budget faster, causing some in-batch requests to 429 — which feeds directly into finding #1 (silently swallowed, watermark still advances). Can't verify without live traffic data; flagging for awareness given the 7-day cadence makes it unlikely at realistic list sizes.

Verified clean, no issues found: `selectUnseenReleases` dedup — a release is added to `known` and `fresh` in the same iteration, no path marks-known-without-notifying; the `platforms[]`/`status`/`offerSummary` field names match the live `check-releases.ts` response shape exactly; the Firefox `"type": "module"` manifest fix is correct and matches Chrome's existing declaration with no privilege change.

---

brandonlucasgreen (commented):

**Security**

No issues found. Notification/popup text is written via `.textContent` and `chrome.notifications.create({title, message})` — neither renders HTML, so untrusted release titles/artist names from the server can't inject markup. Auth (`getStoredSession`/`getAccessToken`/`signOut`) is untouched by this diff. `"type": "module"` on the Firefox manifest only changes script-loading semantics (enables `import`, implies strict mode) — no change to permissions, host access, or CSP. No new `chrome.storage.local` keys are exposed to content scripts.

**Regression**

1. `apps/extension/lib/release-alerts.js` (`selectUnseenReleases`) + `service-worker.js` (`checkForNewReleases`), `plausible` — `knownReleases[artist]` now accumulates every release ever seen with no pruning or size cap, unlike `newReleases` which is time-pruned in `setupReleaseAlerts`. A saved artist with a large/fast-growing catalog steadily grows an unbounded array in `chrome.storage.local`; nothing in this PR (or pre-existing code) trims it, so over months this could approach the storage quota.

2. `service-worker.js`, the block from notification-send through the final `storage.local.set`, `needs-manual-verification` — notifications fire before the mutated `knownReleases` is persisted, so an MV3 service-worker termination in that window re-notifies the same releases next cycle. This ordering pre-dates the PR, but the blast radius is larger now that a single artist can surface multiple releases per check instead of one, so it's worth flagging given the PR explicitly could not be verified in a live browser this session.

3. `apps/extension/lib/release-alerts.js` (dedup key), `confirmed` — normalization is `.toLowerCase()` only, no trim/unicode normalization. This correctly avoids the feared false-collapse (a "Deluxe Edition" title stays distinct from the base title), but a title reported with incidental leading/trailing whitespace between two checks would fail to match and produce a duplicate alert rather than a silent drop. Low severity — noise, not data loss.

4. `service-worker.js` (platform-gate removal) vs. `check-releases.ts`, `confirmed, by design` — every saved artist now POSTs to `check-releases` each cycle, including ones with zero scrapeable links. Verified server-side this degrades gracefully (catalog path answers first; an uncatalogued artist with no bandcamp/faircamp/mirlo link hits a `400` the worker's try/catch logs and continues past) — not a crash, just additional request/log volume, consistent with the PR's stated intent.

Traced and confirmed correct, no bug: the `sinceDays` clamp (`Math.min(Math.max(asked, 31), 365)`, non-finite/≤0 elapsed correctly falls back to 31) and field-name parity between the extension's payload and the server's `CatalogReleaseResult` interface.

---

brandonlucasgreen (commented):
@brandonlucasgreen

Reviewed the diff (`service-worker.js`, `lib/release-alerts.js`, both manifests, `popup.js`, and the two new test files) against `origin/main`, with full surrounding file context, not just the hunks.

## Security

No security concerns found in this diff. Specifically checked and clean:

- **No new manifest permissions.** `manifest.json` / `manifest-firefox.json` diffs are only the version bump (2.6.0 → 2.7.0) and adding `"type": "module"` to Firefox's `background.scripts`. No new `permissions`/`host_permissions`.
- **No injection risk in the popup.** `buildReleaseItem` in `popup/popup.js` builds all release rows with `document.createElement` + `.textContent` (artist name, release name, `releaseSummaryLine(release)` for the subtitle) — no `innerHTML`/`insertAdjacentHTML` touched by this diff, so server-supplied release strings can't inject markup.
- **No secrets introduced.** No new Supabase calls, API keys, or tokens anywhere in the diff; `lib/supabase.js` isn't touched.
- **Server-side defense in depth on the new `sinceDays` param.** The extension now sends `sinceDays` to `/api/check-releases`, but `api/functions/check-releases.ts`'s `resolveWindowDays()` independently validates and clamps it (`api/functions/check-releases.ts:132-135`) — a malicious/garbage value from a compromised client can't widen the query beyond `MAX_WINDOW_DAYS`.

## Regression

- **plausible** — A fully- or partially-failed release check still advances `releaseCheckState.lastCheckDate`, which the new `sinceDays` optimization now uses to compute the *next* check's look-back window — the exact "cache an uncertain answer as a negative" pattern this PR's own commit message calls out and fixes for the browser-closure case. In `checkForNewReleases` (`apps/extension/background/service-worker.js`, the per-artist `try { … } catch (error) { console.error(...) }` around the `checkReleaseAPI` call, followed unconditionally by `releaseCheckState.lastCheckDate = Date.now()` a few lines later), a network error, timeout, or run of server 5xxs for one or more artists is only logged, never surfaced, and never prevents the global `lastCheckDate` update. Because `sinceDaysForCheck` (`apps/extension/lib/release-alerts.js`) floors the window at `MIN_SINCE_DAYS` (31) + 7-day padding, a single bad run is harmless, but a sustained failure streak longer than ~38 days for a given artist (e.g., that artist's specific check keeps erroring for over a month while other artists succeed and keep bumping the shared `lastCheckDate`) would cause the eventual successful check to request a window too narrow to cover the outage, silently losing any release from that gap for good. Worth either not advancing `lastCheckDate` when every artist's check failed, or tracking per-artist last-success rather than one global timestamp. Neither of the two new test files (`extension-release-check.test.ts`, `extension-release-alerts.test.ts`) covers this path — the closest test has one artist fail (400) and one succeed in the same run, but doesn't assert on `lastCheckDate` afterward or exercise a fully-failed run.
- **confirmed** — Minor: the `platforms` dict sent to `checkReleaseAPI` is now built from every link the extension holds (`for (const p of artistData.platforms || []) { if (p && p.sourceId && p.url) platforms[p.sourceId] = p.url; }`), intentionally per the PR description (server ignores it on the catalog path, only the live-scrape fallback reads it) — flagging only to confirm this matches `check-releases.ts`'s documented contract (`api/functions/check-releases.ts:1-21`), which it does. Not a defect, noted as a verified assumption rather than a finding.

No other API-contract drift, off-by-one, or null/empty-state issues found — `releasesFromResponse` correctly falls back from `releases[]` to the legacy singular `release`, `selectUnseenReleases` dedups per-artist (not globally) which is correct, and `sinceDaysForCheck` has explicit tests for the first-run/undefined and clock-skew (`now + 5 days`) cases.

---

PR #420, opened 2026-08-07
Freeze the clock in the payout-pill tests

brandonlucasgreen:
## What

`apps/web/tests/unit/RichArtistProfile.test.tsx > renders the payout percent on each pill` fails on every Bandcamp Friday — confirmed failing on `main` today (2026-08-07), and it recurs roughly monthly.

`SourceBadge` reads the real clock (`isBandcampFriday()` with no argument) to decide whether Bandcamp's pill shows its usual `80-85%` or the `~97%` Bandcamp Friday override. The test asserted `80-85%` unconditionally, so a red `npm run test:unit` was a calendar artifact rather than a real regression — and it blocks the Netlify build on those days.

## The fix

- Freeze the clock to an ordinary Friday (2026-08-14) in the existing payout test, so it asserts the same thing every day of the month.
- Add a companion case that freezes the clock to a real Bandcamp Friday (2026-08-07) and asserts `~97%`, that `80-85%` is gone, and that Patreon's payout is untouched — the override is now covered deliberately instead of by accident of the calendar.
- Each test asserts its date against `isBandcampFriday` before freezing. When `BANDCAMP_FRIDAY_DATES` is updated for a new year and 2026 falls out of the list, the failure points at the stale date rather than at the component.
- `vi.useRealTimers()` in `afterEach` so the frozen clock can't leak into the other 29 tests in the file.

Test-only change; no production code touched.

## Verification

- `npm run test:unit` — 642 passed, 47 files, green.
- `npm run lint` — 71 problems, unchanged from the existing baseline; none in this file.
- Both new assertions were mutation-checked against `SourceBadge.tsx`: forcing `isBCFriday = false` fails only the `~97%` case, forcing `isBCFriday = source.id === 'bandcamp'` fails only the ordinary-Friday case. `SourceBadge.tsx` was restored and is not part of this diff.

### DIFF
--- a/apps/web/tests/unit/RichArtistProfile.test.tsx
+++ b/apps/web/tests/unit/RichArtistProfile.test.tsx
@@ -2,7 +2,16 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { render, screen, fireEvent, cleanup, waitFor } from '@testing-library/react';
 import { RichArtistProfile } from 'src/components/RichArtistProfile';
+import { isBandcampFriday } from 'src/utils/bandcamp-friday';
 import type { ArtistPagePayload } from 'src/types/artist-page';
+
+// SourceBadge reads the real clock to decide whether Bandcamp's payout shows as its usual
+// percentage or the Bandcamp Friday override, so the payout tests below freeze the clock
+// rather than asserting whatever today happens to be. Both dates are Pacific noon, the
+// timezone isBandcampFriday() compares in. Keep them in step with BANDCAMP_FRIDAY_DATES in
+// src/utils/bandcamp-friday.ts when that list is updated for a new year.
+const ORDINARY_FRIDAY = new Date('2026-08-14T12:00:00-07:00');
+const BANDCAMP_FRIDAY = new Date('2026-08-07T12:00:00-07:00');
 
 const basePayload: ArtistPagePayload = {
   artist: {
@@ -54,6 +63,7 @@ describe('RichArtistProfile', () => {
 
   afterEach(() => {
     cleanup();
+    vi.useRealTimers();
   });
 
   it('renders the artist name and image', () => {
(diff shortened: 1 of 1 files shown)


brandonlucasgreen (commented):

**Security**: n/a — test-only change, no production code touched, no external input handled.

**Regression**: `vi.useRealTimers()` is correctly added to the existing `afterEach`, so the frozen clock in the new Bandcamp Friday case can't leak into the other tests in the file. Both new assertions pre-check their fixture date against `isBandcampFriday()` before freezing, so a future update to `BANDCAMP_FRIDAY_DATES` fails loudly at the assertion rather than silently asserting the wrong branch. No edge cases or state-transition concerns — this only fixes test determinism.

No findings.

---

brandonlucasgreen (commented):

**Security:** N/A — confirmed test-only change (`apps/web/tests/unit/RichArtistProfile.test.tsx` is the only file touched, no production code in the diff). None of auth, input validation, secrets, injection, defaults, or data exposure apply. — *confirmed*

**Regression:** Independently re-verified every date/logic claim in the PR description rather than trusting it:

- `2026-08-07` is a real Friday and is present in `BANDCAMP_FRIDAY_DATES` (`apps/web/src/utils/bandcamp-friday.ts`) → `isBandcampFriday()` correctly true. `2026-08-14` is a Friday and absent from that list → correctly false. Both fixtures use `T12:00:00-07:00`, the correct PDT offset for August. — *confirmed*
- The `86-90%`/`80-85%` assertions trace to the real platform registry (`sources.ts`), not an incidental match — `SourceBadge` renders `source.artistPayoutPercent` from the static registry, not from payload data. — *confirmed*
- `afterEach(vi.useRealTimers())` runs regardless of an assertion failure inside the test body, so no leak path into the other 29 tests in the file. — *confirmed*
- No mount-time timers/effects in `RichArtistProfile`, so `vi.useFakeTimers()` can't hang `render()`. — *confirmed*

One unrelated, out-of-scope observation: `BANDCAMP_FRIDAY_DATES` contains `'2026-05-02'`, which is a Saturday, not a Friday — inconsistent with every other entry in that list. Not introduced by this PR and not something this diff could have caught; flagging only since the new test comments explicitly ask future editors to keep in step with that list. Possibly worth a follow-up ticket.

No blocking issues. This PR does what it says.

---

PR #421, opened 2026-08-07
Show fans recent releases from their saved artists on the dashboard

brandonlucasgreen:
## Why

The audit that found **zero fans have ever subscribed to a release feed** also found the reason: the web — the primary client — has no fan-facing release surface at all. Release alerts exist only in the Mac app and the browser extension. `/dashboard`'s one "Releases" link is for *claimed artists* managing their own catalogue. The per-fan ICS/Atom feed shipped in #388 and has had **0 tokens minted ever**, because its only entry point is a section on `/settings`.

Brandon: *"there is very little fan surface area … something like 'Recent Releases' which is a shortlist of releases from the past month from your saved artists. Let's start there."*

## What

A **Recent Releases** section on `/dashboard`, above Saved Artists: up to six releases from the past 30 days plus everything upcoming, newest first, each linking to its buying guide at `/a/{artist}/{release}`.

### One question, one answer

The section reuses `getFeedReleasesForUser` rather than asking the database the same thing a second way. The feed and the dashboard disagreeing about what counts as recent would read as a bug in whichever one you happened to be looking at. That query already gets three things right that must not regress, and they carry over untouched:

- filters `is_hidden` and `needs_review`, so an artist's suppressed release stays invisible and a tier-3 fuzzy match is never presented to a fan as a distinct record
- includes **upcoming** releases, not just past ones
- bounded window, so the row cap can't hide anything

The one difference is ordering. The feed returns ascending, because a calendar and a reader both want the *next* thing at the top. A dashboard shortlist wants the opposite: descending puts anything still to come above everything already out, and then the most recent above older ones — so the cap trims the oldest rather than silently dropping the news. That sort lives in the new endpoint (`toRecentReleases`), not in the shared query.

### Payout ordering is the product

Platforms and the price line come from `orderedSourcePlatforms` / `leadingOfferSummary` in `api/shared/release-display.ts` — the same helpers the release page, the artist page and the feeds use. Artist-paying-first, so a Discogs secondhand copy never leads over a Bandcamp direct purchase, and no ninth copy of the payout figures.

### Feed discovery moves to where the releases are

Once a fan can see recent releases, *"subscribe to this as a calendar or a feed"* belongs here rather than buried in `/settings`. The heading carries the same calendar + RSS marks as the artist page, stroked at the same weight.

**One control, not two.** Both formats come from the same feed token, so a pair of buttons doing the same thing would be two treatments for one action. It's a button rather than a link (unlike the artist page's) because a fan's feed URL contains a credential that doesn't exist until they ask for one — pressing it only reveals the existing `ReleaseFeedControls` panel; minting still takes the explicit press inside it.

### Empty states

- **No saved artists** → the section doesn't render at all. An empty releases box above an empty saved-artists box is two empty states saying the same thing.
- **Saved artists, nothing new** → it says so, and explains the window. This is the common case (36 live saved-artist rows across all users) and a section that silently vanished would read as a feature that isn't working. The subscribe control stays — a fan with nothing out this month is exactly who wants telling when something is.
- **Read failed** → says so, rather than claiming there's nothing new. It reports to Sentry and into the section, not the page-level error, so a failure here doesn't take down the rest of the dashboard.

## Changes

| File | |
|---|---|
| `api/functions/me-recent-releases.ts` | New `GET /api/me/recent-releases`. Follows the `me-*` pattern: bearer auth against the anon client, hand-rolled permissive CORS, rate limited, `Cache-Control: private, no-store`. Added to `api/tsconfig.json`'s include, so it is actually typechecked. |
| `apps/web/src/components/RecentReleasesSection.tsx` | Presentational; the page fetches and passes the auth-dependent subscribe panel in. Follows `ReleasesSection.tsx` — plain `<a>` rows, `sr-only` platform names, same icon decisions. |
| `apps/web/src/pages/DashboardPage.tsx` | Wires it in. |
| `api/functions/db.ts` | `date_precision` joins the shared feed row (see below). |
| `netlify.toml` | The route. |

### One shared-type change

`FeedReleaseRow` gains `datePrecision`. The dashboard prints dates in words, and without it `formatReleaseDate` either returns nothing or — worse — would state "1 January 2023" for a release we only know the year of. The feeds place an event *on* the date and have nothing finer to say, so they ignore it.

## A bug caught in the browser, not in review

At 375px the rows overflowed: `scrollWidth` 636 inside a 375px viewport, so the whole page scrolled sideways. A grid item's automatic minimum size is its content, so `min-w-0` on the inner text column isn't enough — it has to be on the grid item itself. Fixed and re-measured: 375/375 at phone width, 430/430 at large-phone width.

`ReleasesSection.tsx` has structurally identical markup without that `min-w-0` and may well have the same defect on artist pages. Not touched here — flagged separately rather than widened into this PR.

## Verification

Rendered in the Browser pane at desktop light, desktop dark, 375px and 430px, with fixtures covering the cases that go quietly wrong: a name-your-price release (`$0` is not "free"), a year-precision date, a sold-out-only release with no quotable price, a Discogs-cheaper-than-Bandcamp release, and titles long enough to truncate. `npm run dev` can't serve this page (it doesn't run the Netlify functions, and the dashboard is auth-gated), so this used the throwaway Vite component harness rather than changing `apps/web/server/` — **the dev shim is untouched and still has no route for this endpoint.**

- `npm run lint` — 71 problems, the known baseline, unchanged
- `npm run test:api` — 969 passing, including 11 new
- `npm run test:unit` — 651 passing, including 11 new. One pre-existing failure in `RichArtistProfile.test.tsx` (`'86-90%'` not found), verified present on `main` before this branch.
- `npm run typecheck:api` — clean


brandonlucasgreen (commented):

**Security**
- Auth/authorization: `me-recent-releases.ts` authenticates via `anonClient.auth.getUser(token)` against a real bearer token before any DB read, and 401s otherwise — matches the established `me-*` pattern (compared against `me-feed-token.ts`).
- Input validation / injection: no user-supplied parameters reach the query — `getFeedReleasesForUser(userId)` is keyed only on the authenticated user's id, and `userId` comes from Supabase's own verified session, not a client-controlled field. No injection risk.
- Secret handling: n/a, nothing new touched.
- Unsafe defaults / data exposure: `Access-Control-Allow-Origin: '*'` on an endpoint returning per-user data initially looked worth flagging, but it's the same hand-rolled-permissive-CORS pattern already used by `me-feed-token.ts` and friends — safe here specifically because the authorization is a bearer header the browser won't attach cross-origin on its own (unlike a cookie), so a third-party origin can't ride an existing session. Confirmed consistent with existing convention, not a new gap. `Cache-Control: private, no-store` is correctly set so a shared CDN/cache can't leak one fan's list to another.

**Regression**
- Edge cases: `toRecentReleases` sorts on the raw ISO `releaseDate` string via `localeCompare`, which is correct for `YYYY-MM-DD` values and is covered by a test that also checks the cap trims the *oldest* row rather than the newest. Handles a mix of past/upcoming dates correctly in the test fixture.
- Null/empty behavior: `datePrecision` is threaded through from the new `date_precision` column end-to-end (`db.ts` → endpoint → component), with a test asserting a year-precision release never gets rendered as a fabricated day-level date. Good.
- State/race check: `loadRecentReleases()` is awaited inside the same `Promise.all` that gates `loading`, and it swallows its own errors into `releasesError` rather than rejecting — so there's no risk of a flash of "nothing new" before the fetch resolves, and no risk of one failing fetch aborting the other two.
- API contract drift: none — `getFeedReleasesForUser`'s existing filtering (`is_hidden`, `needs_review`) and window bounds are reused rather than re-implemented, so this can't drift from the feed's definition of "recent."
- The 375px overflow bug the PR describes fixing is specific to the new component; the PR correctly notes (but doesn't fix) that `ReleasesSection.tsx` may have the identical pre-existing defect — flagged by the author for separate follow-up, not scope creep here.

---

brandonlucasgreen (commented):

**Security:** No high/medium issues. Auth is sound — `me-recent-releases.ts` validates the bearer token via `anonClient.auth.getUser()` and scopes `getFeedReleasesForUser(userId)` by that validated `user_id` before touching any release data, so there's no path to read another fan's saved-artist releases despite the service-role client bypassing RLS. No injection risk (parameterized Supabase queries, no `dangerouslySetInnerHTML`, React escaping on all rendered release fields). No secrets touched. Response is a whitelisted field set, `Cache-Control: private, no-store`. — *confirmed*

**Regression:**

1. **[medium, confirmed]** `getFeedReleasesForUser` (`api/functions/db.ts:4289-4327`) swallows a DB-level failure and returns `[]`, which `me-recent-releases.ts` then serves as a normal `200 { releases: [] }`. `RecentReleasesSection` renders that as "Nothing new from your saved artists this month" — but the PR description explicitly promises "Read failed → says so, rather than claiming there's nothing new." That promise only holds for network/HTTP failures caught client-side in `DashboardPage.tsx`; a Postgres-level error inside `getFeedReleasesForUser` never reaches that path and produces no Sentry signal. This is the exact "never cache uncertainty" bug class called out in CLAUDE.md — worth a `shouldCache`-style distinction (or at minimum a Sentry report) between "genuinely empty" and "query failed." Inherited from the pre-existing function (also affects the ICS/Atom feed today), but this PR is the first place product copy makes an explicit claim the code doesn't back up.

2. **[low, confirmed]** `/api/me/recent-releases` is fetched unconditionally in `DashboardPage.tsx`'s `Promise.all`, even though the section only renders when `savedArtists.length > 0`. Wasted authenticated round trip for fans with no saved artists — not a correctness bug, just avoidable load.

3. **[low, confirmed, pre-existing]** `getFeedReleasesForUser` orders ascending and caps at `FEED_MAX_RELEASES = 200` before `toRecentReleases` re-sorts descending and caps at 6. If a fan's saved-artist catalogue within the window ever exceeds 200 rows, the ascending limit would keep the oldest rows and could truncate exactly the far-future "upcoming" releases the dashboard most wants to show. Not reachable at current scale (36 saved-artist rows total per the PR), flagging as a latent assumption if usage grows.

No off-by-one issues found in the 30-day window, 6-item cap, or upcoming/past sort boundary — all covered by the new tests and verified against `release-display.ts`. `FeedReleaseRow.datePrecision` addition doesn't leak into the feed generator (`toFeedReleases` builds an explicit field list, ignoring the new field), so no API contract drift there.

---

brandonlucasgreen (commented):

**Security**

No confirmed or plausible issues. Verified: `userId` is derived only from `anonClient.auth.getUser(token)` (server-verified JWT), never from a client param — no IDOR. `getFeedReleasesForUser` (`api/functions/db.ts:4280-4318`) scopes through `saved_artists.eq('user_id', userId)` before touching `releases`, and the `is_hidden`/`needs_review` filters are actually present in code (not just claimed in the PR description). No raw SQL/injectable params (endpoint takes no query params), no secrets in code.

One item, `confirmed` but not a new issue — noted for completeness: `Access-Control-Allow-Origin: '*'` in `me-recent-releases.ts:11-16` matches every other `me-*.ts` endpoint's documented "permissive CORS is fine because auth is bearer-only, no cookies" pattern. This file just omits the explanatory comment its siblings carry — a consistency nit, not a vulnerability.

**Regression**

1. `db.ts:4280-4317`, `confirmed` — `getFeedReleasesForUser` swallows every Supabase error path (`savedError`, `error`, the `catch`) and returns `[]`, so `me-recent-releases.ts` always responds `200 { releases: [] }` even during a real outage. The dashboard's `loadRecentReleases` only shows the "read failed" state on a thrown exception or non-2xx status, neither of which this path can produce — so a genuine DB outage renders as "Nothing new from your saved artists this month," the exact silent-failure-as-negative-result pattern this PR's own design section (and CLAUDE.md's "never cache uncertainty" principle) calls out. Pre-existing in `db.ts`, but this PR is the first fan-facing surface for it and is also the first place that added an "error" UI state this failure mode can never reach — worth deciding whether to distinguish DB errors from empty results here.

2. `db.ts:4298-4306` / `me-recent-releases.ts:59-66`, `plausible` — `getFeedReleasesForUser` orders ascending and caps at `FEED_MAX_RELEASES = 200` *before* the new endpoint re-sorts descending and takes the top 6. If a fan's saved artists ever produce >200 qualifying rows, the DB-level cap could drop a far-future upcoming release in favor of one nearer the 30-day-ago boundary, undercutting the endpoint's own "cap trims the oldest, never hides the news" comment — that guarantee only holds within the pre-truncated 200. Per CLAUDE.md's dated measurement (36 total `saved_artists` rows repo-wide, 2026-08-07) this looks unreachable today — flagging as latent, not live.

No issues found in: `datePrecision` addition to `FeedReleaseRow` (confirmed the ICS/Atom feed builder only consumes the type, never constructs it, so the new field can't break existing feed consumers), or the `api/tsconfig.json` include claim (both the new file and its test are actually present in the diff).

---

brandonlucasgreen (commented):
@brandonlucasgreen

Reviewed the diff (`api/functions/me-recent-releases.ts`, `db.ts`, `RecentReleasesSection.tsx`, `DashboardPage.tsx`, `netlify.toml`, `api/tsconfig.json`, and both new test files) against `main`.

## Security

- **Auth/authorization**: `authenticateRequest` in `me-recent-releases.ts` derives `userId` from `anonClient.auth.getUser(token)` (Supabase-verified JWT) — never from a client-supplied param — so there's no IDOR path. `getFeedReleasesForUser` (pre-existing, only gains the `date_precision` column here) scopes strictly to that `userId` via `saved_artists.user_id`. Fails closed: missing/invalid token, missing env vars, or auth error all return 401 before any DB read. Verified with a test asserting `getFeedReleasesForUser` is never called on 401.
- **CORS wildcard (`Access-Control-Allow-Origin: '*'`)**: matches the existing `me-*` convention (`me-feed-token.ts`, `me-settings.ts`, etc.) exactly — safe here because the endpoint is bearer-token gated, not cookie-gated, so a cross-origin page can't ride a victim's session. Not a new pattern.
- **SSRF**: no new outbound fetches; no raw `fetch(userUrl)` introduced.
- **Secrets/PII**: no tokens or PII logged; error responses are generic strings, no stack traces leaked to the client.
- **New DB surface**: no new table/column beyond exposing the already-existing `releases.date_precision` (added in migration `20260731120000_releases.sql`) through the existing service-role query path — no migration needed, RLS posture unchanged.
- **Rate limiting**: `checkRateLimit(getClientIp(...), 'standard', ...)` applied before the method/auth checks, consistent with sibling `me-*` functions.

## Regression

- `needs-manual-verification` — **`saved_artists` lookup in `getFeedReleasesForUser` has no explicit page/limit and could theoretically hit PostgREST's silent 1000-row cap** if a single fan has 1000+ saved artists (`api/functions/db.ts`, the `.from('saved_artists').select('artist_id').eq('user_id', userId)` call feeding `getFeedReleasesForUser`, exercised by this PR's new dashboard path). Not introduced by this PR and the releases query itself is safely capped at `FEED_MAX_RELEASES = 200`, but CLAUDE.md flags this exact gotcha (`readAllPages`/`.range()`) and this PR is the first fan-facing surface that exercises this code path from the dashboard — worth a human sanity check on whether any fan is anywhere near that row count today.
- `plausible` — **`loadRecentReleases()` in `DashboardPage.tsx` swallows a 401 into the section-local error state (`"Couldn't load recent releases. Try refreshing."`) instead of redirecting to `/login` like the sibling `claimedResponse` 401 check does.** (`apps/web/src/pages/DashboardPage.tsx`, `loadRecentReleases`). Benign today since the two requests share the same freshly-validated session token, but if the token expires mid-session this section would show a generic error rather than prompting re-auth like the rest of the page — a minor UX inconsistency, not a security issue.
- No other regressions found: sort/cap logic in `toRecentReleases` is covered by dedicated tests (immutability, cap trims oldest not newest, date-precision passthrough), the "coming" boundary is tested at exactly today's date, and the empty-state / error-state / loading paths are each asserted separately in `RecentReleasesSection.test.tsx`.

---

PR #422, opened 2026-08-07
Fix three release-alert defects in the Apple app

brandonlucasgreen:
Three defects in the shipped release-alert client, plus the macOS gap that showed up while fixing them.

## 1. The iCloud dismissal sync was a no-op

`NewRelease.id` was a fresh `UUID()` — generated independently on each device — and `saveDismissedId` published *those* through `NSUbiquitousKeyValueStore` as the dismissed set. Dismiss "Infinite Normal" on one Mac, sync UUID X; the other Mac holds the same release under UUID Y; nothing ever matches. The one cross-device feature the app claimed did nothing at all.

The id is now derived from the release. Two schemes, prefixed so they can't be confused:

| Alert source | `releaseUrl` | id |
|---|---|---|
| Catalogue (`getReleasesForAlerts`) | `https://unstream.stream/a/{artist}/{release}` | `release:kid-lightbulbs/infinite-normal` |
| Live-scrape fallback | the platform's own page | `name:kid lightbulbs/infinite normal` |

The fallback degrades honestly rather than pretending: a scraped alert has no slugs, so it keys on the artist and title the server hands both devices — which is also exactly the identity `isKnownReleaseByName` already dedups on — and the `name:` prefix says out loud that it isn't a catalogue key.

Three things this had to not break, all covered by tests:

- **v3.4.0's stored alerts still decode.** `init(from:)` keeps its defaults-supplying shape and simply stops reading `id`, recomputing it instead. That's deliberate, not incidental: a stored id is either a worthless random UUID or a key this function would produce anyway, and recomputing is what makes two devices that upgrade separately converge on the same identity for a release they both already had.
- **Ids stay unique within the list**, since `id` is the SwiftUI `ForEach` key in `ReleasesTab`.
- **The notification identifier follows** — `releaseAlert-{id}`.

The dismissed set moves to a new KVS key (`dismissedReleaseKeys`). The old one holds per-device UUIDs that can never match a stable id; reusing it would only let dead entries push real ones out of the 200-entry cap. It's left in place rather than deleted so an older device still syncing against it isn't fighting this one.

## 2. Both clients gated on three platforms while the catalogue covers five

`ReleaseAlertManager` built its platforms dictionary from bandcamp/faircamp/mirlo and then `continue`d when it came out empty — but `check-releases` tries the catalogue first and ignores `platforms` entirely on that path. Only the live-scrape fallback needs URLs. So an artist known only through Discogs (1,953 links in production, against Bandcamp's 1,810) was skipped for a question one DB read answers.

Now every link the artist has is sent, keyed by platform id, and nobody is skipped. The server reads only the keys it can scrape, so the extra ones cost nothing.

## 3. `sinceDays` was supported but never sent

`ReleaseCheckAPI` has taken `sinceDays: Int? = nil` since the parameter was added for exactly this; the manager never passed it. A laptop closed for six weeks woke up, asked for the default 31 days, and permanently lost anything that had aged past that window while it slept (spec §5 defect 5).

The window is now the gap since the last check plus a week of padding — a release can be *dated* days before it appears anywhere we can see it — floored at the 31 days we used to get and capped at the server's 365, so the request means what it says instead of being silently clamped. Computed before the loop, because the `defer` moves `lastCheckDate` to now.

## Also: macOS had less than the description suggested

`SupportListView` used `newRelease(for:)`, which returns `.first`. But the bigger hole: a **signed-in** macOS user sees `SyncedArtistsView`, which has no release UI at all — so on the flagship client, release alerts appeared nowhere outside notifications and a DEBUG count in Settings.

Two changes, both small because the scaffolding existed:

- The saved-artists list shows **every** unread release for an artist, not `.first`.
- The popover gains a **New Releases** drill-down: a bell + count beside the segmented control (shown only when there are unread releases), ⌘3, and rows that push each release's buying guide onto the existing route stack. It reuses the back header, Escape and ⌘[ that `.releaseGuide` already had; rows are draggable and carry a context menu with Where to Buy / Open / Copy Link / Share / Dismiss, since the ✕ is hover-dimmed and keyboard and VoiceOver users need a non-hover path to it.

**What I did not build:** a permanent Releases tab alongside Search and Saved, mirroring iOS. That changes the app's top-level navigation, needs a real empty state, and is a product call rather than a bug fix — the drill-down is additive and doesn't foreclose it.

I also didn't add a `shipped-features.json` entry: those are coarse feature-level items and `announced: false` drives a social post, which a bug-fix release probably shouldn't. Say the word if you'd like one.

## Verification

- **71 XCTests pass** (`ReleaseAlertTests` +17), macOS and iOS both build.
- **Each new test was proved to have teeth** by reverting the behaviour it guards and confirming the failure:
  - `stableId` → `UUID().uuidString`: 6 identity tests fail, the legacy-decode test still passes.
  - decode the stored `id` instead of recomputing: `testDecodingRecomputesTheIdRatherThanTrustingAStoredUUID` fails.
  - re-add the bandcamp/faircamp/mirlo filter: both `platformUrls` coverage tests fail.
  - `lookbackDays` → always nil: the three window tests fail.
- The new macOS row was rendered to PNG in light and dark and checked for truncation and the announced/released variants.
- Decisions are pure statics — `NewRelease.stableId`, `ReleaseAlertManager.platformUrls(for:)`, `ReleaseAlertManager.lookbackDays(since:now:)` — following the `selectUnseen` precedent.

Version 3.4.0 → 3.5.0, build 14 → 15, across all four files; `xcodegen generate` re-run for the new source file.

### DIFF
--- a/apps/mac/Unstream.xcodeproj/project.pbxproj
+++ b/apps/mac/Unstream.xcodeproj/project.pbxproj
@@ -50,6 +50,7 @@
 		8916C6D428EE94E16E491566 /* ReleasesTab.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2C8F1E48E643DCC1EECD4A93 /* ReleasesTab.swift */; platformFilters = (ios, ); };
 		8961565967B20B3F28A7877E /* GlobalHotkeyManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6B52C116B5CC91BC40D028AF /* GlobalHotkeyManager.swift */; platformFilters = (macos, ); };
 		8A4181366EB270C58A708C65 /* SearchBarView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 45AC0E531396E8C63A2001CF /* SearchBarView.swift */; platformFilters = (macos, ); };
+		8AE1365921AE81ECF06371F4 /* NewReleasesListView.swift in Sources */ = {isa = PBXBuildFile; fileRef = BAF3C1FFEBBA8D8B753BDB0B /* NewReleasesListView.swift */; platformFilters = (macos, ); };
 		8C94768621D55383151FD59F /* SearchResponse.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1581BE8FAF07C7C9D3CD99F5 /* SearchResponse.swift */; };
 		8CD55AA9054F310B1DBD51DA /* ServicesProvider.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1DEF52280B036F1EB34F4127 /* ServicesProvider.swift */; platformFilters = (macos, ); };
 		917C407BBA300BE70B9BDEFB /* ScrobbleManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = EEEE83E95439F19BFC679216 /* ScrobbleManager.swift */; platformFilters = (macos, ); };
@@ -173,6 +174,7 @@
 		B245CC3DBF71C4B066A96632 /* AuthService.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AuthService.swift; sourceTree = "<group>"; };
 		B4745357EE3AD119D3F7B059 /* BandcampFriday.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BandcampFriday.swift; sourceTree = "<group>"; };
 		B733B6764210C3CAE08F12E0 /* SavedArtistsSync.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SavedArtistsSync.swift; sourceTree = "<group>"; };
+		BAF3C1FFEBBA8D8B753BDB0B /* NewReleasesListView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = NewReleasesListView.swift; sourceTree = "<group>"; };
 		BC8C8A39D8800BDBA893E0EA /* NowPlaying.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = NowPlaying.swift; sourceTree = "<group>"; };
 		BE9E59C365D8F116C435B8BC /* EmptyStateView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = EmptyStateView.swift; sourceTree = "<group>"; };
 		BE9F88505AF840EC94B5C0BC /* FaircampReleaseChecker.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = FaircampReleaseChecker.swift; sourceTree = "<group>"; };
@@ -389,6 +391,7 @@
 		F284A719009A53D4D7A04F14 /* macOS */ = {
 			isa = PBXGroup;
 			children = (
+				BAF3C1FFEBBA8D8B753BDB0B /* NewReleasesListView.swift */,
 				9B08A02BD24EAF475A2689D6 /* NowPlayingView.swift */,
 				EF4E58F91C106A97A5179125 /* PopoverView.swift */,
 				45AC0E531396E8C63A2001CF /* SearchBarView.swift */,
@@ -572,6 +575,7 @@
 				7FA2720BFA00323DC8E66A9A /* ListenBrainzService.swift in Sources */,
 				DAF5726D4B93EA321E5D4643 /* MediaObserver.swift in Sources */,
 				47ADDB550C3A545314574336 /* MirloReleaseChecker.swift in Sources */,
+				8AE1365921AE81ECF06371F4 /* NewReleasesListView.swift in Sources */,
 				F20D83D3642078C9B2C45ED0 /* NowPlaying.swift in Sources */,
 				64514560D09214DE420ED24D /* NowPlayingNotificationManager.swift in Sources */,
 				27A98FF82168ECEF095C04D6 /* NowPlayingView.swift in Sources */,
--- a/apps/mac/Unstream/Info-iOS.plist
+++ b/apps/mac/Unstream/Info-iOS.plist
@@ -21,7 +21,7 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>3.4.0</string>
+	<string>3.5.0</string>
 	<key>CFBundleURLTypes</key>
 	<array>
 		<dict>
@@ -34,7 +34,7 @@
 		</dict>
 	</array>
 	<key>CFBundleVersion</key>
-	<string>14</string>
+	<string>15</string>
 	<key>ITSAppUsesNonExemptEncryption</key>
 	<false/>
 	<key>LSRequiresIPhoneOS</key>
--- a/apps/mac/Unstream/Info-macOS.plist
+++ b/apps/mac/Unstream/Info-macOS.plist
@@ -17,7 +17,7 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>3.4.0</string>
+	<string>3.5.0</string>
 	<key>CFBundleURLTypes</key>
 	<array>
 		<dict>
@@ -30,7 +30,7 @@
 		</dict>
 	</array>
 	<key>CFBundleVersion</key>
-	<string>14</string>
+	<string>15</string>
 	<key>ITSAppUsesNonExemptEncryption</key>
 	<false/>
 	<key>LSApplicationCategoryType</key>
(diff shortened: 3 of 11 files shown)


brandonlucasgreen (commented):

**Security**
- Auth/authorization: n/a — client-only Swift, no new network endpoints or server calls.
- Input validation / injection: `NewRelease.stableId` derives a string key from server-supplied `releaseUrl`/`artistName`/`releaseName` and uses it as an iCloud KVS entry and a `UNNotificationRequest` identifier. Both are plain string APIs — no interpolation into SQL, HTML, or a shell — so no injection surface. No issue.
- Unsafe defaults / data exposure: none spotted; `NewReleasesListView` only renders data already available elsewhere in the app.

**Regression**
- Confirmed, low severity: the dismissed-set migration (`dismissedReleaseIds` UUID → `dismissedReleaseKeys` string) is a one-time behavior change, not just a cross-device fix. Before this PR, `NewRelease.id` was decoded back from disk (`id = try c.decode(UUID.self, forKey: .id)`), so a *same-device* dismiss did persist across restarts via the old key. After this PR, `id` is always recomputed and the old key is never read, so any release a user had already dismissed under v3.4.0/pre-this-build will resurface once after upgrading (it's still in the persisted alert list, and its new stable id won't be in the now-empty `dismissedReleaseKeys` set). This looks like an accepted, understood cost of the migration rather than an oversight — the PR is explicit that the old key is "left in place... isn't fighting this one" — but it's worth confirming that a one-time resurfacing of already-dismissed alerts is acceptable, since it isn't called out explicitly.
- Edge cases: `stableId` folding (lowercase, trim, `/`→`-`) and the `release:`/`name:` prefixing look correct and are well covered by the new XCTests, including the legacy-decode and uniqueness-within-list cases.
- State transitions: `lookbackDays` is computed once before the artist loop (correctly, since the `defer` moves `lastCheckDate` afterward), is floored/capped as described, and falls back to `nil` on a future `lastCheckDate` rather than deriving a negative window. Looks right.
- API contract drift: `platformUrls(for:)` now forwards every saved link instead of gating on bandcamp/faircamp/mirlo — matches the stated server behavior (catalogue path ignores `platforms` entirely; only the live-scrape fallback reads it), consistent with the parallel extension fix in #419.

No other findings.

---

brandonlucasgreen (commented):

**Security**

No issues found. The widened `platformUrls` dict is inert on the server: `check-releases.ts`'s live-scrape path still hard-filters to `bandcamp`/`faircamp`/`mirlo`, so extra keys (Discogs, Instagram, etc.) sent by the client aren't a new SSRF surface — the existing allowlist/stored-link gate is unchanged. No auth, keychain, token, or credential handling is touched anywhere in this diff. iCloud KVS carries only artist/title-derived strings, scoped to the user's own private iCloud account, so the predictable id scheme isn't a data-exposure concern.

**Regression**

1. `apps/mac/Unstream/Models/ReleaseAlert.swift:100-106` (`idComponent`), `confirmed` (mechanism) / `plausible` (real-world trigger) — `idComponent` only folds `/`→`-` and lowercases, so two structurally different artist/title strings can collapse onto the same `name:` id (e.g. artist "AC/DC" title "X" → `ac-dc/x`, identical to artist "AC-DC" title "X"). Since `id` is the sole key for `dismissRelease`'s removal and the SwiftUI `ForEach`, a collision between two live-scraped (uncatalogued) releases means dismissing one silently hides the other, and the folded id gets pushed to `dismissedReleaseKeys` in iCloud KVS — permanently suppressing the unrelated release everywhere. Low-probability (needs a literal `/`↔`-` naming coincidence) but the collision path itself is real, not speculative.

2. `apps/mac/Unstream/Models/ReleaseAlert.swift:75-97` (`stableId`) vs. the server-derived catalog slug, `plausible` — the `name:` fallback id normalizes only via `.lowercased()` (no diacritic/unicode folding), while the `release:` id inherits the server's ASCII slug, so the same real release can carry two different "stable" ids depending on whether it's first seen via live-scrape or the catalog (e.g. "Sigur Rós" → `name:sigur rós/...` pre-catalog vs `release:sigur-ros/...` once catalogued). In practice this doesn't cause a duplicate re-notification today — `ReleaseCheckState.isKnownReleaseByName` is a separate, name-only dedup layer that catches the second sighting before the new id is even built — but it means the "cross-device stable id" this PR is built around isn't actually stable across the live→catalog transition, and the new tests (e.g. `testCatalogIdComesFromTheReleasePageNotTheDisplayNames`) don't cover that transition case. Worth a regression test pinning the current (accidental) safety net so a future change to `isKnownReleaseByName` doesn't quietly reintroduce a duplicate.

Verified with no bug found: the `sinceDays` clamp agrees between client (`min(365, max(31, ...))`) and server (`resolveWindowDays`), including the future-dated-`lastCheckDate` fallback case; and the decode-recompute migration path correctly discards v3.4.0's random UUIDs on every legacy alert.

---

brandonlucasgreen (commented):
@brandonlucasgreen

Reviewed the full diff against `main` (via `api/functions/check-releases.ts` on the server side too, to check the client's assumptions about `sinceDays`/window defaults against what the server actually does).

## Security

- `stableId`'s `release:` scheme only accepts a slug when `URL.host` is *exactly* `unstream.stream` (`ReleaseDetailService.slugs`), so a crafted `releaseUrl` can't spoof a catalog id via a lookalike host.
- The expanded `platformUrls(for:)` now sends every saved platform link (not just bandcamp/faircamp/mirlo) to `check-releases`, but I checked the server side: it only reads keys it can scrape and gates each one through `mayFetch` (the SSRF allowlist), unchanged by this PR — so the extra keys sent are inert, not a new attack surface.
- No secrets, tokens, or Keychain-adjacent data touched anywhere in the diff; the only new persisted/synced value is a content-derived, human-readable string id (artist/release slugs or lowercased names) — nothing sensitive.
- `print("API check failed for \(artistName): \(error)")` (pre-existing, unchanged) logs an artist name, not PII/secrets.

## Regression

1. **plausible** — The `sinceDays` fix computes its lookback window from `checkState.lastCheckDate` (`ReleaseAlertManager.swift:209`, `lookbackDays` at `:299`), but that checkpoint is advanced unconditionally in the `defer` block of `checkForNewReleases` (`:197-203`) regardless of whether any per-artist call actually succeeded — `checkViaAPI` (`:250-265`) catches all errors and silently returns `[]` with no signal back to the caller. If every saved artist's request fails in one scheduled run (server outage, no connectivity at the scheduled Friday check, etc.), `lastCheckDate` still moves to "now," so the *next* successful run computes its window from that false checkpoint instead of the last real success — silently narrowing the safety net the `sinceDays` fix was built for. It's bounded by the `max(31, ...)` floor (any true gap ≤31 days is unaffected either way), so it only bites after a longer outage, but it's a version of the exact "closed laptop" defect this PR fixes, reintroduced via a different path. Confirmable in the code; how often it's hit in practice depends on real-world network/server reliability, hence "plausible" rather than "confirmed."

2. **plausible**, low severity — The dismissed-ids iCloud KVS key changes from `dismissedReleaseIds` to `dismissedReleaseKeys` (`ReleaseAlertManager.swift:21-24`), which is called out and intentional in the PR description (old UUIDs can never match a stable id). During the rollout window before *all* of a fan's devices are on this build, a release dismissed on an old-build device won't sync to a new-build device and vice versa, so it can reappear until every device upgrades. Likely an acceptable, temporary rollout artifact — flagging so it's a conscious call rather than a surprise in support tickets.

3. **plausible**, low severity — The new macOS surfaces (`PopoverView.swift:362` bell count, `:424` ⌘3 shortcut gate; `NewReleasesListView.swift:24,42` empty-state/list) read the raw `releaseAlertManager.newReleases` array directly, with no `.isActive` (30-day) filter — so a release that ages past its active window can keep showing in the bell badge/drill-down list/keyboard shortcut until the next `setupScheduling()` prune, which only runs on app launch or toggling the alerts setting, not on a recurring basis. This is **not a new pattern** — it exactly mirrors the pre-existing iOS `ReleasesTab` and the DEBUG `SettingsView` count, both unfiltered the same way before this PR — so it's an extension of an existing latent staleness quirk to a new surface rather than a fresh regression. Worth a follow-up but not blocking.

4. **needs-manual-verification** — The PR body claims 71 passing XCTests and clean macOS/iOS builds; I reviewed the test file's logic and it looks like it genuinely exercises the claimed behaviors (each test paired with a stated "proved to fail when reverted" check), but I can't run Xcode in this environment to confirm the suite actually passes or that both targets build. Recommend confirming via CI/local build before merge if that hasn't already run green.

No findings on double-firing notifications or resurrected/tombstoned state: `selectUnseen`'s single dedup point (keyed by artist+release name in `checkState.knownReleases`) is unchanged by this PR and correctly prevents re-alerting on anything already recorded, dismissal removes from the display list but not from `knownReleases` (so a dismissed release is never re-notified), and the release-alert dismissal sync (iCloud KVS) is entirely separate from the Supabase saved-artists tombstone/GC sync — this PR doesn't touch `SavedArtistsSync.swift` or the server-side tombstone model at all.

---

PR #423, opened 2026-08-07
Resolve retired artist slugs on the release page

brandonlucasgreen:
Stacked on #418 — **base is `fix/release-detail-slug-aliases`, not `main`.** GitHub will retarget this to `main` automatically once #418 merges. It's stacked rather than parallel because #418's comments in `release-detail.ts` describe this gap as an open one; merging them independently would land a knowingly-wrong load-bearing comment on `main` whichever went first.

## The bug

`/a/{artist}/{release}` renders **nothing** when the artist slug is a retired one — 44 rows in `artist_slug_aliases` as of today, most left by the accent-folding reslug in #410 (`beyonc` → `beyonce`).

`api/edge/release-page.ts` looked the artist up with a plain `.eq('slug', …)` and on a miss called `context.next()`. The next matching edge function is `artist-page-static`, which hands non-crawlers straight to the SPA — and the SPA has no route for a two-segment `/a/` path (`main.tsx` only has `/a/:slug`). So a fan following a release alert minted before the reslug gets a blank page.

#418 closed the same gap in the JSON twin (`/api/release/{artist}/{release}`) and the `/a/{artist}/releases.{ics,xml}` feeds. The HTML page — the one a person actually looks at — was the last surface still behind them.

## Two decisions

**301, not render-in-place.** `artist-page-static.ts` already 301s for a retired artist slug, and rendering identical content at two URLs is duplicate content plus a second page competing with the canonical one. The redirect is about the artist slug *alone*: if the release doesn't exist under the canonical slug either, that request answers with the existing 302 bounce to the artist page. `Cache-Control` caps at an hour rather than the year a permanent redirect invites — browsers cache a 301 aggressively, and a slug later reassigned to a real artist has to be able to stop redirecting. Same cap `artist-page-static` uses.

**The alias query is duplicated in Deno.** Edge functions can't import from `api/functions/`, so this copies the shape of the block in `artist-page-static.ts` and keeps its comment saying why the duplication exists. Resolved **only after the live slug misses**, so a real artist who later takes that slug always wins.

Also updates the two comments in `release-detail.ts` that described this as a gap in the page, since it no longer is.

## Verification

`api/edge/*` isn't in `api/tsconfig.json`'s typecheck include, so this was verified by rendering locally with `npm run preview:release`. That harness couldn't reach the new branch as written — its Supabase stub ignored `.eq()` filters, so the artists read always hit and the miss path was unreachable. It now records the filters, answers an alias row, and links a retired slug from the index page so the redirect is one click away.

```
--- retired slug ---
HTTP/1.1 301 Moved Permanently
cache-control: public, max-age=3600
location: /a/explosionsinthesky/end

--- retired slug, following the redirect ---
final=http://localhost:8845/a/explosionsinthesky/end status=200

--- canonical slug ---
status=200

--- unknown, non-alias slug ---
fell through to the SPA
```

`npm run test:api` (968 passing), `npm run typecheck:api`, `vite build` and `npm run lint` (71 problems, unchanged from the repo baseline) all clean. `npm run build` stops on one pre-existing failure — `RichArtistProfile.test.tsx` expecting `80-85%` where Bandcamp Friday shows `~97%`. Today is the first Friday of the month; it failed identically on `main` at 56a88f7 when I ran it. #420 merged three minutes later and fixes it — this branch just predates it and will pick it up on rebase.

### DIFF
--- a/api/edge/release-page.ts
+++ b/api/edge/release-page.ts
@@ -172,9 +172,45 @@ export default async function handler(request: Request, context: Context) {
         .maybeSingle()
         .abortSignal(controller.signal);
 
+      // A retired slug still resolves. Most of the aliases in production (44 of them as of
+      // 2026-08-07) came from the accent-folding reslug in #410 (`beyonc` -> `beyonce`), and a
+      // release URL minted before that — in an alert, a shared link, a Mac app cache — is a URL
+      // a fan can still be holding.
+      // Falling through instead would hand it to artist-page-static, which hands non-crawlers to
+      // the SPA, which has no route for a two-segment /a/ path: a blank page.
+      //
+      // Checked only *after* the live slug misses, so a real artist who later takes that slug
+      // always wins.
+      //
+      // A 301 to the canonical URL rather than rendering here, matching artist-page-static:
+      // rendering the same page at two URLs is duplicate content. The redirect is about the
+      // artist slug alone — if the release doesn't exist under the canonical slug either, that
+      // request answers with the usual bounce to the artist page.
+      //
+      // Duplicates resolveArtistSlugAlias in api/functions/db.ts — edge functions run on Deno and
+      // cannot import from api/functions.
       if (!artistData) {
+        const { data: alias } = await supabase
+          .from('artist_slug_aliases')
+          .select('artists!inner(slug)')
+          .eq('alias', artistSlug.toLowerCase())
+          .maybeSingle()
+          .abortSignal(controller.signal);
+
+        const canonical = (alias as { artists?: { slug?: string } } | null)?.artists?.slug;
         clearTimeout(timeoutId);
-        return context.next();
+        if (!canonical) return context.next();
+
+        // max-age=3600, not the year a permanent redirect invites: browsers cache a 301
+        // aggressively, and a slug that gets reassigned to a real artist has to be able to stop
+        // redirecting. Same cap artist-page-static uses.
+        return new Response(null, {
+          status: 301,
+          headers: {
+            Location: `/a/${encodeURIComponent(canonical)}/${encodeURIComponent(releaseSlug)}`,
+            'Cache-Control': 'public, max-age=3600',
+          },
+        });
       }
       artist = artistData;
 
--- a/api/functions/release-detail.ts
+++ b/api/functions/release-detail.ts
@@ -12,13 +12,11 @@
 // query (`getReleaseDetail`) and apply the same payout rules, so a native client can never
 // describe a release differently from the web page a fan would see for it.
 //
-// **Retired artist slugs resolve here and not on the HTML page.** This endpoint reads
-// `artist_slug_aliases` on a miss, the way /api/artist and the artist page do. `release-page.ts`
-// does not: on an unknown artist slug it falls through to the SPA, which has no route for a
-// two-segment /a/ path, so /a/{retired-slug}/{release} renders nothing in a browser today. That
-// is a gap in the page, not a reason to reproduce it here — the answers still agree wherever the
-// page produces one, and `pageUrl` below is built from the canonical slug so a client's "open the
-// full guide" lands on a URL that renders.
+// **Retired artist slugs resolve, here and on the HTML page.** This endpoint reads
+// `artist_slug_aliases` on a miss, the way /api/artist and the artist page do; `release-page.ts`
+// does the same and 301s to the canonical URL. The two answer a retired slug differently on
+// purpose — a redirect is right for a page and wrong for a client that asked a question — but
+// neither leaves a fan holding an alias URL with nothing behind it.
 //
 // **`payoutPercent` is computed here, per source, on purpose.** Payout figures are duplicated by
 // hand across eight files in this repo, and that drift is what let the Discord bot quote an
@@ -244,9 +242,8 @@ export async function handler(event: {
         },
         // The page URL, so a native client can offer "open the full guide" without rebuilding
         // the URL shape — and so the two surfaces stay tied together if it ever changes. Built
-        // from the canonical slug rather than the requested one: the HTML page does *not* resolve
-        // aliases (see the file header), so echoing a retired slug back here would hand a client
-        // a link that renders nothing.
+        // from the canonical slug rather than the requested one: echoing a retired slug back
+        // would hand a client a link that only works by way of a redirect.
         pageUrl: `https://unstream.stream/a/${encodeURIComponent(artistSlug)}/${encodeURIComponent(release)}`,
         bandcampFriday,
       }),
--- a/scripts/preview-release-page.ts
+++ b/scripts/preview-release-page.ts
@@ -24,8 +24,9 @@
  * the type/date/offer mapping, and the entire rendered page including payout estimates and
  * Bandcamp Friday handling.
  *
- * Faked: only the two database reads. Nothing is written anywhere — there is no database
- * connection at all, which is the point.
+ * Faked: only the database reads, including one alias row so the retired-slug redirect can be
+ * seen (there is a link for it on the index page). Nothing is written anywhere — there is no
+ * database connection at all, which is the point.
  *
  * One Bandcamp request for the grid, then one per release page you open. Be a good neighbour.
  */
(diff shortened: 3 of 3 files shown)


brandonlucasgreen (commented):

**Security**
- Auth/authorization: n/a — public, unauthenticated route, unchanged.
- Input validation / injection: `artistSlug` (`decodeURIComponent(segments[1])`) is passed unvalidated into `.eq('alias', artistSlug.toLowerCase())`, same as the existing live-slug lookup a few lines above it (`.eq('slug', artistSlug)`) — both go through Supabase's parameterized query builder (PostgREST), not string-concatenated SQL, so there's no injection risk despite the missing `SLUG_PATTERN` check that `release-detail.ts`/`db.ts` use. Confirmed this isn't a regression either: I checked `artist-page-static.ts`'s existing alias branch (lines 246–263) and it's the identical shape — raw `slug` for the live lookup, `slug.toLowerCase()` for the alias lookup, no pattern validation there either. This PR is a faithful duplicate of an already-shipped, already-reviewed pattern, not a new gap. *confirmed, no issue*
- The new `Location` header is built as `` /a/${encodeURIComponent(canonical)}/${encodeURIComponent(releaseSlug)} ``. `canonical` comes only from the DB join (`artist_slug_aliases` → `artists.slug`), never from request input, so it can't be attacker-controlled. `releaseSlug` is user input but is `encodeURIComponent`-escaped, which percent-encodes CR/LF, so no header-injection/response-splitting vector. The path is always relative (`/a/...`), so this can't become an open redirect to an external host. *confirmed, no issue*
- Secret handling: n/a, nothing touched.
- Unsafe defaults / data exposure: none — this only ever resolves an alias to *that same artist's* current slug (`artists!inner(slug)` join), so it can't be used to enumerate or leak a different artist's data.

**Regression**
- Verified the ordering claim: the alias branch only runs after the live-slug `.eq('slug', artistSlug)` lookup misses (`if (!artistData)`), so a real artist who later claims a formerly-retired slug always wins over the alias — matches the stated intent and the `artist-page-static.ts` precedent.
- Verified no redirect loop / no third query: on an alias hit this 301s immediately without checking whether the release exists under the canonical slug; if it doesn't, the *next* request's release lookup returns null and correctly 302s to the artist page (existing code, unchanged) rather than looping.
- `Cache-Control: public, max-age=3600` on the 301 (not a permanent-redirect-length cache) — matches `artist-page-static.ts`'s cap for the same reason (a slug can be reassigned later), so this doesn't create a new risk of a stale redirect outliving a slug reassignment.
- The `release-detail.ts` comment updates are consistent with the code — I checked that endpoint on `main` already resolves aliases (predates this PR, per #418) — so the updated comment is a correction of stale text, not a functional change.
- `scripts/preview-release-page.ts` changes are dev-tooling only (no DB connection, per its own header) — no production risk regardless of correctness.

---

brandonlucasgreen (commented):

**Security**

No open-redirect, header-injection, or XSS issues. The `Location` header is a relative path built from a DB-sourced canonical slug and a properly re-encoded release slug, so CRLF/host injection isn't reachable, and there's no HTML body to escape into. The whole alias-resolution block sits inside the existing outer `try { ... } catch { return context.next(); }`, so any thrown exception (timeout, network failure) correctly falls through instead of producing a bad redirect.

One `plausible` item: `api/edge/release-page.ts:176-180`'s `.eq('alias', artistSlug.toLowerCase())` interpolates user-supplied path input without the `/^[A-Za-z0-9-]+$/` format guard that the Node twin `resolveArtistSlugAlias` in `api/functions/db.ts` applies and explicitly documents as keeping the value "safe to interpolate into a PostgREST filter." `.eq()` should URL-encode the value as a query parameter (normally injection-safe), so this reads as a hardening/consistency gap rather than a demonstrated vulnerability — and it's copied verbatim from the existing pattern in `artist-page-static.ts` rather than introduced here, just now duplicated a third time. Worth confirming exploitability directly rather than by inspection.

**Regression**

`api/edge/release-page.ts:175-182`, `confirmed` — both the initial `artists` lookup and the new alias lookup destructure only `{ data }`, never checking `error`. A genuine Supabase/PostgREST error (RLS misconfig, transient 5xx — anything returning `{data: null, error}` without throwing) is treated identically to "artist not found," so it falls into the alias branch as if the slug were unknown. Unless a matching alias row happens to exist (unlikely for a live canonical slug), the request falls through to `context.next()` → the SPA, which has no two-segment `/a/` route → a blank page for a real, non-retired artist during what was just a transient DB hiccup. This contradicts the PR's own stated invariant ("never after a failed/errored lookup") — that guarantee, as implemented, only covers lookups that *throw*, not in-band PostgREST errors. This exact gap already exists in `artist-page-static.ts`'s copy of the pattern (pre-existing, not introduced here), but this PR duplicates it into a second surface instead of closing it — the Node twin `resolveArtistSlugAlias` in `db.ts`, for comparison, does check `error` and logs it.

Everything else checked out: the only-after-miss ordering is correctly implemented for the exception path, the 301 doesn't leak anything `artist-page-static` wasn't already exposing to crawlers, and a release missing under the successfully-resolved canonical slug still degrades to the existing 302-to-artist-page rather than looping or dead-ending.

---

brandonlucasgreen (commented):
@brandonlucasgreen

Reviewed against the actual base (`fix/release-detail-slug-aliases` @ fa2b615), not `main` — the PR is stacked, so the real diff is 3 files / +95/-27: `api/edge/release-page.ts` (the functional change), `api/functions/release-detail.ts` (comment-only, no logic change), and `scripts/preview-release-page.ts` (dev-only harness).

## Security

- **plausible** — `release-page.ts:192-198` builds the `artist_slug_aliases` query from `artistSlug.toLowerCase()` with no format check, unlike the function it's explicitly duplicating: `resolveArtistSlugAlias` in `api/functions/db.ts` (line ~3047) guards with `/^[A-Za-z0-9-]+$/` before querying, and its own comment says that guard exists to "keep the value safe to interpolate into a PostgREST filter." This copy drops it. In practice I don't think it's exploitable — `.eq()` goes through supabase-js's parameterized PostgREST filter encoding, not string-built SQL — so I'm not calling it confirmed, but it's worth adding for defense-in-depth and to stop the duplication from silently drifting from the function it's copying. Note this exact gap already exists in `artist-page-static.ts` (lines 246-262) — this PR copies an existing pattern rather than introducing a new one, so it's pre-existing exposure, not new.
- **confirmed: no open-redirect** — `release-page.ts:210` builds `Location: /a/${encodeURIComponent(canonical)}/${encodeURIComponent(releaseSlug)}` as a relative, same-origin path. `canonical` comes only from the `artists.slug` column via the `artists!inner(slug)` join (trusted DB data, not attacker-controlled), and `releaseSlug` is re-encoded so it can't break out of the path or inject into the header. No absolute/attacker-supplied origin is ever used.
- **confirmed: no redirect loop** — `artist_slug_aliases.artist_id` is a foreign key into `artists`, and the alias query always resolves through the join to that artist's *current* `slug`. A redirect target is therefore always a live `artists.slug`, so the next request always hits the primary `.eq('slug', …)` lookup (`release-page.ts:171`) on the first try — the alias branch can't chain into another alias lookup. Verified via the migration's own lookup-order note in `20260803180000_artist-slug-aliases.sql`.
- **needs-manual-verification / low severity** — a retired slug now behaves observably differently from an unrelated unknown slug (301 vs. falling through to the SPA/404), which lets a caller distinguish "this used to be an artist" from "never was one." This is the intended behavior of the feature and mirrors the already-shipped `artist-page-static.ts` pattern, so it isn't new exposure from this PR, and the leaked information (a retired slug string) is low sensitivity. Flagging only because enumeration was explicitly in scope for this review.
- RLS: no new table/column in this diff — `artist_slug_aliases` already ships RLS-enabled with no policies (server-only, service-role reads only) from a prior migration, untouched here. No finding.
- No secrets, no new outbound `fetch()` of a user-controlled URL (this is a Supabase client call, not an SSRF-relevant HTTP fetch), no auth/authorization surface touched.

## Regression

- **plausible** — the single 5-second `AbortController` timeout (`release-page.ts:154`) now covers up to two sequential Supabase round trips (`artists` lookup, then `artist_slug_aliases` lookup) whenever the artist slug misses, instead of one. Under slow-DB conditions this raises the odds of hitting the abort mid-alias-lookup. It fails safe either way — the outer `catch` (unchanged) falls through to `context.next()` on abort — so this is a latency/availability nit, not a broken-page risk.
- **no finding** — the case-sensitivity mismatch between the primary lookup (`artistSlug`, case-sensitive) and the alias lookup (`artistSlug.toLowerCase()`) only ever makes the alias match a superset of what case-sensitive matching would, so it can't regress a previously-working slug; it mirrors `artist-page-static.ts`'s existing behavior.
- **no finding** — `release-detail.ts` changes are comment-only (updating stale prose that described `release-page.ts` as not resolving aliases, since it now does); no API contract or logic changed.
- **no finding** — `scripts/preview-release-page.ts`: the stub Supabase client now tracks `.eq()` filters so the artists-table read can actually miss (previously it always hit, making the alias branch unreachable in the harness) — this is a dev-only fix, not shipped code, and the change looks correct (Bandcamp fetch is correctly gated to only the canonical-slug request path via `requestedArtist === artistSlug`, so the retired-slug demo link doesn't cause a spurious fetch).
- No test coverage was added for the new `release-page.ts` branch, but this matches the existing project pattern — edge functions aren't in `api/tsconfig.json`'s typecheck include or the Vitest suite, and the structurally-identical alias logic in `artist-page-static.ts` has no test either. Verification here relies on the manual `npm run preview:release` run described in the PR body, consistent with how this codebase already treats edge functions.

---

PR #424, opened 2026-08-07
Let the deadline, not a fetch count, stop the release-detail pass

brandonlucasgreen:
~1,050 release sources in the catalogue have no price. This is why.

## What was actually happening

`MAX_DETAIL_FETCHES_PER_RUN = 100` is invocation-wide and shared across a batch of up to `MAX_ARTISTS_PER_RUN = 25` — an average of **4** release-page reads per artist, against a per-artist cap of **40**. So the artists late in every sweep were reached and then left unpriced.

Meanwhile the invocation was nowhere near its own deadline. Reconstructing per-invocation throughput from `release_catalog_state` (clustering artists by `last_attempted_at`), the last 20 scheduled sweeps show a **median of 102 releases detailed in a 5–7 minute span**, against a 9-minute detail deadline and a 15-minute Netlify ceiling. It stopped because it ran out of *count*, not time — we were using about two thirds of the wall clock already allotted.

Cross-tabbing every `release_sources` row by "has `detail_checked_at`" × "has any `release_offers`" (production, 2026-08-07):

| platform | never read | read, no price | read, priced |
|---|---|---|---|
| bandcamp | **867** | 31 | 3,947 |
| faircamp | **121** | 0 | 55 |
| discogs | **43** | 0 | 796 |
| mirlo | 7 | 0 | 32 |
| jamcoop | 0 | 0 | 136 |

Once a source is read it gets a price 99%+ of the time, so nothing here is a parser problem and no re-scrape of already-read pages is needed. The 988 unread bandcamp+faircamp sources sit on just 141 artists, mean 7 each — and the largest catalogue in the whole database is 25 releases, so the per-artist cap of 40 has never once been the binding constraint. It is left alone.

## The change

**Bandcamp: `MAX_DETAIL_FETCHES_PER_RUN` 100 → 300**, so the 9-minute deadline is what stops the pass — which is what it was always meant to be. 300 is a backstop, not a target: the deadline leaves ~325s for detail work once the batch's fixed overhead is paid (~215s of inter-artist sleeps, grid fetches, MusicBrainz's mandatory ~1/sec, Discogs at 2.6s), which is ~158 fetches at the measured pace. 300 sits above that and well below the 540 the 1s spacing alone would permit in 9 minutes.

**The rate is untouched.** Raising a *count* under a fixed deadline cannot extend the invocation or the request rate — the pass stops *starting* fetches at the deadline whatever the count says. Pacing stays at ~1 request/second and there is a test that fails if it doesn't. Raising the rate would be the unsafe axis and isn't done here.

**The log line now names which limit fired** (`deadline` vs `run-cap`), so a future drift back into rationing is visible instead of silent.

**Faircamp: purchase pages get their own allowance.** They were fetched *after* the release-page pass and inherited whatever was left of the shared 150, so prices were structurally last in line for a budget the release pages had already spent. Per-artist release cap 20 → 30 to match `FAIRCAMP_MAX_CANDIDATES` in the parser (at 20, the tail of a 23- or 30-release site could never be reached at all — not reached late, never).

## Worth flagging: Faircamp's 121 is mostly *not* starvation

I probed the fully-unread instances live before changing their budgets, and the numbers don't support the budget story. Five of five fully-unread sites emit **no purchase page at all** on their release pages — `music.bedlamsteps.uk`, `www.willchatham.com`, `music.lminiero.it`, `music.axwax.eu`, `music.futzle.com` — while both *partially*-read sites do have one. Faircamp emits no purchase page for a free, unlisted or code-unlocked release, so for ~73 of those 121 sources there is no price to read and never will be. Two more sites (`www.base.at`, `sknob.fr`, ~32 sources) now return zero release candidates from their homepage at all, which is separate markup drift and out of scope here.

Only ~13 sources — `music.kidlightbulbs.com` and `futurecabaret.com` — are the genuinely budget-shaped slice. The budget split above is still right, but it will not move 121.

So this PR also **records the no-purchase-page case as checked**. Leaving `detail_checked_at` null filed a settled fact under "we haven't looked yet", and that conflation is exactly what made these read as starvation in the cross-tab. An unreadable purchase page is still left unstamped for retry — "we couldn't read it" and "there is nothing to read" are different claims, and only the second is cacheable.

## What this costs the sites we crawl

Stated explicitly so it's reviewable:

| | before | after |
|---|---|---|
| Bandcamp release pages per invocation | 100 (count-bound) | ~160 (deadline-bound), 300 hard ceiling |
| Bandcamp requests/day, 4 scheduled sweeps | ~470 | **~700**, worst case **1,300** |
| sustained request rate | ~1/sec, never faster | unchanged |

Worst case 1,300/day is under one request a minute averaged, and never more than one a second instantaneously, against robots-permitted paths.

The extra cost is also self-limiting: a release page is re-read only every 30 days, so once the backlog drains the raised cap goes largely unspent — steady state is roughly 170 re-reads a day, which the old 470/day already covered. The spend is the backlog, not a permanent increase.

## Why not the cron cadence

`recatalog-sweep.ts` argues "the lever for total throughput is the cron cadence, not this number", and that's still true — for **artists reached**. It was not true for **releases priced**, which is a different quantity starved somewhere else entirely. Raising the cadence would have re-reached the same artists sooner and starved them again, at four times the request rate. The cadence stays at every 6 hours; I've recorded the distinction in both files so the next person doesn't have to re-derive it.

## Verification

- `npm run lint` — clean for these files (the repo's 71-problem baseline is pre-existing and untouched)
- `npm run test:api` — 965 passing, including 7 new tests in `catalog-detail-budget.test.ts`
- `npm run typecheck:api` — clean (`catalog-artist-background.ts` is in the include list)
- `npm run ingest:try -- kidlightbulbs --detail=3` — real fetch/parse/map path, prices read correctly
- Every new test was checked for teeth by reverting each constant and confirming the matching failure

The tests drive the real handler with real parsers and the real allowlist, faking only the network, the database and the clock — so the budget arithmetic and pacing run for real.

**Heads up, unrelated:** `apps/web/tests/unit/RichArtistProfile.test.tsx` ("renders the payout percent on each pill") fails on `main` too — verified by checking main out clean. Since `npm run build` runs the unit tests, main's build is currently red independently of this PR. Not touched here.

## After deploy

Re-run the cross-tab in a few days; the number that matters is **never read**, not "no price". The drain is gradual by design — a budget change only takes effect for artists whose 7-day cooldown has expired, and 796 of the 2,672-artist pool are inside it right now.

### DIFF
--- a/.github/workflows/recatalog-sweep.yml
+++ b/.github/workflows/recatalog-sweep.yml
@@ -22,10 +22,23 @@ name: Release catalogue sweep
 #   - The ~2,400 artists with no catalogue at all drain in about three weeks.
 #
 # What it costs the sites we crawl: the background function paces itself at one second between
-# artists, at most 100 Bandcamp release-page fetches per run on top of one grid page each. Four
-# runs is on the order of 400-500 sequential Bandcamp requests a day — well under one a minute
-# averaged, against robots-permitted paths. The lever if the tail needs to be tighter is this
-# cadence, not the batch size, which is capped by what one invocation can do.
+# artists and one second between release-page fetches, and stops starting them nine minutes into
+# a fifteen-minute invocation. At the response times measured on 2026-08-07 that deadline allows
+# roughly 160 release pages per run, on top of one grid page per Bandcamp artist (~17 of a
+# 25-artist batch). Four runs is therefore on the order of 700 sequential Bandcamp requests a day,
+# with a hard ceiling of 1,300 if Bandcamp ever answers fast enough to reach the 300-fetch
+# backstop — around one request a minute averaged, and never more than one a second, against
+# robots-permitted paths.
+#
+# That was ~470/day until the detail cap was raised from 100, which had been rationing one
+# invocation's release-page reads across the whole batch and leaving the artists late in each
+# sweep unpriced. The extra cost is also self-limiting: a release page is re-read only every 30
+# days, so once the backlog drains the raised cap goes largely unspent and the steady state is
+# roughly 170 re-reads a day.
+#
+# The lever if the *artist* tail needs to be tighter is this cadence, not the batch size, which is
+# capped by what one invocation can do. Note the two are different problems — see the
+# SWEEP_BATCH_SIZE comment in api/functions/recatalog-sweep.ts.
 #
 # Safe to over-run. A duplicate invocation finds the same artists claimed and does nothing, so
 # re-running by hand is free.
--- a/api/functions/__tests__/catalog-detail-budget.test.ts
+++ b/api/functions/__tests__/catalog-detail-budget.test.ts
@@ -1,0 +1,345 @@
+// What stops the release-detail pass — and what must not.
+//
+// The pass reads one page per release for dates, formats and prices, so it is the one part of
+// ingest whose cost scales with catalogue size rather than artist count. Two limits bound it: a
+// wall-clock deadline (`DETAIL_BUDGET_MS`) and an invocation-wide fetch count
+// (`MAX_DETAIL_FETCHES_PER_RUN`). **The deadline is meant to be the one that fires.**
+//
+// It wasn't. The count sat at 100 and was shared across a batch of up to 25 artists — 4 fetches
+// each against a per-artist cap of 40 — so the artists late in every batch got no detail pass at
+// all, while the invocation finished in a 5-7 minute span against a 9-minute deadline. Measured
+// against production on 2026-08-07 that left 1,057 `release_sources` never read, 867 of them
+// Bandcamp. Once a source is read it gets a price 99%+ of the time, so nothing here was a parser
+// problem; it was rationing.
+//
+// These tests pin both halves: the count no longer rations a batch, and the deadline still stops
+// the pass. Pacing is asserted too, because the fix was to raise a *count* while leaving the
+// request *rate* alone — the reverse would be the outage this codebase has had before.
+//
+// Timers are faked so the 1s-per-request pacing costs no real seconds. That means Date.now()
+// advances by exactly the pacing the code asks for, which is what makes the deadline assertions
+// meaningful rather than incidental.
+
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+
+const mocks = vi.hoisted(() => ({
+  claimArtistForCatalog: vi.fn(),
+  getArtistForCatalog: vi.fn(),
+  persistReleases: vi.fn(),
+  persistFaircampReleases: vi.fn(),
+  persistReleaseDetail: vi.fn(),
+  recordCatalogOutcome: vi.fn(),
+  attachDiscoveredSource: vi.fn(),
+  persistDiscogsReleases: vi.fn(),
+  persistJamcoopReleases: vi.fn(),
+  persistMirloReleases: vi.fn(),
+  persistMusicBrainzEnrichment: vi.fn(),
+  safeFetch: vi.fn(),
+  fetch: vi.fn(),
+}));
+
+vi.mock('../db', () => ({
+  claimArtistForCatalog: mocks.claimArtistForCatalog,
+  getArtistForCatalog: mocks.getArtistForCatalog,
+  persistReleases: mocks.persistReleases,
+  persistFaircampReleases: mocks.persistFaircampReleases,
+  persistReleaseDetail: mocks.persistReleaseDetail,
+  recordCatalogOutcome: mocks.recordCatalogOutcome,
+  attachDiscoveredSource: mocks.attachDiscoveredSource,
+  persistDiscogsReleases: mocks.persistDiscogsReleases,
+  persistJamcoopReleases: mocks.persistJamcoopReleases,
+  persistMirloReleases: mocks.persistMirloReleases,
+  persistMusicBrainzEnrichment: mocks.persistMusicBrainzEnrichment,
+}));
+
+vi.mock('../safe-fetch', () => ({
+  safeFetch: mocks.safeFetch,
+  safeHostname: (u: string) => {
+    try {
+      return new URL(u).hostname;
+    } catch {
+      return '<unparseable>';
+    }
+  },
+}));
+
+// The parsers, the allowlist and the budget arithmetic are all real. Only the network and the
+// database are stubbed.
+import { handler } from '../catalog-artist-background';
+
+const SECRET = 'test-secret-value';
+
+/** A grid page carrying `count` releases, in the shape ingestBandcampGrid actually reads. */
+function gridPage(count: number): string {
+  const items = Array.from({ length: count }, (_, i) => {
+    const n = i + 1;
+    return `<li data-item-id="album-${n}" data-band-id="203035041" class="music-grid-item">
+      <a href="/album/release-${n}">
+        <div class="art"><img src="https://f4.bcbits.com/img/a105950668${n}_2.jpg" alt="" /></div>
+        <p class="title">Release ${n}</p>
+      </a>
+    </li>`;
+  });
+  return `<html><body><ol class="editable-grid music-grid">${items.join('\n')}</ol></body></html>`;
+}
+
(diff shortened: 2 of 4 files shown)


brandonlucasgreen (commented):

**Security**
- Auth/authorization: n/a — `catalog-artist-background.ts` is the existing internal background function, still gated the same way (`INTERNAL_FUNCTION_SECRET` bearer check, unchanged by this diff). *confirmed*
- Input validation / injection: no new external input handling. All changes are to numeric constants, budget bookkeeping, and control flow in an already-authenticated internal path. *confirmed*
- Secret handling: nothing touched.
- Unsafe defaults / data exposure: the new "record a release with no purchase page as checked" branch (`catalogFaircampPrices`) writes `{ releaseDate: null, datePrecision: 'unknown', status: null, offers: [] }` rather than fabricating a price — verified this can't be read as "free"/"name your price" by anything downstream; it's an explicit empty-offers state, matching the existing empty-offer rendering on the release page (`release-page.ts`'s "Price information not available" vs "No formats listed" branches, keyed off `detail_checked_at`). No issue.

**Regression**
- Verified `MAX_FAIRCAMP_RELEASES_PER_ARTIST` (raised 20→30) matches `FAIRCAMP_MAX_CANDIDATES` in `release-ingest.ts:588` exactly, as the comment claims — checked directly, they're both 30.
- The Faircamp purchase-page budget split (`priceFetchesLeft` separate from `fetchesLeft`) is correct: `catalogFaircampPrices` now decrements its own counter and only after the `needsDetail`/no-purchase-page short-circuits, so a release that doesn't need a check or has no purchase page never touches the budget at all. No off-by-one against the old combined-budget behavior.
- The `deadline` vs `run-cap` log-line distinction in `catalogDetails` is a plain ternary on `pastDeadline` computed before the OR — no logic change to *when* the loop stops, only to what gets logged, so it can't itself introduce a regression in stopping behavior.
- State transition risk I looked for and didn't find: raising `MAX_DETAIL_FETCHES_PER_RUN` to 300 while `DELAY_BETWEEN_DETAIL_FETCHES_MS` stays at 1,000ms could only extend wall-clock spend, not request rate — and the new test (`still paces release-page requests about a second apart`) asserts the minimum gap directly with faked timers, so this isn't just asserted in the PR description, it's pinned.
- Edge case covered by tests: a batch whose due-release count would have exceeded the old cap of 100 (3×40=120) now completes every artist, including the last one — the actual failure mode described in the PR body.

No confirmed or plausible issues found. The one thing I'd flag as **needs-manual-verification** rather than a defect: this raises sustained Bandcamp request volume to a worst-case ~1,300/day (vs ~470 today) if Bandcamp ever answers fast enough to hit the 300-fetch backstop instead of the 9-minute deadline. The PR's own math argues this is self-limiting (30-day re-read cadence, steady state ~170/day) and stays within robots-permitted paths at ≤1 req/sec — I can't verify actual Bandcamp response times from here, but the reasoning is sound and the rate-limiting invariant is test-enforced, so this reads as an accepted, well-documented tradeoff rather than a gap.

---

brandonlucasgreen (commented):
@brandonlucasgreen

Reviewed the diff (`api/functions/catalog-artist-background.ts`, `api/functions/recatalog-sweep.ts`, `.github/workflows/recatalog-sweep.yml`, `api/functions/__tests__/catalog-detail-budget.test.ts`) against the local checkout with full file context, not just the hunks.

## Security

- No new outbound-fetch paths: every fetch in the touched code still goes through the existing `safeFetch(url, 10_000)` / `isFetchableReleaseUrl` (host-match-or-allowlist) pattern already in the file — unchanged by this PR.
- `safeFetch` (`api/functions/safe-fetch.ts`) uses a real `AbortController` timeout, so a hung upstream response can't block the deadline check indefinitely — each iteration is bounded to ~11s worst case (1s pacing + 10s timeout), not unbounded.
- No new Supabase tables/columns, so no RLS gap.
- No secrets touched; `MIRLO_API_KEY` handling is pre-existing and untouched.
- Auth (`isInternalRequest` shared-secret check, `isCatalogEnabled` gate) is untouched.
- `artistIds` is still server-side bounded to `MAX_ARTISTS_PER_RUN` (25) via `.slice(...)` regardless of what the caller sends, so the 3x rise in `MAX_DETAIL_FETCHES_PER_RUN` (100→300) doesn't create a new caller-controlled amplification vector — it only changes how much of an already-bounded, already-authenticated batch gets spent, and the ~1 req/s pacing (`DELAY_BETWEEN_DETAIL_FETCHES_MS`) is explicitly left untouched and is asserted by a new test (`'still paces release-page requests about a second apart'`).

## Regression

- **needs-manual-verification** — Faircamp "no purchase page" now gets **stamped as checked** (`persistReleaseDetail` with empty offers) instead of left unstamped forever, in `catalogFaircampPrices` (~line 774 in the PR version, the `if (!purchaseUrl) { ... continue; }` branch). This is deliberate and well-argued (PR body cites live probing of 5 production instances), and the diff correctly keeps the *fetch-failure* case (unreadable purchase page) unstamped for retry — covered by the new test `'leaves a release unstamped when its purchase page cannot be read'`. The residual risk is that this classification rests entirely on `ingestFaircampReleasePage`'s `purchaseHref` extraction (in `release-ingest.ts`, not touched by this PR) being correct — if that parser ever silently misses a purchase link that does exist, the release would be marked "checked, no price" for a 30-day cooldown instead of retried next run, the same "cache uncertainty as certainty" failure mode this codebase has been bitten by before. Worth a human sanity check against a live example, since it can't be fully verified from the diff alone.
- **plausible, low severity** — Same branch: `persistReleaseDetail` is called for the no-purchase-page case with no budget/deadline check guarding it (it's a DB write, not a fetch, so this is intentional and correct), but it means an artist with many such releases writes N rows in a tight loop with no pacing between them. Almost certainly fine (DB writes, not the resource being rate-limited), flagging only because it's a new code path that wasn't exercised before.
- **No other regressions found.** Verified specifically, per your review brief:
  - **Deadline hit mid-loop**: `catalogDetails`'s per-iteration budget check (`budget.fetchesLeft <= 0 || Date.now() > budget.deadline`) runs *before* starting each fetch, and `persistReleaseDetail` is called per-release inside the loop, so a break mid-loop just leaves the remaining `due` releases with `detail_checked_at` unset — picked up by `needsDetail` next run. No partial/inconsistent row state.
  - **Off-by-one / unit mismatch**: `DETAIL_BUDGET_MS = 9 * 60_000` (ms) compared consistently against `Date.now()` (ms); `fetchesLeft` decrements from 300 to exactly 0 over 300 iterations, matching the new test's `expect(detailFetchCount()).toBe(300)`. No mismatch found.
  - **Hung fetch blocking past deadline**: covered above under Security — `safeFetch`'s `AbortController` timeout bounds this.
  - **Faircamp budget split** (`fetchesLeft` for release pages vs. new `priceFetchesLeft` for purchase pages): both are decremented independently and initialized once in the handler, not reset per-artist — verified the shared-budget-across-batch invariant still holds, matching the existing pattern for the other budgets (Discogs, Jam.coop, Mirlo) that weren't touched.
  - `recatalog-sweep.ts` and the workflow YAML changes are comment-only (verified with `git diff ... | grep -v '^[+-]#'`) — no functional/cron-schedule drift.
  - `MAX_FAIRCAMP_RELEASES_PER_ARTIST` raised to 30 to match `FAIRCAMP_MAX_CANDIDATES` in `release-ingest.ts` — confirmed that constant is in fact 30, so the comment's claim is accurate.
  - The response contract (`{ catalogued, skipped }`, status codes) is unchanged.
  - `catalog-artist-background.ts` and `recatalog-sweep.ts` are both in `api/tsconfig.json`'s narrow typecheck `include` list, so this diff is covered by `npm run typecheck:api`, not just function tests.

Note: I could not execute the test suite in this sandbox (no `node_modules` available), so the above is a static/manual trace of the budget arithmetic and the new `catalog-detail-budget.test.ts` assertions rather than a confirmed test run — the test file itself (7 new tests) reads as a solid regression suite for exactly the failure modes it targets.

---

PR #425, opened 2026-08-07
Stop a long release title scrolling the artist page sideways

brandonlucasgreen:
Follow-up to the same defect fixed in `RecentReleasesSection` (#421). `ReleasesSection` — the release list on `/a/:slug` — has structurally identical markup and the same bug.

## What was wrong

The release rows are grid items inside `grid gap-2 sm:grid-cols-2`. A grid item's automatic minimum size is its content, and the title carries `truncate` (which sets `white-space: nowrap`), so a long title sized the whole row to its own text rather than shrinking inside its track. `min-w-0` was already on the inner text column; that isn't enough — it has to be on the grid item itself.

## Measured, not eyeballed

Rendered through a throwaway Vite harness in the artist page's real container chrome (`main.px-4` → `max-w-4xl mx-auto`), inside fixed-width iframes so each one is genuinely its own viewport (`resize_window` doesn't change `documentElement.clientWidth` in the browser pane). Fixture: a long release title plus a full meta line (`Album · 1 June 2024 · from $28 · ≈$22.40–$23.80 to artist`).

| viewport | | `clientWidth` | `scrollWidth` | grid track | row width |
|---|---|---|---|---|---|
| 375px | before | 375 | **569** | 343 | **553** |
| 375px | after | 375 | 375 | 343 | 343 |
| 414px | before | 414 | **569** | 382 | 553 |
| 414px | after | 414 | 414 | 382 | 382 |

Titles still truncate with an ellipsis and the meta line wraps as before; only the sideways page scroll is gone.

## Test

A regression test in `ReleasesSection.test.tsx` asserts the class, since jsdom does no layout and reports `scrollWidth` as 0 either way. Verified it has teeth: removing `min-w-0` from the component fails it.

The harness files were deleted; only the component and its test are in the diff.

## Checks

- `npx vitest run apps/web/tests/unit/ReleasesSection.test.tsx` — 17 passed
- `npm run test:unit` — 641 passed, 1 failed: `RichArtistProfile › renders the payout percent on each pill`, which fails identically on `main` with this branch stashed. Pre-existing, not from this change.
- `npm run lint` — 71 problems, the repo's existing baseline; none in the files touched here.

### DIFF
--- a/apps/web/src/components/ReleasesSection.tsx
+++ b/apps/web/src/components/ReleasesSection.tsx
@@ -157,10 +157,14 @@ function ReleaseRow({ release, artistSlug }: { release: Release; artistSlug: str
   // itself, so the platform a fan sees leading the row is also the one the price came from.
   const platforms = orderedSourcePlatforms(release.sources);
 
+  // `min-w-0` on the row itself, not only on the text column inside it. This anchor is a grid
+  // item, and a grid item's automatic minimum size is its content — the title's `truncate` sets
+  // `white-space: nowrap`, so a long title made the row as wide as its own text and pushed the
+  // whole page sideways: 569px of scroll in a 375px viewport before this class.
   return (
     <a
       href={`/a/${encodeURIComponent(artistSlug)}/${encodeURIComponent(release.slug)}`}
-      className="flex items-center gap-3 px-3 py-2 rounded-xl border border-border hover:bg-bg-hover transition-colors"
+      className="flex items-center gap-3 px-3 py-2 rounded-xl border border-border hover:bg-bg-hover transition-colors min-w-0"
     >
       {release.artworkUrl ? (
         <img
--- a/apps/web/tests/unit/ReleasesSection.test.tsx
+++ b/apps/web/tests/unit/ReleasesSection.test.tsx
@@ -166,6 +166,19 @@ describe('ReleasesSection', () => {
     show([release()], 1);
     expect(screen.queryByText(/more$/)).toBeNull();
   });
+
+  // Measured in a real browser at a 375px viewport: without `min-w-0` on the row anchor the
+  // document's scrollWidth was 569px against a 343px grid, so the whole artist page scrolled
+  // sideways on a phone. The row is a grid item, and a grid item's automatic minimum size is its
+  // content — the title's `truncate` sets `white-space: nowrap`, so a long title sized the row to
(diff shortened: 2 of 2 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen

Reviewed the full diff (`apps/web/src/components/ReleasesSection.tsx` + `apps/web/tests/unit/ReleasesSection.test.tsx`, +18/-1) against the local checkout, including the surrounding file for context beyond the hunk.

## Security

No security concerns found in this diff. The change is a single Tailwind class (`min-w-0`) added to an existing `<a>` element's `className` string, plus an explanatory comment and a unit test asserting the class is present. No network fetches, no new/changed data flow, no HTML string interpolation (this is JSX, auto-escaped), no auth/authorization paths, no Supabase schema or query changes, no secrets touched. Nothing here goes near `isUrlHostnameAllowed`/SSRF surface, edge-function HTML escaping, or caching logic.

## Regression

- **needs-manual-verification**: The fix relies on `min-w-0` overriding the grid item's automatic minimum size so the title (`truncate`, `white-space: nowrap`) shrinks instead of forcing row/page width. This is a real, well-known CSS mechanism and the PR description includes real-browser measurements (`scrollWidth` 569→375 at 375px), but the accompanying unit test (`ReleasesSection.test.tsx`) only asserts the class string is present in `className` — jsdom does no layout, so the test can't itself catch a regression in the actual overflow fix (e.g. if a future change reordered/removed a sibling class in a way that reintroduced the bug via specificity or a conflicting utility). This is explicitly acknowledged in the PR body/comment, so it's not a diff defect, just a residual gap a human should be aware of — worth a quick manual check in a real browser/deploy preview at a narrow viewport, since that's the only thing that can actually confirm it.
- **confirmed (non-issue)**: The class is added only to the row-level `<a>` (`ReleaseRow`), consistent with the existing `min-w-0` already present one level down on the inner `<span className="flex-1 min-w-0">` (line 183 in the current file) — no conflicting or duplicate state, and `shrink-0` on the artwork/badge siblings is preserved, so the artwork and "Coming" badge should still hold their fixed size while only the text column shrinks. No behavioral drift beyond the intended width fix.
- **confirmed (non-issue)**: Scope is narrow — `ReleaseRow` is only rendered from `ReleasesSection`, which itself is only used by `RichArtistProfile.tsx` (per a repo-wide grep for `ReleaseRow`/`ReleasesSection`), so there's no other call site whose layout assumptions this could silently break.

---

PR #426, opened 2026-08-07
Link saved artists to the artists table, and see the ones already unlinked

brandonlucasgreen:
Brandon saved Rodney Owl, whose newest record came out yesterday, and Recent Releases stayed empty while his artist page listed it.

The save recorded `artist_slug: 'rodneyowl'` with `artist_id: null`. The artists row is `rodney-owl`. Nothing linked them, and every feature keyed on the artist was blind to the save.

**Measured on production: 25 of 37 live saved rows are in that state — 68%.**

## Three symptoms, one cause

1. **Recent Releases and the ICS/Atom feeds can't see those artists.** `getFeedReleasesForUser` selected `saved_artists.artist_id` and nothing else. This is **not** a bug in #421 — it faithfully reuses the feed query; the hole has silently limited the feeds since #388. The dashboard is just where it first became visible.
2. **Saving never triggers cataloguing for them.** `saved-artists.ts` guards `if (artistId) await requestArtistCatalog(...)`, so "a save is the strongest signal that someone wants this artist's releases" was dead for two thirds of saves.
3. **The artist page shows "Save" for an artist you just saved** — `savedArtistIds` holds `rodneyowl`, the page asks `isArtistSaved('rodney-owl')`.

## The write side: one card, two identities

`ResultCard` saved under `result.claimedSlug || result.id`. `result.id` is a search-pipeline key — `nameonly-…`, `known-…`, `qobuz-…`, or a platform handle like `rodneyowl` — and matches no row in the artists table, so `findExistingArtist` missed and the row was written unlinked.

**The slug was already in the payload.** `attachArtistPageSlugs` puts `knownSlug` on every placeable result, server-side, precisely so no client has to derive it — and `ResultCardHeader` has always used it to build the `/artist/` link. So the same card linked to `/artist/rodney-owl` and saved under `rodneyowl`.

It is now **one `const` rather than two matching expressions**, because it was two: the save handler and the saved-state effect each derived the key, so fixing one and not the other leaves the heart dark on an artist you just saved. Reverting the fix fails 4 of the 6 new tests, including that one.

**Neither other client needed changing** — checked rather than assumed. The Mac's `SavedArtistsSync.saveArtist` has no callers (it only pulls), and `popup.js:154` already had the three-way expression.

## The read side: recover the rows that are already wrong

Fixing the client does nothing for 25 existing rows, and a fan's calendar shouldn't stay wrong until a backfill runs. `savedArtistIdsForUser` now resolves `artist_id` → exact `artist_slug` → `artistSlug(artist_name)` → `artist_slug_aliases`.

### Reviewer note: my first cut was wrong, and only production data showed it

I designed this as "stored slug + alias table". Twelve green tests — and it recovered **zero** artists for the reporting account. **Not one** unlinked slug matches an artists row or an alias directly; they are squashed *names* (`rodneyowl`, `seoulmetro`, `modelactriz`) and prefixed keys (`qobuz-robertlogan`), so `artistSlug(artist_name)` is the only thing that recovers them.

**The guard that makes the name path safe:** a name is a far weaker key than a slug, so a name-derived hit must prove itself — the found artist's own name has to normalize identically. Without it a row with a generic name ("Music") adopts an unrelated artist's releases, putting someone else's record in a fan's calendar. Worse than showing nothing.

## `deleted = false`, which was also missing

`saved_artists` uses tombstones (migration 017), so an artist you *unsaved* kept feeding your dashboard and your calendar. Kid Lightbulbs was doing exactly that.

## Verified against production, not only mocks

`getFeedReleasesForUser` for the reporting account: **3 releases before, 4 after** — the new one being Rodney Owl — *Loneliness Is Contagious* (2026-08-06). The tombstoned artist is correctly gone.

## Not in this PR

Backfilling the 25 rows' `artist_id`. That is a production data write and wants a dry run printing every proposed link first.

## Verification

`test:api` **998** ✅ · `test:unit` **692** ✅ · `typecheck:api` ✅ · lint unchanged at the 71-problem baseline.

Five load-bearing behaviours mutation-tested, each failing exactly its intended test: the `deleted` filter, slug resolution, alias resolution, name derivation, and the name-agreement guard.

New tests drive db.ts against a recording fake Supabase client (following `recatalog-sweep-selection.test.ts`) so the real query body runs — a module mock would assert my own fake.

### DIFF
--- a/api/functions/__tests__/saved-artist-resolution.test.ts
+++ b/api/functions/__tests__/saved-artist-resolution.test.ts
@@ -1,0 +1,226 @@
+// Which saved artists a fan's feed can actually see.
+//
+// The bug this guards, found 2026-08-07 the hour /dashboard's Recent Releases shipped: the feed
+// keyed on `saved_artists.artist_id` alone, and **25 of 37 live rows have it NULL** because a save
+// made from a search result sent a synthetic key (`rodneyowl`, `qobuz-pearljam`, `nameonly-…`)
+// that matched no artists row. Two thirds of every fan's saved list was invisible to their feed,
+// their calendar and their dashboard — and it looked exactly like "those artists have nothing new".
+//
+// Driven against a recording fake Supabase client rather than a module mock, following
+// recatalog-sweep-selection.test.ts, so the real query body runs — the `deleted` filter and the
+// fallback chain are the things under test, and a module mock would assert my own fake.
+
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+
+interface Filter { kind: 'eq' | 'in' | 'gte' | 'order' | 'limit'; column?: string; value?: unknown }
+interface Query { table: string; columns: string; filters: Filter[] }
+
+const queries: Query[] = [];
+
+const tables: Record<string, Record<string, unknown>[]> = {
+  saved_artists: [],
+  artists: [],
+  artist_slug_aliases: [],
+  releases: [],
+};
+
+/** Set to a table name to make that table's read fail. */
+let failingTable: string | null = null;
+
+function makeClient() {
+  return {
+    from(table: string) {
+      return {
+        select(columns: string) {
+          const query: Query = { table, columns, filters: [] };
+          queries.push(query);
+
+          const builder = {
+            eq(column: string, value: unknown) { query.filters.push({ kind: 'eq', column, value }); return builder; },
+            in(column: string, value: unknown) { query.filters.push({ kind: 'in', column, value }); return builder; },
+            gte(column: string, value: unknown) { query.filters.push({ kind: 'gte', column, value }); return builder; },
+            order(column: string, value: unknown) { query.filters.push({ kind: 'order', column, value }); return builder; },
+            limit(value: number) { query.filters.push({ kind: 'limit', value }); return builder; },
+            then(resolve: (r: unknown) => unknown, reject: (e: unknown) => unknown) {
+              if (failingTable === table) {
+                return Promise.resolve({ data: null, error: { message: 'connection reset' } }).then(resolve, reject);
+              }
+              let rows = tables[table] ?? [];
+              for (const f of query.filters) {
+                if (f.kind === 'eq') rows = rows.filter(r => r[f.column as string] === f.value);
+                if (f.kind === 'in') {
+                  const wanted = new Set(f.value as unknown[]);
+                  rows = rows.filter(r => wanted.has(r[f.column as string]));
+                }
+              }
+              return Promise.resolve({ data: rows, error: null }).then(resolve, reject);
+            },
+          };
+          return builder;
+        },
+      };
+    },
+  };
+}
+
+vi.mock('@supabase/supabase-js', () => ({ createClient: () => makeClient() }));
+
+process.env.SUPABASE_URL = 'https://example.supabase.co';
+process.env.SUPABASE_SERVICE_KEY = 'service-key';
+
+const { getFeedReleasesForUser } = await import('../db');
+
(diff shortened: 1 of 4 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Staff-eng security + regression pass on this one.

## Security

**Auth/authorization — no issues found.** Traced all three `getFeedReleasesForUser` call sites: `me-recent-releases.ts` requires a valid bearer token and uses the resulting `data.user.id` (never client-supplied); `feed-releases.ts`'s token route resolves `userId` server-side via `getFeedTokenOwner` before calling; the `/u/{handle}` route gates on `usernames.saved_artists_public` and 404s uniformly for unknown-or-unshared handles. `savedArtistIdsForUser` only ever resolves slugs against rows already scoped to `user_id = userId` — it can't leak another user's saved list.

**Input validation — no issues found.** All new Supabase calls (`.eq()`, `.in()`) go through the query builder, not string interpolation, so there's no PostgREST-filter injection surface. `artistSlug()` normalizes to `[a-z0-9-]` before use as a lookup key.

**Injection risks — no issues found**, same reasoning as above.

**Secret handling — not applicable**, no secrets touched in this diff.

**Data exposure — plausible, low severity, needs-manual-verification.** `savedArtistIdsForUser` queries the public `artists` table by slug candidates derived from a user's own saved rows and returns ids merged into one list with no signal about which candidate resolved which row. Since `artists` rows are already public data (rendered on `/a/:slug`), this isn't cross-user exposure, but worth a second look if `artists` ever grows a field that shouldn't be broadly queryable — the `.in('slug', [...candidates])` has no cap today because the comment reasons "one person's saved list, not a table scan." That's true now (measured 37 live rows) but isn't structurally enforced.

**Unsafe defaults — no issues found.** On `slugError` the function returns the ids already resolved rather than swallowing the whole batch or guessing (`api/functions/db.ts` ~L4355-4361), consistent with the codebase's "never cache uncertainty" rule — nothing here writes a negative result anywhere.

## Regression

- **Bot flagged a possible semantic revert** (8 lines removed from `db.ts` matching additions from #415/#406/#404/#388/#376/#360). Checked by hand: the removed block was the old inline `saved_artists` read + error guard in `getFeedReleasesForUser`; the replacement (`savedArtistIdsForUser`, ~L4407-4413) preserves the same guard semantics — a read failure still short-circuits to `[]` via the `null` return — while adding the slug/alias/name fallback and the `deleted = false` filter. **Confirmed false positive**, not an actual revert.
- **Tombstone filter was genuinely missing before this PR** (`deleted = false` on the `saved_artists` read) — an unsaved artist was silently still feeding feeds/calendar. Now fixed and covered by the "ignores an unsaved (tombstoned) artist" test. Good catch.
- **Name-derived match guard** (`artistSlug(byName.name) === derived`) correctly prevents a weak/generic saved name from adopting an unrelated artist's release calendar — confirmed by the "refuses a name-derived match whose artist name does not agree" test. This is the right shape of defense for a fallback this permissive.
- **Client saveKey unification** (`result.claimedSlug || result.knownSlug || result.id`) — confirmed the same expression now drives save, remove, and the saved-state read effect in `ResultCard.tsx`, closing the "heart stays dark" desync the PR description calls out. `knownSlug` is server-populated via `attachArtistPageSlugs`, not client-derived, so no new client-trust surface here.
- **Edge case not covered by this PR (acknowledged in the description):** the 25 existing production rows with `artist_id: null` stay unlinked until a separate backfill runs; this PR only stops the bleeding for new saves and recovers read-path visibility. No action needed here, just flagging it's still open.

Nothing blocking. Nice work tracking the mismatch down to a single `const`.

---

PR #427, opened 2026-08-07
Backfill the saved_artists rows that were never linked to an artist

brandonlucasgreen:
Follows #426, which fixed both live paths — the card now saves under `knownSlug`, and `getFeedReleasesForUser` resolves the slug at read time — but left the rows already in the database unlinked.

> ⚠️ **This script has already been run against production.** `25` NULL `artist_id`s became `6`. The PR is the record of how the data got that way, and the tool for next time; merging it changes no behaviour. Following the `backfill-published-artist-rows.ts` precedent of keeping these in the repo.

## Result

| | before | after |
|---|---|---|
| `artist_id` NULL | 25 | **6** |
| linked | 13 | 32 |
| total rows | 38 | 38 |

19 linked, none created or destroyed. The reporting account's feed still returns Rodney Owl — *Loneliness Is Contagious*, now through `artist_id` rather than the read-time fallback.

## Why 6 remain, deliberately

Three resolution rules need no corroboration — an exact slug, an alias, or a saved slug that **is** an `artists.id` (the client sent the UUID; `ArtistPage`'s own comment records that as a past bug, and `5c58da01-…` really is `james-snelling`).

The fourth rule, re-deriving `artistSlug(artist_name)`, is the one that recovers the synthetic keys — and the dry run showed the name-agreement check is **not enough on its own, because the saved name can be junk too**:

| saved slug | saved name | why it's junk |
|---|---|---|
| `nameonly-blixbyrd` | `Music` | resolves to a *real* artists row also called "Music", whose only link is `https://bonkwave.org/music` — a Faircamp page path from the artist-as-release parsing bug |
| `qobuz-mikeposner` | `Qobuz Mikeposner` | `saved-artists.ts` title-cases the slug when the client sends no name, so the name **is** the slug and adds no evidence |

So a name-derived match now also requires the saved slug and saved name to **agree with each other**: strip the synthetic prefix, drop hyphens, and `qobuz-robertlogan` / "Robert Logan" agree while `nameonly-blixbyrd` / "Music" do not. That check rejects exactly the six, and nothing else.

**Not attempted:** re-hyphenating the leftover (`pearljam` → `pearl-jam`). It would recover four more rows by guessing where the words divide, and a wrong guess puts another artist's records in someone's feed. A row left unresolved is a reportable outcome; a wrong link is silent.

### Reviewer note: an earlier cut used the wrong axis

I first refused every `unverified` target. That dropped Puscifer, Cigarettes After Sex and Model/Actriz — all real artists, `unverified` only because nobody had searched them into confidence — while the actual junk was identifiable without it. Confidence measures our certainty about the *artist*; self-agreement tells you whether *this row* is coherent. Worth knowing, because `unverified` is a tempting filter.

## Safety properties

- **Update-only, one column, by primary key**, with `WHERE artist_id IS NULL` so a concurrent link is never overwritten. No inserts, no deletes.
- **`artist_slug` is not touched.** `(user_id, artist_slug)` is unique and several users hold both spellings, so canonicalizing could collide — and it would change the key the client already holds. This leaves the cosmetic "the artist page still offers Save for an artist you saved" symptom open.
- **Triggers no cataloguing.** 19 artists must not become 19 crawls; the scheduled sweep already covers them.
- **`artistSlug` is imported from db.ts**, not reimplemented, so the script and the shipped read path cannot drift.
- Dry run is the default; `--write` applies. A rollback list of the 25 row ids was captured before writing.

## Also reported, not acted on

One user saved Bird Streets under both spellings, so they now hold two rows for the same artist. Harmless — ids are de-duplicated at read time — but merging rows is a deletion, so it wants its own decision. The script prints it rather than fixing it.

## Verification

`lint` unchanged at the 71-problem baseline · root `tsc` clean. Verified against production: 38 rows before and after, 19 linked, and the end-to-end feed read re-run afterwards.

### DIFF
--- a/scripts/backfill-saved-artist-ids.ts
+++ b/scripts/backfill-saved-artist-ids.ts
@@ -1,0 +1,247 @@
+#!/usr/bin/env npx tsx
+/**
+ * Link the `saved_artists` rows whose `artist_id` is NULL to the artist they actually name.
+ *
+ * Why this exists: saving an artist from a search result used to send a synthetic search-pipeline
+ * key (`rodneyowl`, `qobuz-robertlogan`, `nameonly-…`) rather than the artist's page slug, so
+ * `findExistingArtist` missed and the row was written unlinked. Measured 2026-08-07: **25 of 37
+ * live rows**. Every feature keyed on the artist was blind to them — the release feeds, the
+ * /dashboard shortlist, and the `requestArtistCatalog` call that makes a save the strongest signal
+ * to go and crawl someone.
+ *
+ * #426 fixed both live paths: the card now saves under `knownSlug`, and `getFeedReleasesForUser`
+ * resolves the slug at read time so the existing rows already work in the feeds. This script
+ * repairs the stored data itself, so the read-time fallback stops being load-bearing and the
+ * *catalogue* trigger (which has no fallback) starts working for these artists on their next save.
+ *
+ * Usage:
+ *   npx tsx scripts/backfill-saved-artist-ids.ts            # dry run (default) — prints every proposal
+ *   npx tsx scripts/backfill-saved-artist-ids.ts --write    # apply
+ *
+ * Safety properties, in order of how much they matter:
+ *
+ *  - **Update-only, one column.** It sets `artist_id` on rows where it is currently NULL, by
+ *    primary key, one row at a time. It never inserts, never deletes, and never touches any other
+ *    column — including `artist_slug`, see the note below. A row it cannot resolve is left exactly
+ *    as it is. Given the artist_links wipe of 2026-07-29, nothing here removes anything.
+ *  - **The same resolution rules as the shipped read path** (`savedArtistIdsForUser` in db.ts):
+ *    exact slug, then `artistSlug(artist_name)` *if the found artist's own name agrees*, then the
+ *    alias table. `artistSlug` is imported rather than reimplemented so the two cannot drift.
+ *  - **A name-derived match must prove itself.** A name is a far weaker key than a slug; without
+ *    the name-agreement check a row with a generic name ("Music") would adopt an unrelated
+ *    artist's releases, which puts someone else's record in a fan's calendar.
+ *  - **No cataloguing is triggered.** Linking 25 artists must not dump 25 crawls on Bandcamp. The
+ *    scheduled sweep already covers them.
+ *
+ * Deliberately NOT done here: rewriting `artist_slug` to the canonical form. It would fix the
+ * cosmetic "the artist page still offers Save for an artist you saved" symptom, but the table has
+ * a unique key on (user_id, artist_slug) and several users already hold *both* spellings — so a
+ * rewrite can collide, and it would also change the key the client currently has in memory. The
+ * dry run reports those rows so the decision can be made with the list in hand.
+ */
+
+import { createClient } from '@supabase/supabase-js';
+import { artistSlug } from '../api/functions/db';
+
+const WRITE = process.argv.includes('--write');
+
+const url = process.env.SUPABASE_URL;
+const key = process.env.SUPABASE_SERVICE_KEY;
+if (!url || !key) {
+  console.error('SUPABASE_URL and SUPABASE_SERVICE_KEY must be set (source the repo .env).');
+  process.exit(1);
+}
+const client = createClient(url, key);
+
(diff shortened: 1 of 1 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Security/regression review of `scripts/backfill-saved-artist-ids.ts`.

## Security

- **Secret handling — no issue.** `SUPABASE_SERVICE_KEY`/`SUPABASE_URL` come from env, the script hard-fails if unset, and nothing logs the key or raw row PII (only artist slug/name are printed). `confirmed`
- **Injection — no issue.** Every query goes through the supabase-js builder (`.eq`/`.in`/`.is`/`.range`); no string-built SQL. `confirmed`
- **Unsafe defaults — no issue.** Dry-run is the default; `--write` is opt-in, and writes are scoped `WHERE artist_id IS NULL` by primary key (update-only, no inserts/deletes). `confirmed`
- **Auth/authorization — no issue.** This is a local CLI maintenance script gated by possession of the service-role key, not a served endpoint; no authz surface is added.
- **Data exposure — real, but pre-existing and *not* introduced by this diff.** See the "resolution rules have already drifted" finding below — it's a cross-user data-exposure bug in the *already-shipped* `savedArtistIdsForUser`, which this PR's own script logic implicitly reveals. `plausible` (root cause is outside this diff; flagging because this PR is the evidence trail for it)

## Regression

1. **The script's name-match rule is stricter than the live read path it claims to mirror — and the live path has the bug the script works around.**
   The docstring/PR body say the script uses "the same resolution rules as the shipped read path (`savedArtistIdsForUser` in `db.ts`)" and that importing `artistSlug` means "the script and the shipped read path cannot drift." That's only true for the slug-derivation helper — the *matching logic* around it has already drifted:
   - Script (new file, ~line with `slugAndNameAgree`): a name-derived match additionally requires the **saved row's own slug and name to self-agree** (`slugAndNameAgree(slug, row.artist_name)`), which is exactly what correctly rejects `nameonly-blixbyrd` / `"Music"` when a real artist is *also* named "Music".
   - Live path, `api/functions/db.ts` (`savedArtistIdsForUser`, ~lines 4376–4381): only checks `artistSlug(byName.name) === derived` — i.e., does the *found artist's* name match the derived slug. It has no equivalent self-agreement check on the saved row.
   - Net effect: for a saved row shaped like `artist_slug='nameonly-blixbyrd', artist_name='Music'` matching a real artist whose own name is genuinely `"Music"`, the backfill script correctly refuses to persist the link (one of the 6 left unresolved) — but `savedArtistIdsForUser` will still accept that exact match on every live request. That means today, right now, in production, whichever fan holds that row is being shown an unrelated artist's releases in their feed/dashboard/calendar. This PR fixes the data for 19 rows but leaves the shipped code path carrying the bug its own analysis found. `confirmed` (via direct comparison of the two implementations)
   - Note the existing unit test `refuses a name-derived match whose artist name does not agree` in `api/functions/__tests__/saved-artist-resolution.test.ts` uses an artist named `"The Music Tapes"`, not `"Music"` — so it doesn't reproduce the real production shape from the PR description and currently masks this gap.

2. **The UUID/`id` resolution rule can't actually resolve a row on its own — it depends on coincidence within the batch.**
   `byId` is built only from the `artists` rows returned by `.in('slug', [...candidates])`. A raw UUID is never a real `slug` value, so an artist referenced only via `artist_slug = <uuid>` is never fetched by that query, and `byId.get(slug)` can only hit if that *same* artist also happens to be pulled in through a different candidate's real/derived slug elsewhere in the batch. The one production case cited (`5c58da01-… → james-snelling`) most likely worked because another unlinked row in the same run referenced `james-snelling` by its normal slug. Re-run this script later against a batch where a UUID-keyed row is the *only* reference to that artist, and the `id` rule silently produces no match — it falls into "UNRESOLVED — left untouched" with no error, which is safe (no wrong link gets written) but quietly breaks the "tool for next time" claim for that rule. `confirmed` (via code reading — no direct id-based query exists)

Both are fail-safe (no data gets corrupted or wrongly linked by *this script*), so they don't block the already-executed production run described in the PR. But #1 in particular is worth a follow-up: the live `savedArtistIdsForUser` should get the same self-agreement guard the script has, or the cross-user data-exposure gap stays open for the 6 rows this PR deliberately left NULL (and any future row shaped like them).

---

PR #428, opened 2026-08-08
Split the dashboard's releases into Upcoming and Recent

brandonlucasgreen:
## What

`/dashboard`'s "Recent Releases" section listed releases that aren't out yet. This splits it into two sections — **Upcoming Releases** above **Recent Releases** — with releases dated in the future in the first and the past 30 days in the second.

## Why

An album that has been announced but hasn't been released is not a *recent release*, so the heading was wrong on its own terms. The per-row "Coming" badge was papering over it. Brandon:

> They include upcoming releases. We should split this into 2 sections: Upcoming Releases and Recent Releases. Upcoming should include releases with a release date in the future; Recent should include releases in the past 30 days.

## How

**`api/functions/me-recent-releases.ts`** — `toRecentReleases` becomes `splitRecentReleases`, and the response body is `{ upcoming, recent }` rather than `{ releases }`. The window is unchanged: still `getFeedReleasesForUser`'s past 30 days plus everything upcoming, so the feed and the dashboard can't disagree about what counts as recent.

Three decisions worth a reviewer's attention:

- **Split on the date, not `releases.status`** — the same rule as the "Coming" badge it replaces. A row dated next month is upcoming whatever a status column nothing keeps ticking over happens to say. Today counts as *out*.
- **Each list capped at six separately**, not a shared budget of six. A shared cap would let a run of announcements hide everything that actually came out — which is this bug again in miniature.
- **Opposite sorts.** Upcoming stays ascending so its cap trims the furthest off; recent flips to descending so its cap trims the oldest rather than the news.

**`RecentReleasesSection.tsx`** — renders up to two sections, upcoming first. Only a section with rows in it renders; when neither has any, one box explains the silence under the "Recent Releases" heading rather than two boxes saying the same thing. The subscribe control stays a single button riding whichever heading comes first, since both lists come out of the same feed token.

The per-row **"Coming" badge is removed**: it existed to stop an unreleased album vanishing into a list that read as history, and a section heading now says that for every row at once.

## Reviewer notes

- **Response shape changed.** `/api/me/recent-releases` has exactly one consumer (`DashboardPage`), which is why the key could just change. A browser holding a stale pre-deploy SPA chunk would read neither key and show the empty state until it reloads — graceful, but wrong-looking for that short window.
- `apps/web/server/` (the dev shim) still has no route for this endpoint, so `npm run dev` can't show the section. Verified with a throwaway Vite harness instead.

## Testing

- 692 web unit tests, 1002 API tests, `typecheck:api` and the web `tsc -b` all pass. Lint is clean on the touched files (the 71 remaining problems are the repo's existing baseline).
- New coverage: the split itself (an unreleased album stays out of `recent`), the today boundary, both sort directions under a cap, and that the caps are independent. On the web side, that the two lists get their own headings in the right order, that an empty list doesn't render a heading, and that there is still exactly one subscribe button across both sections.
- Rendered all three states (both sections, recent-only, empty) in a throwaway harness. No console errors, and no horizontal overflow at 375px (`scrollWidth === clientWidth === 375`).

### DIFF
--- a/api/functions/__tests__/me-recent-releases.test.ts
+++ b/api/functions/__tests__/me-recent-releases.test.ts
@@ -18,7 +18,7 @@ vi.mock('../ratelimit', () => ({
   getClientIp: mocks.getClientIp,
 }));
 
-import { handler, toRecentReleases, RECENT_RELEASE_LIMIT } from '../me-recent-releases';
+import { handler, splitRecentReleases, RECENT_RELEASE_LIMIT } from '../me-recent-releases';
 import type { FeedReleaseRow } from '../db';
 
 function row(overrides: Partial<FeedReleaseRow> = {}): FeedReleaseRow {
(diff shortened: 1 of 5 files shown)


brandonlucasgreen (commented):

**Security**

- Auth/authorization — `authenticateRequest` and the bearer-token check in `me-recent-releases.ts` are untouched by this diff; `getFeedReleasesForUser(userId)` still scopes to the authenticated user only. **confirmed**: no change to the authz surface.
- Input validation / injection — no new external input is parsed. The only new parameter is `now: Date`, which defaults to `new Date()` and is never taken from the request; the DB query (`db.ts`) is unchanged. **confirmed**: no injection surface added.
- Secret handling — not touched.
- Data exposure — `toShortlistRow` returns the same field set as the old `toRecentReleases` (artist/release metadata + per-source offers), just factored out; no new fields, no PII added. **confirmed**: no new exposure.
- Unsafe defaults — none found.

No security findings.

**Regression**

- **API contract drift — needs-manual-verification.** `/api/me/recent-releases` response body changes from `{ releases }` to `{ upcoming, recent }`. You've already called this out in the PR description: `DashboardPage` is the only consumer (confirmed via repo-wide grep — no other caller of this route), and the client destructures with `data.upcoming || []` / `data.recent || []`, so a stale pre-deploy SPA chunk degrades to the empty state rather than throwing. Worth a manual check that the empty-state copy still reads sensibly during that brief window, but nothing crashes.
- **Date-boundary logic — confirmed correct.** Split is `releaseDate > today` (upcoming) vs `<= today` (recent), using `now.toISOString().slice(0,10)` against the ISO date strings from `FeedReleaseRow.releaseDate` (non-null, filtered at the DB layer in `toFeedRows`). "Today counts as recent" is covered by the boundary test, matches the PR's stated intent, and string comparison on `YYYY-MM-DD` is safe (no parsing footguns).
- **Cap logic — confirmed correct.** Independent caps per list (not a shared budget of 6) with opposite sort directions, verified by the new tests (`caps each list separately`, `orders upcoming soonest first`, `orders recent newest first`). Matches the stated intent that a run of announcements shouldn't hide what's already out.
- **Pre-existing, not introduced here — needs-manual-verification.** `getFeedReleasesForUser` orders ascending and applies `FEED_MAX_RELEASES` (200) at the DB layer *before* this split runs. Since the ordering starts from the oldest date in the trailing-30-day window, a saved-artist set with a very large combined past+future release count could have future releases truncated by that upstream cap before `splitRecentReleases` ever sees them — i.e. "Upcoming" could under-report for a fan with an unusually large catalogue. This query is unchanged by this PR, so it's out of scope here, but flagging since the split makes the upcoming/recent distinction more visible to users than it was before.
- **CI bot note.** The semantic-revert-check bot flagged deletions across all 5 changed files as reverting #421 (merged the same day, "Show fans recent releases..."). Looking at the diff, this reads as the expected shape for a same-day follow-up PR that reworks the feature #421 just shipped (renaming `toRecentReleases`→`splitRecentReleases`, restructuring the component), not an actual functional revert — the new tests cover strictly more behavior than the old ones (split, boundary, independent caps) rather than removing coverage. Flagging as a false positive rather than a real regression, but worth a quick look since the bot fires on exactly the pattern this repo's retro docs warn about.

No confirmed regressions blocking merge; the two needs-manual-verification items above are worth a glance but neither is a functional bug I could reproduce from the code.

---

PR #371, opened 2026-08-01
Fix FAQ card styling to match Changelog/Guides

brandonlucasgreen:
## Summary
- FAQ's accordion cards used `border-border-primary`, a Tailwind class that doesn't exist in the theme (only `--color-border`/`border-border` is defined), so cards rendered with a bright default border instead of the subtle one Changelog/Guides use. Switched to `bg-surface-secondary`/`border-border` and matched hero spacing (`pt-6`/`mb-2`) and item spacing (`space-y-4`) to those pages.
- Added the missing `font-display` (Darker Grotesque) to GuidePage's "Not found" heading — it was the only heading across FAQ/Changelog/Guides/GuidePage not using it.

## Test plan
- [x] Verified in-browser (dark + light) that FAQ card borders/spacing now match Changelog and Guides
- [x] Verified GuidePage's not-found state renders in Darker Grotesque
- [x] `npm run lint` clean on changed files
- [x] `npm run test:unit` — 492 passed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #429, opened 2026-08-08
Let people subscribe to what ships, by email or by feed

brandonlucasgreen:
The changelog and the guides were both write-only — somebody who cared about either had to remember to come back and check. This gives each one a feed and a way to be told.

## RSS feeds

Two RSS 2.0 feeds, generated at build time and wired into `npm run build`:

- **`/changelog.xml`** — one item per shipped feature, linking to its anchor on the changelog page. Entries now carry `id`s, and the page scrolls to the target once the entry it names actually exists (verified landing 96px down, clear of the 69px sticky header).
- **`/guides.xml`** — full rendered post in `content:encoded`. A feed carrying only the one-line description would give a reader nothing to read and Buttondown nothing to send.

Both share the new `scripts/rss.ts` with the dispatch feed, which held the only copy of the escaping and date handling until there was a second feed to keep in step. **The dispatch output is byte-identical apart from `lastBuildDate`** — checked with a diff before and after the refactor.

Feed auto-discovery `<link>` tags added to `index.html`.

## Newsletter signup

Signup POSTs to `/api/newsletter/subscribe` (`api/functions/newsletter-subscribe.ts`), which calls Buttondown server-side, rather than embedding Buttondown's own form. The embed would need third-party script and frame hosts in the CSP and can't be styled to match the site. Proxying keeps the CSP untouched, keeps the API key out of the browser, gets the form the same per-IP rate limiting as everything else, and tags each signup `changelog` / `guides` / `settings` — anything off that list is dropped rather than forwarded, since Buttondown creates tags on demand.

**Double opt-in.** No subscriber `type` is sent, so Buttondown creates the subscriber as `unactivated` and emails a confirmation link. The form is public, so anyone can type in an address that isn't theirs. The UI says "check your inbox", never "you're subscribed". An already-subscribed address reports as success, not an error.

**Fails loudly.** A missing API key returns 503 and a missing-config Sentry event; a network or upstream failure returns 502. None of them read as a successful signup — that failure mode loses subscribers silently, which nobody notices until the list is a month short.

## Also here

- The guides index gains published dates and pillar labels, so it reads like a blog index.
- Page titles on Guides, Changelog, FAQ and Support move from `font-semibold` to `font-extrabold` (Darker Grotesque is a variable font at `wght 300..900`, so 800 renders genuinely heavier — confirmed in computed style, not just the class name).
- Dev-server stub in `apps/web/server/api.ts` mirrors the response shapes and logs to the console, so styling the form locally can't put test addresses on the live list.

## Deploy requirement

`BUTTONDOWN_API_KEY` must be set in Netlify — already done. No CSP change, no new SSRF allowlist host, no migration.

## Testing

Full build green: 1,013 API tests, 710 web tests, lint at its existing baseline with nothing new.

New coverage: `newsletter-subscribe.test.ts` (15 cases — double opt-in, tag validation, duplicate handling, upstream/network/missing-key failures, rate limiting), `NewsletterSignup.test.tsx`, and `rss-feed.test.ts` (escaping, RFC 822 dates, frontmatter, CDATA-terminator escape, XML parseability).

Both key guards were mutation-tested: adding `type: 'regular'` and turning an upstream 500 into a success each fail the suite.

### DIFF
--- a/api/functions/__tests__/newsletter-subscribe.test.ts
+++ b/api/functions/__tests__/newsletter-subscribe.test.ts
@@ -1,0 +1,177 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+
+// Mirrors ratelimit.ts's RateLimitResult, whose `response` is optional — without the
+// annotation the mock's type is inferred from its default and can't be given a 429.
+interface RateLimitResult {
+  limited: boolean;
+  response?: { statusCode: number; headers: Record<string, string>; body: string };
+}
+
+const mocks = vi.hoisted(() => ({
+  checkRateLimit: vi.fn(() => Promise.resolve({ limited: false } as RateLimitResult)),
+  getClientIp: vi.fn(() => '127.0.0.1'),
+  captureException: vi.fn(),
+  captureMessage: vi.fn(),
+}));
+
+vi.mock('../ratelimit', () => ({
+  checkRateLimit: mocks.checkRateLimit,
+  getClientIp: mocks.getClientIp,
+}));
+vi.mock('../../lib/sentry', () => ({
+  Sentry: { captureException: mocks.captureException, captureMessage: mocks.captureMessage },
+}));
+
+import { handler } from '../newsletter-subscribe';
+
+function post(body: unknown) {
+  return { httpMethod: 'POST', headers: {}, body: JSON.stringify(body) };
+}
+
+/** Stand in for a Buttondown response. Only the bits the handler reads. */
+function buttondownReply(status: number, payload: unknown) {
+  return {
+    ok: status >= 200 && status < 300,
+    status,
+    text: () => Promise.resolve(payload === undefined ? '' : JSON.stringify(payload)),
+  } as unknown as Response;
+}
+
+describe('newsletter-subscribe handler', () => {
+  const fetchMock = vi.fn();
+
+  beforeEach(() => {
+    vi.resetAllMocks();
+    mocks.checkRateLimit.mockResolvedValue({ limited: false });
+    mocks.getClientIp.mockReturnValue('127.0.0.1');
+    process.env.BUTTONDOWN_API_KEY = 'test-key';
+    vi.stubGlobal('fetch', fetchMock);
+  });
+
+  afterEach(() => {
+    vi.unstubAllGlobals();
+    delete process.env.BUTTONDOWN_API_KEY;
+  });
+
+  it('subscribes a valid address and reports it as pending confirmation', async () => {
+    fetchMock.mockResolvedValue(buttondownReply(201, { id: 'abc' }));
+
+    const res = await handler(post({ email: 'fan@example.com', source: 'changelog' }));
+
+    expect(res.statusCode).toBe(200);
+    expect(JSON.parse(res.body)).toEqual({ status: 'pending' });
+  });
+
+  it('does not send a subscriber type, so Buttondown runs its double opt-in', async () => {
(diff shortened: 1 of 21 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Senior-staff-style pass on this PR (security + regression only, no style/quality nits).

## Security

**`api/functions/newsletter-subscribe.ts` — email-bombing via the public signup form** — *plausible*
The endpoint is unauthenticated by design (anyone can type in an address that isn't theirs — that's the point of double opt-in). But abuse protection is only the `strict` per-IP rate limit (10 req/min, 500 req/day, shared with `me-password.ts`/`resolve-url.ts`/etc.). There's no per-target-address throttling and no CAPTCHA, so a single IP can still trigger up to 500 Buttondown confirmation emails/day to arbitrary victim addresses (mail-bombing/harassment vector), and that's before considering IP rotation. The double opt-in itself caps the blast radius (only one email per address until confirmed, no account takeover), so I'd call this an accepted trade-off rather than a blocker — flagging in case Buttondown-side abuse signals aren't enough on their own.

**Secret handling, CORS, input validation, injection, logging — confirmed clean.** `BUTTONDOWN_API_KEY` stays server-side, sent only via the `Authorization: Token` header, never echoed in responses or Sentry payloads (Sentry only gets `response.status`/`payload.code`, never the raw body or email). `source` is checked against an allowlist `Set` before being forwarded as a Buttondown tag, so arbitrary tag injection is closed off (`newsletter-subscribe.test.ts` covers this). `Access-Control-Allow-Origin: '*'` matches the existing pattern for unauthenticated, no-credential POST endpoints elsewhere in `api/functions/` (`me-username.ts`, `saved-artists.ts`, etc.) — consistent, not a new hole. RSS generation (`scripts/rss.ts`) escapes XML entities and the `]]>` CDATA terminator, with dedicated parseability tests — no injection surface from guide/changelog content into the feed XML.

## Regression

**Branch is stale relative to `main` by one merged PR** — *needs-manual-verification*
`git log` on the head branch (`newsletter-subscription-integration-e9dcf5`) shows its parent is `05a28619` (#427), not `eb8ef48a` (#428, "Split the dashboard's releases into Upcoming and Recent") — even though the PR's reported base sha is already `eb8ef48a`. That's what's driving the semantic-revert-check bot comment about deleted `me-recent-releases.ts`/`RecentReleasesSection.tsx`/`DashboardPage.tsx` lines: none of those files are actually touched by this PR's own commit, so it's a false positive caused by the branch simply not having #428 yet, not an actual revert. `mergeable_state: unstable` on the PR is consistent with this. Recommend updating/rebasing the branch onto latest `main` before merging so CI and the bot are diffing against what will actually land — low risk either way since a normal three-way merge won't drop #428's changes, but worth confirming before merge rather than after.

**`FaqPage.tsx`/`SupportPage.tsx` font-weight lines flagged by the same bot are non-issues** — the "deleted" line is just this PR's own `font-semibold` → `font-extrabold` change on the heading `#371` had touched; intentional, matches the PR description.

**Buttondown re-subscribe state transition — untested** — *needs-manual-verification*
Tests cover a brand-new address and an already-`unactivated`/subscribed address (`email_already_exists` → `already_subscribed`), but not what happens for an address that previously unsubscribed. Depending on Buttondown's API behavior, that could either correctly restart double opt-in or silently no-op — worth a manual check against the real API since it's not something a mock can settle.

No auth/authorization, unsafe-default, or data-exposure issues found beyond the above.

---

PR #430, opened 2026-08-08
Add Resend transactional emails for claim approval/rejection

brandonlucasgreen:
Wires up the first, cheapest notification trigger identified in the Buttondown-vs-Resend
strategy discussion: artist claim approved (auto link-back and manual admin review) and
claim rejected. Both paths already had a verified email in scope, making this the lowest-
effort starting point before tackling release-found alerts, which need new delta-detection
plumbing in the recatalog pipeline first.

- api/lib/resend.ts: thin fetch-based Resend client, mirrors the Buttondown pattern in
  newsletter-subscribe.ts rather than adding the SDK as a dependency.
- api/functions/notifications.ts: sendNotificationOnce wraps the send with an idempotency
  guard backed by the new email_log table, so a retried request or a second admin click
  can't double-send.
- supabase/migrations/20260808000000_email-log.sql: server-only email_log table (RLS on, no
  policies) with a unique index on (notification_type, reference_id) as the send-once guard.
- Hooked into claim-artist.ts's auto-verify path and admin-verify.ts's approve/reject path.
- Added claim-artist.ts and admin-verify.ts to api/tsconfig.json's typecheck include — both
  came back clean under strict mode, so they're now covered going forward.

Requires RESEND_API_KEY in Netlify and a verified sending domain in the Resend dashboard
before this can send in production; unset, sendTransactionalEmail fails loudly to Sentry
rather than silently dropping notifications.

### DIFF
--- a/api/functions/claim-artist.ts
+++ b/api/functions/claim-artist.ts
@@ -7,7 +7,7 @@ import { createClient } from '@supabase/supabase-js';
 import { Sentry } from '../lib/sentry';
 import { getClient } from './db';
 import { checkRateLimit, getClientIp } from './ratelimit';
-import { sendNotificationOnce } from './notifications';
+import { sendNotificationOnce, notifySavedArtistsOfNewLinks } from './notifications';
 import { escapeHtml } from '../lib/html';
 
 const PLATFORM_PATTERNS: [string, RegExp][] = [
@@ -680,6 +680,21 @@ export async function handler(event: {
         subject: `Your claim of ${artist.name} on Unstream is verified`,
         html: `<p>Your claim of <strong>${escapeHtml(artist.name)}</strong> on Unstream has been verified — your profile is now live at <a href="https://unstream.stream/a/${escapeHtml(slug)}">unstream.stream/a/${escapeHtml(slug)}</a>.</p><p>You can edit your bio, photo, and links any time from your dashboard.</p>`,
         text: `Your claim of ${artist.name} on Unstream has been verified — your profile is now live at https://unstream.stream/a/${slug}.\n\nYou can edit your bio, photo, and links any time from your dashboard.`,
+      });
+    }
+
+    if (discoveredLinks.length > 0) {
+      // Excludes the claimant themselves — they already got the claim-approved email above and
+      // just added these links, so it isn't news to them.
+      void notifySavedArtistsOfNewLinks({
+        client,
+        artistId: artist.id,
+        artistName: artist.name,
+        artistSlug: slug,
+        platforms: discoveredLinks.map(l => l.platform),
+        excludeUserId: userId,
+      }).catch(err => {
+        Sentry.captureException(err, { extra: { context: 'notifySavedArtistsOfNewLinks', artistId: artist.id } });
       });
     }
 
--- a/api/functions/claim-artist.ts
+++ b/api/functions/claim-artist.ts
@@ -7,6 +7,8 @@ import { createClient } from '@supabase/supabase-js';
 import { Sentry } from '../lib/sentry';
 import { getClient } from './db';
 import { checkRateLimit, getClientIp } from './ratelimit';
+import { sendNotificationOnce } from './notifications';
+import { escapeHtml } from '../lib/html';
 
 const PLATFORM_PATTERNS: [string, RegExp][] = [
   ['bandcamp', /([a-z0-9-]+)\.bandcamp\.com/i],
@@ -665,6 +667,21 @@ export async function handler(event: {
     }
 
     console.log(`[Claim] Verified "${artist.name}" — discovered ${discoveredLinks.length} platform links from website`);
+
+    const notifyEmail = (profile.email || authUser.email || '').trim();
+    if (notifyEmail) {
+      // Fire-and-forget: a failed notification email must never fail a successful claim.
+      // sendNotificationOnce logs its own failures to Sentry and is a no-op on retry.
+      void sendNotificationOnce({
+        client,
+        notificationType: 'claim_approved_auto',
+        referenceId: artist.id,
+        recipientEmail: notifyEmail,
+        subject: `Your claim of ${artist.name} on Unstream is verified`,
+        html: `<p>Your claim of <strong>${escapeHtml(artist.name)}</strong> on Unstream has been verified — your profile is now live at <a href="https://unstream.stream/a/${escapeHtml(slug)}">unstream.stream/a/${escapeHtml(slug)}</a>.</p><p>You can edit your bio, photo, and links any time from your dashboard.</p>`,
+        text: `Your claim of ${artist.name} on Unstream has been verified — your profile is now live at https://unstream.stream/a/${slug}.\n\nYou can edit your bio, photo, and links any time from your dashboard.`,
+      });
+    }
 
     // Build full link list for the review step (existing + newly discovered, deduplicated)
     const allLinksMap = new Map<string, { platform: string; url: string }>();
--- a/api/functions/db.ts
+++ b/api/functions/db.ts
@@ -17,6 +17,7 @@ import {
   uniqueReleaseSlug,
 } from './release-utils';
 import { requestArtistCatalog } from './request-catalog';
+import { notifySavedArtistsOfNewRelease } from './notifications';
 import { Sentry } from '../lib/sentry';
 import { isNonArtistSlug } from '../lib/non-artist-names';
 import { isExcludedArtistSlug } from '../lib/excluded-artists';
@@ -577,25 +578,37 @@ export async function recordCatalogOutcome(
         .maybeSingle();
       const prev = (data as { consecutive_failures: number } | null)?.consecutive_failures ?? 0;
       patch.consecutive_failures = prev + 1;
-    } else if (outcome.releasesFound === 0) {
-      // A run that finds 0 releases where it previously found 20 is a parser break or a bot
-      // challenge, not an artist deleting their catalogue — and it is recorded as a *success*,
-      // so nothing else about it looks wrong. Report the transition, because the alternative
-      // is a pipeline that degrades into finding nothing and never says so.
-      //
-      // Only the transition: an artist who has always had 0 releases is not news, and this
-      // fires on every trigger, not just the scheduled sweep.
+    } else {
+      // Read once and branch: a drop to zero is a parser/bot-challenge alert (see below), an
+      // increase is news for anyone who saved this artist. Fetched for every successful run,
+      // not just the sweep, since demand-driven catalog triggers (a save, a search) are exactly
+      // where a fan is most likely to be waiting on this.
       const { data } = await client
         .from('release_catalog_state')
         .select('releases_found')
         .eq('artist_id', artistId)
         .maybeSingle();
       const previous = (data as { releases_found: number | null } | null)?.releases_found ?? 0;
-      if (previous > 0) {
+
+      if (outcome.releasesFound === 0 && previous > 0) {
+        // A run that finds 0 releases where it previously found 20 is a parser break or a bot
+        // challenge, not an artist deleting their catalogue — and it is recorded as a *success*,
+        // so nothing else about it looks wrong. Report the transition, because the alternative
+        // is a pipeline that degrades into finding nothing and never says so.
+        //
+        // Only the transition: an artist who has always had 0 releases is not news, and this
+        // fires on every trigger, not just the scheduled sweep.
         Sentry.captureMessage('[catalog] release count dropped to zero', {
           level: 'warning',
           tags: { area: 'release-catalog', kind: 'releases-dropped-to-zero' },
           extra: { artistId, previousReleasesFound: previous },
+        });
+      } else if (outcome.releasesFound > previous) {
+        // Fire-and-forget: notifySavedArtistsOfNewRelease bails out immediately if nobody saved
+        // this artist (true for nearly everything the sweep touches) and logs its own failures
+        // to Sentry — a notification email must never affect cataloging's own success/failure.
+        void notifySavedArtistsOfNewRelease({ client, artistId, releasesFound: outcome.releasesFound }).catch(err => {
+          Sentry.captureException(err, { extra: { context: 'notifySavedArtistsOfNewRelease', artistId } });
         });
       }
     }
--- a/api/functions/notifications.ts
+++ b/api/functions/notifications.ts
@@ -115,6 +115,41 @@ async function resolveEmails(client: SupabaseClient, userIds: string[]): Promise
   return emails;
 }
 
+export type NotificationPreferenceColumn = 'new_release' | 'new_platform_link' | 'weekly_analytics_recap';
+
+/**
+ * Drops user ids who have explicitly turned this notification off. A user with no row in
+ * notification_preferences has never touched their settings and is treated as enabled — see
+ * the migration comment. A lookup failure fails *open* (keeps everyone) rather than closed:
+ * silently suppressing a notification someone opted into by default is the "cache uncertainty
+ * as a negative" mistake this codebase specifically avoids elsewhere, and it would look
+ * identical to nothing being wrong.
+ */
+export async function filterByPreference(
+  client: SupabaseClient,
+  userIds: string[],
+  column: NotificationPreferenceColumn,
+): Promise<string[]> {
+  if (userIds.length === 0) return userIds;
+
+  const { data, error } = await client
+    .from('notification_preferences')
+    .select(`user_id, ${column}`)
+    .in('user_id', userIds);
+
+  if (error) {
+    Sentry.captureException(error, { extra: { context: 'filterByPreference', column } });
+    return userIds;
+  }
+
+  const optedOut = new Set(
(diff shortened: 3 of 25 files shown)


api/functions/notifications.ts
```diff
@@ -0,0 +1,248 @@
...
+ * Sends a transactional email at most once per (notificationType, referenceId). Failures —
+ * duplicate send, missing API key, Resend error — are logged to Sentry and swallowed; a
+ * notification email is never allowed to fail the request that triggered it.
+ */
+export async function sendNotificationOnce(params: NotifyOnceParams): Promise<void> {
+  const { client, notificationType, referenceId, recipientEmail, subject, html, text } = params;
+
+  const { data: logRow, error: insertError } = await client
+    .from('email_log')
+    .insert({
```
brandonlucasgreen, line 37: **Regression / reliability — plausible.** This insert (status defaults to `'sending'`) happens *before* `sendTransactionalEmail` is called below. The email_log migration comment says explicitly: "There's no retry path today, so a `'failed'` row staying failed is correct — a future retry mechanism would need to delete the row first." That's true for `'failed'`, but there's no code path that ever revisits a row stuck at `'sending'` — the guard only checks for a unique-violation (`23505`) on retry, not `status`. If the call to `sendTransactionalEmail` never runs (see the fire-and-forget comment on `claim-artist.ts`/`admin-verify.ts`), the row is permanently stuck at `'sending'` and the unique index on `(notification_type, reference_id, recipient_email)` blocks every future attempt for that exact event — silently dropping the notification forever. This is the same "cache uncertainty as a negative" failure shape CLAUDE.md calls out as the codebase's most-repeated bug class (#317–#328); worth at least a `status = 'sending'` cleanup/retry path, or reordering so the row is only written once the send outcome is known.

---

api/functions/claim-artist.ts
```diff
@@ -666,6 +668,36 @@ export async function handler(event: {
 
     console.log(`[Claim] Verified "${artist.name}" — discovered ${discoveredLinks.length} platform links from website`);
 
+    const notifyEmail = (profile.email || authUser.email || '').trim();
+    if (notifyEmail) {
+      // Fire-and-forget: a failed notification email must never fail a successful claim.
+      // sendNotificationOnce logs its own failures to Sentry and is a no-op on retry.
+      void sendNotificationOnce({
```
brandonlucasgreen, line 675: **Regression / reliability — plausible, needs-manual-verification.** `claim-artist.ts` is a regular (foreground) Netlify function, not one of this repo's `*-background.ts` functions — the pattern this codebase already uses elsewhere (`catalog-artist-background.ts`, `discord-search-background.ts`) specifically for "must keep running after the response is sent" work, with a comment explaining exactly why: the response is returned immediately and the function then gets up to 15 minutes to finish. Here, `sendNotificationOnce(...)` (and `notifySavedArtistsOfNewLinks` below) is kicked off with a bare `void` and never awaited before the handler returns its HTTP response. Netlify Functions run on Lambda-style infrastructure with no `waitUntil`-equivalent guarantee for unawaited promises — once the response is flushed, the execution environment can freeze before this promise resolves, so the notification (and the claim-approval email a claiming artist is expecting) may simply never send, with no error surfaced anywhere. Combined with the `email_log` insert-before-send ordering (see notifications.ts comment), an interrupted send here is not just delayed, it's dropped permanently. Same pattern at `admin-verify.ts:327` and `:356`. Recommend either `await`ing these sends in the request path (small latency cost) or moving them into a background function.

---

brandonlucasgreen (commented):
@brandonlucasgreen Security + regression pass on this PR (Resend transactional emails, saved-artist alerts, weekly analytics recap, notification preferences). Reviewed the full diff — 3 commits, 21 files. Two inline comments below on the highest-severity finding; summary of everything checked follows.

## Security

- **Auth/authorization** — no issues found. `me-notifications.ts` authenticates via Supabase `auth.getUser(token)` and scopes all reads/writes to `user.userId`. `weekly-analytics-recap.ts` gates on `isInternalRequest`, which uses `timingSafeEqual` against `INTERNAL_FUNCTION_SECRET` (existing helper, reused correctly). `admin-verify.ts`'s new email sends ride on the existing `authenticateAdmin` gate and don't introduce a new privilege path.
- **Input validation** — `me-notifications.ts` POST validates each toggle is a boolean and 400s otherwise. *(Note: like several existing `me-*` functions, e.g. `me-location.ts`/`me-password.ts`, a non-object JSON body such as `null` or `true` would throw on the `'x' in body` check rather than 400 cleanly — but that's a pre-existing pattern across the codebase, not something this PR introduces, so I'm not flagging it as a PR regression.)*
- **Secret handling** — `RESEND_API_KEY` and `INTERNAL_FUNCTION_SECRET` are read from env only, never logged. A missing `RESEND_API_KEY` fails loudly to Sentry (`resend.ts`) instead of silently no-op'ing, matching CLAUDE.md's "never cache uncertainty" guidance. Good.
- **Injection risks** — email bodies interpolate `artist.name` / `reviewerNotes` / platform lists through `escapeHtml` (new shared helper in `api/lib/html.ts`, consistent with the edge-function XSS defense already covered by `xss-defense.test.ts`). The Resend call is JSON-encoded via `fetch`, not raw SMTP, so CRLF header injection via artist names isn't reachable. No SSRF surface added — `RESEND_SEND_URL` is a fixed constant, not user input.
- **Unsafe defaults** — `notification_preferences` defaults every toggle to enabled, with the stated rationale that saving/claiming implies consent; `filterByPreference` fails *open* (keeps recipients) on a lookup error rather than silently suppressing a notification someone opted into. Documented and consistent with the "fail open on uncertainty, not closed" principle used elsewhere in this codebase — reasonable call.
- **Data exposure** — `weekly-analytics-recap.ts` and `notifications.ts` use the service-role client to read `artist_profiles.email` / resolve auth emails via the admin API, which is expected for server-side sends and never returned to a client. `email_log` and the Resend client aren't reachable from any client-facing endpoint.

## Regression

- **Confirmed correct**: the `recordCatalogOutcome` restructuring in `db.ts` (single read of `releases_found`, then branch on drop-to-zero vs. increase) preserves the original zero-drop Sentry alert behavior exactly and correctly adds the new "notify on increase" branch — verified against the pre-PR version, no behavior change for the existing alert path.
- **State transition — see inline comments.** The two flagged spots (`notifications.ts`'s insert-before-send ordering in `email_log`, and the unawaited `void sendNotificationOnce(...)` calls in `claim-artist.ts`/`admin-verify.ts`) combine into a real risk: if the foreground function's execution environment is reclaimed before the fire-and-forget promise resolves, the notification is dropped *and* the `email_log` unique-index guard permanently blocks any future retry for that same event — because the guard only checks for a duplicate insert, not whether the row ever reached `'sent'`/`'failed'`. Flagged as **plausible / needs-manual-verification** since it depends on Netlify's exact Lambda-freeze timing, but the codebase already has the correct pattern for this (`*-background.ts` functions) and doesn't use it here.
- **Edge case / null-empty**: `notifySavedArtistsOfNewLinks`'s `referenceId` (`artistId:sortedPlatforms`) correctly re-fires when the discovered platform set changes and correctly dedupes when it's identical — checked this against a re-claim scenario and it holds up.
- **API contract drift**: none — `/api/me/notification-preferences` is a new route, doesn't touch existing contracts. `netlify.toml` redirect addition is additive only.

---

brandonlucasgreen (commented):
@brandonlucasgreen Automated security/regression review of this PR (full diff read, cross-checked against `middleware.ts`, `newsletter-subscribe.ts`, `admin-verify.ts`, `claim-artist.ts`).

### Security

**1. Recipient email is never format-validated before being used as the Resend `to` field — confirmed (unvalidated input), plausible/needs-manual-verification (injection consequence).**
`api/functions/claim-artist.ts` (~L435, pre-existing): `body.email` is accepted from the client with no format check — unlike `newsletter-subscribe.ts`, which validates via `EMAIL_REGEX`. That value is persisted to `artist_profiles.email`, then this PR's new code (~L668) does `const notifyEmail = (profile.email || authUser.email || '').trim();` and passes it straight into `sendNotificationOnce({ recipientEmail: notifyEmail, ... })` → `sendTransactionalEmail({ to: recipientEmail, ... })` → the Resend `to` field. An authenticated user completing a claim could set `body.email` to a third party's real address (Unstream would then email a stranger a "your claim is verified" message from `notifications@unstream.stream`) or to a value containing control characters, if Resend doesn't sanitize `to` itself. Worth adding the same `EMAIL_REGEX` check used in `newsletter-subscribe.ts` at the point `body.email` is accepted.

**2. Email `subject` lines interpolate artist names without escaping or length limits — needs-manual-verification.**
`admin-verify.ts` (~L861, L878), `notifications.ts` (~L1375, L1419), `weekly-analytics-recap.ts` (~L1558) escape the HTML `body` but not `subject`. `artist.name` originates from scraped/Wikidata/MusicBrainz data upstream of this diff and isn't guaranteed free of `\r\n`. If Resend passes `subject` through without stripping control characters, this is a header-injection vector. Can't confirm Resend's own sanitization from the diff — worth a quick check against their docs/behavior.

**3. `RESEND_API_KEY` handling, SSRF surface, admin gating, data exposure — no issues found.** Key is read from `process.env` only, sent as a Bearer header, never logged (asserted by `resend.test.ts`). `RESEND_SEND_URL` is a hardcoded constant, not built from request input, so no new SSRF allowlist entry is needed. `admin-verify.ts` gates the entire handler (including both new email sends) behind `authenticateAdmin(...)`, and blocks re-approving/re-rejecting an already-decided request (`request.status !== 'pending'`). `email_log` ships with RLS enabled and zero policies (service-role only), correctly commented as deliberate.

### Regression

**4. Fire-and-forget email sends with no `.catch()` — confirmed as a code-pattern inconsistency, plausible as a real bug (needs-manual-verification for Netlify runtime behavior).**
`admin-verify.ts` L856 (`void sendNotificationOnce({...})`, approve) and L873 (reject), plus `claim-artist.ts` L906, have no `.catch()` and aren't awaited — unlike the sibling call at `claim-artist.ts` L920-929 in the *same file*, which does chain `.catch(err => Sentry.captureException(...))`. In a serverless function, the execution context can freeze once the HTTP response returns, before an un-awaited promise resolves. Failure scenario: the `email_log` row is inserted (claiming the send as "in flight"), the function returns and freezes before `sendTransactionalEmail` actually completes, and the email is never sent — but the unique constraint on `(notification_type, reference_id, recipient_email)` then permanently blocks any retry for that claim (the migration's own comment notes "there's no retry path today"). Net effect: an artist can be silently left with no approval/rejection email and no way to get one. Suggest matching the existing `.catch()` pattern at minimum, and ideally awaiting the send before returning the response.

**5. `newsletter-subscribe.ts` `ALLOWED_SOURCES` shrank from `['changelog','guides','settings']` to `['changelog','guides']`, and `NewsletterSignup`'s `defaultEmail` prop was removed — confirmed, low risk.** `SettingsPage.tsx` and its test were updated in the same PR to match, and no other `source: 'settings'` caller was found in the diff — but since local `main` isn't this PR's branch, worth a repo-wide grep on the actual branch for any other caller before merge.

**6. Edge cases — handled well.** Missing/empty email is guarded before sending; a missing artist slug in the weekly recap is skipped rather than producing a broken link; duplicate sends are blocked by the unique index; preference-lookup failures fail open (by design) rather than silently dropping a wanted notification. No RLS/migration gaps — both new tables (`email_log`, `notification_preferences`) ship RLS with appropriate policies in this PR.

---

api/functions/claim-artist.ts
```diff
@@ -666,6 +668,36 @@ export async function handler(event: {
 
     console.log(`[Claim] Verified "${artist.name}" — discovered ${discoveredLinks.length} platform links from website`);
 
+    const notifyEmail = (profile.email || authUser.email || '').trim();
```
brandonlucasgreen, line 671: **[Security — data exposure / notification abuse] confirmed**

`notifyEmail` here is `profile.email`, which is set earlier in this handler from `body.email` (client-supplied) falling back to `authUser.email` only when empty — there's no check that the submitter actually controls the address they typed in. The manual-review path (`verification_requests.email`, further down this file) has the same shape: `body.email || authUser.email`, unvalidated.

Before this PR those fields were inert — shown to an admin, or copied into `artist_profiles` — nothing auto-sent to them. This PR is what turns them into a live send target: this call, plus the two new `sendNotificationOnce` calls in `admin-verify.ts` on approve/reject.

Net effect: a user who can legitimately verify an artist (real site link-back) can still set the contact email to a third party's address, and Unstream's Resend account will then deliver a "claim approved/verified" email to that address — someone who never interacted with the site. `checkRateLimit(ip, 'strict', …)` bounds volume per IP and the manual-review path additionally requires a human admin to click approve/reject, so this isn't a mass-spam primitive, but it is a real send-to-unverified-address path using the verified sending domain. Also worth checking (**needs-manual-verification**): `recipientEmail` isn't validated as a single well-formed address before reaching `sendTransactionalEmail`'s `to` field — if Resend accepts a comma-separated string there, one submission could target multiple third parties at once.

Suggest validating the submitted email matches `authUser.email`, or at minimum enforcing single-address syntax before it reaches `sendNotificationOnce`.

@brandonlucasgreen

---

api/functions/db.ts
```diff
@@ -577,26 +578,38 @@ export async function recordCatalogOutcome(
...
+        // is a pipeline that degrades into finding nothing and never says so.
+        //
+        // Only the transition: an artist who has always had 0 releases is not news, and this
+        // fires on every trigger, not just the scheduled sweep.
         Sentry.captureMessage('[catalog] release count dropped to zero', {
           level: 'warning',
           tags: { area: 'release-catalog', kind: 'releases-dropped-to-zero' },
           extra: { artistId, previousReleasesFound: previous },
         });
+      } else if (outcome.releasesFound > previous) {
```
brandonlucasgreen, line 606: **[Regression — edge case / state transition] confirmed**

`previous` defaults to `0` when there's no `release_catalog_state` row yet (line 591). On an artist's *first-ever* successful catalog run, `outcome.releasesFound > previous` is true for any nonzero count, so `notifySavedArtistsOfNewRelease` fires for every current saver — even though nothing is actually new, the artist just got indexed for the first time.

Per CLAUDE.md's "Keeping catalogues fresh" section, "never catalogued" is one of the priority tiers `getStaleCatalogCandidates` sweeps, and saving an artist before their first catalog run is explicitly the common case the sweep exists to backfill (saved-but-never-catalogued). So this isn't a hypothetical: a fan saves a brand-new-to-Unstream artist who already has, say, 20 releases on Bandcamp; the next sweep or demand-trigger catalogs them for the first time; every current saver gets a "new release from X" email listing what is really just the artist's existing back catalogue.

Worth guarding the notify branch on `previous > 0` (i.e., only notify on an actual increase from a previously-known nonzero count, not from "unknown" to "known") — mirroring how the drop-to-zero branch already requires `previous > 0` to avoid firing on an artist that's simply never had releases.

@brandonlucasgreen

---

brandonlucasgreen (commented):

**Security**
1. **[confirmed]** Unverified recipient email used for automated transactional sends (`claim-artist.ts` line 671, and the two new `sendNotificationOnce` calls in `admin-verify.ts`). `body.email` is client-supplied and only falls back to `authUser.email` when empty — nothing checks the submitter controls the address. Before this PR that field was inert (display-only / stored, never auto-emailed); this PR is what makes it a live send target. Bounded by IP rate-limiting and (for the manual-review path) a human admin action, but it's a real send-to-arbitrary-address path riding the verified sending domain. See inline comment for detail, incl. a needs-manual-verification note on whether `to` accepts comma-separated addresses.
2. Auth/authz, secret handling, injection, unsafe defaults: nothing found. `RESEND_API_KEY` handling fails loudly rather than silently dropping sends (good — matches the "never cache uncertainty" principle). Email bodies consistently run user-influenced strings (artist name, admin `reviewerNotes`) through `escapeHtml` before interpolating into HTML. `email_log` and `notification_preferences` migrations both ship RLS correctly (server-only-no-policies vs. `auth.uid() = user_id`, matching existing patterns). `me-notifications.ts` scopes every read/write to the authenticated `user.userId` — no way to touch another user's preferences. `weekly-analytics-recap.ts` is gated by the existing constant-time `isInternalRequest` check, used correctly.

**Regression**
1. **[confirmed]** First-time cataloging of an artist notifies every current saver of a "new release" for what's actually just their pre-existing back catalogue, because `previous` defaults to `0` when no `release_catalog_state` row exists yet (`db.ts`, the `recordCatalogOutcome` rewrite). See inline comment — CLAUDE.md's own sweep-priority description confirms saved-before-first-catalog is the common case, not an edge case.
2. Everything else checked out: the `email_log` unique-index dedup is correctly composite on `(notification_type, reference_id, recipient_email)` so no cross-notification-type collisions; `filterByPreference` fails open (matches "never cache uncertainty"); the `newsletter-subscribe` `ALLOWED_SOURCES` narrowing and the `NewsletterSignup` prop removal are consistent across component, tests, and the settings page swap-out — no dangling `source: 'settings'` caller left behind.

No blocking issues beyond the two flagged above — both are fixable without touching the overall design.

---

PR #431, opened 2026-08-08
Add terms of use, rewrite the privacy policy against what the code actually does

brandonlucasgreen:
Unstream had a privacy policy and nothing else. That was arguably fine when it was a search box, but people now create accounts, save artists to our servers, claim artist profiles on their own behalf, and publish a list with their username and location on it. Writing terms for that meant auditing what we actually collect — and the audit turned up more than a missing page.

## Terms of use (`/terms`)

23 numbered sections with a plain-language summary box at the top. Written around what the product does rather than a generic template, so the load-bearing parts are the honest ones:

- **§5** — we're a directory, not a shop; we're not party to your purchases; results can be wrong; payout percentages are estimates that don't account for processing fees, label splits, or deals we can't see.
- **§6** — exactly what goes public when you enable sharing, that email never does, and that turning it off can't retrieve copies already made.
- **§7** — what you're asserting when you claim a profile, and that verification is a best-effort check rather than an endorsement.
- **§8** — a narrow content licence: host and display to run the service, in the places you chose to publish. No sale, no sublicensing.
- **§12** — DMCA notice, counter-notice, repeat-infringer policy.

Consent lives in one `LegalConsent` component rather than three copies of a sentence. A magic link creates an account when there isn't one, so every sign-in form is also a sign-up form — all five carry the line: web login, the save-artist interstitial, the artist claim step, the extension popup, and the Apple sign-in sheet. The last two keep their own copy and are commented to point back at the component.

"Make public" in `SharingControls` now says what it publishes *before* you press it, rather than after.

Choices worth a second opinion: Massachusetts governing law, 13+ minimum age, courts and small claims rather than binding arbitration, and no named legal entity. If Unstream is ever incorporated, the name goes in §2 and §20 and nowhere else.

## Privacy policy rewrite

The old one had drifted badly enough to be a liability. It still said *"for listeners, all user preferences and saved artists are stored locally on your device"* — untrue since sync shipped — and predated public sharing, the newsletter, the release feeds and the public API. A policy that understates collection is worse than no policy, because it's a promise already being broken.

Every claim in the new one was checked against the migrations and the functions. The evidence is written down in `docs/specs/data-collection-audit.md` so it can be checked rather than trusted.

The substantive change is refusing to call pseudonymous data anonymous. Three categories, plainly labelled:

- **Anonymous** — the artist dashboard counters. Daily totals per artist per metric, no user, no session, no IP.
- **Pseudonymous** — product events carry a daily-rotating keyed session hash; rate limiting holds raw IPs for up to 24 hours. Not nothing, so not filed under "anonymous".
- **Account-linked** — email, saved artists, username/location, feed token, artist profile, verification messages, API key owner.

Searches get their own section, because "we don't store searches" was the comfortable answer and not the true one. We never record **who** searched for what — there's no search history and signing in doesn't create one — but the terms persist in the results cache, the Bandcamp probe table, and server logs, detached from any person. Disclosed rather than glossed.

Also adds what was simply missing: GDPR legal bases, per-item retention periods, international transfer disclosure, the processor list, and how to exercise your rights.

## Search terms no longer go to GoatCounter

GoatCounter's default path is `location.pathname + location.search`, `count.js` was loaded with no settings, and a search puts the artist name straight in the URL as `/?q=artist`. So every search was being recorded as a page path — on a dashboard advertised as public metrics.

Setting the path explicitly before `count.js` loads fixes it. Verified in a browser: the URL still reads `/?q=some+private+artist` while `goatcounter.path()` returns `/`. Events fired from `services/analytics.ts` pass an explicit path and were never affected; the edge-rendered pages don't load GoatCounter at all.

Sentry is deliberately left alone — a search that breaks is only diagnosable if you know what was searched for — and that exception is disclosed in the policy.

This stops new records. The `/?q=` paths already in GoatCounter need clearing by hand under Settings → Manage pageviews.

## Two migrations

**`20260808120000` — service policies that weren't.** A `CREATE POLICY` with no `TO` clause applies to `PUBLIC`, and `PUBLIC` includes `anon` — the role a request carrying the public anon key runs as. Several policies named "Service …" omit it, so they granted to everyone the access they were named for.

Reproduced by applying the deployed policies verbatim to a local Postgres with Supabase's role model. As `anon`, with no JWT: read every verified artist's claim email, read/approve/delete verification requests, set `verified_at` on any profile, rename catalogue artists, insert merge overrides. All seven denied after the migration.

They're deleted rather than re-scoped to `service_role`, because the service role has `BYPASSRLS` and never consulted them — they granted nothing to the role they named and everything to the role they forgot. `20260731120000_releases.sql` already had this right, and its comment is the model.

Nothing outside the server talks to PostgREST: functions and edge functions use the service key, the extension only calls `/auth/v1`, and every anon-key client in `api/functions/` exists solely to call `auth.getUser()`. Verified against the same database that `service_role` keeps full access, public catalogue reads still work, an owner still reads their own rows, and a different signed-in user still sees nothing.

**`20260808130000` — `app_events` retention.** The table had none, so every event since it shipped still carried its `session_hash`. That hash is a keyed HMAC that rotates daily and only ever deduplicated within a single day, so past 90 days it has no analytical value left — only liability. A nightly sweep nulls it at 90 days rather than deleting the row, keeping every dashboard count intact while ending the row's life as personal data. Boundary verified: 91 days scrubbed, 89 days kept, event context untouched. The first run clears the backlog.

Both migrations are re-runnable, and both RLS lessons — *row-level isn't column-level*, and *a policy with no `TO` clause means everyone* — are written into the audit doc, since this repo has now hit each of them twice.

## Deploy notes

- Migrations auto-apply via `supabase-migrate.yml` on merge to `main`. `pg_cron` is confirmed enabled, so the retention migration's `CREATE EXTENSION` is a no-op.
- The RLS migration describes the exposure it closes. Merge-to-deploy is under a minute, which keeps that window short.
- `/terms` is SPA-only and added to the sitemap; no `netlify.toml` changes.

## Testing

`npm run build` green, `npm run lint` unchanged (71 pre-existing problems before and after), 710 web unit tests and 1017 API tests passing. Both legal pages rendered in a browser; both migrations validated against a real Postgres, including idempotency and the regression checks above.

https://claude.ai/code/session_01MVxipMUsXsnABZ51k21sZJ

---
_Generated by [Claude Code](https://claude.ai/code/session_01MVxipMUsXsnABZ51k21sZJ)_

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,4 +123,5 @@ When adding a new migration, create a file in `supabase/migrations/` named `YYYY
 - [Roadmap](https://github.com/users/brandonlucasgreen/projects/4)
 - [Public metrics](https://unstream.goatcounter.com)
 - [Privacy policy](https://unstream.stream/privacy-policy)
+- [Terms of use](https://unstream.stream/terms)
 - [Support Unstream](https://unstream.stream/support)
--- a/api/edge/artist-page-static.ts
+++ b/api/edge/artist-page-static.ts
@@ -525,6 +525,8 @@ export default async function handler(request: Request, context: Context) {
         <a href="https://letterbird.co/hi-d2078591" target="_blank" rel="noopener noreferrer" style="color:var(--muted);text-decoration:none">Contact</a>
         <span style="color:var(--muted);opacity:0.4;font-size:10px">&#x2022;</span>
         <a href="/privacy-policy" style="color:var(--muted);text-decoration:none">Privacy policy</a>
+        <span style="color:var(--muted);opacity:0.4;font-size:10px">&#x2022;</span>
+        <a href="/terms" style="color:var(--muted);text-decoration:none">Terms of use</a>
       </nav>
     </div>
   </footer>
@@ -628,6 +630,8 @@ export default async function handler(request: Request, context: Context) {
         <a href="https://letterbird.co/hi-d2078591" target="_blank" rel="noopener noreferrer" style="color:var(--muted);text-decoration:none">Contact</a>
         <span style="color:var(--muted);opacity:0.4;font-size:10px">&#x2022;</span>
         <a href="/privacy-policy" style="color:var(--muted);text-decoration:none">Privacy policy</a>
+        <span style="color:var(--muted);opacity:0.4;font-size:10px">&#x2022;</span>
+        <a href="/terms" style="color:var(--muted);text-decoration:none">Terms of use</a>
       </nav>
     </div>
   </footer>
--- a/api/edge/noscript-search.ts
+++ b/api/edge/noscript-search.ts
@@ -259,6 +259,8 @@ function renderPage(query: string, results: SearchResult[], error?: string): str
         <a href="https://liberapay.com/brandonlucasgreen/donate" target="_blank" rel="noopener noreferrer">Donate</a>
         <span class="footer-dot">&#x2022;</span>
         <a href="/privacy-policy">Privacy</a>
+        <span class="footer-dot">&#x2022;</span>
+        <a href="/terms">Terms</a>
       </nav>
     </div>
   </footer>
--- a/api/edge/release-page.ts
+++ b/api/edge/release-page.ts
@@ -437,6 +437,8 @@ export default async function handler(request: Request, context: Context) {
         <a href="https://letterbird.co/hi-d2078591" target="_blank" rel="noopener noreferrer" style="color:var(--muted);text-decoration:none">Contact</a>
         <span style="color:var(--muted);opacity:0.4;font-size:10px">&#x2022;</span>
         <a href="/privacy-policy" style="color:var(--muted);text-decoration:none">Privacy policy</a>
+        <span style="color:var(--muted);opacity:0.4;font-size:10px">&#x2022;</span>
+        <a href="/terms" style="color:var(--muted);text-decoration:none">Terms of use</a>
       </nav>
     </div>
   </footer>
--- a/api/edge/u-handle.ts
+++ b/api/edge/u-handle.ts
@@ -219,6 +219,8 @@ export default async function handler(request: Request, context: Context) {
         <a href="/support" style="color:var(--muted);text-decoration:none">Support</a>
         <span style="color:var(--muted);opacity:0.4;font-size:10px">&#x2022;</span>
         <a href="/privacy-policy" style="color:var(--muted);text-decoration:none">Privacy policy</a>
+        <span style="color:var(--muted);opacity:0.4;font-size:10px">&#x2022;</span>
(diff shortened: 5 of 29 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Automated security/regression review of this PR (full diff read, cross-checked against `supabase/schema.sql`).

### Security

This PR is not pure content — alongside the terms/privacy pages it also ships two real code changes worth calling out:

**1. `supabase/migrations/20260808120000_*.sql` tightens RLS on `artist_profiles`, `verification_requests`, `artists`, `artist_links`, `artist_merge_overrides`, `app_events` — confirmed, correct, and a genuine fix.** I checked `supabase/schema.sql`: several existing policies (`"Public read verified profiles"`, `"Service insert/update profiles"`, `"Service role full access"`, etc.) had no `TO` clause, meaning they applied to `PUBLIC` including `anon` — the key shipped in the client bundle. This migration's `DROP POLICY` statements match those policy names exactly, and no client code (web, extension) queries these tables directly via the anon-key client — everything goes through the API. Good catch and correctly scoped. One nit: `supabase/schema.sql` itself isn't updated to drop the old policy text, so a fresh read of that file still shows the pre-fix (permissive) policies as the "source of truth" — worth a follow-up so the logical schema doesn't drift from the migrations, especially since schema/migration drift is exactly the kind of thing this PR's own privacy-policy rewrite is trying to make accurate.

**2. `apps/web/index.html` GoatCounter path override — confirmed correct.** Sets `window.goatcounter.path` before `count.js` loads so `/?q=<search term>` isn't recorded on the public GoatCounter dashboard. Script ordering is correct.

**3. New/changed pages (`TermsOfUsePage.tsx`, `PrivacyPolicyPage.tsx`, `LegalConsent.tsx`) — no XSS surface.** Static JSX only, no `dangerouslySetInnerHTML`, no interpolation of dynamic or user-supplied values.

No findings on CORS/auth middleware, SSRF allowlist, secrets handling, or CSP/`netlify.toml` headers — none of those were touched by this PR.

### Regression

**4. Routing — no collision or breakage.** `/terms` is a new SPA route covered by the existing catch-all redirect in `netlify.toml`. Anchor links (`/terms#section-6`) resolve to the correct section. No existing route/slug was restructured, and `scripts/generate-sitemap.ts` was updated to include the new URL.

**5. `LegalConsent` is informational only, not a signup gate — needs-manual-verification, low severity.** None of the sign-in/sign-up forms it's attached to (web, extension, macOS) require acknowledging it before proceeding — no checkbox, nothing disabled until read. This matches the pre-existing pattern (not a regression), but flagging in case the intent here was to create binding consent at signup.

Overall: clean PR, one confirmed pre-existing RLS gap correctly fixed, no new vulnerabilities introduced.

---

PR #432, opened 2026-08-08
Fix search hanging in private windows, and move the migrations off a used timestamp

brandonlucasgreen:
Two follow-ups to #431. The first is a production bug I introduced there and should be merged as soon as it's reviewed.

## Search hung in private windows — regression from #431

**Symptom:** search sat on its skeletons forever, and `/api/search/sources` was never requested at all. Only in private windows; signed-in browsing was fine.

**Cause.** The GoatCounter change in #431 sets `window.goatcounter` to a settings object in `index.html`, so `count.js` reads the path from it on load. That means the object now exists from the first paint, and `count.js` is what adds `count` to it later — if it arrives. Private windows and content blockers stop it arriving, and

```js
window.goatcounter?.count({ path, event: true })
```

guards the object, not the method. Previously a blocked `count.js` left `window.goatcounter` undefined and the optional chain short-circuited harmlessly. Now the object is always there, so the call reached `.count` and threw `TypeError: window.goatcounter?.count is not a function`.

`App.tsx` calls `analytics.trackSearch()` on the line *above* the `try` block wrapping the search, so the throw escaped `handleSearch` entirely — after `setIsLoading(true)`, before `searchPlatforms()` was ever called. That's the permanent skeleton and the empty network tab.

Signed-in browsing was unaffected only because `count.js` loads normally there. Nothing about it was auth-related, which is what made it look like a backend problem.

**Fix.** Guard `count`, not just the object. The `try/catch` around it is the real lesson: analytics is best-effort and must never be able to reach the caller, whatever the next blocker does differently.

**Verified.** New tests cover all four shapes `window.goatcounter` can take — absent, settings-only, throwing, and loaded normally. The settings-only case fails with the exact production error without the fix, and the whole file passes with it. Also reproduced in a real browser with `count.js` blocked at the network layer: before, no search request; after, the request fires and there are no page errors.

## Migrations moved off a timestamp production has already used

`supabase db push` failed on the merge of #431 before applying anything:

```
Remote migration versions not found in local migrations directory.
  supabase migration repair --status reverted 20260808000000
```

Production is ahead of `main`. `supabase-migrate.yml` was run by manual `workflow_dispatch` against the unmerged `claude/email-capture-newsletter-strategy` branch at 12:45, applying that branch's `20260808000000_email-log` and `20260808120000_notification-preferences` straight to the production database. Neither file is on `main`, so the next push from `main` found a remote version it couldn't account for and refused to continue. **Nothing from #431 reached the database — the RLS fix is not live.**

It also means `20260808120000` is spent. Supabase keys history on the timestamp prefix alone, so the RLS migration sharing that prefix would have been read as already-applied and skipped silently once the history mismatch was resolved — the worst outcome for a migration whose job is closing a hole. Both files are renamed to `20260808160000` / `20260808170000`, strictly after everything production has and after `notification_preferences` whenever that branch merges.

**This unblocks the ordering but not the history mismatch.** `main` still needs the two migration files production has already run before `db push` will proceed — cleanest by merging the newsletter PR, since its migrations are live in production already and `main` is the thing that's out of date. I'd avoid `migration repair --status reverted`, which would tell Supabase that a table which exists doesn't.

Worth considering separately: `workflow_dispatch` on that workflow runs against the **production** database from whatever branch is selected, which is how the two got out of step.

## Testing

`npm run build` green, `npm run lint` unchanged (71 pre-existing problems), 714 web unit tests (4 new) and 1017 API tests passing.

https://claude.ai/code/session_01MVxipMUsXsnABZ51k21sZJ

---
_Generated by [Claude Code](https://claude.ai/code/session_01MVxipMUsXsnABZ51k21sZJ)_


brandonlucasgreen (commented):

**Security:** nothing found. No new auth/authorization, input-validation, secret-handling, injection, or data-exposure surface — this is a client-side try/catch around a best-effort analytics call, plus a pure filename rename of two already-reviewed migrations (only a header comment added, policy SQL unchanged per the diff).

**Regression:** none identified.
- The `trackEvent` fix (guard `count`, not just the settings object) matches the actual failure mode described in the PR body, and the new test file exercises all four `window.goatcounter` shapes (absent, settings-only/no-count — the regression case, throwing, and normal) plus the working case, so the fix is verified rather than just plausible.
- The migration rename is a pure rename (`git diff` shows `similarity index 93%`/`100%`) — no SQL content changed, so it carries no independent correctness risk; it only resolves the timestamp collision with what's already live in production, which is exactly what it needs to do given `20260808120000` was already consumed by the out-of-band `workflow_dispatch` run described in the PR body.

One thing worth double-checking outside the code (not a finding, just a process note): the PR body says merging the newsletter PR (#430) first is the cleanest path to resolving the migration-history mismatch before `main`'s `db push` will proceed — worth sequencing these two merges accordingly.

---

PR #435, opened 2026-08-09
Add Guides, FAQ, and Contact links to the logged-out header

brandonlucasgreen:
## Summary
- The logged-out header only had a Login link. Adds Guides, FAQ, and Contact to its left, matching what the footer already links to.
- Responsive: same shared `navItems` array feeds both the desktop inline nav and the existing mobile hamburger drawer, so at narrow widths it collapses automatically — no new responsive logic needed.
- Contact points offsite to the same Letterbird address used in the footer. Since `MobileNav`'s `NavItem`/render loop only supported internal router `Link`s, added an `external` flag so both the desktop nav and drawer render it as a plain `<a target="_blank" rel="noopener noreferrer">` instead.

## Test plan
- [x] `npx tsc -b apps/web --noEmit` — clean
- [x] `npm run lint` — no new errors (pre-existing baseline noise unchanged)
- [x] `npm run test:unit` — 717 passing
- [x] Verified in browser at desktop width: Guides / FAQ / Contact / Login render inline, correct hrefs, Contact opens in a new tab
- [x] Verified in browser at mobile width (375px): hamburger opens a drawer with Guides / FAQ / Contact / Login stacked in order

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #437, opened 2026-08-09
Build artist locations from one source instead of merging fields

brandonlucasgreen:
The Foo Fighters artist page read **"Seattle, California"**. Spotted in a screenshot of the search result view on 2026-08-09.

It is not a bad record. Two independent defects produced it, and 244 production rows are affected.

## What was wrong

**Locations were assembled a field at a time across sources.** `mergeLocations` took the city from MusicBrainz and then filled the empty `country` slot from whichever other source had one. MusicBrainz says Foo Fighters formed in Seattle (their `begin-area`); `foofighters.bandcamp.com` says "California". Each source was right on its own. The splice named a place that does not exist.

**MusicBrainz areas were classified by `type`, which is null on almost all real data.** Foo Fighters' area is `{ name: "United States", type: null, "iso-3166-1-codes": ["US"] }`, so the country was filed as a *city*. That is where the "United States", "Germany" and "Michigan" values in the `city` column came from.

## Blast radius, measured against production

244 `artists` rows carry a location. Confirmed chimeras:

| Artist | Stored | Reality |
|---|---|---|
| Foo Fighters | Seattle, California | Seattle is Washington |
| Modest Mouse | Issaquah, Oregon | Issaquah is Washington |
| Zeal & Ardor | Basel, New York | Basel is Switzerland — the row's own `country_code` says `CH` |
| Honeycrush | Brooklyn, Florida | — |
| Pole | Düsseldorf, Russia | `country_code` says `DE` |
| Boy Harsher | Savannah, Massachusetts | — |
| Citizen | Michigan, Michigan | `city` held a country/state |

Plus 21 rows with `country = "California"`, 11 `"New York"`, 7 `"Ontario"` — US states and Canadian provinces in the country column — and one reading `"Uni"`.

## The fix

In `api/search/enrichment.ts`:

- **`parseMusicBrainzArea`** classifies by the ISO code arrays (`iso-3166-1-codes` = country, `iso-3166-2-codes` = subdivision, neither = city) rather than by `type`, and resolves a bare country code to a name via `Intl.DisplayNames` so pages no longer read "Sacramento, US".
- **`pickLocation`** replaces `mergeLocations`: it takes the most specific source **whole**. Fields are never combined across sources. MusicBrainz is preferred over Bandcamp because the Bandcamp probe can match the wrong account — that is how Honeycrush got its row.

Both search functions carried their own copy of the area block; both now call the shared helper.

No new column. `country` is already a display slot, not a real country — every client renders `city, country ?? countryCode`, including the Mac app's `region` computed properties. A real `region` field would mean a migration plus five clients plus an API shape change that shipped Mac builds would ignore.

## Verification

13 tests in `api/functions/__tests__/artist-location.test.ts`, built on the real API payloads. Checked they have teeth: 5 fail against the old logic.

End to end through the real handler (one artist per process — batching trips MusicBrainz's 1 req/sec limit and silently exercises the fallback branch):

- Foo Fighters → **Seattle, United States**
- Explosions in the Sky → Austin, United States
- Destroy Boys → Sacramento, United States

`npm run test:api` 1066 pass · `npm run test:unit` 717 pass · `typecheck:api` clean · lint at its existing 71-problem baseline.

## Existing rows

`scripts/backfill-artist-locations.ts` (`npm run backfill:locations`, dry run by default, `-- --write` to apply) re-derives the stored rows. Update-only, three columns, by primary key — no inserts, no deletes. It leaves alone any row MusicBrainz cannot confidently match, and any row where MusicBrainz is *less* specific than what is stored, so AZALI's "Seattle, Washington" is not traded for a bare "United States".

Being run against production alongside this PR.

## Trade-off

MusicBrainz usually stops at the country, so US artists resolved from it now read "Austin, United States" rather than "Austin, Texas" — roughly half the affected rows lose the state. Recovering it means taking the region from a second source, which is the bug this PR removes. That is only safe when both sources name the same city, and it still breaks on Portland, Columbus and Springfield. Putting "Portland, Maine" on a Portland, Oregon band is worse than "Portland, United States". Correctness over specificity — easy to revisit.


scripts/backfill-artist-locations.ts
```diff
@@ -0,0 +1,225 @@
...
+    'begin-area'?: MusicBrainzArea;
+  };
+  return parseMusicBrainzArea(data.area, data['begin-area'], data.country) ?? null;
+}
+
+async function main() {
+  const rows = await readAllPages<ArtistRow>((from, to) =>
+    client
+      .from('artists')
+      .select('id, slug, name, city, country, country_code')
```
brandonlucasgreen, line 174: **Regression — confirmed.** This query has no `match_confidence` filter, so it selects (and, under `--write`, updates) claimed artists along with unclaimed ones.

That breaks an invariant enforced everywhere else in the write path: `db.ts` skips claimed rows for both profile persistence (`if (existing?.match_confidence === 'claimed') { ... skip }`, ~db.ts:3509) and link enrichment (`if (artist.match_confidence === 'claimed') { ... skip }`, ~db.ts:3619), and this migration's own header comment says location specifically is included in that guarantee: *"claimed artists manage their own location via the claim/edit flow (Phase 3) and are skipped by the enrichment path (match_confidence='claimed' short-circuit in db.ts)"* (`supabase/migrations/20260424070000_artist-location.sql:5-7`).

`wouldLoseACity` (line ~104) doesn't help here — it only refuses to overwrite when the derived location has *no* city at all. If a claimed artist set "Portland, Oregon" by hand and MusicBrainz resolves the band to a different or less precise city (or just disagrees), the script logs `CHANGE` and, under `--write`, silently overwrites the artist-curated value.

Since the PR description says this script is being run against production alongside this PR, suggest adding `match_confidence` to the `ArtistRow` select/interface and skipping (or at minimum flagging) any row where it's `'claimed'`, the same way the rest of the codebase does.

---

api/search/enrichment.ts
brandonlucasgreen, line 1: **Plausible, low severity.** `pickLocation`'s doc comment says it takes "the most specific source that answered," but the implementation actually prefers *argument order* first, specificity second: `answered.find(s => s.city) ?? answered.find(s => s.country) ?? answered[0]` picks the first source in the call (MusicBrainz) if it has *any* city, even a bare one, ahead of a later source that might have a fuller `city + country` record. That matches the "correctness over specificity" trade-off described in the PR body and is covered by the passing tests, so this reads as intentional — just flagging that the doc comment overstates it as picking the most complete record rather than the first-in-priority-order one with a city.

---

brandonlucasgreen (commented):
@brandonlucasgreen Staff-eng-style security + regression pass on this PR.

## Security

Nothing in the core diff (`api/search/enrichment.ts`, `search-musicbrainz.ts`, `search-sources.ts`) changes auth, CORS, or SSRF surface — the fetch paths are unchanged, only the area-parsing/merge logic is. No injection, secret-handling, or unsafe-default issues found there.

- **Auth/authz** — no change. **needs-manual-verification**
- **Input validation** — `parseMusicBrainzArea`/`pickLocation` only reshape already-fetched MusicBrainz/Bandcamp/Mirlo data; no new external input parsed. **confirmed** (no issue)
- **Secret handling** — `scripts/backfill-artist-locations.ts` reads `SUPABASE_SERVICE_KEY` from env, matching the convention used by every other backfill script in `scripts/`; never logged. **confirmed** (no issue)
- **Injection risks** — the script's MusicBrainz query is `encodeURIComponent`-escaped before going into the URL. **confirmed** (no issue)
- **Unsafe defaults** — the backfill script defaults to dry-run (`--write` is opt-in), which is the right default for a script that hits production. **confirmed** (no issue)
- **Data exposure** — see the claimed-artist finding below; it's a data-*integrity* issue rather than exposure, but worth being aware it touches production artist rows. **confirmed**, inline comment on `scripts/backfill-artist-locations.ts`

## Regression

1. **`scripts/backfill-artist-locations.ts` overwrites claimed artists' locations — confirmed, high severity.** Inline comment on the file. The row-select has no `match_confidence` filter, so it includes claimed profiles; `wouldLoseACity` only guards against losing a city outright, not against clobbering a claimed artist's hand-set location with a different MusicBrainz-derived one. This breaks the invariant this migration's own comment describes ("claimed artists manage their own location... skipped by the enrichment path") and that `db.ts` enforces everywhere else (`match_confidence === 'claimed'` short-circuits for both profile persistence and link enrichment). Since the PR says this script runs against production alongside the PR, this is live-data risk, not hypothetical.

2. **Branch is stale relative to `main` — confirmed.** `git merge-base origin/main origin/magical-keller-514eef` resolves to `f40bd55` (#435), one commit behind current `main` tip `5c92f86` (#436, "Add a press kit at /press and a contact page at /contact"). That's exactly what the repo's own `semantic-revert-check` bot flagged in its comment on this PR — it diffs against current `main` and sees the press-kit/contact-page files as deletions because this branch was cut before they existed, not because this PR touches them (the actual diff is only the 6 files listed above). `mergeable_state` is `"unstable"`, not `"dirty"`, so GitHub isn't reporting a conflict, and a real merge commit or squash-merge should reconcile cleanly — but there's no upside to leaving it stale. Rebase onto `main` before merging so CI runs against current `main` and the bot warning clears.

3. **`pickLocation` doc says "most specific," code says "first-in-order-with-a-city" — plausible, low severity.** Inline comment on `api/search/enrichment.ts`. Looks intentional given the PR's stated correctness-over-specificity trade-off and is covered by tests; just a doc-vs-behavior mismatch worth a one-line tweak.

4. **Edge cases / null-empty behavior** — `parseMusicBrainzArea`'s classification-by-ISO-code (vs. the old `type`-based check) is well covered by the 13 new tests, including the null-`type` cases that caused the original bug. No gaps found. **confirmed** (no issue)
5. **State transitions** — `pickLocation`'s "one source, whole" replacement of `mergeLocations` is a real behavior change for every artist whose location previously came from two sources spliced together; that's the intended fix, and the trade-off (losing US state-level specificity in about half of affected rows) is explicitly documented in the PR description. **confirmed** (intentional, documented)
6. **API contract drift** — none; `ArtistLocation`'s shape is unchanged, only how it's populated.

No blocking security issues. The one thing I'd want resolved before this merges (or before the backfill script is run with `--write` again) is #1 — it's a straightforward fix (add `match_confidence` to the select and skip `'claimed'` rows) and matches the pattern already used throughout `db.ts`.

---

PR #438, opened 2026-08-10
Support Loop, web half: Bandcamp collection import, auto-supported, collection pages (Steps 1 + 3)

brandonlucasgreen:
## What this is

Steps 1 and 3 of the Support Loop (`support-loop-spec.md`), plus Brandon's answers to open questions 5 and 6: the full web half of the loop, from connecting a Bandcamp collection to a shareable collection page.

## Step 1 — Bandcamp collection import

1. `/settings` → **Bandcamp collection** → generate a Subsonic credential in Bandcamp's Fan Settings, paste username + credential.
2. `me-bandcamp` verifies the credential against Bandcamp inline (`ping`), converts it to the Subsonic salted-token pair, encrypts it, stores it, and kicks off `bandcamp-sync-background`.
3. The background function pages through `getAlbumList2`, upserts items into the new `collection_items` table as `provenance: purchased` (idempotent on `user_id, source, external_id`), and matches albums to Unstream artists/releases by normalized name — skipping ambiguous names rather than guessing.
4. The settings section polls while syncing, then offers Re-sync and Disconnect (with an explicit keep-or-delete-items choice).

## OQ6 — buying is supporting

The import marks every unambiguously matched artist as **saved + supported**. Guardrails: tombstoned rows are never resurrected (permanent dismissal is a locked spec decision), an already-supported row keeps its original `supported_at`, and nothing can ever un-support or touch notes. Newly saved artists get catalogued by the six-hourly sweep (saved artists sort first) rather than an immediate crawl burst.

## Step 3 — the collection page (OQ5: two views of the same data)

`/u/:handle` now leads with an **album-art grid** of purchased, non-hidden items — counts in the header ("N releases collected · M artists supported"), matched items linking to their release page so a viewer can buy the same record — with the artists list below. Same content on the crawler render (capped at 60 tiles). `/dashboard` gains a **Your Collection** section: the owner's full grid including hidden items, per-item hide/unhide, and an empty state that points at the import.

## Decisions worth reviewing

- **Credential encryption (spec OQ1):** app-level AES-256-GCM via `node:crypto`, key in Netlify env `BANDCAMP_CREDENTIAL_KEY`. pgcrypto puts the key in SQL text (Postgres logs); Vault keeps key and ciphertext in one trust domain. With this split, a leak of either store alone reveals nothing.
- **`bandcamp_connections` is server-only** (RLS, no policies): even the owner's session can't read the ciphertext.
- **`collection_items` INSERT is service-role only** so nothing client-side can mark itself `purchased` — the provenance gate is what keeps the public page honest.
- **Never cache uncertainty, sync edition:** a mid-pagination failure records `sync_status='error'`; a partial import is never recorded as a completed sync. Likewise a failed collection read on the public page is a 500, not an empty collection.
- **`listening_signals` ships now** (schema only) as the landing zone for Steps 2/5.

## Verification

- 50 new tests across crypto, the Subsonic client, connect/disconnect, sync + auto-mark rules, collection endpoints, and both UIs; full suites pass (1,120 API + 726 web) and `npm run build` is green. Lint at the pre-existing 71-problem baseline.
- Migration validated against a throwaway Postgres 16: clean apply, idempotent re-apply, upsert dedup, NULL-distinct manual items, CHECK rejections, RLS shape.

## Before this works in production

1. Set `BANDCAMP_CREDENTIAL_KEY` in Netlify (Functions scope): `node -e "console.log(require('crypto').randomBytes(32).toString('base64'))"`
2. `INTERNAL_FUNCTION_SECRET` is already set for the catalog functions and is reused here.

Not testable end-to-end until a real Bandcamp account connects — spec OQ2 (what the beta actually returns) still needs that.

Pairs with #439 (Apple Music library import, Step 2). Independent code paths; either can merge first.

### DIFF
--- a/api/functions/me-bandcamp.ts
+++ b/api/functions/me-bandcamp.ts
@@ -213,10 +213,10 @@ export async function handler(event: {
     const username = typeof body.username === 'string' ? body.username.trim() : '';
     const password = typeof body.password === 'string' ? body.password : '';
     if (!username || username.length > 200) {
-      return { statusCode: 400, headers: CORS_HEADERS, body: JSON.stringify({ error: 'A Bandcamp username is required' }) };
+      return { statusCode: 400, headers: CORS_HEADERS, body: JSON.stringify({ error: 'The username Bandcamp generated is required' }) };
     }
     if (!password || password.length > 500) {
-      return { statusCode: 400, headers: CORS_HEADERS, body: JSON.stringify({ error: 'The credential from Bandcamp’s Fan Settings is required' }) };
+      return { statusCode: 400, headers: CORS_HEADERS, body: JSON.stringify({ error: 'The password Bandcamp generated is required' }) };
     }
 
     if (!isCredentialKeyConfigured()) {
@@ -240,7 +240,7 @@ export async function handler(event: {
         return {
           statusCode: 400,
           headers: CORS_HEADERS,
-          body: JSON.stringify({ error: 'Bandcamp rejected that username or credential. Check both in Bandcamp’s Fan Settings → Subsonic.' }),
+          body: JSON.stringify({ error: 'Bandcamp rejected that username or password. Check both against Fan Settings → Subsonic on Bandcamp.' }),
         };
       }
       console.warn('[me-bandcamp] Bandcamp ping failed:', error instanceof Error ? error.message : String(error));
--- a/api/functions/me-bandcamp.ts
+++ b/api/functions/me-bandcamp.ts
@@ -1,0 +1,336 @@
+// API endpoint: /api/me/bandcamp
+// GET    — connection status: connected, username, sync state, item count, last sync.
+// POST   — { username, password }: connect. Verifies against Bandcamp's Subsonic API,
+//          stores the encrypted salted-token pair (never the password), starts a sync.
+//          { resync: true }: re-run the import for an existing connection.
+// DELETE — { deleteItems? }: disconnect. Deletes the credential row; optionally deletes
+//          imported collection items too.
+//
+// Support Loop Step 1 (support-loop-spec.md). Credential rules from collection-spec.md §4:
+// the password exists only inside the POST request — it is converted to the Subsonic
+// (t, s) pair, encrypted (credential-crypto.ts), and discarded. It must never be logged,
+// stored, echoed back, or attached to a Sentry event.
+
+import { createClient } from '@supabase/supabase-js';
+import { Sentry } from '../lib/sentry';
+import { getClient } from './db';
+import { checkRateLimit, getClientIp } from './ratelimit';
+import { encryptCredential, isCredentialKeyConfigured } from './credential-crypto';
+import {
+  deriveSubsonicToken,
+  subsonicPing,
+  subsonicArtistCount,
+  SubsonicError,
+} from './bandcamp-subsonic';
+
+const CORS_HEADERS = {
+  'Content-Type': 'application/json',
+  'Access-Control-Allow-Origin': '*',
+  'Access-Control-Allow-Headers': 'Content-Type, Authorization',
+  'Access-Control-Allow-Methods': 'GET, POST, DELETE, OPTIONS',
+};
+
+interface ConnectionRow {
+  bandcamp_username: string;
+  sync_status: string;
+  sync_error: string | null;
+  item_count: number | null;
+  last_synced_at: string | null;
+}
+
+function toStatusShape(row: ConnectionRow | null) {
+  if (!row) return { connected: false as const };
+  return {
+    connected: true as const,
+    username: row.bandcamp_username,
+    syncStatus: row.sync_status,
+    syncError: row.sync_error,
+    itemCount: row.item_count,
+    lastSyncedAt: row.last_synced_at,
+  };
+}
+
+async function authenticateRequest(authHeader: string | undefined): Promise<{ userId: string } | null> {
+  if (!authHeader?.startsWith('Bearer ')) return null;
+  const token = authHeader.slice(7);
+
+  const url = process.env.SUPABASE_URL;
+  const anonKey = process.env.SUPABASE_ANON_KEY;
+  if (!url || !anonKey) return null;
+
+  const anonClient = createClient(url, anonKey);
+  const { data, error } = await anonClient.auth.getUser(token);
+  if (error || !data.user) return null;
+  return { userId: data.user.id };
+}
+
+/**
+ * Ask the background function to run the import. Returns false when it couldn't be asked
+ * (missing config, network) — the caller records that as a sync error rather than leaving
+ * the row claiming a sync that will never happen. Same handshake as request-catalog.ts:
+ * Netlify answers 202 immediately, so there is no result to read back.
+ */
+async function requestSync(userId: string): Promise<boolean> {
+  const secret = process.env.INTERNAL_FUNCTION_SECRET;
+  // `URL` and not `DEPLOY_PRIME_URL`: only URL, SITE_NAME and SITE_ID reach a function at
+  // runtime — see request-catalog.ts.
+  const siteUrl = process.env.URL;
+  if (!secret || !siteUrl) {
+    console.log('[me-bandcamp] sync not requested — INTERNAL_FUNCTION_SECRET or site URL not configured');
+    return false;
+  }
+
+  try {
+    const controller = new AbortController();
+    const timeoutId = setTimeout(() => controller.abort(), 3_000);
+    try {
+      await fetch(`${siteUrl}/.netlify/functions/bandcamp-sync-background`, {
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+          Authorization: `Bearer ${secret}`,
+        },
+        body: JSON.stringify({ userId }),
+        signal: controller.signal,
+      });
+    } finally {
+      clearTimeout(timeoutId);
+    }
+    return true;
+  } catch (error) {
+    console.warn('[me-bandcamp] sync request failed:', error instanceof Error ? error.message : String(error));
+    return false;
+  }
+}
+
+const SYNC_START_FAILED = 'Sync could not be started. Use Re-sync to try again.';
+
+export async function handler(event: {
+  httpMethod: string;
+  headers: Record<string, string | undefined>;
+  body: string | null;
+}) {
+  if (event.httpMethod === 'OPTIONS') {
+    return { statusCode: 204, headers: CORS_HEADERS, body: '' };
+  }
+
+  const ip = getClientIp(event.headers);
+  const rl = await checkRateLimit(ip, 'standard', CORS_HEADERS);
+  if (rl.limited) return rl.response;
+
+  const client = getClient();
+  if (!client) {
+    return { statusCode: 500, headers: CORS_HEADERS, body: JSON.stringify({ error: 'Database not configured' }) };
+  }
+
+  const user = await authenticateRequest(event.headers.authorization);
+  if (!user) {
+    return { statusCode: 401, headers: CORS_HEADERS, body: JSON.stringify({ error: 'Not authenticated' }) };
+  }
+
+  if (event.httpMethod === 'GET') {
+    const { data: row, error } = await client
+      .from('bandcamp_connections')
+      .select('bandcamp_username, sync_status, sync_error, item_count, last_synced_at')
+      .eq('user_id', user.userId)
+      .maybeSingle();
+
+    if (error) {
+      console.error('[me-bandcamp] Error fetching connection:', error.message);
+      return { statusCode: 500, headers: CORS_HEADERS, body: JSON.stringify({ error: 'Failed to load connection status' }) };
+    }
+
+    return { statusCode: 200, headers: CORS_HEADERS, body: JSON.stringify(toStatusShape(row)) };
+  }
+
+  if (event.httpMethod === 'POST') {
+    let body: Record<string, unknown>;
+    try {
+      body = JSON.parse(event.body || '{}');
+    } catch {
+      return { statusCode: 400, headers: CORS_HEADERS, body: JSON.stringify({ error: 'Invalid JSON' }) };
+    }
+
+    // Re-sync an existing connection.
+    if (body.resync === true) {
+      const { data: existing, error } = await client
+        .from('bandcamp_connections')
+        .select('sync_status')
+        .eq('user_id', user.userId)
+        .maybeSingle();
+      if (error) {
+        console.error('[me-bandcamp] Error checking connection for resync:', error.message);
+        return { statusCode: 500, headers: CORS_HEADERS, body: JSON.stringify({ error: 'Failed to start sync' }) };
+      }
+      if (!existing) {
+        return { statusCode: 404, headers: CORS_HEADERS, body: JSON.stringify({ error: 'No Bandcamp connection to sync' }) };
+      }
+      if (existing.sync_status === 'syncing') {
+        return { statusCode: 409, headers: CORS_HEADERS, body: JSON.stringify({ error: 'A sync is already running' }) };
+      }
+
+      // Mark syncing before asking: the background function reads this row, so the
+      // status write must land first (same ordering as connect below).
+      const { error: markError } = await client
+        .from('bandcamp_connections')
+        .update({ sync_status: 'syncing', sync_error: null })
+        .eq('user_id', user.userId);
+      if (markError) {
+        console.error('[me-bandcamp] Error updating sync status:', markError.message);
+        return { statusCode: 500, headers: CORS_HEADERS, body: JSON.stringify({ error: 'Failed to start sync' }) };
+      }
+
+      const started = await requestSync(user.userId);
+      if (!started) {
+        await client
+          .from('bandcamp_connections')
+          .update({ sync_status: 'error', sync_error: SYNC_START_FAILED })
+          .eq('user_id', user.userId);
+      }
+
+      const { data: row, error: readError } = await client
+        .from('bandcamp_connections')
+        .select('bandcamp_username, sync_status, sync_error, item_count, last_synced_at')
+        .eq('user_id', user.userId)
+        .single();
+      if (readError) {
+        console.error('[me-bandcamp] Error reading connection after resync:', readError.message);
+        return { statusCode: 500, headers: CORS_HEADERS, body: JSON.stringify({ error: 'Failed to start sync' }) };
+      }
+      return { statusCode: 200, headers: CORS_HEADERS, body: JSON.stringify(toStatusShape(row)) };
+    }
+
+    // Connect.
+    const username = typeof body.username === 'string' ? body.username.trim() : '';
+    const password = typeof body.password === 'string' ? body.password : '';
+    if (!username || username.length > 200) {
+      return { statusCode: 400, headers: CORS_HEADERS, body: JSON.stringify({ error: 'A Bandcamp username is required' }) };
+    }
+    if (!password || password.length > 500) {
+      return { statusCode: 400, headers: CORS_HEADERS, body: JSON.stringify({ error: 'The credential from Bandcamp’s Fan Settings is required' }) };
+    }
+
+    if (!isCredentialKeyConfigured()) {
+      // Configuration, not user error — and worth a Sentry event because the feature is
+      // silently unusable until the env var lands.
+      console.error('[me-bandcamp] BANDCAMP_CREDENTIAL_KEY is not configured');
+      Sentry.captureMessage('me-bandcamp: BANDCAMP_CREDENTIAL_KEY not configured', 'error');
+      return { statusCode: 500, headers: CORS_HEADERS, body: JSON.stringify({ error: 'Bandcamp connections are not available right now' }) };
+    }
+
+    // Derive the salted token; the password is not referenced again after this line.
+    const { t, s } = deriveSubsonicToken(password);
+    const credential = { username, t, s };
+
+    // Verify before storing: a stored credential that never worked would surface as a
+    // confusing background sync failure instead of an inline "check your credential".
+    try {
+      await subsonicPing(credential);
+    } catch (error) {
+      if (error instanceof SubsonicError && error.isAuthFailure) {
+        return {
+          statusCode: 400,
+          headers: CORS_HEADERS,
+          body: JSON.stringify({ error: 'Bandcamp rejected that username or credential. Check both in Bandcamp’s Fan Settings → Subsonic.' }),
+        };
+      }
+      console.warn('[me-bandcamp] Bandcamp ping failed:', error instanceof Error ? error.message : String(error));
+      return {
+        statusCode: 502,
+        headers: CORS_HEADERS,
+        body: JSON.stringify({ error: 'Bandcamp didn’t respond. Their Subsonic support is in beta — try again in a minute.' }),
+      };
+    }
+
+    // A nice-to-have confirmation number; a failure here shouldn't block the connect.
+    let artistCount: number | null = null;
+    try {
+      artistCount = await subsonicArtistCount(credential);
+    } catch {
+      // The sync will establish real counts.
+    }
+
+    const ciphertext = encryptCredential(JSON.stringify({ t, s }));
+
+    // Store first, then ask for the sync: the background function reads this row, so
+    // requesting before the upsert lands would race it into "no connection found".
+    const { error: upsertError } = await client
+      .from('bandcamp_connections')
+      .upsert(
+        {
+          user_id: user.userId,
+          bandcamp_username: username,
+          credential_ciphertext: ciphertext,
+          sync_status: 'syncing',
+          sync_error: null,
+        },
+        { onConflict: 'user_id' }
+      );
+
+    if (upsertError) {
+      console.error('[me-bandcamp] Error storing connection:', upsertError.message);
+      return { statusCode: 500, headers: CORS_HEADERS, body: JSON.stringify({ error: 'Failed to save the connection' }) };
+    }
+
+    const started = await requestSync(user.userId);
+    if (!started) {
+      await client
+        .from('bandcamp_connections')
+        .update({ sync_status: 'error', sync_error: SYNC_START_FAILED })
+        .eq('user_id', user.userId);
+    }
+
+    return {
+      statusCode: 200,
+      headers: CORS_HEADERS,
+      body: JSON.stringify({
+        connected: true,
+        username,
+        syncStatus: started ? 'syncing' : 'error',
+        syncError: started ? null : SYNC_START_FAILED,
+        artistCount,
+      }),
+    };
+  }
+
+  if (event.httpMethod === 'DELETE') {
+    let body: Record<string, unknown> = {};
+    try {
+      body = JSON.parse(event.body || '{}');
+    } catch {
+      return { statusCode: 400, headers: CORS_HEADERS, body: JSON.stringify({ error: 'Invalid JSON' }) };
+    }
+
+    const { error: deleteError } = await client
+      .from('bandcamp_connections')
+      .delete()
+      .eq('user_id', user.userId);
+
+    if (deleteError) {
+      console.error('[me-bandcamp] Error deleting connection:', deleteError.message);
+      return { statusCode: 500, headers: CORS_HEADERS, body: JSON.stringify({ error: 'Failed to disconnect' }) };
+    }
+
+    let itemsDeleted = 0;
+    if (body.deleteItems === true) {
+      const { count, error: itemsError } = await client
+        .from('collection_items')
+        .delete({ count: 'exact' })
+        .eq('user_id', user.userId)
+        .eq('source', 'bandcamp');
+      if (itemsError) {
+        console.error('[me-bandcamp] Error deleting collection items:', itemsError.message);
+        return {
+          statusCode: 500,
+          headers: CORS_HEADERS,
+          body: JSON.stringify({ error: 'Disconnected, but deleting imported items failed. Try again from Settings.' }),
+        };
+      }
+      itemsDeleted = count ?? 0;
+    }
+
+    return { statusCode: 200, headers: CORS_HEADERS, body: JSON.stringify({ connected: false, itemsDeleted }) };
+  }
+
+  return { statusCode: 404, headers: CORS_HEADERS, body: JSON.stringify({ error: 'Not found' }) };
+}
--- a/api/functions/me-bandcamp.ts
+++ b/api/functions/me-bandcamp.ts
@@ -83,8 +83,9 @@ async function requestSync(userId: string): Promise<boolean> {
   try {
     const controller = new AbortController();
     const timeoutId = setTimeout(() => controller.abort(), 3_000);
+    let response: Response;
     try {
-      await fetch(`${siteUrl}/.netlify/functions/bandcamp-sync-background`, {
+      response = await fetch(`${siteUrl}/.netlify/functions/bandcamp-sync-background`, {
         method: 'POST',
         headers: {
           'Content-Type': 'application/json',
@@ -95,6 +96,14 @@ async function requestSync(userId: string): Promise<boolean> {
       });
     } finally {
       clearTimeout(timeoutId);
+    }
+    // Netlify's dispatcher answers 202 when it accepts a background invocation. Anything
+    // else means the sync will never run — notably a 404 when this code runs on a deploy
+    // preview, whose `URL` points at production before the function has shipped there.
+    // Claiming "syncing" on that response would spin forever.
+    if (!response.ok) {
+      console.warn(`[me-bandcamp] sync dispatch refused: HTTP ${response.status}`);
+      return false;
     }
     return true;
   } catch (error) {
--- a/api/functions/me-bandcamp.ts
+++ b/api/functions/me-bandcamp.ts
@@ -65,16 +65,46 @@ async function authenticateRequest(authHeader: string | undefined): Promise<{ us
 }
 
 /**
- * Ask the background function to run the import. Returns false when it couldn't be asked
- * (missing config, network) — the caller records that as a sync error rather than leaving
- * the row claiming a sync that will never happen. Same handshake as request-catalog.ts:
- * Netlify answers 202 immediately, so there is no result to read back.
+ * The origin to dispatch the background job to: the deploy that is actually serving this
+ * request, when we can prove it's one of ours.
+ *
+ * Netlify exposes only `URL`, `SITE_NAME` and `SITE_ID` to a function at runtime, and `URL`
+ * always names *production* — so a deploy preview would hand its work to production, which
+ * on a branch-only function is a 404 and on a shipped one is the wrong deploy running it.
+ * The `Host` header does name the current deploy.
+ *
+ * But `Host` is client-supplied, and dispatching to it unchecked would POST
+ * INTERNAL_FUNCTION_SECRET to whatever hostname an attacker sent. So it is honoured only
+ * when it is one of this site's own Netlify hostnames; anything else falls back to `URL`,
+ * which is the safe (if less useful) old behaviour.
+ *
+ * Unlike catalog dispatch — which writes shared artist data and spends a shared crawl
+ * budget, and so stays pinned behind request-catalog.ts's production gate — a Bandcamp sync
+ * writes only the rows of the user who asked for it, so letting a preview run its own is
+ * both safe and the point.
  */
-async function requestSync(userId: string): Promise<boolean> {
+export function selfDispatchOrigin(host: string | undefined): string | null {
+  if (!host) return null;
+  const hostname = host.split(':')[0].trim().toLowerCase();
+  if (hostname === 'unstream.stream' || hostname === 'www.unstream.stream') {
+    return `https://${hostname}`;
+  }
+  // Deploy previews and branch deploys: <deploy-name>--unstream.netlify.app
+  if (hostname === 'unstream.netlify.app' || /^[a-z0-9-]+--unstream\.netlify\.app$/.test(hostname)) {
+    return `https://${hostname}`;
+  }
+  return null;
+}
+
+/**
+ * Ask the background function to run the import. Returns false when it couldn't be asked
+ * (missing config, network, a non-202 answer) — the caller records that as a sync error
+ * rather than leaving the row claiming a sync that will never happen. Same handshake as
+ * request-catalog.ts: Netlify answers 202 immediately, so there is no result to read back.
+ */
+async function requestSync(userId: string, host: string | undefined): Promise<boolean> {
   const secret = process.env.INTERNAL_FUNCTION_SECRET;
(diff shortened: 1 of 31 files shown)


supabase/migrations/20260809120000_bandcamp-collection.sql
```diff
@@ -0,0 +1,192 @@
...
+CREATE POLICY "Users can view own collection items"
+  ON collection_items FOR SELECT
+  USING (auth.uid() = user_id);
+
+CREATE POLICY "Users can update own collection items"
+  ON collection_items FOR UPDATE
+  USING (auth.uid() = user_id);
+
+CREATE POLICY "Users can delete own collection items"
+  ON collection_items FOR DELETE
```
brandonlucasgreen, line 148: **[confirmed] Authorization / data exposure — the UPDATE policy has no column scoping, so it doesn't actually protect `provenance`.**

Line 146's comment says INSERT is service-role-only "so a client that could insert its own rows could mark anything 'purchased'" — but that reasoning stops at INSERT. This `FOR UPDATE USING (auth.uid() = user_id)` policy (no `WITH CHECK`, matching the pattern from `saved_artists`/`usernames`/`notification_preferences`) grants the row owner unrestricted column-level UPDATE via direct PostgREST calls — the browser already holds both the anon key and the user's session JWT, so this doesn't require going through `me-collection.ts` at all.

`me-collection.ts` only ever writes `hidden`, but that's an app-level convention, not a DB-level restriction. RLS is row-scoped, not column-scoped: a signed-in user can `PATCH .../collection_items?id=eq.<their own row>` and rewrite `release_id`, `title`, `artist_name`, `art_url`, or `acquired_at` on any row they already own (e.g. a real $0 Bandcamp import), while `provenance` stays `'purchased'`. That fabricates "I purchased this" for an arbitrary release/artist they never paid for — exactly the property this table exists to protect (per the migration's own header comment: *"only 'purchased' is ever public... Inserts are service-role only so provenance stays honest"*), and it'll be user-facing once the public collection page (Step 3) reads this table.

This table's UPDATE exposure is materially worse than the existing tables that share this pattern, because none of `saved_artists`/`usernames`/`notification_preferences` carry a "this row asserts a real-world fact shown publicly" guarantee — `collection_items` explicitly does.

Suggested fix: restrict the client-facing UPDATE grant to the `hidden` column only (`REVOKE UPDATE ON collection_items FROM authenticated; GRANT UPDATE (hidden) ON collection_items TO authenticated;`), or drop this policy entirely and route hide/unhide exclusively through the service-role `me-collection.ts` endpoint (which already does the "you can only touch your own rows" check itself).

cc @brandonlucasgreen

---

api/functions/me-bandcamp.ts
```diff
@@ -0,0 +1,336 @@
...
+  httpMethod: string;
+  headers: Record<string, string | undefined>;
+  body: string | null;
+}) {
+  if (event.httpMethod === 'OPTIONS') {
+    return { statusCode: 204, headers: CORS_HEADERS, body: '' };
+  }
+
+  const ip = getClientIp(event.headers);
+  const rl = await checkRateLimit(ip, 'standard', CORS_HEADERS);
```
brandonlucasgreen, line 127: **[plausible] Unsafe default — connect is rate-limited more loosely than the analogous password endpoint.**

This endpoint accepts a live credential and verifies it against Bandcamp in real time (`subsonicPing`), the same shape of risk `me-password.ts` guards against — and that file uses `checkRateLimit(ip, 'strict', ...)` (10/min, 500/day) for exactly that reason. Here it's `'standard'` (30/min, 1000/day), 2-3x more permissive, on an endpoint whose own background-function comment elsewhere in this PR explicitly worries about becoming "a credential oracle." Since this is also acting as a live auth proxy against a third party (Bandcamp), not just protecting local state, I'd expect it to be at least as strict as `me-password.ts`, arguably more so given it's IP-keyed and an authenticated attacker with disposable Unstream accounts could use it to test Bandcamp credential lists at a materially higher rate than the codebase's existing convention allows elsewhere.

Worth a deliberate call on whether `'standard'` was intentional here, or should be `'strict'`.

---

api/functions/me-bandcamp.ts
```diff
@@ -0,0 +1,336 @@
...
+
+export async function handler(event: {
+  httpMethod: string;
+  headers: Record<string, string | undefined>;
+  body: string | null;
+}) {
+  if (event.httpMethod === 'OPTIONS') {
+    return { statusCode: 204, headers: CORS_HEADERS, body: '' };
+  }
+
```
brandonlucasgreen, line 125: **[regression / state-transition, confirmed via code] Non-atomic check-then-mark on `sync_status` — TOCTOU between concurrent resync requests.**

`resync` reads `sync_status`, checks it isn't `'syncing'`, then writes `'syncing'` in a separate request. Two concurrent calls (double-click, two tabs) can both pass the read-check before either write lands, starting two concurrent `bandcamp-sync-background` invocations against the same `user_id`. Both will independently upsert into `collection_items` and then race to write `sync_status`/`item_count`/`last_synced_at` on completion — last writer wins, so the row can end up `idle` with an `item_count` that doesn't match either sync's actual result.

Same shape of gap on connect (line ~194 below): the `upsert(..., { onConflict: 'user_id' })` on a *new* connect happens unconditionally even if a prior sync for the same user is still in flight — the old background job (holding the old decrypted credential already) will still run to completion and unconditionally overwrite `sync_status`, potentially clobbering the newer connect's state or writing results computed under the *old* Bandcamp identity after the row now claims the new username.

Not exploitable for privilege escalation (still scoped to the user's own row), but a real data-integrity edge case worth a conditional update (`UPDATE ... WHERE sync_status <> 'syncing'`, check rows affected) instead of select-then-update, and/or the background job re-validating it's still the current sync for that connection before writing results.

---

brandonlucasgreen (commented):
@brandonlucasgreen Security + regression pass on this one (credential handling, new RLS surface, background sync). Full diff read for `credential-crypto.ts`, `bandcamp-subsonic.ts`, `bandcamp-sync-background.ts`, `me-bandcamp.ts`, `me-collection.ts`, `BandcampConnect.tsx`, the migration, and `netlify.toml`/`api/tsconfig.json`.

**Auth/authorization**
- **[confirmed]** `collection_items` UPDATE RLS policy has no column scoping — see inline comment on the migration. This is the one I'd block on: it lets any signed-in user rewrite `release_id`/`title`/`artist_name` on their own rows via direct PostgREST calls, undermining the "provenance stays honest" guarantee the migration explicitly documents for INSERT but doesn't carry through to UPDATE.
- App-level ownership checks elsewhere (`me-collection.ts`'s `.eq('user_id', user.userId)` on every query, `me-bandcamp.ts` scoping DELETE to the caller) look correct.
- `bandcamp-sync-background.ts`'s `isInternalRequest` gate on the background function matches the existing `catalog-artist-background.ts` pattern — fine.

**Input validation**
- Username/password length caps, JSON parse guards, and Subsonic response parsing (`toAlbum`'s `typeof` checks before trusting any field) all look solid. No obvious under-validation.

**Secret handling**
- `credential-crypto.ts`'s AES-256-GCM design (key in Netlify env, ciphertext in Supabase, password discarded after token derivation, decrypt failures fail closed) is well-reasoned and matches the documented trade-off analysis. Didn't find anywhere the plaintext password or derived token leaks into logs/Sentry — `BandcampConnect.tsx` explicitly skips Sentry on the connect failure path to avoid carrying form contents into an exception, and server-side errors only ever include method names/error codes, never the request URL (which carries `u`/`t`/`s`).

**Injection risks**
- No raw SQL, no user-controlled URLs reaching `fetch` (Subsonic host is a hardcoded constant, explicitly reasoned about in `bandcamp-subsonic.ts`'s header comment re: SSRF). Nothing to flag.

**Unsafe defaults**
- **[plausible]** `me-bandcamp.ts` connect uses the `'standard'` rate-limit tier for a credential-verifying endpoint where `me-password.ts` uses `'strict'` for the analogous case — see inline comment.

**Data exposure**
- `bandcamp_connections` is correctly server-only (RLS, no policies) and `me-bandcamp.ts` never echoes the credential back. `me-collection.ts`'s `ITEM_COLUMNS` whitelist doesn't leak anything from the connections table. No issues found here beyond the UPDATE-scope gap above, which is a data-*integrity* exposure (fabricated purchase claims) rather than a confidentiality one.

**Regression**
- **[confirmed via code]** Non-atomic check-then-mark on `sync_status` (resync and connect both) — TOCTOU race between concurrent requests; see inline comment on `me-bandcamp.ts`.
- Edge cases around empty/ambiguous artist-name matching in `matchReleases` are handled conservatively (ambiguous → skip, matching-read failure → degrade to unmatched rather than fail the sync) — consistent with the "never cache uncertainty" principle, no issue.
- `subsonicFetchAllAlbums`'s `MAX_PAGES` hard-fail instead of silent truncation is the right call per the same principle.
- No existing API contracts changed; this is all new surface (new tables, new routes), so no drift to flag there.

Nice writeup in the PR description on the crypto trade-off analysis — that reasoning is sound, my pushback is specifically that it doesn't extend to the UPDATE path on `collection_items`.

---

supabase/migrations/20260809120000_bandcamp-collection.sql
```diff
@@ -0,0 +1,192 @@
...
+DROP POLICY IF EXISTS "Users can update own collection items" ON collection_items;
+DROP POLICY IF EXISTS "Users can delete own collection items" ON collection_items;
+
+CREATE POLICY "Users can view own collection items"
+  ON collection_items FOR SELECT
+  USING (auth.uid() = user_id);
+
+CREATE POLICY "Users can update own collection items"
+  ON collection_items FOR UPDATE
+  USING (auth.uid() = user_id);
```
brandonlucasgreen, line 145: **[security: authorization] confirmed** — the `"Users can update own collection items"` UPDATE policy has no `WITH CHECK` and no column restriction, only `USING (auth.uid() = user_id)`. Per Postgres RLS semantics, when `WITH CHECK` is omitted on an UPDATE policy, `USING` is reused for it — so it only restricts *which rows* an owner can touch, not *which columns or values* they may write.

`bandcamp_connections` and the (missing) INSERT policy on this same table are deliberately server-only "so provenance stays honest" — the migration's own comment says a client that could **insert** rows could mark anything `purchased`. This UPDATE policy lets an authenticated user achieve the identical outcome via **update**: using their own Supabase session (the anon key is public in the client bundle, and any user's JWT is trivially obtainable), they can call PostgREST directly — bypassing `me-collection.ts` entirely — and do e.g.

```
supabase.from('collection_items')
  .update({ provenance: 'purchased', release_id: '<any-uuid>', title: '...', artist_name: '...' })
  .eq('id', '<their own row id>')
```

`auth.uid() = user_id` still holds (they're not changing ownership), so the write succeeds. That flips their own `listened`/`owned` rows (once Steps 2/5 start writing those) to `purchased`, or repoints `release_id` at an arbitrary release — directly undermining "provenance stays server-asserted" on the one field the public collection page gates on.

This is the same bug class this repo already fixed once in `20260808160000_scope-service-policies-to-service-role.sql` ("RLS is row-level, not column-level — the lesson from migration 023"), and the fix used there is the same one that applies here: since the only client-facing mutation (`hidden`) already goes through `me-collection.ts` with the service-role client, drop this UPDATE policy (and rely on server-only writes, like `collection_items` INSERT already does) rather than trying to hand-roll a column-safe `WITH CHECK`.

---

api/functions/me-bandcamp.ts
brandonlucasgreen, line 127: **[security: unsafe defaults / credential oracle] plausible** — this handler's `'standard'` tier (30 req/min, 1000/day per IP, `api/functions/ratelimit.ts`) also gates the POST-connect branch, which verifies a `{username, password}` pair against Bandcamp via `subsonicPing`. That makes this endpoint a credential-testing oracle: an attacker can use it to test Bandcamp username/credential pairs (e.g. from a breach dump) through Unstream's IP rather than Bandcamp's directly.

`me-password.ts`, the closest existing analog (also verifies a credential inline), uses `'strict'` (10 req/min, 500/day). Worth using `'strict'` for the connect path here too, or splitting the tier by branch (GET/resync can stay `'standard'`, POST-connect goes `'strict'`) — the current per-IP limit is also the only throttle (no per-username lockout), so it's the main lever available.

---

api/functions/me-bandcamp.ts
```diff
@@ -0,0 +1,336 @@
...
+    try {
+      body = JSON.parse(event.body || '{}');
+    } catch {
+      return { statusCode: 400, headers: CORS_HEADERS, body: JSON.stringify({ error: 'Invalid JSON' }) };
+    }
+
+    const { error: deleteError } = await client
+      .from('bandcamp_connections')
+      .delete()
+      .eq('user_id', user.userId);
```
brandonlucasgreen, line 316: **[regression: state transitions] needs-manual-verification** — `DELETE` removes the `bandcamp_connections` row (and, with `deleteItems`, the `collection_items` rows) unconditionally, including while `bandcamp-sync-background.ts` has an in-flight sync for that user. That background function only checks for the connection row *once*, at the start (`bandcamp-sync-background.ts` lines ~1207-1221) — if a disconnect-and-delete races a running sync, the sync can finish afterward and re-upsert `collection_items` rows for a connection (and credential) the user just asked to delete, since its final `.update(...).eq('user_id', userId)` on a since-deleted row is a silent 0-row no-op rather than an error. Net effect: "Disconnect and delete items" doesn't reliably leave the collection empty if a sync was mid-flight. Worth either having the background job re-check the connection still exists before the final upsert/status-update, or having disconnect flip a cancellation flag the sync checks between pages.

---

brandonlucasgreen (commented):
@brandonlucasgreen Security + regression pass on this one — inline comments have the details.

**Security**

- **[confirmed]** `collection_items`'s `"Users can update own collection items"` RLS policy has no `WITH CHECK`/column restriction, so an authenticated user can flip their own rows' `provenance` to `purchased` (or repoint `release_id`) directly via PostgREST, bypassing the server-asserted-provenance guarantee the migration explicitly built the INSERT restriction to protect. Same bug class this repo already fixed in `20260808160000_scope-service-policies-to-service-role.sql`.
- **[plausible]** `me-bandcamp.ts`'s POST-connect branch (verifies a password against Bandcamp) shares the `'standard'` rate-limit tier with GET/resync, unlike the closest analog `me-password.ts`, which uses `'strict'` for its own inline credential check. Makes this a somewhat generous credential-testing oracle for Bandcamp accounts.
- Everything else checked out: `isInternalRequest` constant-time compare is reused correctly for the background function, `credential-crypto.ts`'s AES-256-GCM + separate-trust-domain key design is sound and well-reasoned, the Subsonic client never logs URLs/credentials, ownership checks (`user_id` in every query against the service-role client) are present on both `me-bandcamp.ts` and `me-collection.ts`, CORS wildcard on these is consistent with every other `me-*` endpoint in this codebase (bearer-token auth, not cookie-based), and the AppleScript-adjacent SSRF surface doesn't apply here since `SUBSONIC_SERVER` is a hardcoded constant, not user input.

**Regression**

- **[needs-manual-verification]** Disconnect-during-active-sync race: deleting the connection row doesn't stop an in-flight `bandcamp-sync-background` run, which can silently re-write `collection_items` after "Disconnect and delete items" was clicked. Detail on the inline comment.
- Edge cases around pagination (`subsonicFetchAllAlbums`'s `MAX_PAGES` backstop, mid-pagination throw instead of partial-success), duplicate Subsonic ids, and ambiguous-artist-name matching all look deliberately handled and are covered by tests.
- `api/tsconfig.json` was correctly updated to include all five new source files plus their tests, so this doesn't fall into the "unlisted `api/` file, type errors won't fail the build" trap called out in `CLAUDE.md`.

No findings on secret handling beyond the rate-limit note above (the credential is genuinely never logged, echoed, or Sentry-attached, and the encryption design is well-justified), and no injection risks — everything hitting Bandcamp goes through parameterized `URLSearchParams`.

---

brandonlucasgreen (commented):
@brandonlucasgreen Automated senior-staff-engineer review (scheduled security/regression pass). This is the biggest and most security-relevant of the three open PRs (first at-rest credential storage in the codebase), so it got the deepest read: credential-crypto.ts, bandcamp-subsonic.ts, me-bandcamp.ts, bandcamp-sync-background.ts, me-collection.ts, public-saved-artists.ts, u-handle.ts, and the new migration, plus a comparison against the existing `me-*.ts` CORS/auth conventions.

## Security

- **Subsonic credential in outgoing-request URLs vs. Sentry breadcrumbs — needs-manual-verification, potentially high severity.** `bandcamp-subsonic.ts` puts the Subsonic auth token/salt (`u`, `t`, `s`) in the request **query string**, and the file's own comment states these URLs "must therefore NEVER be logged, thrown, or attached to Sentry events." However, `bandcamp-sync-background.ts` calls `Sentry.captureException(...)` on every sync failure path (decrypt failure, any `SubsonicError`, etc.), and `api/lib/sentry.ts` initializes `@sentry/node` (v10) with no `beforeBreadcrumb` scrubbing and no explicit disabling of the default HTTP/fetch instrumentation. That default instrumentation has historically recorded outgoing request URLs as breadcrumbs, which — if not stripped of query strings by this SDK version/config — would attach the plaintext credential to the very Sentry event this code is trying to keep it out of. I could not verify the exact breadcrumb behavior for this SDK version without `node_modules` installed, so this needs a manual check (or an explicit `beforeBreadcrumb` scrub / fetch-integration exclusion added defensively regardless of the SDK's default).
- **SSRF-allowlist bypass — needs-manual-verification, low current risk.** `bandcamp-subsonic.ts`'s `subsonicRequest` calls `fetch()` directly instead of going through `isUrlHostnameAllowed()` / `ALLOWED_OUTBOUND_HOSTNAMES`, which CLAUDE.md calls mandatory for "any code fetching an external URL." The target (`SUBSONIC_SERVER`) is a hardcoded constant today, not derived from user input, so this isn't currently exploitable — but it's a policy exception, and the code comment ("if generic servers ever come back into scope... SUBSONIC_SERVER a parameter") flags that a future change would need to add the allowlist check at that point, not skip it.
- **Credential encryption design — reviewed, no issues found.** `credential-crypto.ts`: AES-256-GCM, random 12-byte IV per call, GCM auth tag verified on decrypt (tamper-evident), key/ciphertext kept in separate trust domains (Netlify env vs. Supabase), any decrypt failure is treated as "credential unusable" rather than silently degrading — consistent with the repo's "never cache uncertainty" principle.
- **Authorization — reviewed, no issues found.** `me-bandcamp.ts` / `me-collection.ts` derive `userId` from a verified Supabase bearer token and scope every query by it explicitly (the code even comments that the service-role client bypasses RLS, so the `user_id` filter *is* the ownership check). `bandcamp-sync-background.ts` — the only code path that decrypts a credential and calls out to Bandcamp — is gated by `isInternalRequest`, the same shared-secret pattern used by `catalog-artist-background.ts`, so it can't be triggered by an external caller to make Unstream hit Bandcamp with an arbitrary stored credential.
- **Data exposure — reviewed, no issues found.** `public-saved-artists.ts` and `u-handle.ts` both filter the public collection to `provenance = 'purchased' AND hidden = false` (matching the new RLS/spec intent) and project only non-sensitive fields. The new HTML interpolation in `u-handle.ts` (title, artist name, art URL, release URL) is consistently run through the file's `escapeHtml` copy, including inside the ``'&lt;img src="..."&gt;'`` attribute, so no new XSS surface from the collection tiles.
- **CORS — consistent with existing debt, not a new issue.** `me-bandcamp.ts`/`me-collection.ts` use the same wildcard-origin `CORS_HEADERS` as `me-settings.ts`/`me-username.ts`. Those existing files carry a comment acknowledging this as known "CORS debt" (Bearer-auth means a wildcard origin isn't directly exploitable for CSRF); the new files don't repeat that comment, which is a minor documentation-consistency nit, not a new vulnerability.

## Regression

- **API/behavior contract, confirmed intentional.** `markArtistsSupported` in `bandcamp-sync-background.ts` will flip `saved_artists.supported` back to `true` on a re-sync even if a user had explicitly un-supported (not deleted) that artist — only `deleted = true` tombstones are respected. This is explicitly called out in-code as a locked spec decision ("Brandon, 2026-08-09, spec open question 6"), so it's a deliberate contract rather than a bug — flagging only so it's confirmed as still-current product intent, since "un-support is not durable against a Bandcamp re-sync" is a real, user-visible behavior.
- **Future contract risk — needs-manual-verification.** The `collection_items` upsert (`onConflict: 'user_id,source,external_id'`) deliberately omits `hidden` so re-syncs don't un-hide user-hidden items, but it does unconditionally overwrite `title`/`artist_name`/`art_url` from the latest Bandcamp data on every re-sync. No manual-edit endpoint exists yet, so this is inert today — but it's a constraint any future "edit collection item" feature will need to respect (or it will be silently reverted on the next sync).
- **Migration deployed ahead of merge — noted, not a defect.** The migration's header comment states it was already applied to production via `workflow_dispatch` ahead of this PR merging, to let the (production-sharing) deploy preview exercise the connect flow. The change is purely additive (`CREATE TABLE IF NOT EXISTS`, `RLS enabled with no policies` on the credential table), so this doesn't look unsafe, but it's a deviation from the repo's normal "migration merges land its own deploy" ordering worth being deliberate about going forward.
- **State/edge cases — reviewed, no issues found.** `subsonicFetchAllAlbums` throws (rather than returning a partial result) on hitting `MAX_PAGES`, and `bandcamp-sync-background.ts`'s catch block records `sync_status = 'error'` rather than a false "idle/complete" — correctly avoids recording a partial import as a successful sync. Dedup-by-Subsonic-id before the upsert avoids a "same row twice in one batch" Postgres error.

## Summary

The credential-handling design itself (encryption choice, key separation, never-log discipline, ownership checks) is solid and clearly deliberate. The one item worth checking before this ships broadly is the Sentry-breadcrumb question above — it's the only place a real credential leak could plausibly slip through, and it's exactly the failure mode the code's own comments say must not happen.

---

brandonlucasgreen (commented):
@brandonlucasgreen Staff-eng review — security + regression pass. This is the highest-risk PR of the three open right now (credentials, encryption, new user data, public-facing rendering), so went deep on the security-critical primitives.

## Security — checked and clean (no finding)
Traced these directly against the code rather than taking the PR description at its word, since they're exactly the things this codebase has been burned by before:
- **SSRF**: `bandcamp-subsonic.ts` fetches a hardcoded `SUBSONIC_SERVER` constant; user-supplied `username` only ever appears as a `URLSearchParams` value (`u=`), never in host/path. No injection path from username to request destination.
- **Secrets/logging**: no thrown error or log statement includes the built Subsonic URL (which carries the token/salt). `BANDCAMP_CREDENTIAL_KEY` appears nowhere as a literal — tests generate a random key at runtime.
- **RLS**: `bandcamp_connections` — enabled, zero policies (deny-all to anon, service-role bypasses). `collection_items` — SELECT/UPDATE/DELETE scoped to `auth.uid() = user_id`, **no INSERT policy** (so INSERT is denied to everything but service-role) — confirms the PR's "nothing client-side can mark itself purchased" claim. `listening_signals` — same shape.
- **IDOR**: `userId` on every `me-*` endpoint comes solely from `auth.getUser(token)`; no client-suppliable `user_id` param anywhere. Hide/unhide chains `.eq('id', id).eq('user_id', user.userId)`.
- **Public data exposure**: both `/u/:handle` and the crawler render independently filter `.eq('provenance', 'purchased').eq('hidden', false)` server-side; neither response includes email/user_id.
- **AES-256-GCM**: IV is `randomBytes(12)` per call (test confirms non-deterministic ciphertext for identical plaintext); auth tag is verified on decrypt (tampered ciphertext/wrong key throws, test-covered).
- **OQ6 guardrails** (tombstone exclusion, preserved `supported_at`, ambiguous-match skip) — all three enforced in the actual upsert/update SQL, not just described, and unit-tested.

## Security — finding

**1. [confirmed, pattern deviation — not currently exploitable] Subsonic fetch bypasses the mandated `isUrlHostnameAllowed()` allowlist.** `api/functions/bandcamp-subsonic.ts` uses a raw `fetch()` against a hardcoded host rather than routing through the allowlist CLAUDE.md requires for "any code that fetches an external URL." Safe today since the host is a constant and username never reaches it — but the file's own comment flags a future "generic Subsonic servers" extension, and that's exactly the change that would turn this into a real SSRF surface with no allowlist in the way. Worth an allowlist entry now for defense-in-depth/consistency rather than after the extension lands.

## Regression

**2. [plausible] No staleness recovery if the background sync dies without hitting its `catch`.** `sync_status='error'` is only ever written from the explicit `try/catch`. A Netlify background function killed by its own execution/memory ceiling never reaches it, leaving the row at `sync_status='syncing'` forever — and the resync endpoint hard-refuses (`409`) whenever status is `'syncing'`, with no time-based override. Only recovery is Disconnect + Reconnect. The catalog-job equivalent of this problem already has a fix pattern in this codebase (`release_catalog_state.last_attempted_at`/`consecutive_failures`); this table has no analogous staleness signal.

**3. [plausible] Race between Disconnect and an in-flight background sync.** The sync handler reads the `bandcamp_connections` row once at the top and never re-checks it still exists. Disconnecting (optionally with delete-items) while a sync for that user is mid-flight lets the sync finish anyway — re-inserting `collection_items` and re-marking artists supported after the user asked to remove them — and the final status update on the now-gone row no-ops silently, so this is invisible in logs.

**4. [plausible] TOCTOU on the resync guard.** The `sync_status` check and the `sync_status='syncing'` write in `me-bandcamp.ts` are separate round trips with no lock between them. Two near-simultaneous resync requests can both pass the check before either commits, dispatching two concurrent background syncs for the same user. Low severity given idempotent upsert keys, but a real gap.

**5. [confirmed, process note not a code defect] Migration applied to production ahead of merge.** `20260809120000_bandcamp-collection.sql`'s own header states it was applied via `workflow_dispatch` on 2026-08-09 ahead of review, so the deploy preview (which shares the production DB) could exercise the connect flow. Low-risk given the new tables are additive/RLS-locked/inert until app code writes to them, but if this PR changes shape substantially or doesn't land, the schema is already live and needs explicit cleanup rather than just not merging.

---

brandonlucasgreen (commented):
@brandonlucasgreen Security/regression review of this PR (full diff read, all hunks — credential crypto, Subsonic client, sync background function, edge/API surfaces, migration).

## Security

**Auth/authorization** — no findings. Ownership checks are correctly scoped everywhere a client can act on `collection_items` (`me-collection.ts`, `me-bandcamp.ts` both filter on `user_id`), and `bandcamp-sync-background.ts` is gated by the existing `isInternalRequest()` (fails closed if the secret is unset).

**`plausible`** — `api/functions/me-bandcamp.ts`, POST "Connect" handler. This authenticated endpoint accepts an arbitrary `{username, password}` and calls `subsonicPing()` against Bandcamp, returning a distinct 400 (bad credentials) vs 502 (unreachable). Any logged-in Unstream user can use it as a Bandcamp-login oracle for credentials that aren't their own — e.g. testing a breached-credential list — laundered through Unstream's IP, with no protection beyond the generic 30/min/IP rate limit (no per-target-username limiting). Real-world exploitability against Bandcamp's own defenses isn't verifiable from the diff (`needs-manual-verification`), but the code-level gap — nothing beyond the standard IP rate limit — is `confirmed`. Worth at least rate-limiting by submitted username, not just caller IP.

**Input validation** — no findings. `username`/`password` are length-capped and type-checked; all Supabase access goes through the parameterized query builder, no raw SQL interpolation anywhere in the diff.

**Secret handling** — no findings. Verified against the PR's own claims:
- `confirmed` — AES-256-GCM in `api/functions/credential-crypto.ts`, fresh `randomBytes(12)` IV per call, auth tag checked via `setAuthTag()`/`decipher.final()` (throws on tamper, test-covered). `getKey()` throws on a missing or wrong-length `BANDCAMP_CREDENTIAL_KEY` — no default/fallback key anywhere.
- `confirmed` — `bandcamp-subsonic.ts` never lets the Subsonic URL (which embeds the token) reach an error message, a thrown exception, or Sentry; `me-bandcamp.ts`'s connect-failure path skips Sentry entirely so form contents can't ride along in a captured exception.

**Injection risks (SSRF)** — `confirmed` (low severity, mitigated in practice): `bandcamp-subsonic.ts`'s `SUBSONIC_SERVER` fetch does not go through `isUrlHostnameAllowed()`/`ALLOWED_OUTBOUND_HOSTNAMES`, which CLAUDE.md states is mandatory for any outbound fetch. Not exploitable today — the host is a hardcoded constant, not derived from request input, and `*.bandcamp.com` is already allowlisted anyway — but it's a real deviation from the codified pattern, and the file's own comment says the host is planned to become a parameter later ("deferred, not discarded"). Worth an explicit `isUrlHostnameAllowed(SUBSONIC_SERVER)` assertion now so that future change can't silently reintroduce real SSRF. No XSS found — `u-handle.ts`'s new collection tiles escape every interpolated value via `escapeHtml()`.

**Unsafe defaults** — no findings. Encryption and internal-request auth both fail closed.

**Data exposure** — `confirmed`: `api/edge/u-handle.ts`, new collection-read hunk (`@@ -107,6 +118,21 @@`):
```
const { data: collectionData } = await supabase.from('collection_items').select(...)...
collectionItems = collectionData || [];
```
This discards `error`. Supabase-js resolves `{ data: null, error }` rather than throwing on a query-level failure (RLS denial, transient Postgres error), so a failed read here silently renders as "no collection" on the public/crawler-facing page instead of surfacing a failure — exactly the "never cache uncertainty" anti-pattern the codebase's own conventions call out repeatedly, and one this PR gets right in the two JSON API surfaces it touches (`me-collection.ts` and `public-saved-artists.ts` both correctly 500 on a read error, both test-covered) but not here. Note the adjacent pre-existing `savedArtists = savedData || []` line has the same latent pattern already — this isn't new to the file, but the new code reproduces it in the one place the PR explicitly claims to have handled correctly. No PII/account-metadata leak found in the new `collection` field on `public-saved-artists.ts` (title/artist_name/art_url/acquired_at/url only, explicitly test-covered).

## Regression

**`plausible`** — `api/functions/bandcamp-sync-background.ts`, `matchLibrary()` releases lookup (`.from('releases').select(...).in('artist_id', chunk).eq('is_hidden', false)`) is a single unpaged `.select()`, not `readAllPages`/`.range()`. Per CLAUDE.md, PostgREST silently caps any response at 1,000 rows. If a 50-artist chunk collectively has >1,000 non-hidden releases, the tail is silently dropped and those albums import unmatched (graceful degradation, not data loss, but silent) — inconsistent with this same PR's correct paging elsewhere (`me-collection.ts`, `public-saved-artists.ts`, the artist lookup in the same function).

**`plausible`, minor** — same file, the chunked `collection_items` upsert followed by `markArtistsSupported()` and the final `sync_status` update: if a later upsert chunk throws, earlier chunks are already durably committed, so a user can see a partial collection while the connection simultaneously reports `sync_status='error'`. The core claim — a partial sync is never recorded as `success` — holds (`confirmed`), but the visible data and reported status can disagree, and nothing surfaces that "Re-sync" is safe to re-run on top of a partial import (it is, per the upsert dedup key).

No off-by-one issues found in Subsonic pagination or chunk sizing (all test-covered), and no API contract drift — response shapes for the three new/changed endpoints match what their consuming components expect.

## Explicit claims, verified against the code

| Claim | Result |
|---|---|
| `collection_items` INSERT is service-role only | **PASS** — migration defines SELECT/UPDATE/DELETE policies scoped to `auth.uid() = user_id` but no INSERT policy at all; with RLS on, that locks INSERT to the service-role client. Comment confirms this is deliberate. |
| `bandcamp_connections` RLS enabled, no policies | **PASS** — confirmed, matches the `bandcamp_slug_probes` precedent with an explanatory comment. |
| Real authenticated encryption, no default-key fallback | **PASS** — see Secret handling above. |
| Outbound Bandcamp fetch through the SSRF allowlist | **FAIL on process / PASS on outcome** — see Injection risks above. |
| Re-sync never resurrects a tombstoned row | **PASS** — `markArtistsSupported()` excludes any slug already present (including tombstoned/`deleted:true`) from `toInsert`, and filters `!row.deleted` for `toSupport`; directly test-covered. |
| Partial sync never recorded as success | **PASS** — pagination throws mid-read rather than returning a partial list; any upsert failure throws before reaching the `sync_status: 'idle'` update. |
| Failed collection read never silently renders as empty | **PASS for `me-collection.ts`/`public-saved-artists.ts`** (both 500 on read error, test-covered) / **FAIL for `api/edge/u-handle.ts`** — see Data exposure above. |

Nothing here blocks the merge on its own — the confirmed items are a hardcoded-host SSRF process gap and one edge-function error path that should 500 instead of rendering empty — but I'd fix the `u-handle.ts` error handling and the `matchLibrary()` paging before shipping, and think about the credential-oracle rate limiting before this is used at scale.

---

brandonlucasgreen (commented):
@brandonlucasgreen — senior-staff security + regression pass over the whole branch (`origin/main...support-loop-implementation-39d9e6`), with depth on credential handling, auth, SSRF, and the public-data path rather than the UI files.

**The credential path is genuinely well built.** Worth saying before the findings: the key comes from `BANDCAMP_CREDENTIAL_KEY` and is never hardcoded (`credential-crypto.ts`); the plaintext password is converted to the Subsonic `(t, s)` pair and never persisted, logged, or echoed; Subsonic request URLs (which carry auth in the query string) never reach an error message or a Sentry event; every one of the three new endpoints derives `user_id` from the verified bearer token and never from the request body, and every DB read and write is scoped with `.eq('user_id', ...)` against the service-role client. **I found no IDOR, no SQL string concatenation, and no unescaped interpolation on the edge render.** Five inline comments carry the detail for the findings below.

## Security

- **[confirmed]** Consent scope silently expanded — `api/functions/public-saved-artists.ts:99` and `api/edge/u-handle.ts:122`. The collection is gated on the existing `usernames.saved_artists_public` flag, whose toggle copy (`SharingControls.tsx:123`, unchanged in this PR) promises only "your username, the artists you've saved, and your location". Anyone who enabled sharing months ago has their purchase history published the moment they connect Bandcamp, with no re-consent and no warning in `BandcampConnect.tsx`.
- **[confirmed]** `collection_items` UPDATE RLS policy defeats the server-asserted-provenance invariant — `supabase/migrations/20260809120000_bandcamp-collection.sql:143`. `FOR UPDATE USING (auth.uid() = user_id)` with no column restriction lets a browser holding the anon key PATCH the table directly and set `provenance`, `release_id`, `art_url`, `title` — bypassing `me-collection.ts`, which deliberately restricts itself to `hidden`. No IDOR (omitted `WITH CHECK` defaults to `USING`) and no live exploit while every row is already `'purchased'`, but it becomes one the moment `'owned'`/`'listened'` rows exist in Step 2/5. (Flagging here rather than on #444 because the gap is only visible against this PR's application code.)
- **[plausible]** `/api/me/bandcamp` POST is a Bandcamp credential-verification oracle — `api/functions/me-bandcamp.ts:202-232`. It returns a distinguishable 400 ("Bandcamp rejected that username or credential") vs 502, rate-limited only per-IP at the `standard` tier (30/min, 1000/day), so an authenticated attacker can check Bandcamp credentials through Unstream from Netlify's IPs. `strict` tier plus a per-user rather than per-IP bucket would close it.
- **[plausible, low]** Possible credential fragment to Sentry — `api/functions/bandcamp-sync-background.ts:283`. The decrypt catch captures the raw exception, which may be a `JSON.parse` error whose message embeds a snippet of the *decrypted* `{"t":…,"s":…}` plaintext. Only reachable if GCM authenticates but the plaintext is malformed, so near-impossible in practice — but `captureMessage` with a fixed string costs nothing.
- **[not a finding, noted deliberately]** SSRF: `subsonicRequest` fetches a hardcoded `SUBSONIC_SERVER` constant with all user input confined to `URLSearchParams`, so it doesn't need `isUrlHostnameAllowed()`, and `bandcamp.com` is on the allowlist regardless. The "keep the server a constant, not a parameter" comment at the top of `bandcamp-subsonic.ts` is the right call and should survive any future Funkwhale/Navidrome work.

## Regression

- **[confirmed]** A degraded match writes over good data — `api/functions/bandcamp-sync-background.ts:321,325`. `matchLibrary` degrades to empty matches on a failed `artists` read (line 76) or a `releases` chunk error (line 114), and the upsert then persists `release_id: null` / `art_url: null` over rows a previous sync matched correctly. One transient Supabase hiccup during a re-sync strips artwork and release links from the whole collection and kills the public page's "buy this record" links, while reporting success. Textbook never-cache-uncertainty; both degrade paths are also `console.warn`-only, so a systematic break is invisible.
- **[confirmed]** Re-sync reverts a deliberate un-support — `api/functions/bandcamp-sync-background.ts:198-201`. The filter re-supports any non-deleted row with `supported === false`, which is exactly what `saved-artists.ts`'s `unsupport` action produces. The tombstone case is handled and tested; this near-neighbour isn't.
- **[needs-manual-verification]** Short-page termination can record a truncated import as complete — `api/functions/bandcamp-subsonic.ts:200`. If Bandcamp clamps `size` below the requested 500, the first page comes back short and pagination stops, writing a partial collection as a finished sync — the failure the `MAX_PAGES` backstop guards in the other direction.
- **[plausible]** Unbounded public collection read — `api/functions/public-saved-artists.ts:99`. `readAllPages` with no cap (up to 50,000 rows) on an anonymous 30/min/IP endpoint; the edge render caps at 60, this path has no equivalent.
- **[confirmed, minor]** Capped count reported as the real one — `api/edge/u-handle.ts:167`. `${collectionItems.length} releases collected` reads from the 60-item crawler query, so a 500-release collection's OG/social card says "60 releases collected". Since these tags exist to be shared, that number should come from a count query.
- **[confirmed, minor]** Release match key drops `release_type` — `api/functions/bandcamp-sync-background.ts:118,127`. Keys on `${artist_id}:${match_key}` while `releases` identity is `(artist_id, release_type, match_key)` (unique index, `20260731120000_releases.sql:108`), so an album and a single sharing a title collapse to whichever row the paged read returned last and an item can link to the wrong release.
- **[confirmed, minor]** Unvalidated UUID → 500 instead of 400 — `api/functions/me-collection.ts:108`. A non-UUID `id` makes Postgres raise 22P02 and the handler answers 500 "Failed to update item". Add a shape check next to the existing `hidden` boolean check.
- **[confirmed, minor]** Missing concurrency guard on connect — `api/functions/me-bandcamp.ts:239`. The resync branch 409s when `sync_status === 'syncing'`, but the connect branch doesn't, so re-submitting the form mid-sync dispatches a second concurrent background run. Upserts are idempotent so the blast radius is a raced `item_count`/`last_synced_at`, but the guard already exists a few lines up.

**Checked and correct, for the record:** re-sync idempotency (`onConflict: 'user_id,source,external_id'` matches the migration's UNIQUE constraint); `hidden` preservation across re-syncs (omitting the column from the payload is the right mechanism, since PostgREST's merge-duplicates only sets columns present in the body — worth a test, as it's currently the one load-bearing claim in that file with no coverage); in-batch dedupe by Subsonic id, which correctly avoids "cannot affect row a second time"; unmatched items (`release_id` NULL) rendering without a link on both the SPA and edge paths; timing-safe `isInternalRequest` on the background function; a mid-pagination failure landing in the catch and recording `sync_status='error'` rather than a zero-item success. No breaking API contract changes — `public-saved-artists` gains an optional `collection` field and the client type matches.

Nothing here blocks the concept; the two I'd want resolved before merge are the sharing-consent scope and the `collection_items` UPDATE policy.

---

PR #439, opened 2026-08-10
Add Apple Music library import to the Mac app (Support Loop Step 2)

brandonlucasgreen:
## What this is

Step 2 of the Support Loop (`support-loop-spec.md`): the Mac app can now read the whole Music.app library — play counts plus each track's `cloud status` — and roll it up into a local per-artist snapshot. This is the data the Support List (Step 4) renders: *artists you actually listen to, with owned-vs-streamed provenance attached*.

**Settings → Integrations → Music Library**: Import button, progress, a summary line ("1,240 tracks · 340 you own · 900 subscription-only · 62 artists"), and a Forget Imported Data button. First import triggers the macOS Automation permission prompt.

## The part that must not drift

The provenance mapping is spec §4's table, implemented conservatively:

| Music.app `cloud status` | Unstream provenance |
|---|---|
| `matched`, `uploaded`, `purchased` | `owned` — an iTunes purchase is a copy you own, **not** direct support |
| `ineligible`, `not uploaded`, `duplicate`, `error` | `owned` — all describe a local file |
| `subscription` and anything uncertain | `listened` — when unsure, claim less |

Nothing from Apple Music ever counts as support. Tests pin every row of this table.

## Implementation notes

- Four batched `… of every track` reads — one Apple event per property, so large libraries take seconds, not thousands of round trips.
- Cloud-status enums are decoded by four-char code (`kPur`, `kSub`, …) in Swift rather than compared as keywords in AppleScript source — `is not uploaded` parses as a negation there.
- A misaligned read (property arrays of different lengths) is refused outright instead of zipped — wrong provenance attached to the right artist would be worse than no data.
- Local-first: the snapshot lives in UserDefaults JSON on this Mac only (deliberately not iCloud KVS — 1 MB quota). "The data never leaves this Mac" appears in the UI copy and is true.
- No entitlement changes needed — `com.apple.Music.library` was already in the sandbox's scripting-targets. `NSAppleEventsUsageDescription` broadened.

## Verification

- 18 new XCTests (provenance table, zip/refuse behaviour, rollup aggregation, diacritic-insensitive artist merging, gap derivation, Codable round-trip, OSType decoding); full Mac suite passes on-device.
- App builds clean (one pre-existing provisioning-profile hiccup on this machine worked around with `CODE_SIGNING_ALLOWED=NO`; no signing-relevant changes in this PR).
- Not yet exercised against the real library end-to-end — the first live import needs the Automation permission prompt accepted interactively.

Pairs with #438 (Bandcamp collection import, Step 1). Independent code paths; either can merge first.


brandonlucasgreen (commented):
@brandonlucasgreen Security + regression pass on this one.

**Security**

No findings. This PR is local-only (Music.app → UserDefaults on-device, nothing sent to a server), which rules out most of the usual categories:

- **Injection**: `musicLibrarySettingsSource` (the AppleScript) is a static string constant with no interpolated data — nothing here builds script text from a variable, so there's no AppleScript-injection surface, unlike a naive "compare enum keyword in script source" approach the comments explicitly say was avoided.
- **Auth/authorization, secret handling, data exposure**: N/A — no network calls, no credentials, and the snapshot deliberately stays in local UserDefaults (not iCloud KVS), matching the existing `ReleaseAlertManager` persistence pattern.
- **Unsafe defaults**: the provenance mapping defaults *uncertain* statuses to `.listened` rather than `.owned` — "claim less" is the safe-by-default direction here, and `testUncertainStatesClaimLessNotMore` pins it.

**Regression**

- The length-mismatch guard in `MusicLibrary.makeRecords` (refuse rather than zip misaligned arrays) and the `-1743` permission-denied vs. generic-failure error split in `AppleMusicLibraryService.readLibrary` both look correct and are covered by tests/tested behavior.
- Diacritic/case-insensitive artist merging, empty-artist-name handling (counted in totals, no row), and the enum-code decode fallback to `"unknown"` on an unrecognized four-char code are all exercised by `MusicLibraryTests`.
- Nothing here changes shared/server-side assumptions, so no API contract drift to check.

One non-blocking note: `AppleMusicLibraryService.fourCharCode` decodes via `.macOSRoman` and falls back to `"????"` only when `code == 0`; a non-ASCII/garbage OSType from a future Music.app version would still attempt the decode rather than falling back, though Apple's cloud-status codes are documented as ASCII four-char codes today so this is theoretical, not a real gap.

---

brandonlucasgreen (commented):
@brandonlucasgreen Automated senior-staff-engineer review (scheduled security/regression pass). Read the full diff for this PR (MusicLibrary.swift, AppleMusicLibraryService.swift, MusicLibrarySettingsSection.swift, the Info.plist usage-description change, and the test file).

## Security

- **No issues found.** This feature is entirely local and offline: `AppleMusicLibraryService` executes a **fixed, non-interpolated** AppleScript string against Music.app — no user input is ever concatenated into script source, so there's no AppleScript-injection surface. The resulting snapshot is persisted only to `UserDefaults` and, per the code's own comment, never leaves the device (no network call anywhere in this diff).
- `NSAppleEventsUsageDescription` was updated to accurately disclose the new library read ("Library data never leaves this Mac") rather than silently piggybacking on the existing now-playing permission — correct handling of an expanded permission's usage string.
- **Input validation — no issues found.** `MusicLibrary.makeRecords` explicitly refuses to zip Music.app's parallel result arrays when their lengths don't match (which would otherwise misattribute cloud-status/play-count data to the wrong track) and clamps negative play counts to zero rather than propagating them.

## Regression

- **No issues found** in edge-case, off-by-one, state-transition, or API-contract terms. `provenance(forCloudStatus:)` defaults any unrecognized/future Music.app status to `.listened` (the "claim less" side of the mapping), so an unexpected status degrades safely instead of overstating support — directly covered by `testUncertainStatesClaimLessNotMore`.
- **Minor, non-functional:** in `rollup`, the display name stored per normalized artist key is whichever track's artist string is encountered first (`if acc.name.isEmpty { acc.name = track.artist }`). If the underlying track order isn't stable across imports, casing/whitespace variants of the same artist name could display inconsistently between runs. Cosmetic only — aggregation correctness (counts, play totals) is unaffected.

Clean PR — no security or regression concerns block this.

---

brandonlucasgreen (commented):

## Security
No confirmed vulnerabilities. Specifically checked and clean:
- **AppleScript injection**: `musicLibraryScriptSource` is a static literal string, zero interpolation of dynamic/user input.
- **Entitlements**: `com.apple.Music.library` was already present pre-PR; entitlements file untouched in this diff.
- **"Uncertain → listened" default**: implemented in exactly one place (`MusicLibrary.provenance(forCloudStatus:)`), unmapped/unknown codes fall through to `.listened` via the `default:` branch. Single source of truth, test-pinned.
- **Local-only storage**: persistence uses `UserDefaults.standard` only, never the existing `NSUbiquitousKeyValueStore` iCloud path already used elsewhere in the app (`SupportListManager.swift`). No network/analytics/Sentry call touches the snapshot anywhere in this diff — "the data never leaves this Mac" holds for the code as shown.

**[plausible, security-adjacent] Misalignment guard checks length, not order.** The "refuse outright on misalignment" guard in `MusicLibrary.makeRecords` only verifies the four separately-fetched Apple Event lists (artist/album/play-count/cloud-status — four separate statements in one `tell` block) are the same *length*, not that they're still in the same *order*. If track ordering shifts between these four sequential round trips, same-length arrays could be silently misaligned — attaching the wrong cloud status to the wrong artist, which is the exact failure mode this guard is meant to prevent. Can't confirm Apple's actual ordering guarantees across sequential `of every track` fetches without an on-device test against a live, syncing library. Consider fetching all four properties off a single `every track` record instead, which would make misalignment structurally impossible rather than length-checked.

## Regression
Provenance table, empty-library/no-artist edge cases, Forget-Imported-Data state clearing, re-import replace-not-merge semantics, and macOS/iOS target wiring all traced correctly and match their test coverage — no findings.

**[needs-manual-verification] Latent normalization mismatch, not yet a bug.** `MusicLibrary.normalizeArtistName` folds diacritics/case/width; the existing `SupportListManager.isArtistSaved` only lowercases. `unsupportedArtists(...)` (this PR, for the not-yet-built Support List step) has no production call site yet, so nothing is broken today — but when Step 4 wires the two together, "Sigur Rós" vs "Sigur Ros" would be treated as different artists by `SupportListManager` even though `MusicLibrary`'s own normalization would unify them. Worth aligning the two normalization schemes before that integration lands rather than after.

As the PR itself notes, real-device behavior (Automation permission flow, actual timing/ordering of the four Apple Events on a large library) hasn't been exercised end-to-end yet.

---

brandonlucasgreen (commented):
@brandonlucasgreen Security/regression review of this PR (full diff read — `MusicLibrary.swift`, `AppleMusicLibraryService.swift`, the settings UI, and the test suite).

## Security

No findings in any category:
- **Auth/authorization** — N/A, local-only feature, no network or server calls in this diff.
- **Input validation** — the only "external input" is Music.app's own AppleScript response; it's decoded defensively (`atIndex(...) ?? 0` / `?? ""` throughout, `max(0, playCounts[i])` clamps negative play counts) rather than force-unwrapped.
- **Secret handling** — N/A, no secrets.
- **Injection risks** — the AppleScript source (`musicLibraryScriptSource`) is a static string constant with no user-input interpolation, so no script-injection surface.
- **Unsafe defaults** — a length mismatch between the four parallel arrays AppleScript returns is refused outright (`makeRecords` returns `nil`) rather than zipped and guessed at, avoiding silently misattributed provenance.
- **Data exposure** — the imported snapshot (artist names, play counts, album names) is stored unencrypted in local `UserDefaults`, consistent with the existing `ReleaseAlertManager` pattern in this codebase and explicitly never transmitted off-device; not flagging this as a finding since it's local, sandboxed, user-owned listening data, not credentials or PII shared with a server.

## Regression

No findings:
- The provenance mapping (spec §4's table: `matched`/`uploaded`/`purchased`/`ineligible`/`not uploaded`/`duplicate`/`error` → `owned`, everything else → `listened`) is pinned by tests for every row, including the "claim less when uncertain" default.
- Edge cases are explicitly handled and tested: empty library, tracks with no artist name (counted in totals, no artist row), diacritic/case-insensitive artist merging, unknown status codes falling back to `"unknown"` → `.listened`, negative play counts clamped to zero, and a Codable round-trip for the persisted snapshot.
- No off-by-one issues in the descriptor decoding (`(1...n).map` matches AppleScript's 1-indexed lists correctly) or in the malformed-result guard (`result.numberOfItems == 4` plus all four `atIndex` calls checked before use).
- No API contract drift — this is new functionality wired into `UnstreamApp.swift`/`SettingsView.swift` with no changes to existing call sites.

Clean PR from a security/regression standpoint.

---

brandonlucasgreen (commented):
@brandonlucasgreen — reviewed `git diff main...apple-music-library-import` in full (8 files), plus the adjacent files it depends on (`MediaObserver.swift`, `Unstream-macOS.entitlements`, `ReleaseAlertManager.swift`, `SupportListManager.swift`, `project.yml`, `PrivacyInfo.xcprivacy`).

The privacy story holds up under tracing, and the pure half is well-tested. The substantive concerns are all in `readLibrary()`'s failure modes — several ways for a broken read to look like a successful import.

## Security

- **[confirmed — no issue]** No AppleScript injection. `musicLibraryScriptSource` (`AppleMusicLibraryService.swift:39`) is a static constant with zero interpolation, and it is the only new `NSAppleScript(source:)` in the diff (line 89). The other two in the app (`MediaObserver.swift:65,347`) are also static/caller-supplied-from-constants. Track metadata flows *out* of AppleScript into Swift structs, never back into script source.
- **[confirmed — no issue]** "The data never leaves this Mac" is true as written. I traced every reference to `LibrarySnapshot`, `LibraryArtistStats`, and the `appleMusicLibrarySnapshot` key: the only sinks are `UserDefaults.standard` (`AppleMusicLibraryService.swift:159-171`) and SwiftUI `Text`. `NSUbiquitousKeyValueStore` is used in `ReleaseAlertManager.swift:20` and `SupportListManager.swift:23` with explicit per-key writes — there is no blanket defaults→iCloud mirror that could sweep the new key up. No network path touches the snapshot.
- **[confirmed — no issue]** No secrets, and no PII in logs: the two `print` calls (lines 77, 80) emit counts and an error description, not artist names.
- **[confirmed — no issue]** No privilege escalation. `com.apple.Music.library` is already in the `com.apple.security.scripting-targets` group in `Unstream-macOS.entitlements`, so the comment's claim is accurate and no entitlement was widened. `init()` only reads UserDefaults, so no Apple event (and no permission prompt) fires at launch — the import is strictly user-initiated. `-1743` is mapped to a distinct `.permissionDenied` with actionable copy (`AppleMusicLibraryService.swift:99`). Whether macOS actually returns `-1743` versus `-600`/`-10004` for a *sandbox scripting-target* denial as opposed to a TCC denial is **needs-manual-verification** — worth a real denial test, since the fallback message ("Make sure Music is installed") would be wrong.
- **[confirmed — acceptable]** UserDefaults rather than Keychain is the right call and is consistent with the codebase: `ReleaseAlertManager.swift:441` persists `releaseCheckState` the same way (JSON → `UserDefaults.standard`). The data is aggregate listening/ownership metadata inside the app-sandbox container, so other sandboxed apps can't read it. Two things worth being deliberate about rather than fixing: it does encode a coarse purchase signal (`ownedTrackCount` per artist) and it will be captured by Time Machine / device backups; and `PrivacyInfo.xcprivacy` keeps `NSPrivacyCollectedDataTypes` empty, which stays accurate only as long as nothing transmits it — Step 3 should revisit that file if the gap list ever syncs.
- **[confirmed — no issue]** Track metadata handling is safe. Strings go into Swift `String`, `JSONEncoder`, and SwiftUI `Text`; nothing is rendered as HTML, passed to a shell, or fed back into a script. A crafted artist name is at worst a weird-looking row.

## Regression

- **[confirmed]** A total decode failure is indistinguishable from a legitimate import. If `enumCodeValue` yields `0` for every track — which happens when `cloud status` is `missing value`, plausible for a user with iCloud Music Library off — `fourCharCode` returns `"????"`, the map lookup misses, `makeRecords` substitutes `"unknown"` (`MusicLibrary.swift:109`), and `provenance` defaults to `.listened`. The import then *succeeds* with `ownedTotal == 0` and `subscriptionTotal == 0`, and `summaryLine` hides both because it gates each on `> 0` (`MusicLibrarySettingsSection.swift:74-81`). The user sees "12,431 tracks · 604 artists", the snapshot is persisted, and the single field the feature exists to capture is silently absent. This is the "silent failure that looks like success" pattern CLAUDE.md calls out. Concrete fix: count unresolved status codes in `makeRecords`/`rollup` and treat "no track resolved to a known status" (or >90% unknown) as `.malformedResult` rather than a successful import.
- **[needs-manual-verification]** The twelve four-char codes at `MusicLibrary.swift:58-71` can't be verified from this environment; they need checking against `sdef /System/Applications/Music.app`. Note the failure mode is invisible: one wrong code degrades that status to `"unknown"` → `.listened`, and no test catches it because the tests assert the *name*→provenance mapping, not the *code*→name mapping. A test that asserts every value in `cloudStatusNamesByCode` is a key the mapping handles wouldn't help either — only the sdef comparison will.
- **[confirmed]** An unexpected reply shape also decodes to a successful empty import, and overwrites good data. `stringList`/`intList`/`enumCodeList` all `guard n > 0 else { return [] }` (`AppleMusicLibraryService.swift:129,135,141`), and `NSAppleEventDescriptor.numberOfItems` returns 0 for a *non-list* descriptor too. So a shape change in Music's reply produces four empty arrays → `makeRecords` succeeds → `trackTotal: 0` → `saveSnapshot` replaces a previously good snapshot with an empty one. Suggest refusing to overwrite a non-empty stored snapshot with a zero-track result, or checking the sub-descriptors are `typeAEList` before decoding.
- **[plausible]** The length guard protects less than its comment claims. The script issues four separate `… of every track of lib` Apple events (the file comment at lines 15-16 says as much), so the library can change between them. Equal-length-but-shifted — one track deleted and another added mid-read, or an iCloud reorder — passes `makeRecords`'s guard (`MusicLibrary.swift:101-103`) and attaches album/plays/status to the wrong artist, which is precisely the misattribution the guard was written to prevent. Low probability and arguably inherent to the batched approach, but the comment ("Refuse rather than zip a misaligned library") reads as a stronger guarantee than the check provides; worth softening or adding a `count of tracks` sanity re-read.
- **[confirmed]** `readLibrary()` will launch Music.app, breaking an established invariant. Every other Apple event in the app is gated on `isAppRunning("com.apple.Music")` (`MediaObserver.swift:196-201`) specifically so the app never launches a player; the new `tell application "Music"` has no such guard. Compounding it, there's no `with timeout of …` wrapper, so a cold Music launch plus library load on a large library can exceed AppleScript's 60-second default event timeout and surface as `.scriptFailed(-1712)` with the misleading "Make sure Music is installed" copy. Suggest wrapping the `tell` in `with timeout of 300 seconds` and either gating on `isAppRunning` or giving `-1712` its own message.
- **[plausible]** `NSAppleScript` is documented thread-unsafe, and this widens an existing exposure. `readLibrary()` runs on a detached task while `MediaObserver.fetchNowPlayingAsync` runs `NSAppleScript` on `DispatchQueue.global` every poll (`MediaObserver.swift:117-121`). The pre-existing overlap window was milliseconds; a full-library read holds the component for seconds, so concurrent execution is now likely rather than theoretical. Serializing both through one dedicated queue would remove the question. **needs-manual-verification** for whether it manifests as a crash in practice.
- **[plausible — probably acceptable]** `normalizeArtistName` (`MusicLibrary.swift:171`) folds diacritics, case, and width, so two genuinely distinct artists differing only by a diacritic merge into one row with summed stats. Real but rare, and the consequence is a merged local count, never an overstated support claim — I'd accept it. Two related nits worth a line: the surviving display name is whichever variant appears first in library order (`MusicLibrary.swift:139`), so a merged artist can render as "sigur ros"; and `albumCount` de-dupes with `.lowercased()` (line 145) instead of the same folding, so "Ágætis Byrjun" and "Agaetis Byrjun" count as two albums — inconsistent with how the artist key is built.
- **[confirmed — no issue]** Provenance mapping is complete. All twelve names in `cloudStatusNamesByCode` reach a branch in `provenance(forCloudStatus:)`; the seven owned-file states match exactly, and `unknown`/`removed`/`subscription`/`prerelease`/`no longer available` plus any unrecognized code fall through to `.listened`, the lesser claim, per spec §4. I found no path where `owned` is claimed on uncertain input.
- **[confirmed — no issue]** `fourCharCode` (`AppleMusicLibraryService.swift:145`) extracts bytes big-endian in the same order the test constructs them; no off-by-one. `String(bytes:encoding: .macOSRoman)` cannot fail for four arbitrary bytes, so the `?? "????"` fallback is unreachable — harmless, but it means an unknown code becomes a garbage 4-char string rather than `"????"`, which is fine since both miss the map.
- **[confirmed — no issue]** Cross-platform wiring is consistent. Both `SettingsView` stored properties are optionals, which get implicit `nil` defaults in the memberwise init, so the `SettingsView()` preview at `SettingsView.swift:622` still compiles; `UnstreamApp.swift:133` is the only other call site and passes both. `SettingsView.swift`, `AppleMusicLibraryService.swift` and `MusicLibrarySettingsSection.swift` all carry `platformFilters = (macos, )` in the pbxproj, matching what `project.yml`'s `destinationFilters: [macOS]` directory globs would regenerate — so editing the pbxproj directly here doesn't drift from XcodeGen. `MusicLibrary.swift` builds on iOS (Foundation only, no `#if`), and `UnstreamTests` is `platform: macOS`, so the test's reference to `AppleMusicLibraryService.fourCharCode` is fine.
- **[minor]** `unsupportedArtists` and `LibraryProvenance`'s `.listened` case are exercised only by tests — Step 3 groundwork, per the comment, but CLAUDE.md's "no speculative exports" applies if Step 3 slips.

---

brandonlucasgreen (commented):
Security + regression pass (auth/authz, input validation, secret handling, injection, unsafe defaults, data exposure; then changed assumptions, edge cases, state transitions, API contract drift).

**Security — no findings.** This diff is entirely local and offline:
- `AppleMusicLibraryService.swift`'s AppleScript source (`musicLibraryScriptSource`) is a static string with no interpolated user input, so there's no AppleScript-injection surface.
- Neither `MusicLibrary.swift` nor `AppleMusicLibraryService.swift` makes any network call — the "data never leaves this Mac" claim in the Settings UI copy checks out against the code.
- The snapshot is stored via `UserDefaults.standard` (app-sandboxed), not iCloud KVS, matching the stated reasoning (1 MB quota).
- The provenance mapping (`MusicLibrary.provenance(forCloudStatus:)`) defaults every uncertain/unrecognized cloud status to `.listened` rather than `.owned`, and nothing maps to "supported" — conservative in the correct direction, and pinned by tests.
- `makeRecords` refuses to zip mismatched-length property arrays rather than guessing at alignment — correct call given a silent misattribution would put the wrong provenance on the right artist.

**Regression — no findings.** `atIndex` access across the four parallel-array decodes (`stringList`/`intList`/`enumCodeList`) is consistently 1-based, matching `NSAppleEventDescriptor`'s convention — no off-by-one. Empty-library, no-artist-name, and mismatched-array edge cases are all exercised by `MusicLibraryTests.swift`. `readLibrary()`'s hardcoded `-1743` permission-denied code is a minor fragility (falls into the generic `.scriptFailed` message if Apple ever changes it) but not a functional regression.

@brandonlucasgreen

---

brandonlucasgreen (commented):

**Security**

- Auth/authorization — N/A / confirmed fine. This is local-only (Music.app via Apple events); the only gate is the existing macOS Automation permission prompt, and the required `com.apple.Music.library` scripting-access group is already present in `Unstream-macOS.entitlements` (not added by this PR, verified it predates it).
- Input validation — confirmed fine. `MusicLibrary.makeRecords` checks all four parallel arrays are equal length before zipping and returns `nil` on mismatch rather than misattributing a status to the wrong track; play counts are clamped to `max(0, ...)`.
- Secret handling — N/A, no secrets/credentials involved in this feature.
- Injection risks — confirmed none. `musicLibraryScriptSource` is a static string literal with no user/dynamic input interpolated into the AppleScript source.
- Unsafe defaults — confirmed fine. Unrecognized cloud-status codes fall back to `"unknown"` → `LibraryProvenance.listened`, matching the stated "when unsure, claim less" policy — an unmapped status can't accidentally overstate ownership or support.
- Data exposure — **plausible (minor)**: the imported snapshot (artist/album names, play counts, cloud status) is persisted unencrypted via `UserDefaults` rather than Keychain (`AppleMusicLibraryService.saveSnapshot`). Severity is low — no credentials, sandboxed app container, and the code comment already calls out avoiding iCloud KVS for size reasons — but flagging since it's a new on-disk store of listening-history-shaped data. Worth confirming this is an accepted risk vs. Keychain/other at-rest protection, especially once this data starts feeding the "unsupported artists" gap report.

**Regression**

- Checked whether the new `SettingsView(libraryService:)` parameter breaks existing call sites (e.g. the `#Preview { SettingsView() }` in `SettingsView.swift`): it doesn't — `libraryService` is `Optional`, so the synthesized initializer defaults it to `nil`.
- Checked whether `MusicLibraryTests.swift` referencing the macOS-only `AppleMusicLibraryService` (guarded by `#if os(macOS)` in its own file) could break a cross-platform test build: `UnstreamTests` is declared `platform: macOS` in `project.yml`, so this is fine.
- No off-by-one, state-transition, or API-contract issues found in the rollup/aggregation logic (`MusicLibrary.rollup`, `unsupportedArtists`) — covered reasonably well by the new XCTest cases (diacritic/case merging, empty-artist handling, Codable round-trip).

No confirmed vulnerabilities found.

---

brandonlucasgreen (commented):

This is a local-only macOS app feature (Apple Music library read via AppleScript, rolled up into a local snapshot) — it adds no new server endpoint and no network call to Unstream's backend, so most of the checklist doesn't apply to it.

**Security**
- **Auth/authorization** — n/a, no server-side surface touched.
- **Input validation** — `MusicLibrary.makeRecords` refuses (`nil`) on any length mismatch between the four parallel arrays returned by AppleScript rather than zipping a misaligned read, which correctly prevents attaching the wrong provenance/play-count to the wrong track. Negative play counts are clamped to 0. No finding.
- **Secret handling** — n/a, nothing new is stored beyond the library snapshot itself.
- **Injection risks** — cloud-status values are decoded by exact 4-char OSType code lookup (`cloudStatusNamesByCode`) rather than compared as free-text keywords, avoiding the "`is not uploaded` parses as a negation" class of bug the PR description itself calls out. The AppleScript source is a static literal, not built from any external input. No finding.
- **Unsafe defaults** — an unrecognized status code falls back to `"unknown"` → `.listened`, the least-claims outcome, matching the "when unsure, claim less" rule this feature is built around — correct default direction. No finding.
- **Data exposure** — `needs-manual-verification`: the snapshot (aggregated per-artist play counts and owned/streamed provenance) is stored unencrypted in `UserDefaults` rather than Keychain. The code comment notes this matches the existing `ReleaseAlertManager` pattern and the data never leaves the Mac, but flagging since it's still listening-history data sitting in a plist readable by any process in the sandbox container.

**Regression**
- **Changed assumptions** — none; this is new, additive functionality behind a new settings section, and doesn't modify any existing code path.
- **Edge cases / null/empty** — an empty library produces an empty snapshot (tested); tracks with no artist name are counted in the totals but produce no artist row (tested). Both handled explicitly rather than left implicit.
- **Off-by-one** — descriptor access loops correctly use AppleScript's 1-based indexing (`(1...n)`) consistently across all four decoders.
- **State transitions** — `importLibrary()` guards re-entrancy with `guard !isImporting else { return }` before setting `isImporting = true`, with `defer` resetting it on every exit path — no double-import race.
- **API contract drift** — n/a, no shared/server API is touched by this PR.

No confirmed or plausible vulnerabilities found in this PR.

---

brandonlucasgreen (commented):
Reviewed for security (auth/authz, input validation, secret handling, injection, unsafe defaults, data exposure) and regression risk (changed assumptions, edge cases, off-by-one, state transitions, API contract drift). This PR touches only the Mac app — no server, no network calls, no new attack surface reachable by a remote party.

**Security:** No issues found.
- **Injection:** `musicLibraryScriptSource` is a fixed literal with no user input interpolated into it, so there's no AppleScript-injection surface — the four batched `… of every track of lib` reads are the only script that ever runs.
- **Data exposure:** Snapshot is stored in `UserDefaults` only, deliberately not iCloud KVS (correctly reasoned in the comment: 1 MB quota). The `NSAppleEventsUsageDescription` string was broadened to accurately describe the new library read, matching what's actually being requested — good privacy-string hygiene.
- **Secret handling / auth:** N/A, nothing here touches credentials or backend auth.
- **Unsafe defaults:** `makeRecords` refuses (returns nil) on any length mismatch between the four parallel arrays rather than zipping a misaligned result — the right call, since a partial zip would silently misattribute cloud status to the wrong track.

**Regression:** No issues found. Provenance mapping (the part flagged in the PR description as "must not drift") is exhaustively covered by `MusicLibraryTests`, including the negative cases (`purchased` → `owned`, not `supported`; unmapped/uncertain statuses → `listened`, never the reverse). Edge cases — empty library, artist-less tracks, diacritic/case variants, negative play counts — all have explicit tests.

One thing already flagged in the PR description itself, worth restating since it's a manual-verification item, not a code issue: **not yet exercised against a real library end-to-end** (the Automation permission prompt needs interactive acceptance). Worth doing before this ships, since the AppleScript property names and enum codes were sourced from `sdef` rather than a live run.

@brandonlucasgreen

---

brandonlucasgreen (commented):
@brandonlucasgreen This PR has had nine automated security/regression passes since 2026-08-10, most reaching "no findings." I re-read the current diff rather than repeat that checklist. Security-wise I agree: this is fully local/offline (static AppleScript, no network, `UserDefaults`-only storage), so there's no real attack surface here.

What's worth flagging instead: the 2026-08-11 19:40 pass (the deepest one, cross-referencing `MediaObserver.swift` and friends) found three **confirmed** regression-shaped bugs that are still present in the current diff — none of the four later passes mention them, so they look like they got lost rather than dismissed:

1. **`tell application "Music"` has no run-check guard.** Every other AppleScript call in this app gates on `isAppRunning("com.apple.Music")` (`MediaObserver.swift`) specifically so Unstream never launches the player. `AppleMusicLibraryService.readLibrary()` has no such guard, so a library import launches Music.app — a real behavior change from the app's existing invariant, not just a style nit.
2. **A total cloud-status decode failure looks like a successful import.** If every `cloud status` code fails to map (e.g. `missing value`, plausible with iCloud Music Library off), `makeRecords` falls back to `"unknown"` for every track, `provenance` defaults every one to `.listened`, and `ownedTotal`/`subscriptionTotal` both land at 0. `MusicLibrarySettingsSection.summaryLine` only shows the "you own" / "subscription-only" parts when `> 0`, so the UI just shows "N tracks · M artists" with no signal anything went wrong. This is exactly the "silent failure that looks like success" pattern CLAUDE.md calls out for cached/parsed data elsewhere in this codebase.
3. **`saveSnapshot` unconditionally overwrites a good snapshot with an empty one.** `stringList`/`intList`/`enumCodeList` all treat a non-list or empty descriptor as `[]` rather than an error, so an unexpected AppleScript reply shape produces `trackTotal: 0` and that gets persisted over whatever was previously imported — a successful-looking re-import can silently erase good data.

None of these are security vulnerabilities — this stays entirely on-device — but they're real correctness gaps in the failure paths, and they're the kind of thing that's easy for a fast automated pass to miss on a second read of already-"clean" code. Worth a decision (fix before merge vs. explicitly accept and move on) rather than letting the PR keep collecting "no findings" comments that don't re-derive them.

---

brandonlucasgreen (commented):
@brandonlucasgreen Security review (automated pass)

**Security**
- No confirmed vulnerabilities. `musicLibraryScriptSource` (the AppleScript run against Music.app) is a static string literal, never assembled from user or external input — no injection surface.
- No secrets touched, and this PR makes no network calls: the feature is only `NSAppleScript` execution plus `UserDefaults` read/write. Confirmed by reading the full diff of `AppleMusicLibraryService.swift`.
- Data exposure — needs-manual-verification (judgment call, not a defect): the imported snapshot (artist/album names, play counts, cloud-status) is stored in `UserDefaults.standard` as a plaintext plist, not Keychain. The data isn't credential-like, and the UI/plist copy's claim that it "never leaves this Mac" is accurate — there's no network write path in this diff — but flagging the storage choice for the record since it is a form of personal listening data.
- Verified the `com.apple.Music.library` scripting-access entitlement the code comment claims is "already in Unstream-macOS.entitlements" is in fact present in that file — the comment is accurate, not an undocumented runtime-permission gap.

**Regression**
- `isImporting` guard in `importLibrary()` prevents overlapping imports — confirmed correct.
- `MusicLibrary.makeRecords` refuses to zip mismatched-length parallel arrays (returns `nil` rather than misattributing a cloud status to the wrong track) — confirmed correct, and consistent with this codebase's "never cache/guess on an uncertain read" convention.
- The `guard n > 0 else { return [] }` before each `(1...n)` range in the descriptor decoders (`stringList`/`intList`/`enumCodeList`) correctly avoids a Swift `ClosedRange` runtime trap on an empty descriptor list (e.g. an empty library) — confirmed correct.
- `SettingsView` gained a new optional `libraryService` stored property. Confirmed the struct's synthesized memberwise initializer still defaults both it and `releaseAlertManager` to `nil`, so the existing zero-argument `#Preview { SettingsView() }` call site still compiles — no build regression.
- Provenance mapping (`purchased` → `.owned`, never `.supported`) is covered by unit tests and matches the stated intent of not overstating direct artist support.

No blocking findings.

---

brandonlucasgreen:
@brandonlucasgreen Staff-eng review pass on this one.

## Security

- **Auth/authorization** — N/A. Purely local: reads Music.app via Apple Events under the existing `com.apple.security.scripting-targets` → `com.apple.Music.library` entitlement, no new server endpoint or auth path.
- **Input validation** — `confirmed`, handled well: `MusicLibrary.makeRecords` verifies all four parallel arrays returned by the AppleScript are the same length and returns `nil` (refusing the whole batch) on mismatch, rather than zipping misaligned data and silently mis-attributing cloud status to the wrong track. Play counts are clamped to `max(0, ...)`. Unknown cloud-status codes fall back to `"unknown"` → `.listened` (the "claim less, not more" default), not a crash or a default of `.owned`.
- **Secret handling** — none.
- **Injection risks** — checked: the AppleScript source (`musicLibraryScriptSource`) is a static string literal with no interpolated user or track data, so there's no AppleScript-injection surface here.
- **Unsafe defaults** — `needs-manual-verification`: `LibrarySnapshot` is persisted via plain `UserDefaults.standard` (JSON-encoded, unencrypted), not Keychain. That matches the existing pattern the PR cites (`ReleaseAlertManager`'s state) and the data itself is the user's own local listening stats rather than a credential, so this looks like an intentional, consistent choice — flagging only because it's new disk-resident data outside the sandbox's more-protected stores, worth a quick nod from whoever owns the local-storage conventions.
- **Data exposure** — `confirmed` as correctly scoped: no network calls appear anywhere in this diff (`AppleMusicLibraryService`, `MusicLibrarySettingsSection`, `MusicLibrary`), which matches the updated `NSAppleEventsUsageDescription` string's claim that "library data never leaves this Mac." Worth a rubber-stamp check that this stays true as the Support List feature (referenced as a follow-on) actually ships.

## Regression

- **Changed assumptions** — `SettingsView` gained a second optional stored property (`libraryService: AppleMusicLibraryService?`). Verified this doesn't break the zero-arg `#Preview { SettingsView() }` at `SettingsView.swift:616` — Swift's synthesized memberwise init treats un-initialized `Optional` stored properties as implicitly defaulted to `nil`, so both `releaseAlertManager` and the new `libraryService` remain omittable. No other call site found (`UnstreamApp.swift:135` is the only non-preview constructor, and it was updated).
- **Edge cases / null-empty** — good coverage in `MusicLibraryTests`: empty library, tracks with no artist name (counted in totals, no row), diacritic/case-variant artist merging, unknown status codes, negative play counts. `unsupportedArtists` filtering logic also has directed tests.
- **Off-by-one** — checked the `NSAppleEventDescriptor` decoding loops (`stringList`/`intList`/`enumCodeList`): they use `(1...n)`, which is correct since AppleScript list descriptors are 1-indexed, not 0-indexed. Easy place to get wrong and it's right here.
- **State transitions** — `importLibrary()` guards re-entrancy with `isImporting`; `clearSnapshot()` resets `snapshot` but not `lastError` — minor inconsistency (a stale error message could persist in the UI after "Forget Imported Data" until the next failed/successful import), not a security issue, `plausible` low-severity UI nit.
- **API contract drift** — N/A, no server-side API touched by this PR.

No blocking findings. The provenance mapping (`purchased` → `owned`, never → "supported") is exactly the kind of place I'd expect a bug, and it's both correctly conservative and directly tested against the spec's stated intent.

---

brandonlucasgreen (commented):
@brandonlucasgreen Security/regression pass on this PR (Apple Music library import, Mac app).

**Security**
- Auth/authorization: N/A — local-only feature, no server endpoints added.
- Input validation: `MusicLibrary.makeRecords` refuses to zip AppleScript's parallel result arrays if their lengths mismatch (returns `nil` rather than silently misattributing a status to the wrong track) — confirmed, good defensive handling of an untrusted-shape external read.
- Injection risks: the AppleScript source (`musicLibraryScriptSource`) is a static string constant, never built from user or track data, so there's no script-injection surface. — confirmed, no issue.
- Unsafe defaults / data exposure: the snapshot (play counts, owned-vs-streamed status per artist) is persisted only in local `UserDefaults`, never transmitted — matches the updated `NSAppleEventsUsageDescription` and in-app copy ("Library data never leaves this Mac"). Confirmed by reading `AppleMusicLibraryService` — no network call anywhere in the diff. `com.apple.Music.library` scripting entitlement was already present pre-PR, not newly added. — confirmed, no issue.

**Regression**
- API/contract drift: `SettingsView`'s init gained a second optional param (`libraryService`). Both stored properties are optionals with implicit `nil` default, and the only two call sites (`UnstreamApp.swift`, the `#Preview` using `SettingsView()`) are consistent with that. — confirmed, no issue.
- Provenance mapping (the one place a bug here would misrepresent a real number): `purchased` (iTunes Store) is explicitly mapped to `.owned`, never treated as "supported" — tested directly (`testITunesPurchaseIsOwnedNotSupport`) and matches the spec note in the code comment. — confirmed, no issue.
- Edge cases: empty library, no-artist tracks, negative play counts (clamped to 0), diacritic/case artist-name merging, and unknown cloud-status codes all have direct test coverage in `MusicLibraryTests.swift`. — confirmed, no issue.

No security or regression findings on this PR. It's UI/local-storage-only and reasonably well covered by tests.

---

brandonlucasgreen:
@brandonlucasgreen

Reviewed the diff (8 files: `AppleMusicLibraryService.swift`, `MusicLibrary.swift`, `MusicLibrarySettingsSection.swift`, `MusicLibraryTests.swift`, `Info-macOS.plist`, `UnstreamApp.swift`, `SettingsView.swift`, `project.pbxproj`) against the full file contents on `origin/apple-music-library-import`, plus the entitlements file and `me-listening.ts` on the unrelated branch that implements the server side, for context only.

## Security

No confirmed vulnerabilities in this diff. Two items worth a closer look:

- **needs-manual-verification** — `AppleMusicLibraryService.swift` (`syncToAccount()`, `deleteRemoteSignals()`): both POST/DELETE `https://unstream.stream/api/me/listening` with a bearer token, but that endpoint is not part of this PR and does not exist on `main`. I found an implementation on a separate, unmerged branch (`collection-not-saved-artists`, commit `3f34273`) that does derive the user id server-side from `auth.getUser(token)` rather than trusting a client-supplied id, which is the right pattern — but since that code isn't in this PR, the actual server-side authorization for the endpoint this client talks to can't be verified from this diff and should be checked again against whichever PR actually ships it.
- **plausible** (low severity) — `AppleMusicLibraryService.swift:171-182` (`loadSnapshot`/`saveSnapshot`): the persisted `LibrarySnapshot` (artist names + per-artist play counts — a private listening-history record) is stored as plaintext JSON in `UserDefaults.standard`, not Keychain. It's not a credential and the PR explicitly scopes it to non-iCloud local storage, and it matches the existing `ReleaseAlertManager` UserDefaults pattern the code comment cites, but it's still a persistent unencrypted record of what the user listens to, worth flagging per the project's Keychain-for-sensitive-data guidance rather than treating silently as fine.

## Regression

- **plausible** — `AppleMusicLibraryService.swift:100-135` (`readLibrary`): the four `… of every track` properties (`artist`, `album`, `played count`, `cloud status`) are fetched as four separate, sequential Apple events, not one atomic read. `makeRecords` only refuses when the four returned arrays have *different* lengths; if the library changes between the four reads (track added/removed) in a way that keeps all four counts equal but shifts ordering, cloud status could be silently zipped onto the wrong artist — the exact "wrong provenance on the right artist" failure the surrounding comment says this is designed to prevent, which the length-only check doesn't fully guarantee against.
- **plausible** — `AppleMusicLibraryService.swift:203-206` (`syncToAccount`): when `AuthService.shared.isSignedIn` is `true` but `currentAccessToken()` returns `nil`/throws (stale local session), the guard clause returns silently with `syncError` left `nil`. `MusicLibrarySettingsSection.swift` then shows "Not yet — re-import to sync" with no explanation, so a real failure reads as "hasn't happened yet."
- **plausible** — `AppleMusicLibraryService.swift:186-196` (`clearSnapshot`): `deleteRemoteSignals()` runs as an untracked `Task { }` with no completion guarantee. Quitting the (menu-bar) app immediately after "Forget Imported Data" can cancel the DELETE mid-flight, leaving data on the server despite the button's promise that forgetting locally also forgets remotely, with no retry or surfaced error.
- **needs-manual-verification** — API contract drift: as noted above, `/api/me/listening` doesn't exist in this PR or on `main`; the only implementation is on a separate unmerged branch. If #439 merges first, every sync attempt 404s (shown to the user as "Couldn't sync your library to Unstream (error 404)") until the server-side PR also lands — worth sequencing or noting explicitly in the merge plan.
- **plausible** — `UnstreamApp.swift`: this PR also relocates macOS `unstream://` magic-link handling from `.onOpenURL` on the `Settings` scene to `AppDelegate.application(_:open:)` — unrelated to the library-import feature and un-covered by new tests. It reads like a real fix (the old placement plausibly never fired reliably), but it's a change to the sign-in deep-link path bundled into a feature PR; worth an explicit manual smoke test of magic-link sign-in on macOS before merging, since a regression there would be silent and unrelated to what reviewers would naturally focus on here.

---

brandonlucasgreen:

### Security
- **plausible** — `apps/mac/Unstream/Platform/macOS/AppleMusicLibraryService.swift:244-251` (`deleteRemoteSignals`): the DELETE request's outcome is discarded entirely (`_ = try? await URLSession.shared.data(for: request)`, no status check), so "Forget Imported Data" (`clearSnapshot()`, line 195) can silently fail to remove server-side signals while the UI shows no error and local state is already wiped — the user has no way to know the deletion promise ("forgetting locally has to forget remotely too") wasn't kept.
- **confirmed** — `apps/mac/Unstream/Platform/macOS/AppleMusicLibraryService.swift:203-242` (`syncToAccount`): the request/response cycle to `https://unstream.stream/api/me/listening` only validates the numeric status code, never the response content — which is exactly what makes the regression finding below invisible (see below).

### Regression
- **confirmed** — `apps/mac/Unstream/Platform/macOS/AppleMusicLibraryService.swift:215` and `:247` POST/DELETE `https://unstream.stream/api/me/listening`: no such backend route exists. `netlify.toml` has explicit `/api/me/*` redirects for every other endpoint (`me-settings`, `me-bandcamp`, `me-collection`, `me-feed-token`, `me-recent-releases`, etc.) but none for `listening`, and there's no `me-listening.ts` in `api/functions/`. The only match is the SPA catch-all (`from = "/*" to = "/index.html" status = 200`), so both requests return `index.html` with HTTP 200. Since the Swift success check is only `guard (200..<300).contains(status)` with the body discarded, this is treated as a successful sync — `lastSyncedAt` gets set and `MusicLibrarySettingsSection.swift`'s privacy footer permanently and falsely claims data "is synced to your Unstream account" for every signed-in user, while nothing is ever stored server-side. This is exactly the failure class CLAUDE.md's "never cache uncertainty" principle warns about — a non-answer (SPA fallback HTML) is being treated as a positive result.
- **confirmed** — same file, ~line 90s: `syncToAccount()` is called unconditionally at the end of every `importLibrary()`, so every "Re-import Library" click also fires this always-failing/no-op network call. It's silent to anyone not reading the console log (`print("[AppleMusicLibrary] sync failed: HTTP \(status)")`) — the account-sync half of this feature ships non-functional despite all 18 unit tests passing, because those tests only cover the pure `MusicLibrary` logic, not `AppleMusicLibraryService`'s network path.
- **needs-manual-verification** — `apps/mac/Unstream/UnstreamApp.swift:131-138` and `:298-305`: magic-link auth-callback handling (`unstream://auth/callback`) is moved from a `.onOpenURL` modifier on the `Settings` scene to `AppDelegate.application(_:open:)` — an incidental change to a load-bearing sign-in path bundled into this otherwise-unrelated feature PR. `CFBundleURLTypes`/`CFBundleURLSchemes` registration for `unstream` is present in `Info-macOS.plist`, so the delegate method should fire, but whether AppKit reliably delivers a cold-launch open-URL event before/after `applicationDidFinishLaunching` (i.e. whether a magic-link click that launches the app fresh still completes sign-in) isn't resolvable by reading source alone.

**Bottom line:** the account-sync piece of this feature (POST/DELETE to `/api/me/listening`) needs a real backend endpoint before merge — right now it silently no-ops and misreports success to users.

---

brandonlucasgreen (commented):

**Security:**
- **confirmed (high)** — `apps/mac/Unstream/Platform/macOS/AppleMusicLibraryService.swift:215,227,247`: `syncToAccount()` POSTs and `deleteRemoteSignals()` DELETEs `https://unstream.stream/api/me/listening`. On `main` (this PR's base), no `me-listening.ts` function or `/api/me/listening` route exists. Netlify's catch-all SPA redirect (`netlify.toml:398`, `/* → /index.html`) matches any unmatched route/method and returns 200. Since the client's success check is only `(200..<300).contains(status)` (line 227), the POST "succeeds" against the SPA shell — `lastSyncedAt` gets set and the Settings UI reports "Synced to your account: just now" while nothing is stored server-side. This is the exact "don't report uncertainty as success" failure class called out in CLAUDE.md. **Note:** PR #454 (`collection-not-saved-artists`, based on `batch/aug-16`) adds a real `api/functions/me-listening.ts` with matching POST/GET/DELETE semantics — if that lands on `main` before or alongside this PR, this finding resolves itself; if #439 merges to `main` independently first, it ships a false-positive "synced" UI state with zero backend.
- **needs-manual-verification (deferred)** — server-side auth/RLS scoping for `/api/me/listening` can't be assessed from this diff since the endpoint doesn't exist on this PR's base branch; see PR #454's implementation for that half.
- **confirmed (low)** — `AppleMusicLibraryService.swift:244-251`: `deleteRemoteSignals()` discards the response/error entirely (`_ = try? ...`), so "Forget Imported Data" can't currently distinguish a real remote delete from a no-op. Low impact only because the remote endpoint doesn't exist yet either.
- No secret-handling, injection, or SSRF issues found: the sync destination is a hardcoded constant, the AppleScript source has no data interpolation, and only artist name + play count leave the device.

**Regression:**
- **plausible (low)** — `AppleMusicLibraryService.swift:80-86`: auto-sync fires on every import whenever the user is signed in, with consent limited to descriptive footer text rather than an explicit toggle. Worth a product judgment call given CLAUDE.md's transparency principle, not a hard bug.
- This diverges from the existing Bandcamp `collection_items` pattern (server-fetched via `bandcamp-sync-background.ts`) by necessity — Apple Music library data is only readable client-side — but as shipped in this PR alone, it's a client-only half of that pattern with a dangling API contract until a backend lands (see security finding above).

Flagging as a blocker given the confirmed-high finding: this should merge after/with #454's backend, or gate the "synced" UI state on something more specific than a 2xx from a route that doesn't exist yet on `main`.

---

brandonlucasgreen:
@brandonlucasgreen Reviewed the diff for PR #439 (Apple Music library import). Findings below.

## Security

- **[confirmed] "Forget Imported Data" doesn't reliably clear server-side data.** `AppleMusicLibraryService.clearSnapshot()` always wipes the local UserDefaults snapshot, then fires `deleteRemoteSignals()` async. But `deleteRemoteSignals()` guards on `AuthService.shared.isSignedIn` and a fresh access token — if the user signed out *after* a prior successful `syncToAccount()` (or the token fetch simply fails), the guard short-circuits and no DELETE is ever sent. The button's own code comment says forgetting locally "has to forget remotely too, or 'Forget Imported Data' quietly leaves a copy on the server," but the signed-out path does exactly that. (`Platform/macOS/AppleMusicLibraryService.swift`, `clearSnapshot()` / `deleteRemoteSignals()`)
- **[confirmed] Remote-delete failures are swallowed silently.** `deleteRemoteSignals()`'s network call is `_ = try? await URLSession.shared.data(for: request)` — no status-code check, no error surfaced to the UI, no retry. A failed DELETE (network error, 5xx, expired token → 401) leaves data on the server while the user is told nothing went wrong. This is the one place ("forget my data") where a silent failure matters most. (Same file/function.)
- **[no issues found] AppleScript injection.** The only AppleScript executed is the static `musicLibraryScriptSource` literal — no track-derived or user-controlled strings are interpolated into script source, and reads use `NSAppleScript` directly (no `osascript` shell-out). No injection surface.
- **[no issues found] Provenance-mapping default.** `MusicLibrary.provenance(forCloudStatus:)`'s `default` branch returns `.listened` (the least-claim state) for any unmapped/unknown status; nothing falls through to `.owned` implicitly, and `testUncertainStatesClaimLessNotMore` covers it. Matches the documented intent.
- **[no issues found] Logging / secret handling.** Only aggregate counts are printed (`print("[AppleMusicLibrary] imported N tracks…")`); no artist names, track titles, or the bearer token are logged. Local storage is plain `UserDefaults.standard` (not iCloud KVS), consistent with the "local-only" claim in the Settings copy.

## Regression

- **[needs-manual-verification] Unrelated auth-callback refactor bundled into this PR.** `UnstreamApp.swift` moves macOS `unstream://auth/callback` handling from `.onOpenURL` on the `Settings` scene to `AppDelegate.application(_:open:)`. This is a real behavior change to magic-link sign-in, orthogonal to the Apple Music import feature and untested by this PR's new test file. Worth an explicit manual pass on magic-link sign-in on macOS before merge.
- **[needs-manual-verification] `/api/me/listening` payload shape.** `syncToAccount()`/`deleteRemoteSignals()` POST/DELETE `https://unstream.stream/api/me/listening` with `{"source":"apple_music","signals":[{"artistName","playCount"}]}`. Per the task context this endpoint ships in #454; couldn't confirm the shape matches from this repo alone.
- **[plausible] "Forget" during an in-flight import.** `clearSnapshot()` has no internal guard against a concurrent `importLibrary()`; only the UI disables the button while `isImporting` is true. If the two ever race, the import's completion handler (`snapshot = snap; saveSnapshot(snap)`) would silently resurrect data the user just asked to forget.
- **[needs-manual-verification] Very small library (0–1 tracks).** `makeRecords` correctly refuses misaligned parallel arrays (confirmed, and unit-tested), but whether Music.app's `of every track` batch properties reliably return well-formed lists (vs. a bare scalar) for a 0- or 1-track library is an AppleScript runtime behavior that can't be confirmed from source alone.
- **[no issues found] Array zip / off-by-one.** `makeRecords` fails closed on any length mismatch across the four parallel arrays rather than zipping partially — matches the PR's stated design and is directly covered by `testMakeRecordsRefusesMisalignedArrays`.

---

PR #440, opened 2026-08-10
Announce each release once, and give the notification emails an opt-out footer

brandonlucasgreen:
Three fixes to the product notification emails, plus the migration they need.

## 1. A release is announced exactly once

The release alert worked out what was new from counts: `releases_found` before and after a catalog run, with the difference taken as "the newest N rows by `created_at`". A count can't say **which** records are new. A run that drops one release and adds two leaves the total up by one and re-announces something already sent; a total that dips and recovers repeats the same rows again.

`releases.alert_sent_at` states it directly. The alert claims an artist's unaccounted-for releases with a single `UPDATE ... RETURNING` — atomic, so two catalog runs racing on one artist split the new releases rather than both claiming the lot — emails exactly what it claimed, and **sends nothing at all when the claim comes back empty**.

Two consequences, both deliberate and commented at the call site:

- **Claiming happens before recipients are resolved.** Releases found while nobody was saving the artist are marked seen and never announced, so saving an artist gets you alerts from that point forward rather than a mail-out of everything you missed.
- **An artist's first catalogue is claimed but not announced.** A whole back catalogue arriving at once is a backfill; "34 new releases" would be wrong about every record in the list. This also covers a catalogue that previously found nothing at all — a broken parser since recovered.

The migration backfills every existing row, so nothing already catalogued goes out as new.

## 2. Release alerts say what actually shipped

```
New release from Test Artist: Fine Motor Control

Test Artist, an artist you saved on Unstream, has 5 new releases:
- Fine Motor Control
  12 August 2026 · Available on Bandcamp, Discogs
  https://unstream.stream/a/test-artist/fine-motor-control
- Untitled Sketches
  Release date not listed
  ...
…and 2 more.
```

Dates render at the precision actually stored — "July 2026" for a month-only date, "Release date not listed" when there is none — so no day is ever invented. Platforms are ordered artist-paying-first via `payoutRank`, matching the release pages. Up to five releases are named and the rest counted. A failed platform lookup degrades the email to titles and dates rather than dropping it, since the releases are already claimed by that point.

## 3. An opt-out footer on every subscription email

New releases, new platform links, and the weekly analytics recap now end with why the recipient got it, a **Manage your email notifications** link, and sender identity — plus a `List-Unsubscribe` header (RFC 2369) pointing at the same place. Settings gained an `id="notifications"` anchor for those links to land on.

The claim approved/rejected emails are deliberately left alone: they answer an action the person just took and have nothing to unsubscribe from, which is also what the `notification_preferences` migration already says.

## Verification

`npm run build` passes — 1078 API tests, 717 web unit tests, including new coverage for the claim-once behaviour, the first-catalogue suppression, date/platform rendering, the footer, and the header passthrough.

The migration was exercised against a throwaway Postgres 16: existing rows backfill, new rows arrive `NULL`, a second claim returns zero rows, a re-run doesn't re-backfill pending rows, and the partial index serves the claim lookup.

## Known sharp edge

A re-catalog can discover a genuinely old release the parser missed earlier, and that will be announced as new. The release date in the email makes it visible ("3 March 1998"); if it turns out to be noisy, the fix is to skip claims whose release date is well in the past.

---


brandonlucasgreen (commented):
@brandonlucasgreen Automated senior-staff-engineer review (scheduled security/regression pass). Read the full diff for this PR (notifications.ts, db.ts, resend.ts, the new migration, and the test changes).

## Security

- **Content spoofing via unescaped subject line — plausible, low severity.** In `notifySavedArtistsOfNewRelease` (`api/functions/notifications.ts`), the email `subject` interpolates `artist.name` and `releases[0].title` without `escapeHtml`/sanitization, unlike the `html`/`text` bodies which do escape them. These values come from enrichment/scrape data rather than a live user request in this path, and Resend's JSON API transport rules out classic SMTP header-CRLF injection — but if that title/name pipeline ever admits attacker-influenced text, it lands unescaped in a subject line. Low risk today; flagging as needs-manual-verification rather than confirmed.
- **Auth/authorization, input validation, secret handling, unsafe defaults, data exposure — no issues found.** All new DB access (`claimUnalertedReleases`, `getPlatformsByRelease`) goes through the Supabase client with parameterized filters; `artistId` is always server-derived (from `recordCatalogOutcome`, never from a request body in this path). The migration's `ADD COLUMN ... DEFAULT now()` → `DROP COLUMN DEFAULT` sequencing is a standard, race-free backfill — existing releases are correctly marked "already alerted" so deploy doesn't cause a mass-announce. `SUBSCRIPTION_EMAIL_HEADERS`/List-Unsubscribe correctly points at a login-gated settings page rather than an unauthenticated one-click endpoint, matching the code's own RFC 8058 reasoning.

## Regression

- **Edge case (is_hidden ↔ alert_sent_at) — needs-manual-verification.** `claimUnalertedReleases` only claims rows where `is_hidden = false`. A release hidden at first catalog time is never claimed while hidden (`alert_sent_at` stays NULL); if later un-hidden, the next catalog run will claim and announce it as "new," even if it's actually old. Nothing in this PR triggers an unhide, so it's latent rather than active, but worth confirming it's the intended behavior before an unhide path exists.
- **Changed assumption — confirmed, appears intentional.** The suppression rule moved from "count changed → always email" to "`previousReleasesFound === 0` → never email" inside `notifySavedArtistsOfNewRelease` itself. This correctly stops backfills from being announced (well tested), but it also means a catalog run recovering from a *parser failure* that had driven `releases_found` to 0 is treated identically to a brand-new artist's first backfill — both stay silent rather than announcing the recovered releases. The code comment calls this out as deliberate; flagging so it's a conscious call, not a side effect.
- **API contract drift — none found.** The only caller of `notifySavedArtistsOfNewRelease` (`db.ts`) was updated in the same PR to the new `{ previousReleasesFound }` shape, and no other caller exists (verified via repo-wide search).

No blocking issues found — the two flagged items are worth a look but neither looks like an active bug.

---

PR #441, opened 2026-08-10
Only alert on releases that are actually new

brandonlucasgreen:
Follow-up to #440.

Cataloguing discovers records, it doesn't witness them being released. A parser improvement, a newly linked platform or a Discogs pass can surface an album from 1998 for the first time, and to the alert that looked exactly like an album published this morning. "New release from X" about a record somebody may have owned for twenty years is noise, and noise in the first alerts anyone receives is the expensive kind.

## The rule

An alert now only names releases that **came out in the last week, or are still to come**. Everything else the run turned up is still claimed, so it can't resurface in a later email either, and an artist with nothing recent produces no email at all.

"Still to come" covers both a future release date and an upstream pre-order flag (`status = 'announced'`), which can reach us without a date — that one still gets announced, saying plainly that the date isn't listed rather than inventing one.

## Undated releases stay pending, they aren't written off

An undated release is deliberately left **unclaimed**. "We don't know when this came out" is not "this is old" — the Bandcamp detail pass is budgeted, so a release can arrive dateless in one run and dated in the next — and writing `alert_sent_at` on the strength of a missing date would cache that uncertainty as a permanent no, which is the single most repeated bug class in this codebase.

So an undated release stays pending until a date settles it in either direction. One that never gets a date is never announced, which is the honest outcome: we have nothing to tell anyone about when it happened.

The practical consequence is that `idx_releases_unalerted` now holds a standing population of undated rows rather than briefly-pending ones. It's still a small per-artist lookup and needs no change, but #440's column comment said otherwise, so `20260810180000` restates it. That migration is **comment-only — no schema change**.

## The first-catalogue special case is gone

#440 suppressed an artist's opening crawl wholesale, because it arrives as a whole discography and "34 new releases" was wrong about every record in it. That was only ever true because the records were *old*, which the date window now excludes on their own merits. Keeping both rules would mean a first crawl containing last week's single gets silently dropped — a real alert lost. One honest rule ("is this release actually new") replaces two overlapping heuristics, and `previousReleasesFound` disappears from the call.

## Why the claim reads before it writes

The obvious implementation is a single `UPDATE` with the age rule inlined as `.or('release_date.not.is.null,status.eq.announced')`. That puts the rule in a PostgREST filter string used nowhere else in this repo, and a mistake there wouldn't fail — it would quietly match the wrong rows and mail them out.

So the claim reads the unaccounted-for rows, decides in `canJudgeAge` (typechecked, unit-tested), and writes by id. Atomicity is unchanged: `alert_sent_at IS NULL` is still on the write, and its `RETURNING` is still what decides who won when two catalog runs race on the same artist. The cost is one extra query on a path that only runs when a crawl found something new.

## Verification

`npm run build` passes — 1083 API tests, 717 web unit tests. New coverage for the age window (including the exact 7-day boundary in both directions), upcoming releases, dateless pre-orders, an undated release becoming announceable once its date arrives, mixed batches where only the recent entries are named *and counted*, the claim-write failure path, and losing a race to a concurrent run.

Both migrations were applied in order against a throwaway Postgres 16 to confirm the comment-only change lands cleanly on top of #440's.

---

### DIFF
--- a/api/functions/__tests__/notifications.test.ts
+++ b/api/functions/__tests__/notifications.test.ts
@@ -104,6 +104,7 @@ interface ReleaseRow {
   slug: string;
   release_date: string | null;
   date_precision: string | null;
+  status: string;
   created_at: string;
 }
 
@@ -113,14 +114,15 @@ interface ReleaseRow {
  * `saved_artists`, `artists`, `releases`, `release_sources`, `email_log`,
  * `notification_preferences`, and `auth.admin.getUserById` all in one client.
  *
- * The `releases` update is modelled as a real claim: it hands back the rows that are still
- * unclaimed and empties the pool, so a second call to the same client sees what production
- * would see on the next catalog run — nothing.
+ * `releases` is modelled as the real two-step claim: a select of everything still unclaimed,
+ * then an update over the ids the caller chose, which marks exactly those and reports back which
+ * it won. A second call sees what the next catalog run would see, so anything the caller declines
+ * to claim — an undated release — stays pending exactly as it does in the database.
  */
 function makeFanoutClient(opts: {
   savedArtists?: { data: { user_id: string }[] | null; error?: unknown };
   artistRow?: { data: { name: string; slug: string } | null; error?: unknown };
-  releases?: { data: ReleaseRow[] | null; error?: unknown };
+  releases?: { data: ReleaseRow[] | null; error?: unknown; claimError?: unknown; stolenIds?: string[] };
   releaseSources?: { data: { release_id: string; platform: string }[] | null; error?: unknown };
   emails?: Record<string, string>;
   optedOut?: Record<string, boolean>;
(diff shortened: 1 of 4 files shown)


brandonlucasgreen (commented):

## Security
No findings. `artistId`/release `id`s never originate from user input on this path, all reads/writes go through the Supabase query builder (parameterized), and the PR's own design deliberately keeps the age rule out of a PostgREST filter string (read-then-claim-by-id) specifically to avoid a class of silent-mismatch risk. No auth surface, no secrets, no new data exposure.

## Regression

**1. [confirmed] A previously-undated release that later gets a date is never actually announced, unless a coincidental new release also lands for the same artist.**
Traced end-to-end in `db.ts` and `catalog-artist-background.ts`: `notifySavedArtistsOfNewRelease` is only invoked from `recordCatalogOutcome` when `outcome.releasesFound > previous` (`db.ts` ~L606). `releasesFound` is the *grid size* for that run (`catalogBandcamp` → `found: written.length`), not a delta of "what changed." Backfilling a date onto a release that was already counted in a prior run's grid doesn't change grid size, so this gate never fires for that run — `claimDatedReleases`/`isNewsworthy` never even execute against that row. The release does stay correctly unclaimed in the DB (`alert_sent_at IS NULL`), satisfying the letter of "stays pending until a date settles it," but there's no path that re-examines it once dated unless *another* genuinely new release happens to trip the count gate later, at which point it rides along as a side effect. If that artist gets nothing else, this release is silently never announced — which contradicts the PR's stated guarantee. Worth deciding: either call `notifySavedArtistsOfNewRelease` on every successful run (not just on count growth), or add a narrower signal (e.g. "did this run's detail pass complete a previously-undated release") to the gate in `db.ts`.

**2. [plausible] The claim write no longer re-checks `is_hidden`.** Old `claimUnalertedReleases` was a single `UPDATE ... WHERE is_hidden = false AND alert_sent_at IS NULL`. New `claimDatedReleases` checks `is_hidden` only on the initial `select` (~L308-313); the later `.update(...).in('id', ...).is('alert_sent_at', null)` write has no `is_hidden` condition. A release hidden (e.g. by an admin merge) in the window between read and write can still be claimed and announced despite now being hidden. Narrow timing window, but a real reduction vs. the prior atomic guarantee.

**3. [plausible] `isNewsworthy`/`canJudgeAge` treat all dates as day-precision.** `date_precision` is read into `ClaimedRow` but never consulted. A month/year-precision release stored as day 1 of that month/year can fall outside (or wrongly inside) the 7-day window relative to when it actually shipped. New with this PR — no added test exercises non-`'day'` precision.

**4. [plausible] `status === 'announced'` is trusted unconditionally in `isNewsworthy`**, with no cross-check against `release_date` even when one is present. If upstream data ever has `status='announced'` with a `release_date` well in the past (e.g. a lagged transition to `'released'`), the release bypasses the 7-day window entirely on status alone.

**5. [needs-manual-verification] Boundary-test determinism.** The new 7-day-boundary test computes `now`/`isoDate()` via separate untampered `Date.now()` calls. Extremely low probability, but a UTC-midnight straddle between the two reads could flip the exact-boundary assertion. Worth confirming CI has never flaked here, since the PR explicitly calls out this boundary as a verification highlight.

Migration `20260810180000` verified comment-only as claimed (single `COMMENT ON COLUMN`, no schema change).

---

PR #442, opened 2026-08-10
Pause the weekly analytics recap

brandonlucasgreen:
Follow-up to #440 / #441. The recap, not the release alerts, is what exhausted the day's Resend allowance.

## Why

**Volume.** It's the only email that fans out to the whole roster at once — one send per verified claimed profile, all in the same minute — with nothing throttling or spreading the batch. On Resend's free plan (100/day) a single Monday run can consume the entire day's allowance and starve the saved-artist alerts, which are the ones a fan is actually waiting on. The release alerts are bounded by savers and, after #441, by whether anything was genuinely released; the recap is bounded only by how many artists have claimed a profile, which is the number we want to grow.

**Content.** Most claimed artists don't yet get enough searches, views or clicks in a week for the recap to say anything worth reading. "3 views, 0 clicks" is an honest number and a discouraging email, and it makes Unstream look smaller than it is to exactly the people we most want to convince.

## What changes

The weekly cron comes off `weekly-analytics-recap.yml`. `workflow_dispatch` stays, so it can still be run by hand — a deliberate act by somebody who can see the day's remaining allowance, and safe to over-run because the function's `email_log` guard (artist + that week's Monday) won't send twice.

The `/settings` toggle is hidden while it's paused, for the same reason the saved-artist toggles were hidden while those were admin-only: a switch for an email nobody receives is a promise the product doesn't keep. The stored preference is untouched — still written, still returned by the API, still honored — so restoring the toggle alongside the schedule brings back whatever each artist had already chosen.

**This is a pause, not a removal.** The function, its tests, the `notification_preferences.weekly_analytics_recap` column and the email body all stay exactly as they are.

## What has to be true to turn it back on

Recorded in the workflow header so it doesn't get restored by reflex:

- Enough traffic that a typical recap reports numbers worth reading — and/or a floor below which an artist is skipped rather than sent a discouraging one.
- Either a paid Resend plan, or a send budget that can't crowd out the release alerts.
- A monthly cadence would ease both at once: fewer sends, and numbers with more in them.

## Verification

`npm run build` passes — 1083 API tests, 718 web unit tests. The workflow YAML parses with `workflow_dispatch` as its only trigger. A new UI test asserts the recap offers no switch while the underlying preference still exists.

---

### DIFF
--- a/.github/workflows/weekly-analytics-recap.yml
+++ b/.github/workflows/weekly-analytics-recap.yml
@@ -8,19 +8,36 @@ name: Weekly analytics recap
 # Netlify functions in this repo — same precedent as recatalog-sweep.yml,
 # schedule-social-posts.yml, and upstash-keepalive.yml.
 #
-# ## Cadence: once a week, Monday mornings US time
+# ## PAUSED — no schedule, manual dispatch only
 #
-# '0 13 * * 1' is 13:00 UTC Monday — 6am PT / 9am ET, so the recap lands at the start of an
-# artist's work week rather than overnight. Every run recomputes the "last 7 days" window from
-# when it actually fires, so this is safe to reschedule without a backfill.
+# The weekly cron ('0 13 * * 1' — 13:00 UTC Monday, 6am PT / 9am ET) is deliberately removed,
+# for two reasons that both need to stop being true before it goes back.
 #
-# Safe to over-run: the function's own idempotency guard (email_log, keyed on artistId + the
-# Monday date of that week) means a duplicate invocation the same week sends nothing twice.
+# 1. Volume. This is the only email that fans out to the whole roster at once: one send per
+#    verified claimed profile, all in the same minute. On Resend's free plan (100/day) a single
+#    Monday run exhausts the entire day's allowance and starves the saved-artist alerts, which
+#    are the ones a fan is actually waiting on. Nothing here throttles or spreads the batch.
+#
+# 2. Content. Most claimed artists don't yet get enough searches, views or clicks in a week for
+#    the recap to say anything worth reading. "3 views, 0 clicks" is an honest number and a
+#    discouraging email, and it makes Unstream look smaller than it is to exactly the people
+#    we most want to convince.
+#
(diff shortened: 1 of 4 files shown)


brandonlucasgreen (commented):

**Security**
- Auth/authorization: N/A — no endpoint auth or ownership checks are touched by this diff.
- Input validation: N/A — no new inputs.
- Secret handling: N/A — no secrets involved.
- Injection risks: N/A — no new fetches, queries, or rendered HTML.
- Unsafe defaults: N/A — the cron removal is the safe direction (fails toward not sending, not toward over-sending).
- Data exposure: N/A.

No findings in any category — this is a workflow YAML change plus a client-only UI change (removing one array entry from `TOGGLES` in `NotificationPreferences.tsx`).

**Regression**
- Verified the full `NotificationPreferences.tsx` file, not just the diff hunk: `weeklyAnalyticsRecap` stays in the `Preferences` type and is still fetched/POSTed by key, it's just no longer in the `TOGGLES` array that drives rendering. No other code path iterates `TOGGLES` or assumes 3 entries, so removing the entry doesn't break `handleToggle`, the fetch/save round-trip, or the test file's remaining assertions. — **confirmed, no regression**.
- Backend (`me-notification-preferences` function, email send path) is untouched by this diff, so the stored preference continues to round-trip exactly as before. — **confirmed**.
- State transition (schedule → manual-only): the workflow's own `email_log` idempotency guard (keyed on artistId + week) is unchanged and still prevents double-sends if the manual dispatch is run more than once in a week. — **confirmed**.
- Edge case: an artist who already opted in to the recap keeps `weeklyAnalyticsRecap: true` in storage with no UI to see or change it until the toggle is restored — this matches the PR's stated intent (pause, not remove) and isn't a bug, but flagging it as a **needs-manual-verification** product note: worth confirming no other surface (e.g. an email footer "manage preferences" link) still references the now-hidden toggle in a way that would confuse someone who clicks through.

Overall: safe, self-contained change. No security or regression blockers.

---

PR #444, opened 2026-08-11
Record the migration production already has, unblocking every later migration

brandonlucasgreen:
**Migration deploys have been broken since 2026-08-10. Release alerts have been down for ~30 hours.**
This is the unblock.

## What happened

`20260809120000_bandcamp-collection.sql` was applied to production on 2026-08-09 via
`workflow_dispatch` from **PR #438, which is still open**. That was deliberate and documented in the
migration's own header — deploy previews share the production database, and the migration is purely
additive, so applying it early was the safe side of the migration/deploy race.

The unforeseen consequence: `supabase db push` refuses to run when the remote has an applied version
that isn't in `supabase/migrations/` locally, and it **aborts before applying anything**. So one file
sitting on an unmerged branch silently blocked every migration on `main`.

Three consecutive runs failed with the identical error, applying nothing:

| Run | PR | Result |
|---|---|---|
| 2026-08-10 14:05 | #440 | `Remote migration versions not found` |
| 2026-08-10 14:45 | #441 | same |
| 2026-08-11 19:26 | #443 | same |

## What was stranded in production

- `20260810120000_releases-alert-sent-at` — adds `releases.alert_sent_at`
- `20260810180000_alert-sent-at-comment` — documents it
- `20260811120000_app-events-dashboard-indexes` — the disk-IO indexes from #443

⚠️ **The first one is a live production bug.** `api/functions/notifications.ts` shipped in #440
querying `.is('alert_sent_at', null)` against a column that doesn't exist in production, so
`claimDatedReleases` errors, reports to Sentry, and returns `[]`. **No release alert has been sent
since 2026-08-10.** There should be repeated `claimDatedReleases.read` exceptions in Sentry from that
date.

## The fix

Add the migration file to `main`, so the repo again records what production has. Chosen over the two
alternatives:

- **`supabase migration repair --status reverted`** would leave production's migration history
  contradicting the database (the tables exist), and needs a privileged CLI command run against prod.
- **Merging PR #438** would ship the Support Loop web half, which still needs `BANDCAMP_CREDENTIAL_KEY`
  set in Netlify and has Step 3 blocked on open spec questions.

Why this is safe:

- **No feature code ships.** Only the migration file. Those three tables have existed in production
  for two days; only the repo was out of date.
- **PR #438 still merges cleanly.** The file is byte-identical to the copy on that branch —
  `sha256 5f6b31fb3612fe9c087c3df52f850d4a809c83b14b1c249da13ba4f5df1f79b4`, verified against both
  the original commit and the branch tip.
- **It won't be re-applied.** The version is already in the remote history, so `db push` skips it. If
  it ever were, every statement is guarded (`IF NOT EXISTS`, `DROP … IF EXISTS`).
- **It was the only divergence.** The CLI lists every missing version and named exactly one.

## On merge

This applies the three stranded migrations, which restores release alerts and lands the #443 indexes.
Watch that the Supabase migrations workflow goes green.

## Prevention

The trap is now recorded in `.github/workflows/supabase-migrate.yml` and in CLAUDE.md's migrations
section. The early application was a reasonable call — what was missing was "merge it promptly, or
put the migration in its own PR first, and check the workflow went green," because this failure is
loud in Actions and invisible everywhere else.

### DIFF
--- a/.github/workflows/supabase-migrate.yml
+++ b/.github/workflows/supabase-migrate.yml
@@ -1,5 +1,28 @@
 name: Supabase migrations
 
+# Applies new migrations to production on every push to main that touches
+# supabase/migrations/. Also runnable by hand (workflow_dispatch) from a feature branch, to get a
+# migration into production ahead of its PR merging — the deploy preview shares the production
+# database, so an additive migration sometimes has to land early for the preview to exercise it.
+#
+# ⚠️ If you run it from a branch, MERGE THAT BRANCH PROMPTLY, or split the migration into its own
+# PR and merge that first.
+#
+# `supabase db push` refuses to run at all when the remote has an applied version that isn't in
+# supabase/migrations/ locally — "Remote migration versions not found in local migrations
+# directory" — and it aborts *before* applying anything. So a migration applied from an unmerged
+# branch silently blocks every later migration on main, including everybody else's.
+#
+# That is not hypothetical: 20260809120000_bandcamp-collection.sql was applied this way on
+# 2026-08-09 from the still-open PR #438, and from 2026-08-10 to 2026-08-11 three consecutive runs
+# on main failed without applying a thing. One of the migrations left stranded added
+# `releases.alert_sent_at`, which api/functions/notifications.ts had already shipped code against —
+# so release alerts were broken for about 30 hours while this workflow reported the failure to
+# nobody in particular.
+#
+# The failure is loud in Actions and invisible everywhere else. Check this workflow went green
+# after any merge that adds a migration.
+
 on:
   push:
     branches: [main]
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -364,6 +364,24 @@ When adding a table or column: create a new migration in `supabase/migrations/`,
 
 **Auto-deploy:** `.github/workflows/supabase-migrate.yml` runs `supabase db push --linked` on every push to `main` that changes `supabase/migrations/`. Migrations deploy automatically — no manual SQL editor needed. Required GitHub secrets: `SUPABASE_ACCESS_TOKEN`, `SUPABASE_DB_PASSWORD`.
 
+**Applying a migration to production from an unmerged branch poisons every later migration.** The
+workflow can be run by hand (`workflow_dispatch`) from a feature branch, which is sometimes
+necessary — deploy previews share the production database, so an additive migration has to land
+early for the preview to exercise it. But `supabase db push` refuses to run when production has an
+applied version that isn't in `supabase/migrations/` locally, and it aborts *before applying
+anything*. So one migration applied from a branch that then sits unmerged blocks everybody's
+migrations on `main`, silently.
+
+This happened: `20260809120000_bandcamp-collection.sql` went to production on 2026-08-09 from the
+still-open PR #438, and three consecutive runs on `main` then failed without applying anything.
+One stranded migration added `releases.alert_sent_at`, which `api/functions/notifications.ts` had
+already shipped code against — so release alerts were down for about 30 hours. The fix was to add
+the missing file to `main` so the repo again records what production has.
+
+If you apply a migration ahead of its merge: merge it promptly, or put the migration in its own PR
+and merge that first. Either way, **check this workflow went green** — the failure is loud in
+Actions and invisible everywhere else.
+
 **Local dry-run:** `npm run migrate:link` once per checkout, then `npm run migrate:dry-run`. Both
 that and `migrate:list` use `--linked`, matching the workflow — the CLI dropped `--project-ref`
 from these commands, and passing it makes them print a help blob and exit *without* checking
--- a/supabase/migrations/20260809120000_bandcamp-collection.sql
+++ b/supabase/migrations/20260809120000_bandcamp-collection.sql
@@ -1,0 +1,197 @@
+-- Migration: bandcamp_connections, collection_items, listening_signals
+--
+-- Applied to production ahead of the PR merge (workflow_dispatch, 2026-08-09) so the
+-- deploy preview — which shares the production database — could exercise the connect
+-- flow. Purely additive, so the early application is the safe side of the
+-- migration/deploy race.
+--
+-- The Support Loop, Step 1 (docs: support-loop-spec.md, collection-spec.md §4–5).
+-- Three tables:
+--
+--   bandcamp_connections  one row per user who connected their Bandcamp collection via
+--                         Bandcamp's Subsonic API (open beta, shipped 2026-07-16)
+--   collection_items      releases a user actually acquired — the public collection
+--   listening_signals     streaming-side play counts (Apple Music, Last.fm, the Mac app's
+--                         now-playing monitor). Feeds the private "gap" report. Schema lands
+--                         now so Steps 2 and 5 have a landing zone; nothing writes it yet.
+--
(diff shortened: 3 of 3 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen The mechanical fix is right and I verified it end to end: the failing run's log names `20260809120000` as the *only* missing remote version, and the file added here is byte-identical to the commit that was actually pushed to production. Two things are worth acting on before the tables start carrying data — one RLS gap that quietly defeats the invariant the migration's own comments claim, and one FK interaction with the release-merge path.

## Security

- **[confirmed]** `supabase/migrations/20260809120000_bandcamp-collection.sql:143-145` — the `collection_items` UPDATE policy is row-scoped but not column-scoped, which defeats the "provenance stays honest" reasoning at line 151. Withholding the INSERT policy stops a client minting a `purchased` row, but an authenticated session holding the anon key (the web app creates a browser-side Supabase client at `apps/web/src/services/auth.ts:37`, so `authenticated` can reach PostgREST directly) can `UPDATE` its own rows and set `provenance` to `purchased`, `hidden` to false, and rewrite `title` / `artist_name` / `art_url` / `release_id` to anything. Concrete failure: a user flips an `owned`/`listened` item to `purchased`, or repoints a real purchase's `release_id` and title at a release they never bought, and the public collection page asserts support that didn't happen — exactly the lie the comment says the design prevents. Two clean fixes, in order of fit with the rest of the codebase: (a) drop the UPDATE policy and route the hide toggle through a server endpoint like every other write here, or (b) keep it and add column privileges, since RLS is row-level not column-level (as `20260808160000`'s header already notes) — `REVOKE UPDATE ON collection_items FROM authenticated; GRANT UPDATE (hidden) ON collection_items TO authenticated;`. Note `user_id` reassignment is *not* exposed: Postgres reuses `USING` as the implicit `WITH CHECK` for UPDATE, so a row can't be moved to another user. `listening_signals` is correct here — SELECT + DELETE only, no UPDATE, matching its comment at line 191.
- **[needs-manual-verification]** line 111 — `hidden NOT NULL DEFAULT false` combined with "everything from the Bandcamp import is `purchased`" (line 105) means connecting Bandcamp publishes a user's entire purchase history by default, item by item, unless the public collection page is gated behind an explicit account-level sharing opt-in the way `/u/:handle` is (migration 022 `user_sharing`). The page code isn't in this diff so I can't confirm; if it isn't gated, the default is the wrong way round for a table built out of purchase records.
- **[plausible]** `.github/workflows/supabase-migrate.yml:26-31` (pre-existing, not introduced here) — `workflow_dispatch` has no branch or environment restriction, so anyone with write access can run arbitrary migration SQL against production from an unreviewed branch using `SUPABASE_DB_PASSWORD`. This PR documents that path as sanctioned practice without adding a guardrail. A GitHub `environment:` with a required reviewer on the migrate job would keep the escape hatch and put a human in front of it.
- **Verified clean:** `bandcamp_connections` (line 68) does have RLS enabled with zero policies, and no policy anywhere else in `supabase/migrations/` targets that table — `credential_ciphertext` is unreachable through PostgREST for both `anon` and `authenticated`, as the comment claims. CHECK constraints (`sync_status`, `item_count >= 0`, `source`, `provenance`, `acquisition`, `play_count >= 0`) are all closed enumerations or non-negative bounds with no gaps. FK cascades are right: `auth.users` → CASCADE deletes a user's connection, collection and signals on account deletion. No injection surface (no dynamic SQL, no SECURITY DEFINER; the shared trigger `set_releases_updated_at` is plpgsql with `SET search_path = ''`).

## Regression

- **[confirmed]** Recording the file re-applies nothing. Run 31527833248's log shows `supabase db push` aborting with "Remote migration versions not found in local migrations directory" and naming exactly one version, `20260809120000` — so this file is the whole blockage. Because that version is already in the remote `schema_migrations`, push will treat it as applied and skip it; only the three genuinely pending files get applied, in order: `20260810120000_releases-alert-sent-at.sql`, `20260810180000_alert-sent-at-comment.sql`, `20260811120000_app-events-dashboard-indexes.sql`. All three are newer than `20260809120000`, so there's no out-of-order case for `--include-all` to silently paper over, and all three are additive (`ADD COLUMN IF NOT EXISTS`, `CREATE INDEX IF NOT EXISTS`, no `CONCURRENTLY`, so nothing that breaks inside push's transaction). Nothing destructive.
- **[confirmed]** The recorded SQL matches what production actually got. `supabase db push` keys on the version prefix alone and never compares content, so a file amended after its dispatch run would leave `main` recording SQL production doesn't have, permanently and invisibly. Checked by hand: this file's blob is `cabdcca`, identical to `refs/pull/438/head`, which currently resolves to `64ffb9d8` — the exact `head_sha` of the successful `workflow_dispatch` run at 2026-08-10T02:10Z. No drift.
- **[confirmed]** Fully idempotent if it ever does re-run: `CREATE TABLE IF NOT EXISTS`, `DROP TRIGGER IF EXISTS` before each `CREATE TRIGGER`, `CREATE INDEX IF NOT EXISTS`, `DROP POLICY IF EXISTS` before each `CREATE POLICY`, `ENABLE ROW LEVEL SECURITY` (a no-op when already on). One caveat worth knowing rather than fixing: because the CHECKs and the UNIQUE live inside `CREATE TABLE IF NOT EXISTS`, a database where these tables already exist in a *different* shape would be silently accepted rather than corrected.
- **[plausible]** `supabase/migrations/20260809120000_bandcamp-collection.sql:84` interacts badly with the admin release merge. `mergeReleases` (`api/functions/db.ts:2654`) deletes the dropped release row after explicitly repointing `release_sources` to `keepId` — but `collection_items.release_id` is `ON DELETE SET NULL`, so merging two duplicate releases will silently unmatch every collection item pointing at the loser instead of repointing it, downgrading real purchases to unmatched denormalised rows with no way to notice. Add the same `.update({ release_id: keepId }).eq('release_id', dropId)` for `collection_items` alongside the `release_sources` one, and an index on `collection_items(release_id)` so the FK check isn't a seq scan.
- **[plausible]** line 120 — the comment is correct that Postgres treats NULLs as distinct in a UNIQUE constraint (that's the default; `NULLS NOT DISTINCT` would have to be asked for), so manual items genuinely never collide. But nothing enforces the converse: `external_id` is nullable for *all* sources, so a single Bandcamp or Discogs item that imports without one silently opts out of the upsert key and gets duplicated on every subsequent sync — the exact failure the constraint exists to prevent, and one that grows a row per sync rather than erroring. `CHECK (source = 'manual' OR external_id IS NOT NULL)` closes it.
- **[confirmed]** The prose checks out. `api/functions/notifications.ts:313,328` does query and write `alert_sent_at`, and `20260810120000` adds the column with `DEFAULT now()` before dropping the default — so the "alerts broken for ~30 hours" account and the backfill-avoids-an-alert-storm design are both accurate. One addition worth making: this is at least the second occurrence, not a one-off. `20260808160000_scope-service-policies-to-service-role.sql`'s own header records the same failure mode on 2026-08-08 (`notification_preferences` dispatched from an unmerged branch, forcing that file to be renamed off its original prefix), and the run history shows two `main` failures on 2026-08-08 between that dispatch and its merge. A pattern that has now cost a rename and a 30-hour outage probably wants a CI step that compares local `supabase/migrations/` against the remote history and fails the *PR*, rather than a comment that has to be read at the right moment.

---

PR #447, opened 2026-08-14
Fix magic-link sign-in in the Mac app and browser extension

brandonlucasgreen:
Mac app: sendMagicLink() never sent an OTP email at all — it called
getOAuthSignInURL(provider: .email, ...), Supabase's OAuth-authorize flow
for a provider that doesn't exist. Replaced it with the real
signInWithOTP(email:redirectTo:) call, wired the unstream:// URL scheme
(already registered in both Info.plists) to AuthService.handleAuthCallback
via onOpenURL on both platforms, and dropped the now-unused
ASWebAuthenticationSession plumbing.

Extension: signInWithOtp() fires correctly, but the flow had no way to
complete — nothing ever sent the AUTH_MAGIC_LINK_CALLBACK message the
service worker already listens for, and the OTP request itself dropped
redirectTo. Now the popup requests a link to /login (the same page the
web app uses), and a new content script (content/auth-bridge.js) matching
that page forwards the resulting access_token to the background worker,
which stores the session for the extension to pick up.

### DIFF
--- a/apps/extension/content/auth-bridge.js
+++ b/apps/extension/content/auth-bridge.js
@@ -1,0 +1,17 @@
+// Bridges a magic-link sign-in completed on unstream.stream back into the extension.
+// The extension can't receive the emailed link directly (it opens in whatever tab/app
+// handles the user's email), so signInWithOtp() points redirectTo at /login — the same
+// page the web app itself uses. When that page loads with an access_token in the
+// redirect hash, this script forwards the callback URL to the background service worker,
+// which stores the session (see AUTH_MAGIC_LINK_CALLBACK in background/service-worker.js).
+// Runs at document_start so it reads the hash before the web app's own auth handling
+// strips it from the URL.
+if (window.location.hash.includes('access_token')) {
+  try {
+    if (chrome.runtime?.id) {
+      chrome.runtime.sendMessage({ type: 'AUTH_MAGIC_LINK_CALLBACK', url: window.location.href });
+    }
+  } catch {
+    // Extension context invalidated — ignore
+  }
+}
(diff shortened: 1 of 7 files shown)


brandonlucasgreen (commented):
Security + regression pass on this diff (auth/authz, input validation, secret handling, injection, unsafe defaults, data exposure; then changed assumptions, edge cases, state transitions, API contract drift). Three findings, all inline:

1. **[auth/authorization — plausible]** `AUTH_MAGIC_LINK_CALLBACK` has no sender/origin validation on either end (`service-worker.js`'s message handler, or `handleMagicLinkCallback` in `lib/supabase.js`). This PR is what makes that path reachable for the first time. Not independently exploitable today (no `externally_connectable`, so only the extension's own content-script contexts can send it), but there's no defense-in-depth check. See inline comment on `content/auth-bridge.js`.

2. **[auth/authorization, macOS — needs-manual-verification]** Dropping `ASWebAuthenticationSession` removes the mechanism a previous comment says existed specifically to close a custom-URL-scheme-interception risk via PKCE. Whether the replacement (`signInWithOTP` + bare `unstream://` `onOpenURL`) is equally safe depends on supabase-swift's default `flowType`, which isn't set explicitly here. See inline comment on `AuthService.swift`.

3. **[regression, state transitions — plausible]** `cancelPendingAuth()` no longer cancels the underlying request Task, unlike the code it replaced. See inline comment.

No injection, secret-handling, or data-exposure issues found — the anon-key/RLS model is unchanged. Input validation on the extension's `signInWithOtp(email, redirectTo)` call and the Mac app's email field is unchanged from before this PR.

@brandonlucasgreen

---

PR #448, opened 2026-08-14
Retry Bandcamp's Subsonic 500s during import, and unstick killed syncs

brandonlucasgreen:
Users hit "SubsonicError: getAlbumList2: HTTP 500" partway through the
Bandcamp collection import. The call is the only one that asks Bandcamp for
500 records at once, and it fails for some users while completing for others
— which reads as the beta timing out while building a big page, not as an
answer about the collection. It was believed on the first try and the whole
sync was recorded as failed.

- A 5xx, a 429, a dropped connection or a timeout is now flagged retryable on
  SubsonicError and retried (1s, 4s) before it is believed. A rejected
  credential, a non-JSON body or a Subsonic error envelope is the server's
  actual answer and is still fatal on the first try. Only the collection
  fetch retries; connect stays on the request path where waiting is worse.
- A page that still fails is re-requested smaller from the same offset
  (500 → 250 → 125 → 100), then the sync fails honestly. Pagination now
  advances by rows returned rather than page number, so a shrunken page
  can't leave a gap, and the runaway backstop counts rows consumed.
- The failure names the page it died on ("offset 500, size 500"). Offsets
  aren't sensitive; the URL carrying the credential is still never included.

Also fixes the state a killed sync leaves behind. Netlify kills a background
function at 15 minutes and it dies without reaching the catch block, so the
row claims 'syncing' forever — the settings page hides Re-sync while a sync
runs and the resync endpoint answers 409, stranding the user on "Importing…"
with no way to retry. A row that has claimed 'syncing' for over 20 minutes is
now reported as a retryable error, and Re-sync is accepted instead of refused.

Claude-Session: https://claude.ai/code/session_01SjqyY2Ttd5q8wiSncMwnZW

### DIFF
--- a/api/functions/__tests__/bandcamp-subsonic.test.ts
+++ b/api/functions/__tests__/bandcamp-subsonic.test.ts
@@ -19,6 +19,26 @@ function okEnvelope(extra: Record<string, unknown> = {}) {
   } as Response;
 }
 
+/**
+ * Runs a call that may back off between retries without waiting out the backoff. The
+ * rejection is captured before timers advance so it never surfaces as an unhandled one.
+ */
+async function runWithTimers<T>(run: () => Promise<T>): Promise<T> {
+  vi.useFakeTimers();
+  try {
+    const settled = run().then(
+      value => ({ ok: true as const, value }),
+      error => ({ ok: false as const, error })
+    );
+    await vi.runAllTimersAsync();
+    const result = await settled;
+    if (!result.ok) throw result.error;
+    return result.value;
+  } finally {
+    vi.useRealTimers();
+  }
+}
+
 describe('bandcamp-subsonic', () => {
   const fetchMock = vi.fn();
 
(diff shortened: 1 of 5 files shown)


brandonlucasgreen (commented):

**Security**

- Auth/authorization — confirmed fine. Every `me-bandcamp.ts` query is scoped by `.eq('user_id', user.userId)`; the resync/connect changes don't widen that.
- Input validation — confirmed fine. No new external input surfaces; offset/size in `withPageContext` come from internal loop state, not the client.
- Secret handling — confirmed fine. `withPageContext` only ever appends `offset`/`size`; the credential-bearing URL is still never included in an error, and the new tests explicitly assert `err.message` doesn't contain the token/salt (`api/functions/__tests__/bandcamp-subsonic.test.ts`, "names the page that failed, without leaking the credential").
- Injection risks — confirmed none. All fetch params are static/internal, no string concatenation of untrusted input.
- Unsafe defaults — confirmed fine. `SubsonicError.retryable` defaults to `false` (fail-closed); only 5xx/429/timeout are explicitly marked retryable.
- Data exposure — confirmed fine. The only new Sentry payload field is a boolean (`retryable`), no PII/credential material.

**Regression**

1. **plausible** — `api/functions/bandcamp-subsonic.ts`, `subsonicFetchAllAlbums`: once a page triggers the shrink-on-retry fallback, `pageSize` is never restored to `PAGE_SIZE` afterward. Every subsequent page for the rest of that sync keeps using the smaller size, so a single transient blip partway through a large collection permanently increases the request count against Bandcamp for the remainder of that sync (not just the one bad page). Worth deciding if that's intended or if `pageSize` should reset to `PAGE_SIZE` after a clean fetch.
2. **needs-manual-verification** — `me-bandcamp.ts` POST `resync`: the "already syncing" guard is read-then-write with no lock (`select sync_status/updated_at` → `update sync_status='syncing'`). Two concurrent resync requests for the same user (or a resync racing a stale-sync GET-triggered UI retry) can both pass the check and both dispatch `bandcamp-sync-background.ts`. This TOCTOU predates this PR for the non-stale case, but the new stale-sync bypass adds another window where it's reachable. Worth confirming two concurrent background syncs upserting into `collection_items` for the same user can't produce a corrupted `item_count`/partial-then-overwritten result.
3. **plausible** — worst case, a page can go through up to 4 size tiers (500→250→125→100) × up to 3 attempts each with 1s/4s backoff before giving up, repeated per failing page across up to 200 minimum-sized pages. During a real Bandcamp outage this trades the old fail-fast behavior for a much slower, much noisier failure that can run past the 15-minute background function budget (the new stale-sync detection at 20 min does recover the UI, but the sync itself dies silently killed rather than erroring cleanly). Might be worth an overall wall-clock budget in `subsonicFetchAllAlbums` in addition to the per-page retry/shrink logic.

No confirmed vulnerabilities found.

---

PR #449, opened 2026-08-14
Fix the blank page and broken Back when opening a release from a collection

brandonlucasgreen:
The collection grid linked releases with React Router's <Link>, but
/a/{artist}/{release} is rendered by the release-page edge function and has no
route in main.tsx. So the click pushed the URL and rendered nothing: a blank
page at the release's own URL. The release only appeared once the visitor
reloaded, which is the "takes a long time to load" half of the report, and
Back afterwards restored that release document under the profile's URL, which
is the half that needed a hard refresh.

Reproduced in Chromium against the built SPA: click -> empty body at
/a/kingtriumph/bias; reload -> release page; Back -> URL /u/brandon still
showing the release page. With a plain <a> the click loads the release
directly and Back returns to the profile.

Every other release link in the app (ReleasesSection, RecentReleasesSection,
AdminReleaseReviewPage) was already a plain anchor for this reason; the
collection grid was the one that slipped. The artist link beside it stays a
<Link> because /a/{artist} really is an SPA route.

The new test clicks both links and asserts the release one is not cancelled by
the router, so a <Link> can't come back unnoticed.

Claude-Session: https://claude.ai/code/session_01HWjFj2puBf3jhnX3uNHUMF

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -302,6 +302,35 @@ Artists claim profiles via `/claim/:slug` (email magic link or password auth, wi
 
 Signed-in fans can save artists and mark artists as supported (migrations 013–018), synced to the Apple app via `saved-artists-sync.ts` with tombstones and scheduled GC. Users can claim a public username (migration 021) and opt into sharing their list (migration 022); the public list renders at `/u/:handle` via the `u-handle` edge function, backed by `public-saved-artists.ts`. Reserved handles live in `api/lib/reserved-handles.ts`.
 
+### Collections, and why most items start unlinked
+
+A Bandcamp import (`bandcamp-sync-background.ts`) writes a `collection_items` row per album and
+attaches an Unstream release only when one already exists. Most of a real collection therefore
+arrives unlinked — the fan bought from artists nobody has ever searched, so there is no `artists`
+row to match against.
+
+**There is no source URL to fall back to.** Bandcamp's Subsonic API returns `id`, `name`,
+`artist`, `coverArt`, `year`, `genre`, `created` and nothing else — no album or artist URL — so
+"just link to Bandcamp" is not available, and deriving `<artist>.bandcamp.com/album/<title>` mints
+404s. Don't reach for it.
+
+`collection-matching.ts` closes the gap in two halves that become possible at different times:
+
+- `resolveCollectionArtists(userId)` runs at the end of a sync, *after* the connection is marked
+  idle so a discovery failure is never recorded as a failed import. It probes Bandcamp for each
+  unknown artist name (cached forever, negatives included), stores the artist plus their Bandcamp
+  link, and requests catalogues for up to 25 — matching `MAX_ARTISTS_PER_REQUEST`, which silently
+  slices anything longer. The rest ride the six-hourly sweep, whose pool is every artist with a
+  catalogue-able link.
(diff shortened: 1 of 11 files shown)


brandonlucasgreen (commented):

**Security**
- **Auth/authorization** — no change to entry points. `resolveCollectionArtists`/`linkCollectionItemsForArtist` are only invoked from `bandcamp-sync-background.ts` and `catalog-artist-background.ts`, both already gated behind their existing internal-secret checks; no new externally-reachable trigger is introduced. `needs-manual-verification` — the diff shown doesn't include the auth wrapper itself, just the new call sites.
- **Input validation** — `escapeLikePattern()` correctly escapes `%`, `_`, and `\` before the Bandcamp-supplied `artist_name` is used in an `ilike` filter, preventing wildcard injection into the pattern. Good defensive handling of external input. No finding.
- **Secret handling** — untouched by this PR.
- **Injection risks** — every new query goes through the Supabase query builder (parameterized), no raw SQL concatenation. For `CollectionGrid.tsx`'s `<Link>` → `<a href={item.releaseUrl}>` change: confirmed `releaseUrl` is always server-built as `` `/a/${artist}/${release.slug}` `` (`api/functions/collection-utils.ts:68`) from slug functions that only ever emit `[a-z0-9-]` or a content-hash fallback — never raw Bandcamp/user text — so this isn't an open-redirect or `javascript:`-URI vector. No finding.
- **Unsafe defaults** — the `artist_links` upsert uses `ignoreDuplicates: true` on conflict, so a lower-trust probe-derived link can never overwrite one that came from search/claim. Correct direction. No finding.
- **Data exposure** — `linkCollectionItemsForArtist` matches by `artist_name` across *all* users' `collection_items` by design (shared release catalog); it only ever attaches a public `release_id` to a user's own row and never reads/returns another user's data back to a caller. No finding.

**Regression**
- **Changed assumptions** — `matchLibrary()` now keys on `releaseMatchKey()` instead of `normalizeForComparison()`. Read both implementations (`api/functions/release-utils.ts:29`, `api/functions/search-utils.ts:467`): for ordinary ASCII/Latin titles they produce equivalent keys, so existing matches aren't disturbed — the change only fixes titles that previously collapsed to the empty string (and could over-merge). No regression found in the common path.
- **Edge cases / null/empty** — the deadline-exit path in `resolveCollectionArtists` (`toResolve.length - index`) correctly counts the current + remaining names as deferred; `readAllPages` failure returns an empty summary rather than throwing, matching the repo's "never cache uncertainty" rule.
- **Off-by-one** — none found.
- **State transitions** — artist resolution deliberately runs *after* `sync_status` is set to `idle`, wrapped in try/catch that reports to Sentry without failing the import — verified against the two new tests in `bandcamp-sync.test.ts` that pin this ordering.
- **API contract drift** — `bandcamp-sync-background`'s response gains a `resolved` field; additive only, no break for existing consumers.
- **Plausible, low severity** — `MAX_CATALOG_REQUESTS` in `collection-matching.ts` (25) must stay manually in sync with `MAX_ARTISTS_PER_REQUEST` in `request-catalog.ts` (currently also 25, confirmed). It's enforced only by a comment, not a shared constant or assertion — if one changes without the other, catalog requests get silently truncated below the intended count, which is exactly the "no silent caps" failure mode this same file warns against elsewhere. Worth importing the shared constant or asserting equality.

No confirmed vulnerabilities found in this PR.

---

PR #450, opened 2026-08-14
Resolve collection artists even when Bandcamp refuses the collection fetch

brandonlucasgreen:
## What this fixes

The artist-discovery pass added in #449 sat inside the sync's success path, which coupled it to something it has nothing to do with. It reads `collection_items` that are already stored and probes `<slug>.bandcamp.com/music` — neither the Subsonic API nor the stored credential is involved — so whether the collection fetch succeeded has no bearing on whether the artists behind an earlier import can be found.

That coupling bit the same day it shipped. Bandcamp answered `HTTP 500` to `getAlbumList2` at offset 0, so the sync failed and discovery never ran for items imported days earlier. In practice that meant a Re-sync accomplished nothing: the unlinked releases on a profile stayed unlinked.

## What the Sentry event actually says

`getAlbumList2: HTTP 500 (offset 0, size 100)` means #448's shrink ladder ran all the way to the floor — page sizes 500 → 250 → 125 → 100, each with one attempt plus two retries. Twelve requests, all failing at offset 0.

So the hypothesis behind that ladder (Bandcamp timing out while building a large page) does not explain this event: a 100-album page at the very start fails just as hard. This was the Subsonic endpoint declining to answer at all. Nothing to fix on our side for that part — the retry and shrink behaviour from #448 is untouched here.

## The change

Resolution now runs on both the success and the failure path, after the connection row is written either way, and still cannot fail the invocation. Moved into `resolveArtistsSafely()` so both paths share one behaviour and one Sentry report.

A Bandcamp-wide outage makes the probes inconclusive rather than wrong: the probe cache refuses to record an `undecided` verdict, so the worst case is a wasted pass, never a stored "this artist isn't on Bandcamp".

## Testing

`npm run test:api` (1,217 passing) and `npm run typecheck:api`.

New test drives the exact production failure — a retryable `SubsonicError` from `getAlbumList2` — and asserts the sync still reports failure and marks the connection `error`, while resolution runs anyway. The existing test covering "resolution fails, import still succeeds" is unchanged.

## Not included

Two things noticed while diagnosing, deliberately left out of this PR:

- Production Sentry events are tagged `environment: development`. `api/lib/sentry.ts` falls back to `NODE_ENV || 'development'` and `SENTRY_ENV` isn't set on the Functions environment — a config change, not a code one.
- Now that shrinking is known not to help, the ladder spends twelve requests when the *first* page fails at full size. Shrinking only makes sense mid-collection; a failure at offset 0 reads as "Bandcamp is down". Worth revisiting if this recurs.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HWjFj2puBf3jhnX3uNHUMF)_

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -316,12 +316,14 @@ row to match against.
 
 `collection-matching.ts` closes the gap in two halves that become possible at different times:
 
-- `resolveCollectionArtists(userId)` runs at the end of a sync, *after* the connection is marked
-  idle so a discovery failure is never recorded as a failed import. It probes Bandcamp for each
-  unknown artist name (cached forever, negatives included), stores the artist plus their Bandcamp
-  link, and requests catalogues for up to 25 — matching `MAX_ARTISTS_PER_REQUEST`, which silently
-  slices anything longer. The rest ride the six-hourly sweep, whose pool is every artist with a
-  catalogue-able link.
+- `resolveCollectionArtists(userId)` runs at the end of a sync, on **both** the success and the
+  failure path, and after the connection row is written either way — it reads stored
+  `collection_items` and probes `<slug>.bandcamp.com/music`, touching neither the Subsonic API
+  nor the credential, so a Subsonic 500 (routine in this beta) must not block discovery for
+  items imported days earlier. It stores each artist plus their Bandcamp link and requests
+  catalogues for up to 25 — matching `MAX_ARTISTS_PER_REQUEST`, which silently slices anything
+  longer. The rest ride the six-hourly sweep, whose pool is every artist with a catalogue-able
+  link.
 - `linkCollectionItemsForArtist(artistId, name)` runs at the end of every catalogue pass in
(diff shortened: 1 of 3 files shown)


brandonlucasgreen (commented):
Reviewed for security (auth/authz, input validation, secret handling, injection, unsafe defaults, data exposure) and regression risk (changed assumptions, edge cases, off-by-one, state transitions, API contract drift).

**Security:** No issues found. Internal-secret auth (`isInternalRequest`) is untouched, the credential-decrypt-failure path continues to deliberately withhold the decrypted-credential-in-error-message risk from Sentry (preserved from before this PR), and no new external input or outbound fetch is introduced.

**Regression:** One confirmed finding, inline below — the undecryptable-credential failure path bypasses the new unconditional `resolveArtistsSafely()` call, contradicting the invariant this PR states in both the description and the function's own doc comment.

@brandonlucasgreen

---

PR #451, opened 2026-08-14
Skip deploys that can't change the site, and move the test gate to CI

brandonlucasgreen:
Two changes aimed at the same thing: fewer and cheaper Netlify production deploys. A deploy costs a flat 15 credits on Netlify's credit-based plans and build minutes on the legacy plans, and at ~140 pushes to `main` a month that is the dominant cost.

## Skip deploys for pushes that can't change the site

`netlify.toml` now runs `scripts/netlify-ignore-build.sh` as its `ignore` command. It cancels the build when a push touches only paths Netlify never publishes: the Apple app, the browser extension, migrations, docs, workflows, `README.md` and `CLAUDE.md`. Four of the last fifty commits on `main` qualified — an Apple-app bugfix and a migration file each bought a full production deploy of the website.

Netlify inverts the exit code here — **0 cancels the build, 1 runs it** — so every branch in the script defaults to deploying. A skipped deploy leaves production silently stale, which is worse than a wasted build.

Verified against real commits from history, including the three cases that must *not* skip:

| Case | Expected | Result |
|---|---|---|
| README/CLAUDE.md-only commit | skip | skip |
| Apple-app-only commit (#422) | skip | skip |
| Latest web commit | build | build |
| Range spanning many commits | build | build |
| No cached commit (cleared cache) | build | build |
| Same ref (manual redeploy) | build | build |
| Unknown/unreachable ref | build | build |

`data/` and `scripts/` are deliberately absent from the ignore list: the whole `data/` tree is copied into `dist/` by the `copy-data-to-dist` plugin in `apps/web/vite.config.ts`, and `scripts/` generates the manifests, feeds and sitemap. `docs/` is safe only because the OpenAPI spec the site links to is a separate copy at `apps/web/public/docs/openapi.yaml`.

## Move the test suites out of the build and into CI

The two suites move from `npm run build` into `.github/workflows/ci.yml`, which runs them on every PR and every push to `main`. Actions minutes are free on this public repo; a Netlify deploy is not.

The build drops from **57.5s to 24.5s** and now does only what a deploy needs — generators, typecheck, `vite build`, sitemap. A type error still can't ship.

`npm run verify` runs the same gate locally (typecheck + both suites, ~45s, all 1,961 tests passing). Use it instead of `npm run build` to check work is done.

### The trade-off, stated plainly

**A red CI run no longer blocks a deploy by itself.** Netlify builds whatever lands on `main`, so the tests are now a pre-merge gate rather than a pre-deploy one. Enabling branch protection on `main` with the `verify` check required is what restores the old guarantee. Until then, merging a red PR deploys it.

Lint is deliberately not in the gate: ESLint reports 64 pre-existing errors (unused vars and `any` in test files, plus a few pages), so enforcing it now would fail every PR for reasons unrelated to the change. Worth clearing separately.

## Notes

- `CLAUDE.md` is updated for all of the above — the build pipeline description, the testing section, and the workflow list.
- This PR itself will build when merged: `netlify.toml`, `package.json` and `scripts/` are all outside the ignore list, which is correct since they change how the site is built.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FdzkzTnjdXDBWXHnqgSouY)_


brandonlucasgreen (commented):

## Security

- **Injection — confirmed none.** `scripts/netlify-ignore-build.sh` quotes `$CACHED_COMMIT_REF`/`$COMMIT_REF` everywhere they're used (`git diff --name-only "$CACHED_COMMIT_REF" "$COMMIT_REF"`), and the prefix-match `case "$file" in "$prefix"*)` loop only ever matches against fixed literal strings from `IGNORED_PREFIXES` — no attacker-influenced pattern or unquoted expansion reaches the shell.
- **Secret handling — N/A.** No secrets touched.
- **Auth/authorization — N/A.** Build tooling only.
- **Unsafe defaults — confirmed correct.** Every branch of the ignore script (no cached ref, cache==current ref, `git diff` failure, unreadable diff) defaults to `exit 1` ("build"), matching the script's own stated philosophy that an unproven-safe push must deploy. I didn't find a path where an uncertain state resolves to "skip."
- **Least privilege — plausible, worth fixing.** `.github/workflows/ci.yml` sets no `permissions:` block, so it runs with the default `GITHUB_TOKEN` scope. This repo already has a live convention for scoping workflow tokens down — `.github/workflows/semantic-revert-check.yml` explicitly declares `permissions: pull-requests: write` (checked against `main`). `ci.yml` only needs to check out code and run tests, so it should declare `permissions: contents: read` explicitly rather than inheriting whatever the org/repo default is. Low severity on its own (this is `pull_request`, not `pull_request_target`, so fork PRs still can't reach secrets), but `npm ci` on every PR does execute third-party install scripts, and there's no reason for that job to hold a broader-than-read token if the default happens to be permissive.

## Regression

- **Confirmed false positive on the semantic-revert-check bot's own warning.** It flagged this PR for deleting a `package.json` line added in #429. I diffed against `main`'s current `package.json`: the "deleted" line is just the old single-line `build` script; the replacement `build` line still runs `generate-changelog-feed.ts` and `generate-guides-feed.ts` before `vite build` — #429's actual contribution survives, just reformatted across the new `typecheck`/`verify`/`build` scripts. Nothing from #429 was reverted.
- **Changed assumption, needs-manual-verification.** The PR body is explicit that "a red CI run no longer blocks a deploy by itself" until the `verify` check is required in `main`'s branch protection rules. That's the real safety net this PR is relying on to not regress deploy safety — worth confirming it actually gets turned on (before or right after merge), since until then a red PR merged to `main` deploys exactly as the PR itself warns.
- **Edge cases** — the ignore script's failure paths (missing cached ref, identical ref, failed diff) all match the PR's own verification table; no case found where it silently skips a build it shouldn't.

No blocking security findings. The one actionable item is the missing `permissions:` block on the new workflow, and the one confirmation-needed item is enabling the required-check branch protection the PR describes as the actual fix for the stated trade-off.

---

PR #452, opened 2026-08-15
Make local dev run the real backend, so previews aren't the only way to see it

brandonlucasgreen:
## Why

`npm run dev` was Vite alone: no Netlify functions, no edge functions, no Supabase — and no auth, because three keys sat **blank** in the local `.env`, shadowing the real values from the Netlify site settings. Anything touching data or a real API response had to be checked on a deploy preview or in production.

#451 has since disabled deploy previews outright. Its comment names the exit condition: previews should return *"once local dev can stand in for them."* This is that standing-in.

## What changed

- **`npm run dev` → `netlify dev` on :8888** — real `api/functions/`, real `api/edge/`, `netlify.toml` routing, production Supabase data **and auth**.
- **`npm run dev:fast`** — the old Vite-only mode on :5173, for CSS/layout work, now with `--strictPort`.
- **`[dev]` block in `netlify.toml`** pinning `targetPort`, with the reasoning inline.
- **CLAUDE.md** — a mode-comparison table, the production-write warning, and the two silent-failure traps below.

## Verified against production, not assumed

| Check | Result |
|---|---|
| `/a/808-state` under a Googlebot UA | **byte-identical** to production |
| Uncatalogued release `/a/{artist}/{release}` | 302s to the same target locally and in prod |
| `/api/suggest?query=big thief` | real `artists` row (the dev shim returns `[]`) |
| `/api/me/settings`, `/api/me/collection` | 401 as expected |
| Sign-in with bogus credentials | Supabase's own "Invalid login credentials" (was "Auth not configured") |

`npm run verify` exits 0 — 1217 API + 744 web tests. Lint unchanged at the pre-existing 71-problem baseline.

## Two traps this fixes, both of which produced a *silently wrong page* rather than an error

**A blank value in `.env` overrides the real one** from the Netlify site settings — `netlify dev` reports it as `Ignored project settings env var: X (defined in .env file)`. That is why local auth never worked. Fixed on my machine by deleting the three blank lines; CLAUDE.md now says to delete a key rather than leave it blank, and to read the startup log's `Injected`/`Ignored` lists when something is unconfigured.

**A stale dev server silently answers everything.** `netlify dev` proxies to `targetPort` regardless of what is actually there. A Vite left running two days earlier owned 5173, the Vite this run started moved to 5174, and every page served through :8888 was the stale process's build. `--strictPort` turns that into a startup failure — verified: a second Vite exits code 1 rather than drifting.

## Safety

Functions get `SUPABASE_SERVICE_KEY` and bypass RLS exactly as production does. Reads are free; **writes are real**, and CLAUDE.md now says so plainly. Cataloguing stays off: `RELEASE_CATALOG_ENABLED` is production-context only and `netlify dev` uses the dev context.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -27,7 +27,7 @@ These principles should inform product and engineering trade-offs, not just mark
 
 - **Frontend**: React 19 SPA with React Router 7, Tailwind CSS v4, Vite 7, TypeScript. PWA-enabled (`vite-plugin-pwa`). Most pages are lazy-loaded in `apps/web/src/main.tsx`.
 - **Backend**: Netlify Functions (serverless API, `api/functions/`) + Edge Functions (SSR/SEO, `api/edge/`). Functions are TypeScript on Node; edge functions run on Deno (see `deno.lock`).
-- **Local dev API**: a Vite middleware in `apps/web/server/` — a *separate* implementation from `api/functions/`. See "Local dev API vs production API" below; this trips people up.
+- **Local dev**: `npm run dev` runs `netlify dev` — the real functions and edge functions against production Supabase. `npm run dev:fast` is a Vite-only shortcut backed by a *separate* API implementation in `apps/web/server/`. See "Local dev: the full stack, and the fast shim" below; the difference trips people up.
 - **Database**: Supabase (Postgres with RLS) for artist profiles, analytics, merge overrides, API keys, verification requests, saved/supported artists, public usernames, and the Bandcamp probe cache.
 - **Data enrichment**: MusicBrainz, Wikidata, Wikipedia, Discogs, Linktree, Bandcamp artist pages.
 - **Caching / rate limiting**: Upstash Redis (`@upstash/ratelimit`, `api/functions/cache.ts`, `api/functions/ratelimit.ts`).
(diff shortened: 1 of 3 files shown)


brandonlucasgreen (commented):
@brandonlucasgreen Security review (automated pass)

**Security**
- **Secret handling / blast radius — plausible, already surfaced by this PR's own docs:** `npm run dev` now runs `netlify dev` by default (`package.json`), which injects production Supabase credentials (`SUPABASE_SERVICE_KEY`, RLS-bypassing) plus `RESEND_API_KEY` / `BUTTONDOWN_API_KEY` into every local dev environment, and local writes (saves, claims, settings changes) now hit production data and can send real email. Previously the default `npm run dev` (bare Vite) touched no functions or secrets at all — that's a real posture change, not just a docs update. The PR documents this extensively in `CLAUDE.md` ("It reads and writes PRODUCTION Supabase"), so it reads as a deliberate, disclosed trade-off rather than an oversight — flagging it here so it gets an explicit sign-off rather than landing silently as "just tooling."
- No injection, auth, or input-validation issues — the diff is config (`netlify.toml`, `package.json`) plus docs.

**Regression**
- `dev:fast` adds `--strictPort` (paired with the new `[dev]` block's `targetPort = 5173`), closing the stale-port failure mode the `netlify.toml` comment describes (a leftover Vite silently answering all traffic through :8888). No API or behavior contract changes.
- No other changed assumptions beyond the documented dev-workflow swap.

No blocking findings.

---

PR #454, opened 2026-08-16
Collection ≠ Saved Artists, and the gap endpoint that makes the library import useful

brandonlucasgreen:
Two related corrections from Brandon testing the live features, plus the server half of the Apple Music import.

## 1. Collection imports stop writing Saved Artists

> Collection Imports should always add to the collection NOT saved artists. However, you should get upcoming/recent release information from artists both in the collection and the saved artists list.

Saving is a deliberate act that also subscribes you to release alerts; buying a record belongs in the collection. Conflating them meant one Bandcamp import wrote **35 new saved-artist rows** on his account against 8 genuine hand-made marks.

- `markArtistsSupported` and its ~145 lines are gone. Collection artists still get artist rows and catalogues — `collection-matching.ts` handles that independently — so no release data is lost.
- `getFeedReleasesForUser` now unions saved ∪ collection, resolving collection artists exactly via a matched release's `artist_id` and otherwise by deriving the slug. **This is what makes the removal safe**: without it, dropping the write would have silently emptied the dashboard feed for anyone who had imported.

Blast radius checked: that function has exactly one consumer (the dashboard's Upcoming/Recent sections). Email alerts read `saved_artists` directly and are untouched, so a 96-artist import can't become an alert flood.

## 2. `/api/me/listening` — the gap

The Apple Music import (#439) reads a library and, until now, nothing downstream could use it. Per Brandon's "gap only" scope:

- **POST** stores per-artist play counts in `listening_signals`.
- **GET** derives the gap — artists you play a lot and have never supported — subtracting artists marked supported and artists you own a record by, comparing on derived slugs so "Sigur Rós" and "sigur ros" are one artist, and merging duplicates across sources.
- **DELETE** is the counterpart to the app's "Forget Imported Data".

Deliberately absent and tested for: it never writes `saved_artists`, and none of it can reach a public page — an Apple Music play, and even an iTunes purchase, is `owned`/`listened`, never `purchased`. A saved-but-not-supported artist stays **in** the gap: saving isn't paying.

## Verification

- 1,235 API tests pass; typecheck clean; lint at the pre-existing 71-problem baseline.
- 5 new tests for the feed union against a recording fake Supabase client (real query body, including paging), **mutation-tested** — stubbing the collection lookup fails three of them.
- 14 new tests for the listening endpoint, including the slug-matching and "saved ≠ supported" cases.
- The old auto-mark tests are replaced by their inverse: the import must write no `saved_artists` rows even when every artist matches.

Pairs with #439 (the Mac client that uploads to this endpoint). Independent — either can merge first.

### DIFF
--- a/api/functions/db.ts
+++ b/api/functions/db.ts
@@ -4707,8 +4707,60 @@ async function savedArtistIdsForUser(
 }
 
 /**
- * Every release worth putting in one fan's feed: across all their saved artists, dated within
- * the trailing window or still to come.
+ * The artists a fan owns a record by, resolved to artist ids.
+ *
+ * `collection_items` stores the artist as the source spells it, not a foreign key, so this
+ * resolves two ways: exactly, through a matched release's `artist_id`, and by deriving the
+ * slug with `artistSlug()` — the same key every other writer uses — for items that never
+ * matched a release. Unresolvable artists simply contribute nothing.
+ *
+ * Read failures return an empty list rather than null: the collection is an *addition* to the
+ * saved-artist feed, so a failure here should quietly narrow the feed, never blank it.
+ */
+async function collectionArtistIdsForUser(
+  client: NonNullable<ReturnType<typeof getClient>>,
+  userId: string
+): Promise<string[]> {
+  const read = await readAllPages<{ artist_name: string; releases: { artist_id: string } | null }>(
+    (from, to) =>
+      client
+        .from('collection_items')
+        .select('artist_name, releases!left (artist_id)')
+        .eq('user_id', userId)
+        .range(from, to),
+    'collection_items (feed)'
+  );
+  if (!read.ok) return [];
+
+  const ids = new Set<string>();
+  const slugs = new Set<string>();
+  for (const row of read.rows) {
+    if (row.releases?.artist_id) ids.add(row.releases.artist_id);
+    else if (row.artist_name) {
+      const slug = artistSlug(row.artist_name);
+      if (slug) slugs.add(slug);
+    }
+  }
+
+  const pending = [...slugs];
+  for (let i = 0; i < pending.length; i += 100) {
+    const { data, error } = await client
+      .from('artists')
+      .select('id')
+      .in('slug', pending.slice(i, i + 100));
+    if (error) {
+      console.error('[DB] collectionArtistIdsForUser slug resolution failed:', error.message);
+      break;
+    }
+    for (const row of (data as { id: string }[]) || []) ids.add(row.id);
+  }
+
+  return [...ids];
+}
+
+/**
+ * Every release worth putting in one fan's feed: across their saved artists *and* the artists
+ * in their collection, dated within the trailing window or still to come.
  *
  * Two exclusions match the alert path for the same reasons — hidden releases are suppressed by
  * an artist and must stay invisible, and a `needs_review` tier-3 fuzzy flag means we aren't sure
@@ -4719,8 +4771,13 @@ export async function getFeedReleasesForUser(userId: string, now: Date = new Dat
   if (!client) return [];
 
   try {
-    const artistIds = await savedArtistIdsForUser(client, userId);
-    if (artistIds === null) return [];
+    // Saved artists AND artists you own a record by. Buying somebody's album is at least as
+    // strong a signal that you want their next one as saving them is — but the two lists stay
+    // separate everywhere else, because saving is a deliberate act and an import is not.
+    const savedIds = await savedArtistIdsForUser(client, userId);
+    if (savedIds === null) return [];
+    const collectionIds = await collectionArtistIdsForUser(client, userId);
+    const artistIds = [...new Set([...savedIds, ...collectionIds])];
     if (artistIds.length === 0) return [];
 
     const { data, error } = await client
--- a/api/functions/__tests__/bandcamp-sync.test.ts
+++ b/api/functions/__tests__/bandcamp-sync.test.ts
@@ -213,152 +213,23 @@ describe('bandcamp-sync-background handler', () => {
     expect(byExternal['al-3'].release_id).toBeNull();
   });
 
-  describe('auto-mark supported (spec OQ6: buying is supporting)', () => {
-    const ciphertext = () => encryptCredential(JSON.stringify({ t: 'tok', s: 'salt' }));
-    const albums = [{ id: 'al-1', name: 'Illinois', artist: 'Sufjan Stevens' }];
-
-    function withArtists() {
-      mocks.mockReadAllPages.mockResolvedValue({ ok: true, rows: [SUFJAN] });
-      mocks.mockFetchAllAlbums.mockResolvedValue(albums);
-    }
-
-    it('saves a matched artist as supported when they have no saved row', async () => {
-      const { savedInserts, savedUpdates } = setupDb({
-        connectionRow: { bandcamp_username: 'fan', credential_ciphertext: ciphertext() },
-        savedRows: [],
-      });
-      withArtists();
-
-      await handler(internalEvent({ userId: 'user-1' }));
-
-      const inserted = savedInserts.flat();
-      expect(inserted).toHaveLength(1);
-      expect(inserted[0]).toMatchObject({
-        user_id: 'user-1',
-        artist_id: 'artist-1',
-        artist_slug: 'sufjan-stevens',
-        artist_name: 'Sufjan Stevens',
-        supported: true,
-      });
-      expect(inserted[0].supported_at).toBeTruthy();
-      expect(inserted[0].last_modified).toBeTruthy();
-      expect(savedUpdates).toHaveLength(0);
-    });
-
-    it('upgrades an existing unsupported row instead of inserting', async () => {
-      const { savedInserts, savedUpdates } = setupDb({
-        connectionRow: { bandcamp_username: 'fan', credential_ciphertext: ciphertext() },
-        // No artist_id: the legacy shape, where only the slug identifies the artist.
-        savedRows: [{ id: 'row-1', artist_slug: 'sufjan-stevens', supported: false, deleted: false }],
-      });
-      withArtists();
-
-      await handler(internalEvent({ userId: 'user-1' }));
-
-      expect(savedInserts).toHaveLength(0);
-      expect(savedUpdates).toHaveLength(1);
-      expect(savedUpdates[0].patch).toMatchObject({ supported: true });
-      expect(savedUpdates[0].column).toBe('id');
-      expect(savedUpdates[0].values).toEqual(['row-1']);
-    });
-
-    it('leaves an already-supported row untouched (original supported_at preserved)', async () => {
-      const { savedInserts, savedUpdates } = setupDb({
-        connectionRow: { bandcamp_username: 'fan', credential_ciphertext: ciphertext() },
-        savedRows: [{ id: 'row-1', artist_slug: 'sufjan-stevens', supported: true, deleted: false }],
-      });
-      withArtists();
-
-      await handler(internalEvent({ userId: 'user-1' }));
-
-      expect(savedInserts).toHaveLength(0);
-      expect(savedUpdates).toHaveLength(0);
-    });
-
-    it('never resurrects a tombstoned row — permanent dismissal sticks across re-syncs', async () => {
-      const { savedInserts, savedUpdates } = setupDb({
-        connectionRow: { bandcamp_username: 'fan', credential_ciphertext: ciphertext() },
-        savedRows: [{ id: 'row-1', artist_slug: 'sufjan-stevens', supported: false, deleted: true }],
-      });
-      withArtists();
-
-      await handler(internalEvent({ userId: 'user-1' }));
-
-      expect(savedInserts).toHaveLength(0);
-      expect(savedUpdates).toHaveLength(0);
-    });
-
-    // The two cases the slug-only lookup got wrong. A row saved from a search result is filed
-    // under a synthetic slug (`sufjanstevens`), which the canonical slug never equals, so only
-    // artist_id finds it — measured on production 2026-08-14, one import duplicated three of
-    // Brandon's saved artists this way.
-    it('upgrades a row saved under a synthetic slug rather than duplicating the artist', async () => {
-      const { savedInserts, savedUpdates } = setupDb({
-        connectionRow: { bandcamp_username: 'fan', credential_ciphertext: ciphertext() },
-        savedRows: [
-          { id: 'row-1', artist_id: 'artist-1', artist_slug: 'sufjanstevens', supported: false, deleted: false },
-        ],
-      });
-      withArtists();
-
-      await handler(internalEvent({ userId: 'user-1' }));
-
-      expect(savedInserts).toHaveLength(0);
-      expect(savedUpdates).toHaveLength(1);
-      expect(savedUpdates[0].column).toBe('id');
-      expect(savedUpdates[0].values).toEqual(['row-1']);
-    });
-
-    it('never resurrects a tombstone filed under a synthetic slug', async () => {
-      const { savedInserts, savedUpdates } = setupDb({
-        connectionRow: { bandcamp_username: 'fan', credential_ciphertext: ciphertext() },
-        savedRows: [
-          { id: 'row-1', artist_id: 'artist-1', artist_slug: 'sufjanstevens', supported: false, deleted: true },
-        ],
-      });
-      withArtists();
-
-      await handler(internalEvent({ userId: 'user-1' }));
-
-      expect(savedInserts).toHaveLength(0);
-      expect(savedUpdates).toHaveLength(0);
-    });
-
-    // The shape deduplicating an account leaves behind: the superseded row tombstoned, the
-    // canonical one live. The tombstone must not be read as "artist dismissed" — that would
-    // strand the live row unsupported through every future re-sync — and neither row may be
-    // duplicated by a fresh insert.
-    it('supports the live row when a tombstone for the same artist sits beside it', async () => {
-      const { savedInserts, savedUpdates } = setupDb({
-        connectionRow: { bandcamp_username: 'fan', credential_ciphertext: ciphertext() },
-        savedRows: [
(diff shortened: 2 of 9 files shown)


api/functions/db.ts
```diff
@@ -4719,8 +4771,13 @@ export async function getFeedReleasesForUser(userId: string, now: Date = new Dat
...
 
   try {
-    const artistIds = await savedArtistIdsForUser(client, userId);
-    if (artistIds === null) return [];
+    // Saved artists AND artists you own a record by. Buying somebody's album is at least as
+    // strong a signal that you want their next one as saving them is — but the two lists stay
+    // separate everywhere else, because saving is a deliberate act and an import is not.
+    const savedIds = await savedArtistIdsForUser(client, userId);
+    if (savedIds === null) return [];
+    const collectionIds = await collectionArtistIdsForUser(client, userId);
```
brandonlucasgreen: **Data exposure — confirmed.** `getFeedReleasesForUser` now unions saved-artist releases with collection-artist releases, and this function is also the backing read for `getFeedReleasesForHandle` (db.ts:4869), which powers the *public*, unauthenticated calendar feed at `/u/{handle}/releases.ics|.xml` (`feed-releases.ts`).

That public feed is gated only on `usernames.saved_artists_public` — a flag a user opts into believing it publishes their *saved artists* list, which this very PR is careful to keep a deliberate, opt-in act ("the two lists stay separate everywhere else, because saving is a deliberate act and an import is not," per the comment a few lines above). But downstream of this union, a fan's Bandcamp purchase history (`collection_items`, populated by an unattended background sync — the whole point of this PR is that it's *not* something the user chose to publish) now leaks into that same public feed via artist names/release titles, with no separate opt-in.

`public-saved-artists.ts` (the HTML list at `/u/{handle}`) is unaffected — it doesn't call `getFeedReleasesForUser`. So the exposure is specifically the calendar feed, but it's real: anyone who knows/guesses a handle and has sharing enabled for their saved list is now also disclosing what they own via Bandcamp.

Suggest either: (a) keep `getFeedReleasesForHandle` on saved-artists-only (call `savedArtistIdsForUser` directly rather than through the now-merged `getFeedReleasesForUser`), or (b) add a distinct opt-in for collection-derived releases in the public feed. cc @brandonlucasgreen — this feels like a product call on whether collection data should ever be publicly shareable at all.

---

brandonlucasgreen (commented):

**Security**
- Auth/authorization: `bandcamp-sync-background.ts` handler is unchanged and still gated by `isInternalRequest` — fine. `getFeedReleasesForUser`/`getFeedReleasesForHandle` callers are unchanged auth paths. — confirmed, no issue.
- Data exposure: see inline comment on `api/functions/db.ts:4779` — **confirmed**. Merging collection-derived artists into `getFeedReleasesForUser` also flows into the public `/u/{handle}` calendar feed, which is gated on a flag intended for the saved-artists list only, not collection/purchase data. Worth a product decision before merge.
- Input validation / injection: all new reads use parameterized Supabase `.eq`/`.in` filters, no raw SQL or string interpolation into queries. `artistSlug()` (the correct canonical-slug function, not the lossy `normalizeForComparison`) is reused as-is. — confirmed, no issue.
- Secret handling: no secrets touched by this diff. — confirmed, no issue.
- Unsafe defaults: `collectionArtistIdsForUser` fails safe — a `collection_items` read error returns `[]` and only narrows the feed rather than blanking it, consistent with "never cache uncertainty." — confirmed, no issue.

**Regression**
- Changed assumptions: `MatchedArtist` was trimmed to `{ id }` (dropped `slug`/`name`/`imageUrl`) now that `markArtistsSupported` is deleted — grepped for remaining consumers of those fields in `bandcamp-sync-background.ts`/`collection-matching.ts`, none found. — confirmed, no issue.
- API contract drift: background-function response field renamed `artistsMarked` → `artistsMatched`; grepped the repo, nothing else reads that field (background-function body isn't consumed by the client). — confirmed, no issue.
- Edge cases: dedup between saved/collection artist ids before the `releases` query is covered by a test asserting no duplicate ids in the `IN` list; a `collection_items` read failure is covered by a test asserting saved-artist releases still return. — confirmed, no issue.
- State transitions: removing the auto-mark-supported write is a forward-only behavior change (existing rows already marked stay marked) — matches the stated intent, not flagged as a bug.

Everything above is code-grounded; the one open item is the inline comment (data exposure via the public feed) — recommend resolving that before merge.

---

brandonlucasgreen:
@brandonlucasgreen

## Security

- **confirmed** — `api/functions/me-listening.ts` (whole file): the endpoint reads, writes, and deletes rows in a `listening_signals` table for which **no Supabase migration exists anywhere in the repo**. I checked every file in `supabase/migrations/` (latest is `20260814170000_dedupe-saved-artists.sql`) and the only companion PR (#439) touches only `apps/mac/**` — no SQL. Per the project's RLS convention ("new tables ship RLS policies in a migration"), there is currently no RLS policy for this table because the table doesn't exist yet. Until a migration lands (here or separately), this isn't just a functionality gap — it also means the security posture of a table holding per-user listening history (a form of behavioral PII) is completely unreviewed. Recommend adding the migration to this PR (or confirming one was already `workflow_dispatch`-applied to production per the CLAUDE.md migration-drift warning) so RLS can actually be reviewed before merge.
- No other issues found: auth on all three methods (`POST`/`GET`/`DELETE`) correctly derives `userId` from the verified Supabase JWT and every query is scoped with `.eq('user_id', user.userId)` — no cross-user access path. `parseSignals` bounds array length (≤10,000), caps `artistName` at 300 chars, and clamps `playCount` to a non-negative integer, so the endpoint isn't an obvious injection/DoS vector. Wildcard CORS on `me-listening.ts` matches the existing pattern in `me-settings.ts`/`me-username.ts`/`me-password.ts`, where the bearer token (not origin) is the actual authorization boundary — consistent with convention, not a new hole.

## Regression

- **confirmed** — `api/functions/me-listening.ts`: as a direct consequence of the missing migration above, every call to `/api/me/listening` will fail at runtime with a Postgres "relation does not exist" error until that migration is applied. This PR ships its own new endpoint in a broken state.
- **confirmed** — `api/functions/me-listening.ts:203-209` (GET handler): reads of `saved_artists` and `collection_items` used to compute "already covered" artists are gated only by `if (savedRead.ok)` / `if (collectionRead.ok)`, with no error surfaced on failure — a transient read failure silently makes `covered` empty rather than failing the request. That means an artist the user already marked `supported` (or owns via their collection) can reappear in the "gap" list as if never supported, which is the opposite failure mode from `collectionArtistIdsForUser` in `api/functions/db.ts` (added in the same PR), which deliberately narrows rather than fabricates data on the same kind of failure — the two halves of this PR disagree on how to fail.
- **needs-manual-verification** — `api/functions/bandcamp-sync-background.ts`: removing `markArtistsSupported` stops *future* Bandcamp syncs from auto-marking artists supported, but does nothing to un-mark the `saved_artists` rows that earlier syncs already created (the PR body cites 35 such rows on Brandon's own account). No backfill/cleanup accompanies the reversal, so existing affected users keep the unwanted saved rows unless cleaned up separately.

---

brandonlucasgreen:

**Note on base branch:** this PR targets `batch/aug-16`, not `main`; diffed cleanly against that ref, no merge-conflict artifacts observed. Also confirmed `listening_signals` (RLS: `SELECT`/`DELETE` scoped to `auth.uid() = user_id`, no `INSERT`/`UPDATE` policies) was already created by migration `20260809120000_bandcamp-collection.sql`, present on `batch/aug-16` before this PR — so despite `me-listening.ts` being a brand-new endpoint, this PR correctly needs no migration of its own, and "new table without RLS" doesn't apply here.

### Security
- **confirmed** — `api/functions/me-listening.ts:71,135`: `parseSignals` accepts any string for `lastPlayed` with no date-format validation (`typeof record.lastPlayed === 'string' ? record.lastPlayed : null`), then writes it straight to `last_played` (a `timestamptz` column) in the POST upsert. A single malformed `lastPlayed` value anywhere in a chunk of up to 500 signals (`UPSERT_CHUNK`) throws a Postgres type-cast error for the whole chunk, silently discarding up to 499 otherwise-valid signals along with the bad one, surfaced only as a generic 500.
- No IDOR/auth-bypass found: `userId` is derived exclusively from `auth.getUser(token)` via the bearer token, never client-supplied, and every query (POST/GET/DELETE) is scoped with `.eq('user_id', user.userId)`. The wildcard `Access-Control-Allow-Origin: '*'` is pre-existing CORS debt matching `me-settings.ts`/`saved-artists.ts`, not newly introduced here.

### Regression
- **confirmed** — `api/functions/db.ts:4769-4803` (`getFeedReleasesForUser`) now unions saved + collection artist ids, and `api/functions/feed-releases.ts` (unmodified by this PR) also calls `getFeedReleasesForUser(userId)` for the private per-fan `.ics`/`.xml` calendar feed at `/feed/f/{token}`. The PR description states this function "has exactly one consumer (the dashboard's Upcoming/Recent sections)" — that's incorrect; every fan's already-subscribed calendar app will silently start surfacing releases from every artist in their Bandcamp-imported collection too, a live external-integration behavior change the stated blast-radius check missed.
- **plausible** — `api/functions/me-listening.ts:203-209`: on GET, if the `saved_artists` or `collection_items` read fails, the `covered` set is simply left without those rows rather than the request failing or being retried — unlike the `signalsRead` failure a few lines above, which does return 500. Concrete scenario: a transient Supabase hiccup while reading `saved_artists` mid-request drops an artist marked `supported: true` from `covered`, so they appear in "the gap" as if never supported — the same "don't treat a failed read as a negative result" bug class this codebase explicitly calls out as its most-repeated one.
- **plausible** — `api/functions/me-listening.ts:135,142`: `listening_signals` upserts key on exact-match `(user_id, source, artist_name)`; `parseSignals`'s duplicate-collapse is case-insensitive but only within a single upload batch. If the same artist is re-synced from the same source with different casing across two separate calls (the PR's own test data uses "Sigur Rós" vs "sigur ros" for the same artist), a second row gets inserted instead of the first being updated, and the GET handler's slug-keyed `totals` map sums `play_count` across both rows — inflating the reported play count instead of reflecting the current one.
- **needs-manual-verification** — same file: there's no reconciliation path for artists dropped from a later full-library re-sync (only additive `upsert` or a full `DELETE`-all). If the paired Mac client (#439, not in this diff) re-uploads the entire current library on each sync rather than a delta, artists removed from the library would keep contributing stale rows to "the gap" indefinitely — can't be resolved from this diff alone since it depends on #439's actual call pattern.

**Bottom line:** auth/IDOR posture is solid, but the feed-consumer blast-radius claim in the description is wrong (confirmed regression to the calendar feed) and the `lastPlayed` input isn't validated before hitting a typed column.

---

brandonlucasgreen (commented):

**Security:** No confirmed issues. `api/functions/me-listening.ts` derives `userId` exclusively from `supabase.auth.getUser(token)` and every query is scoped with `.eq('user_id', ...)` — no client-suppliable user-id parameter, no IDOR path found. Matches the established `me-settings.ts` auth pattern. Input validation (`parseSignals`) bounds batch size, trims/caps `artistName`, coerces `playCount`, and validates `source` against an allowlist matching the DB `CHECK` constraint. `listening_signals` RLS (owner-only SELECT/DELETE, no INSERT/UPDATE policy) was verified present via the earlier migration `20260809120000_bandcamp-collection.sql`.

- **plausible (low)** — `api/functions/me-listening.ts:71`: `lastPlayed` is validated only as `typeof string`, not as a parseable date. A malformed value would fail the Postgres `timestamptz` cast, surfacing as a 500 instead of a clean 400.
- **plausible (low, latency only)** — `me-listening.ts:168-200`: the three GET reads (`signalsRead`, `savedRead`, `collectionRead`) run sequentially rather than concurrently, unlike `me-settings.ts`'s pattern.

**Regression:** Checked every new/changed query against the CLAUDE.md 1000-row PostgREST cap — `collectionArtistIdsForUser`, and `me-listening.ts`'s three GET reads, all correctly use `readAllPages`. The `bandcamp-sync-background.ts` `artistsMarked`→`artistsMatched` rename has no other consumers (confirmed via repo-wide grep). Removing `markArtistsSupported` (Bandcamp import no longer writes `saved_artists`) is covered by a union in `getFeedReleasesForUser` plus new tests (`feed-collection-artists.test.ts`), including the collection-read-fails-narrows-not-blanks and empty/empty cases — confirmed intentional and tested. Overall a tight, well-tested PR.

---

brandonlucasgreen:
@brandonlucasgreen

Reviewed the diff (`api/functions/me-listening.ts`, `db.ts`, `bandcamp-sync-background.ts`, `netlify.toml`, `api/tsconfig.json`, and the three test files) against `batch/aug-16`.

## Security

No confirmed vulnerabilities. Two plausible items, both minor.

- **Authorization/scoping — confirmed correct.** `me-listening.ts` derives `user.userId` from `supabase.auth.getUser(token)` (real JWT verification against Supabase Auth, same pattern as `authenticateBearer` in `middleware.ts`), never from client input, and every POST/GET/DELETE query is scoped with `.eq('user_id', user.userId)`. No IDOR path — a caller cannot read or write another user's `listening_signals`.
- **`listening_signals` RLS — confirmed present, not new to this PR.** The table (migration `20260809120000_bandcamp-collection.sql`) already has RLS enabled with SELECT/DELETE policies scoped to `auth.uid() = user_id` and deliberately no INSERT/UPDATE policy (server-role writes only), with a comment explaining why — matches the project's "server-only, no policies" convention. `me-listening.ts` correctly uses the service-role client (`getClient()`) for writes, which is why app-layer scoping (above) is what's actually load-bearing here — worth being aware of, not a bug.
- **Input validation — confirmed adequate.** `parseSignals` bounds the array (`MAX_SIGNALS = 10_000`), bounds name length (300 chars), coerces `playCount` to a non-negative finite integer, and collapses in-payload duplicates by lowercased name (avoiding a Postgres "affect row a second time" error). `source` is checked against a `Set` allowlist. No raw SQL/string interpolation anywhere in the diff — all Supabase calls use parameterized `.eq`/`.in`/`.upsert`.
- **Unsafe defaults — confirmed none.** The PR's central "listening ≠ support" invariant holds in code: nothing in `me-listening.ts` writes `saved_artists`, and the removed `markArtistsSupported` is fully deleted (not just gated).
- *Plausible:* **No per-user cap on total accumulated rows.** `MAX_SIGNALS` bounds one request, but nothing bounds how many distinct `artist_name` rows one account can accumulate over repeated calls (each call can add up to 10,000 new distinct names). Low severity since it only bloats the calling user's own data and is IP-rate-limited, but worth a mental note if abuse shows up.
- *Plausible:* **Rate-limit tier deviates from the established `/api/me/*` pattern.** Every other account endpoint (`me-settings.ts`, etc.) uses `accountRateLimitKey(authHeader, ip)` + the `'account'` tier (60/min, keyed to the authenticated user when available) specifically to avoid the bug class described in `ratelimit.ts`'s comment (Sentry UNSTREAM-WEB-12: shared-IP traffic 429ing someone's own account action). `me-listening.ts` instead calls `checkRateLimit(getClientIp(...), 'standard', ...)` — pure per-IP, 30/min — before authentication. Not exploitable, but it reintroduces the exact shared-IP contention this tier was built to prevent, on the one endpoint that does the heaviest per-call work (up to 10k-row upserts).

## Regression

- **Blast-radius claim in the PR description is incomplete.** The description states `getFeedReleasesForUser` "has exactly one consumer (the dashboard's Upcoming/Recent sections)." Grepping the base branch shows three call sites: `me-recent-releases.ts` (the dashboard, as claimed), `feed-releases.ts` (the private per-fan ICS/Atom calendar feed, `getFeedReleasesForUser` called directly), and `db.ts`'s `getFeedReleasesForHandle` (the **public** `/u/{handle}` calendar feed, gated on `saved_artists_public`). Cross-checked against `public-saved-artists.ts`, which already surfaces `collection_items` under that same public-sharing gate for the JSON page — so extending the calendar feeds to include collection-derived releases looks consistent with existing intent, not a new leak. Still, the "exactly one consumer" claim itself doesn't hold — worth correcting or double-checking the calendar-feed behavior explicitly (needs-manual-verification for intent, confirmed for the consumer count).
- **Silent degrade on `savedRead`/`collectionRead` failure in the GET gap handler** (`me-listening.ts`, the `GET` branch). `signalsRead` failure correctly returns 500 (tested). But `savedRead.ok` and `collectionRead.ok` are checked only to populate the `covered` set — a failed read there is never surfaced as an error, just silently treated as "nothing covered." Net effect: if the `saved_artists` or `collection_items` read fails, an already-supported artist can incorrectly appear in the "gap" list rather than the request failing loudly. This is the inverse of the (correct, intentional) behavior in `collectionArtistIdsForUser` in `db.ts`, where a read failure narrows the feed rather than expanding it — here a failure expands the gap with a false positive. Plausible, moderate-severity correctness bug, not caught by the existing tests (only the primary `signalsRead` failure path is tested).
- **PostgREST 1,000-row truncation — confirmed handled correctly.** All three reads in the GET handler (`listening_signals`, `saved_artists`, `collection_items`) and the new `collectionArtistIdsForUser` in `db.ts` use `readAllPages`, which pages in 1,000-row batches and fails loudly (doesn't silently truncate) up to a 50-page backstop. No regression here.
- **`markArtistsSupported` removal — confirmed clean.** The full ~145-line function and its call site are gone, not just disabled; response field renamed `artistsMarked` → `artistsMatched`. That rename is an API contract change on `bandcamp-sync-background.ts`'s response body — needs-manual-verification against whatever reads that field (Mac app / #439), since it isn't visible in this diff.
- **Contract vs. PR #439 (Mac client) — needs-manual-verification.** Couldn't check the client's expected request/response shape for `/api/me/listening` without fetching #439 separately; the shapes in this PR (`{ source, signals: [...] }` for POST, `{ gap: [...], totalArtists }` for GET, `{ deleted: true }` for DELETE) look internally consistent and are exercised by the new tests, but I can't confirm they match what the Mac app actually sends/expects.
- **Edge cases** — empty collection/saved list, zero play counts, and duplicate-artist-across-sources are all explicitly tested and handled correctly (merge-by-slug, "saved but not supported stays in the gap"). No `RLS`/migration gap: the `listening_signals` table and its unique constraint (`user_id, source, artist_name`, matching the `onConflict` used in the upsert) predate this PR.

---

PR #455, opened 2026-08-16
Fix two Sentry warnings: reset-password 400s and the self-inflicted /settings 429

brandonlucasgreen:
Two one-off Sentry reports. Neither was a fault in the thing it named.

## UNSTREAM-WEB-J — `Auth failed: resetPassword`, 400

The breadcrumbs pin it: two clicks on **Forgot password?** with the field empty (guarded, no network call), then an email typed, retyped, and a third click — the one call that reached Supabase.

The reset path did no email validation. The input has `required` and `type="email"`, but those only run on native form submit, and both "Forgot password?" and "Send sign-in link" are `type="button"` with onClick handlers. So `foo@`, `foo@gmail`, or a stray character went straight to GoTrue, which answers `400 validation_failed` and put `Unable to validate email address: invalid format` in the red error box.

Ruled out: rate limiting returns **429**, not 400, and `/recover` on an unknown address returns **200** so it can't leak account existence. A typo is what's left, and once-from-one-person is its signature.

Both buttons now validate first, using the same pattern as the server-side check in `api/functions/artist-auth.ts`.

## UNSTREAM-WEB-12 — `Failed to load settings`

Every breadcrumb is a **429**, across five different `/api/me/*` endpoints. `/api/me/settings` just happened to be the last one to lose, and `SettingsPage` threw the same generic error for any non-ok status — so a spent rate-limit budget looked identical to a database outage.

The cause is in `AuthContext`. Supabase hands back a **fresh Session object** on every emission — tab focus, cross-tab sync, token refresh — and its `INITIAL_SESSION` event lands right after `init()` has already set an identical one. Setting state unconditionally changed the session's *identity* each time, re-firing every `[session]` effect in the app. `/settings` has six of them, so:

- one page load fired **twelve** requests, not six (this is the duplicate `/api/me/bandcamp` and `/api/me/saved-artists-sharing` in the breadcrumbs)
- every subsequent focus event fired six more — and this user was on Safari, which is aggressive about that

`applySession()` now ignores emissions carrying the same access token. The saved-artists clear had to move inside that guard: `new Set()` mints a new identity on every call, which would have recreated exactly the churn it was meant to stop.

Two follow-ons:

**A 429 is backpressure, not a fault.** All five settings fetches now say what actually happened and stay out of Sentry. One rate-limit episode was raising five separate errors, none of which described the cause, all telling the user to "try again" — the one thing that makes it worse.

**Account endpoints get their own bucket.** They shared `standard` (30/min per IP) with artist page views, artist lookups, and analytics events, so browsing a few artist pages first could spend the budget and 429 the settings page on itself. New `account` tier at 60/min, 2000/day — the same reasoning that already gave typeahead its own `lenient` bucket, and whose comment says precisely this. `me-password` stays on `strict` deliberately.

## Verification

- `npm run verify` green — 1225 API tests, 753 web unit tests.
- The login fix checked against the real stack via `npm run dev`: the message renders and **zero** requests reach `/auth/v1/recover`.
- 11 new tests across four files. Each was proven to have teeth by neutering the fix and confirming only the intended tests failed.

### DIFF
--- a/api/functions/__tests__/collection-art.test.ts
+++ b/api/functions/__tests__/collection-art.test.ts
@@ -13,6 +13,8 @@ const mocks = vi.hoisted(() => ({
 vi.mock('../db', () => ({ getClient: () => ({ from: mocks.mockFrom }) }));
 vi.mock('@supabase/supabase-js', () => ({ createClient: mocks.mockCreateClient }));
 vi.mock('../ratelimit', () => ({
+  // Only a bucket name; the endpoints' own auth is mocked separately.
+  accountRateLimitKey: async () => 'user:test-user',
   checkRateLimit: mocks.mockCheckRateLimit,
   getClientIp: mocks.mockGetClientIp,
 }));
--- a/api/functions/__tests__/me-bandcamp.test.ts
+++ b/api/functions/__tests__/me-bandcamp.test.ts
@@ -13,6 +13,8 @@ const mocks = vi.hoisted(() => ({
 vi.mock('../db', () => ({ getClient: () => ({ from: mocks.mockFrom }) }));
 vi.mock('@supabase/supabase-js', () => ({ createClient: mocks.mockCreateClient }));
 vi.mock('../ratelimit', () => ({
+  // Only a bucket name; the endpoints' own auth is mocked separately.
+  accountRateLimitKey: async () => 'user:test-user',
   checkRateLimit: mocks.mockCheckRateLimit,
   getClientIp: mocks.mockGetClientIp,
 }));
--- a/api/functions/__tests__/me-bandcamp.test.ts
+++ b/api/functions/__tests__/me-bandcamp.test.ts
@@ -415,3 +415,34 @@ describe('me-bandcamp handler', () => {
     });
   });
 });
+
+// POST is the one account route that submits a third-party credential and verifies it live
+// against Bandcamp, which makes it a credential-checking oracle. It therefore does NOT share
+// the reads' generous per-user budget: per-user keying would let an attacker buy more guessing
+// budget by holding more Unstream accounts, which is backwards for a credential path.
+describe('me-bandcamp rate limiting', () => {
+  beforeEach(() => {
+    mocks.mockCheckRateLimit.mockClear();
+    mocks.mockCheckRateLimit.mockResolvedValue({ limited: false });
+  });
+
+  it('limits credential submission per IP on the strict tier', async () => {
+    mocks.mockPing.mockResolvedValue(undefined);
+    mocks.mockArtistCount.mockResolvedValue(1);
+    mocks.mockFrom.mockReturnValue({ upsert: vi.fn(() => Promise.resolve({ error: null })) });
+
+    await handler(authedEvent('POST', { username: 'fan', password: PASSWORD }));
+
+    expect(mocks.mockCheckRateLimit).toHaveBeenCalledWith('127.0.0.1', 'strict', expect.anything());
+  });
+
+  it('limits the status read per user on the account tier', async () => {
+    mocks.mockFrom.mockReturnValue({
+      select: vi.fn(() => ({ eq: vi.fn(() => ({ maybeSingle: async () => ({ data: null, error: null }) })) })),
+    });
+
+    await handler({ httpMethod: 'GET', headers: { authorization: 'Bearer valid-token' }, body: null });
+
+    expect(mocks.mockCheckRateLimit).toHaveBeenCalledWith('user:test-user', 'account', expect.anything());
+  });
+});
--- a/api/functions/__tests__/me-collection.test.ts
+++ b/api/functions/__tests__/me-collection.test.ts
@@ -19,6 +19,8 @@ vi.mock('../db', async importOriginal => {
 });
 vi.mock('@supabase/supabase-js', () => ({ createClient: mocks.mockCreateClient }));
 vi.mock('../ratelimit', () => ({
+  // Only a bucket name; the endpoints' own auth is mocked separately.
+  accountRateLimitKey: async () => 'user:test-user',
   checkRateLimit: mocks.mockCheckRateLimit,
   getClientIp: mocks.mockGetClientIp,
 }));
--- a/api/functions/__tests__/me-location.test.ts
+++ b/api/functions/__tests__/me-location.test.ts
@@ -24,6 +24,8 @@ vi.mock('@supabase/supabase-js', () => ({
   createClient: mocks.mockCreateClient,
 }));
 vi.mock('../ratelimit', () => ({
+  // Only a bucket name; the endpoints' own auth is mocked separately.
+  accountRateLimitKey: async () => 'user:test-user',
   checkRateLimit: mocks.mockCheckRateLimit,
   getClientIp: mocks.mockGetClientIp,
 }));
--- a/api/functions/__tests__/me-notifications.test.ts
+++ b/api/functions/__tests__/me-notifications.test.ts
@@ -10,6 +10,8 @@ const mocks = vi.hoisted(() => ({
 vi.mock('../db', () => ({ getClient: () => ({ from: mocks.mockFrom }) }));
 vi.mock('@supabase/supabase-js', () => ({ createClient: mocks.mockCreateClient }));
 vi.mock('../ratelimit', () => ({
+  // Only a bucket name; the endpoints' own auth is mocked separately.
+  accountRateLimitKey: async () => 'user:test-user',
   checkRateLimit: mocks.mockCheckRateLimit,
   getClientIp: mocks.mockGetClientIp,
 }));
--- a/api/functions/__tests__/me-recent-releases.test.ts
+++ b/api/functions/__tests__/me-recent-releases.test.ts
@@ -14,6 +14,8 @@ vi.mock('@supabase/supabase-js', () => ({
   createClient: () => ({ auth: { getUser: mocks.authGetUser } }),
 }));
 vi.mock('../ratelimit', () => ({
+  // Only a bucket name; the endpoint's own auth is mocked separately.
+  accountRateLimitKey: async () => 'user:test-user',
   checkRateLimit: mocks.checkRateLimit,
   getClientIp: mocks.getClientIp,
 }));
--- a/api/functions/__tests__/me-settings.test.ts
+++ b/api/functions/__tests__/me-settings.test.ts
@@ -32,6 +32,8 @@ vi.mock('@supabase/supabase-js', () => ({
   createClient: mocks.mockCreateClient,
 }));
 vi.mock('../ratelimit', () => ({
+  // Only a bucket name; the endpoints' own auth is mocked separately.
+  accountRateLimitKey: async () => 'user:test-user',
   checkRateLimit: mocks.mockCheckRateLimit,
   getClientIp: mocks.mockGetClientIp,
 }));
--- a/api/functions/__tests__/me-username.test.ts
+++ b/api/functions/__tests__/me-username.test.ts
@@ -26,6 +26,8 @@ vi.mock('@supabase/supabase-js', () => ({
   createClient: mocks.mockCreateClient,
 }));
 vi.mock('../ratelimit', () => ({
+  // Only a bucket name; the endpoints' own auth is mocked separately.
+  accountRateLimitKey: async () => 'user:test-user',
   checkRateLimit: mocks.mockCheckRateLimit,
   getClientIp: mocks.mockGetClientIp,
 }));
--- a/api/functions/__tests__/ratelimit-account-tier.test.ts
+++ b/api/functions/__tests__/ratelimit-account-tier.test.ts
@@ -1,0 +1,65 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+// A signed-in user's account endpoints used to share the 'standard' bucket with artist page
+// views and analytics events. /settings fans out to six of them at once, so browsing a handful
+// of artist pages first could spend the budget and 429 the settings page on itself
+// (Sentry UNSTREAM-WEB-12). These pin the separation.
+
+const limitCalls: { prefix: string; identifier: string }[] = [];
+const built: { prefix: string; window: unknown }[] = [];
+
+vi.mock('../redis', () => ({
+  getRedis: () => ({}),
+  reportRedisFailure: vi.fn(),
+}));
+
+vi.mock('@upstash/ratelimit', () => {
+  class FakeRatelimit {
+    prefix: string;
+    constructor(opts: { prefix: string; limiter: unknown }) {
+      this.prefix = opts.prefix;
+      built.push({ prefix: opts.prefix, window: opts.limiter });
+    }
+    static slidingWindow(tokens: number, window: string) {
+      return { tokens, window };
+    }
+    async limit(identifier: string) {
+      limitCalls.push({ prefix: this.prefix, identifier });
+      return { success: true, reset: 0 };
+    }
+  }
+  return { Ratelimit: FakeRatelimit };
+});
+
+describe('account rate limit tier', () => {
+  beforeEach(() => {
+    limitCalls.length = 0;
+    built.length = 0;
+    vi.resetModules();
+  });
+
+  it('spends a different bucket than browsing traffic from the same IP', async () => {
+    const { checkRateLimit } = await import('../ratelimit');
+
+    await checkRateLimit('1.2.3.4', 'standard', {});
+    await checkRateLimit('1.2.3.4', 'account', {});
+
+    const prefixes = limitCalls.map(c => c.prefix).sort();
+    expect(prefixes).toEqual([
+      'rl:account',
+      'rl:daily:account',
+      'rl:daily:standard',
+      'rl:standard',
+    ]);
+  });
+
+  it('gives the account tier room for a full settings page load', async () => {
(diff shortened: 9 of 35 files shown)


brandonlucasgreen:
@brandonlucasgreen

## Security

- **plausible** — `api/functions/me-bandcamp.ts:186`: moving the POST handler (which takes a Bandcamp username/password and verifies it live against Bandcamp's Subsonic API via `subsonicPing`) from `'standard'` to the new `'account'` tier gives a signed-in caller a *dedicated* per-IP budget of 60 req/min and 2000 req/day for this endpoint alone, versus the previous 30/min shared with all other `'standard'`-tier traffic (search, artist pages, analytics). Since the tier is IP-keyed (not per-user), this concretely doubles the throughput available to use `/api/me/bandcamp` as a credential-testing oracle for stuffing stolen Bandcamp username/password pairs, once an attacker holds one Unstream session. Worth confirming this endpoint should really share the same generous "own account, cheap read" bucket as `me-location`/`user-sharing`/etc. rather than staying on something stricter given it's the one `/api/me/*` route that submits and verifies a third-party credential.
- No other confirmed issues in auth/authorization, input validation, secret handling, injection, or data exposure — the `EMAIL_PATTERN` regex in `LoginPage.tsx` is linear (no ReDoS), client-side validation is additive to (not a replacement for) server-side checks, and the Bandcamp password itself is still never logged/echoed (unchanged in this diff).

## Regression

- **confirmed** — `api/functions/me-feed-token.ts:79` was not migrated to the new `'account'` tier and still calls `checkRateLimit(ip, 'standard', ...)`. `ReleaseFeedControls.tsx` (fetches `/api/me/feed-token`) is one of the six `/settings` panels the PR description itself names as the cause of the original bug ("`/settings` alone fans out to six of them"). So the stated fix — "Account endpoints get their own bucket" — is incomplete: this one panel can still have its `'standard'` budget spent by unrelated browsing traffic and 429 on a routine `/settings` load. The blast radius is reduced (the new `RATE_LIMIT_MESSAGE`/429 handling in `ReleaseFeedControls.tsx` means it'll show a friendly message instead of a raw Sentry-reported error), but the underlying self-inflicted-429 root cause this PR targets is not actually closed for that endpoint.
- **plausible** — `apps/web/src/contexts/AuthContext.tsx:91-106`: `applySession` now no-ops (skips `setUser`) whenever `newSession.access_token === sessionRef.current.access_token`. Supabase's `updateUser()` (used by `updatePassword` in `apps/web/src/services/auth.ts:118-127` to set `has_password: true`) fires a `USER_UPDATED` event that typically carries the *same* access token as before (metadata changes don't rotate the JWT). That means the new identity guard will treat it as a no-op and `user` (and derived `hasPassword`) won't update in `AuthContext` until a genuine token refresh occurs. `PasswordSection.tsx` (rendered from `apps/web/src/pages/ArtistDashboardPage.tsx:185`) reads `hasPassword` straight from `useAuth()`, so after a fan sets their first password, the button/copy can keep showing "Set password" instead of "Change" for the rest of the session even though the success message says it worked. (Note: `SettingsPage.tsx`'s own password section is unaffected — it reads `hasPassword` from a fresh `/api/me/settings` fetch, not from `AuthContext`.)
- **Re: the semantic-revert-check bot's comment on this PR** — checked and this is a false positive, on both points it raises. (1) Stale base: `git merge-base origin/main origin/reset-password-400-error-32a18c` resolves to `e472ddf`, which *is* the current tip of `main` — the branch is built directly on top of `main` and already contains #438, #448, #451, #452, #453 in its history, so this isn't a revert caused by branching off an old commit. (2) Whether it undoes #438's fix: the two flagged `-1` lines in `me-bandcamp.ts`/`me-collection.ts` are exactly the `checkRateLimit(ip, 'standard', ...)` → `checkRateLimit(ip, 'account', ...)` tier-rename hunks, immediately paired with a `+1` replacement line in the same diff (visible above in the `get_diff` output) — not a deletion of rate limiting. #438's own diff (`27c49ac`) shows no comment tying that check to the specific value `'standard'`; it was just the pre-existing default tier at the time those endpoints were added. So this is an intentional, uniform re-tiering across all `/api/me/*` endpoints as described in the PR body, not an accidental revert of #438 — modulo the `me-bandcamp` credential-oracle throughput concern raised in Security above, and the `me-feed-token.ts` gap above, which are the two real follow-ups from this change.

---

brandonlucasgreen:

### Security
- **plausible** — `api/functions/me-bandcamp.ts:186` (POST handler): moving this endpoint from `standard` (30/min, 1000/day) to the new `account` tier (60/min, 2000/day) doubles the per-IP throughput on the one account endpoint that accepts a third-party credential in the body (`{ username, password }`, verified live against Bandcamp's Subsonic API per the file's own header comment). Rate-limiting runs before Bearer-token authentication in this handler, so the endpoint functions as a Bandcamp credential-verification oracle and now tolerates ~2x the per-IP guessing rate it did before. The `account` tier's own comment ("cheap authenticated reads") doesn't really fit this write path, and the PR doesn't call out that `me-bandcamp` POST was swept into the same bucket as read-only panels.
- **plausible** — `api/functions/me-username.ts:47`, `me-collection.ts:57`, `saved-artists.ts:137`, `saved-artists-sync.ts:33`, `user-sharing.ts:66`: same pattern — `checkRateLimit` (IP-based) runs before Bearer auth is checked, so the doubled `account` budget (30→60/min, 1000→2000/day) is spent by anyone hitting the endpoint, not just authenticated owners. Lower individual severity since every write still requires a valid token, but it's a straightforward loosening of the anti-abuse budget on auth-adjacent endpoints (e.g. username-claim collision probing) with no compensating per-user/per-token limit added.

### Regression
- **confirmed** — `api/functions/me-feed-token.ts:79` still calls `checkRateLimit(ip, 'standard', CORS_HEADERS)` — it wasn't migrated to the new `account` tier, even though this PR edits the component that calls it (`apps/web/src/components/ReleaseFeedControls.tsx`) to add the same 429-specific `RATE_LIMIT_MESSAGE` handling as the other five migrated panels. So the "self-inflicted 429" fix is incomplete: clicking "Get feed link" / rotate / revoke on `/settings` still draws from the shared `standard` bucket alongside artist-page views and search — the exact bug class this PR targets (browsing artist pages first exhausts the budget, then an account action 429s) is still reachable through this one endpoint, while the UI now tells the user it's harmless backpressure regardless of whether it actually is.
- **needs-manual-verification** — `apps/web/src/contexts/AuthContext.tsx:96-97`: `applySession`'s dedup guard treats two sessions as identical purely by `access_token` equality. If Supabase ever re-emits a session with the same access token but a changed `user` object (e.g. via `supabase.auth.updateUser()`, used by `updatePassword()` in `apps/web/src/services/auth.ts:122`, without a token rotation, or a cross-tab metadata sync), the new `user` payload would be silently discarded and derived UI state such as `hasPassword` (line 335) would stay stale until a genuine token refresh. Whether Supabase's GoTrue client actually fires such a same-token/different-user event can't be confirmed by reading this repo alone.

**Bottom line:** the rate-limit migration is a real fix for the settings-page 429 but is one endpoint short (`me-feed-token.ts`), and widens the write-side budget on a credential-checking endpoint (`me-bandcamp.ts`) more than the tier's own stated purpose suggests was intended.

---

brandonlucasgreen:
Worked through the review. Three findings were real and are fixed in dae2363; the semantic-revert warning is a false positive; two security points are stale but one of them is stale in an unhelpful direction.

## Fixed

**`me-feed-token.ts` still on `standard` — confirmed, and there were two.** `me-recent-releases.ts:114` had the same gap and nobody caught it. Both are now on the per-user `account` tier.

Worth naming why I missed them: my sweep grepped `checkRateLimit(ip, 'standard'` — a *variable name*, not a call. Both of these pass `getClientIp(event.headers)` inline, so both were invisible to it. Grepping `checkRateLimit(` across the account endpoints found the second one immediately.

**`me-bandcamp` POST as a credential oracle — real, and the user-keying commit made it worse rather than better.** The reviews were written against `068228e`, before the tier was keyed by user, so the stated premise ("the tier is IP-keyed") is out of date. But per-user keying doesn't shrink this attack, it *widens* it: where IP keying capped an attacker at one bucket per network, per-user keying gives them 60/min for every Unstream account they hold.

POST now uses `strict` (10/min) keyed by **IP**, matching `me-password`. Per-user is right for the reads and backwards for a credential path, precisely because accounts are cheap to make. GET and DELETE keep the per-user budget, which is what stops the sync poll competing with a colleague on the same network.

**`applySession` dropping `USER_UPDATED` — real.** Flagged as *plausible* / *needs-manual-verification*; it's a genuine defect regardless of whether Supabase currently rotates the token on `updateUser`, because the access token is simply the wrong key for user identity. `hasPassword` derives from `user.user_metadata` (`AuthContext.tsx:348`) and `PasswordSection.tsx:7` reads it, so the guard could leave someone who had just set a password still being offered "Set password".

`user` now propagates unconditionally; only `session` stays token-guarded. That's the right split — `session` is what the six `/settings` panels fetch on, and nothing keys an effect on `user`, so this costs a render rather than a refetch. Pinned by a test that asserts both halves: metadata updates propagate, and the fetches still don't re-fire.

## Not changed

**Semantic-revert warning: false positive.** The flagged `-1` lines in `me-bandcamp.ts` / `me-collection.ts` are the `'standard'` → `'account'` tier rename, each paired with a `+1` on the same line. `BandcampConnect.tsx`'s six deleted lines are the `setInterval` poll replaced by the self-scheduling one. `git merge-base origin/main HEAD` is `e472ddf`, the current tip of `main`, so the branch isn't on a stale base.

**Rate limiting before authentication.** Deliberate, and it's why `accountRateLimitKey` verifies the JWT *locally* — a garbage token costs a local signature check and no network call, so an unauthenticated flood can't be amplified into auth-server load. Limiting after auth would invert that. Unauthenticated callers land in the `ip:` bucket and get a 401.

## Verification

`npm run verify` green — 1230 API tests, 759 web. Three new tests, each mutation-tested: reverting the `user` propagation, the POST tier split, and the `account` keying each fail exactly the test that covers them.

brandonlucasgreen (commented):
@brandonlucasgreen — automated security + regression pass on this PR (reviewed at head `dae2363`).

**Security:**
- **confirmed, already resolved within this PR** — worth flagging anyway since it's exactly the failure pattern this pass looks for: an intermediate commit (`5c7fdba`) moved the `me-bandcamp.ts` POST (which verifies a submitted Bandcamp username/password live against Bandcamp) onto the new per-user `account` rate-limit bucket, same as the other 8 endpoints in this PR. That would have let an attacker buy more credential-guessing budget just by creating more Unstream accounts. The final commit (`dae2363`) correctly reverted POST specifically to per-IP `strict` (10/min) — `api/functions/me-bandcamp.ts:198-208` — with GET/DELETE staying on the per-user bucket, and added a test asserting the IP key. No outstanding action, just flagging that this is the kind of self-inflicted hole a fix-the-noisy-429 pass can introduce.
- **plausible (low)** — `api/functions/ratelimit.ts:78-99`: the new `account` tier is 2x the old `standard` tier (60/min·2000/day vs 30/min·1000/day). `accountRateLimitKey` falls back to `ip:<addr>` for any request without a verifying token, so an anonymous caller hitting these 9 auth-gated endpoints gets 2x the prior pre-auth request budget before its own 401 kicks in. Minor since auth still rejects it, but it's a real widening of the pre-auth surface.
- **needs-manual-verification** — `apps/web/src/contexts/AuthContext.tsx:92`: the `tokenUnchanged` check now suppresses `session` propagation when `access_token` is unchanged (deliberately keeping `user` propagating, which fixes a real `hasPassword` staleness bug). Not verified from the diff alone: whether Supabase can rotate `refresh_token` while keeping the same `access_token` — if so, the stale `refresh_token` sitting in React state would go unpropagated.

**Regression:**
- **confirmed, no issue** — `apps/web/src/pages/LoginPage.tsx`: the new client-side email-format regex is strictly additive (reduces malformed requests reaching Supabase); doesn't touch server-side validation or the account-existence-leak guarantee.
- **confirmed, no issue** — `middleware.ts:104-149`: `verifyJwtLocally` restricts algorithms per key type and checks issuer/audience, so a forged token can't mint an arbitrary `user:<id>` rate-limit bucket.

---

brandonlucasgreen:
@brandonlucasgreen

Reviewed the full diff (rate-limit tier addition, `applySession` dedup, LoginPage email validation, and the 429-handling changes across Settings components).

## Security

**Auth/session guard (`apps/web/src/contexts/AuthContext.tsx`, `applySession()`) — confirmed sound.**
The guard compares `newSession?.access_token === sessionRef.current?.access_token` and skips `setSession`/saved-artist clearing when equal. This is a sound proxy: an access token is a signed JWT with the user id in `sub`, so two different users can never share one, and a genuine `TOKEN_REFRESHED` for the same user always mints a new token (pinned by the "re-fires session effects on a genuine token refresh" test). `USER_UPDATED` (e.g. after `updatePassword()`) deliberately bypasses the guard for `setUser` only, so `hasPassword` still updates on a same-token event — verified by the dedicated test. No path found where a stale session would be kept when it should be replaced.

**Client-side email validation (`apps/web/src/pages/LoginPage.tsx`) — confirmed UX-only, not a validation-gap fix.**
`EMAIL_PATTERN` runs before `signInWithMagicLink`/`resetPasswordForEmail`, which call the Supabase client SDK directly (not an Unstream API route) — GoTrue itself already rejects malformed addresses server-side (that's the literal source of the original 400 this PR is suppressing). So this closes no gap; it only avoids an avoidable round trip and the resulting Sentry noise, and is trivially bypassable by calling `supabase.auth.*` directly — which is fine only because real enforcement already lives upstream at GoTrue, not here. Worth stating explicitly in case this pattern is later copied onto an endpoint where server-side enforcement *isn't* already guaranteed.

**Secrets / injection — confirmed, none.** No new secret handling or string-built queries anywhere in the diff.

**Rate-limit tier change (`api/functions/ratelimit.ts`, `me-bandcamp.ts`) — mostly confirmed correct, one plausible minor widening.**
- Confirmed `me-password.ts` is untouched by this PR and remains on `strict` (10/min, per-IP) — the PR description's claim checks out.
- Confirmed `me-bandcamp.ts` POST (credential submission against Bandcamp) is explicitly kept on `strict`/per-IP with a clear comment explaining why per-user keying would be backwards for a credential-guessing path; GET/DELETE move to the per-user `account` tier. This is the right split.
- Plausible: on all ~10 migrated endpoints, rate-limiting still runs *before* authentication (unchanged from before this PR), and `accountRateLimitKey` falls back to `ip:<ip>` for missing/invalid tokens. That IP-fallback bucket is now `account` (60/min, 2000/day) instead of the old `standard` (30/min, 1000/day) — i.e. unauthenticated or forged-token traffic against these endpoints gets exactly double the previous budget as a side effect of giving real signed-in users headroom. Impact is bounded (none of these are credential oracles, and they still 401 downstream), but it is a real, diff-confirmed widening worth being aware of.

**Data exposure via 429 handling — confirmed, no over-broad swallowing.** Checked every touched fetch call site (`BandcampConnect.tsx`, `NotificationPreferences.tsx`, `ReleaseFeedControls.tsx`, `SharingControls.tsx`, `SettingsPage.tsx`): each special-cases `status === 429` only, and every other non-ok status still falls through to `throw`/existing error handling, which still reaches `Sentry.captureException`. 401/403/500 are not silently reclassified as backpressure.

## Regression

**`[session]`-keyed effects — confirmed no breakage.** Grepped every `useEffect` depending on `session` (`DashboardPage`, `ArtistDashboardPage`, `ArtistPage`, `SettingsPage`, `LoginPage`, `App.tsx`, `CollectionSection`, `NotificationPreferences`, `BandcampConnect`, `ReleaseFeedControls`, `SharingControls`, plus the four inside `AuthContext.tsx` itself). All of them gate on presence/absence of a session or on `session.access_token` for a fetch header — none rely on object identity independent of the token, so deduping same-token re-emissions is consistent with their intent.

**Saved-artists clear moved inside the token guard — confirmed no regression on the normal path.** A genuine sign-out always transitions `access_token` from a real string to `undefined`, so `tokenUnchanged` is `false` and the clear still runs (pinned by the "propagates a sign-out" test).

**Same scenario, race variant — needs-manual-verification.** Whether Supabase's `onAuthStateChange` can ever emit an intermediate same-token session immediately before the `SIGNED_OUT`/`null` emission (which would leave `sessionRef.current` pointing at the old token right up to the real null event, so the clear still fires correctly) isn't something the diff or tests establish either way — it depends on `gotrue-js` internals outside this repo. Low likelihood, but not provably ruled out here.

**API contract drift — confirmed, none.** `docs/openapi.yaml`'s rate-limit table only documents the public v1 API's anonymous/free/pro/internal tiers (`/api/v1/*`, API-key or anonymous-per-IP). The new `account` tier applies only to internal `/api/me/*` and related web-app endpoints, which aren't part of that documented surface, so nothing there is now inconsistent.

**Two tabs / different accounts — confirmed not affected.** The guard only dedupes literal repeat emissions of the same token within one listener; any real token difference (including a second account signed in elsewhere) still propagates through `setSession` unchanged from prior behavior.

---

brandonlucasgreen:
Superseded by #458, which squash-merged this work into `main` together with #457 so the two halves of the incident cost a single production deploy.

Verified on `main` before closing: the `account` tier is present in `ratelimit.ts`, and both `ratelimit-account-tier.test.ts` and `AuthContext-session-identity.test.tsx` are in place. Nothing here is lost — closing the PR only, leaving the branch in case you want the history.

PR #456, opened 2026-08-16
Fix the Mac app build: .onOpenURL is a View modifier, not a Scene one

brandonlucasgreen:
## main does not compile for macOS

```
error: value of type 'Settings<SettingsView>' has no member 'onOpenURL'
```

The magic-link work attached `.onOpenURL` to the SwiftUI **`Settings` scene**. CI never caught it because `ci.yml` runs `npm run verify` — typecheck plus the web and API suites — and **never builds `apps/mac`**.

I verified this by checking out `origin/main`'s own `UnstreamApp.swift` and rebuilding before changing anything, so it's a pre-existing break, not a merge artifact from another branch.

## Why not just move it inside the closure

Attaching it to a View inside `Settings` would compile, but stay broken in a subtler way: a View modifier only receives URLs while that view is in the hierarchy, and this is a menu-bar accessory whose Settings window is normally closed — exactly when a magic-link callback arrives. So the feature would have been dead either way.

URL handling now lives in `AppDelegate.application(_:open:)`, which fires regardless of which windows exist. One file changed.

## Verification

- `xcodebuild … build` succeeds; the XCTest suite passes.
- Deliberately minimal so it can merge immediately and unblock the Mac app; #439 carries the same fix and will merge cleanly on top.

## Follow-up worth doing

A CI job that builds the Mac app would have caught this the day it landed. Happy to add one if you want it.

### DIFF
--- a/apps/mac/Unstream/UnstreamApp.swift
+++ b/apps/mac/Unstream/UnstreamApp.swift
@@ -128,14 +128,13 @@ struct UnstreamApp: App {
         // The popover is driven by AppDelegate (NSStatusItem), but Settings is a real
         // SwiftUI Settings scene: that's what gives us the standard settings window,
         // its frame persistence, and a working "Unstream ▸ Settings… ⌘," menu item.
-        // It also registers this app's `unstream://` URL-scheme handling, which is
-        // what lets a magic-link email complete sign-in even though nothing opened
-        // the browser from inside the app (see handleAuthCallbackURL below).
+        //
+        // `unstream://` URLs are handled in AppDelegate, NOT here. `.onOpenURL` is a View
+        // modifier, not a Scene one — attaching it to `Settings` did not compile — and even
+        // as a View modifier it would only fire while the Settings window happened to be
+        // open, which is exactly when a magic-link callback is least likely to arrive.
         Settings {
             SettingsView(releaseAlertManager: container.releaseAlertManager)
-        }
-        .onOpenURL { url in
-            _ = handleAuthCallbackURL(url)
         }
         #else
         // iOS: Standard windowed app
@@ -170,6 +169,9 @@ struct UnstreamApp: App {
     /// Routes `unstream://auth/callback` (the redirect target for the magic-link
     /// email) to `AuthService`. Returns whether the URL was handled, so callers
     /// can fall through to other deeplink handling otherwise.
+    ///
+    /// macOS routes the same URLs through `AppDelegate.application(_:open:)` instead —
+    /// see the note on the Settings scene above.
     private func handleAuthCallbackURL(_ url: URL) -> Bool {
         guard url.scheme == "unstream", url.host == "auth" else { return false }
         Task { await AuthService.shared.handleAuthCallback(url: url) }
@@ -288,6 +290,15 @@ class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCenterDele
     private static var hasCreatedStatusItem = false
 
     private var popover: NSPopover!
+
+    /// `unstream://` URLs, chiefly the magic-link callback. This lives in the delegate
+    /// because the app is a menu-bar accessory with no persistent window: a SwiftUI
+    /// `.onOpenURL` needs a view in the hierarchy to receive it, and there usually isn't one.
+    func application(_ application: NSApplication, open urls: [URL]) {
+        for url in urls where url.scheme == "unstream" && url.host == "auth" {
+            Task { await AuthService.shared.handleAuthCallback(url: url) }
+        }
+    }
 
     func applicationDidFinishLaunching(_ notification: Notification) {
         AppDelegate.shared = self


brandonlucasgreen:

### Security
No security-relevant changes in this diff. This is a client-side Swift fix relocating `unstream://` URL routing from a non-compiling `Settings` scene modifier to `AppDelegate.application(_:open:)`. The validation guard (`url.scheme == "unstream" && url.host == "auth"`) is copied unchanged from the pre-existing (never-compiling) code — no new external input surface, no secrets touched, and none of this repo's server-side conventions (SSRF allowlist, RLS, escapeHtml, API-key handling) apply to a Mac-app deep-link handler.

### Regression
- **plausible** — `apps/mac/Unstream/UnstreamApp.swift:295-299`: `application(_:open:)` loops over the entire `urls` array and spawns a separate `Task { await AuthService.shared.handleAuthCallback(url: url) }` per matching URL with no dedup/guard. If macOS ever delivers more than one `unstream://auth/...` URL in a single batch (e.g. a stale queued URL plus a fresh click), `handleAuthCallback` could run concurrently more than once for overlapping/duplicate auth callbacks. Not fully verifiable from this file alone since `AuthService.handleAuthCallback`'s idempotency isn't in this diff.
- **needs-manual-verification** — `apps/mac/Unstream/UnstreamApp.swift:295-299`: this delegate method is now the only path that lets a magic-link email complete sign-in on macOS. Whether AppKit reliably delivers `application(_:open:)` on a cold launch (app not yet running when the user clicks the link) relative to `applicationDidFinishLaunching`/app readiness is standard AppKit lifecycle behavior — not resolvable by reading code alone, needs an actual cold-launch-via-URL test.

Everything else in the diff (comment updates, moving the shared scheme/host check) is a narrow, faithful fix matching the stated intent — no other behavior change detected.

---

brandonlucasgreen (commented):

**Security:**
- **plausible (low)** — `apps/mac/Unstream/UnstreamApp.swift:172-177` vs `294-299`: the `scheme == "unstream" && host == "auth"` allowlist check now exists in two places (the now-mostly-dead `handleAuthCallbackURL`, still used by iOS, and the new inline check in `AppDelegate.application(_:open:)`) instead of one shared call site. Identical today, so no live vulnerability, but a future tightening of this validation could land in one copy and not the other. Consider having the delegate call the existing helper instead of re-implementing the guard.

**Regression:**
- **confirmed** — this is a genuine fix, not a regression: `main` doesn't compile today (`.onOpenURL` isn't a `Scene` modifier), so there's no working baseline to regress from. The `AppDelegate`-based handler is app-lifetime-scoped (registered once via `@NSApplicationDelegateAdaptor`, backed by the app-wide `CFBundleURLTypes` entry in `Info-macOS.plist`), so there's exactly one handler regardless of window/navigation state — better than a view-level handler's duplicate/dropped-link risk.
- **needs-manual-verification** — `UnstreamApp.swift:296-298`: `application(_:open:)` receives `[URL]` and spawns one `Task` per URL in the array. If macOS ever batches two `unstream://auth/...` opens in one delivery, both fire concurrently against `AuthService.shared.handleAuthCallback`, and the resulting writes to `session`/`isLoading` could race. Low likelihood, unverified.
- **needs-manual-verification** — `UnstreamApp.swift:172`: `handleAuthCallbackURL` is no longer called anywhere on macOS (still used by iOS). Should just be a "never used" compiler warning, not a functional bug, but not independently confirmed by compiling.

---

brandonlucasgreen:
@brandonlucasgreen

Reviewed the diff (`apps/mac/Unstream/UnstreamApp.swift`) plus the callers it touches (`Services/AuthService.swift`, `Info-macOS.plist`).

## Security

No issues found. Details:

- **Validation preserved (confirmed).** The old `.onOpenURL { handleAuthCallbackURL(url) }` and the new `AppDelegate.application(_:open:)` both gate on the exact same predicate — `url.scheme == "unstream" && url.host == "auth"` (was `handleAuthCallbackURL`'s guard, now inlined in the `for urls where` filter). No check was dropped or weakened moving from the View-modifier path to the AppDelegate path.
- **Token validation unchanged (confirmed).** Both paths hand the URL to the same `AuthService.handleAuthCallback(url:)`, which calls Supabase's `client.auth.session(from: url)` — the SDK, not app code, validates the token/PKCE code. This function isn't touched by the diff.
- **No secret logging (confirmed).** No `print`/`NSLog`/logging of the URL or session was added in the new `application(_:open:)` method.
- **Scheme-hijacking risk (needs-manual-verification, pre-existing, not introduced here).** Custom URL schemes can in principle be claimed by another app on the same Mac; that risk class is identical before and after this PR since the scheme+host filter is unchanged. Worth a separate look (e.g. relying on Supabase's PKCE/state in the token itself) but out of scope for this diff.
- **No unsafe broadening of what's handled (confirmed).** `application(_:open:)` iterates `urls` but only acts on entries matching `scheme == "unstream" && host == "auth"`; everything else is silently ignored, same as before.

## Regression

- **Multiple URLs / array handling (confirmed as correctly handled).** `application(_:open:)`'s `urls: [URL]` is iterated with a `where` filter, so it won't drop a matching URL if the OS batches several. Minor **plausible** note: each match spawns its own `Task`, so two concurrent `unstream://auth/callback` opens would race two concurrent `AuthService.handleAuthCallback` calls (`isLoading`/`errorMessage` aren't synchronized) — low real-world likelihood since a user normally clicks one magic-link email at a time, flagging only for completeness.
- **No double-handling / no conflicting handler (confirmed).** Grepped the file: the only other `.onOpenURL` is in the `#else` (iOS) branch, mutually exclusive via `#if os(macOS)`. A comment at the top of `applicationDidFinishLaunching` explicitly notes no `NSAppleEventManager` handler is used, so there's no second macOS listener for the same event.
- **URL scheme registration unaffected (confirmed).** `Info-macOS.plist` still declares `CFBundleURLSchemes: [unstream]`; the diff doesn't touch registration, only which Swift method receives the event.
- **Cold-launch vs warm-launch delivery (needs-manual-verification).** `NSApplicationDelegateAdaptor(AppDelegate.self)` (unchanged, line 116) should make AppKit queue the URL-open Apple Event to `application(_:open:)` correctly on a cold launch triggered by clicking a magic-link, same as the standard non-SwiftUI-lifecycle pattern this PR is adopting. I can't run the app to confirm; suggest manually testing "app fully quit → click magic-link email → app launches and completes sign-in" in addition to the warm-app case, since that's precisely the scenario the old Settings-scene approach could never satisfy (window closed) and is the PR's stated motivation.
- **API contract drift:** N/A — Mac client only, no external API surface.

---

PR #457, opened 2026-08-16
Stop the cover-art proxy spending everyone's standard rate-limit budget

brandonlucasgreen:
## The symptom

Reported in production today: settings failing to load, releases not loading, the collection UI apparently missing — **on one laptop, while the same account worked fine on a phone.** That's an IP-scoped 429, not a deploy or code-removal problem (I verified the deployed bundles contain all the UI, and no commit since #438 touched those pages).

## The cause

`/api/collection/art` was on the **`standard`** bucket — 30/min per IP.

A collection page renders 15 tiles, and most of them have no stored `art_url`, because only ~28% of a real import matches an Unstream release. So **one page view fires a dozen or more requests through this endpoint**, and two page views spend the whole budget. Everything else on `standard` then 429s: the release lists, artist pages, artist lookups.

The `lenient` tier's own comment already describes this exact failure for typeahead:

> fires on every debounced typing pause, so it needs its own bucket — sharing 'standard' would let an actively-searching user 429 themselves out of account actions

Cover art is the same shape: high frequency, low cost, and a CDN hit after the first view. It moves to `lenient` (120/min).

## Relationship to #455

Complementary, and both are needed:

- **#455** fixes the larger cause — `AuthContext` re-firing every `[session]` effect on each Supabase emission, doubling request counts and re-firing on every tab focus — and moves account endpoints to their own `account` tier.
- **This PR** covers the gap #455 doesn't: it never touches `collection-art.ts`, and `me-recent-releases` stays on `standard`. So without this, a collection page view can still 429 the release lists.

Merge order doesn't matter. The new test lives in its own file rather than in `collection-art.test.ts`, which #455 also edits, so they merge cleanly either way.

## Verification

- 1,225 API tests pass; typecheck clean; lint at the pre-existing 71-problem baseline.
- 2 new tests: the endpoint spends `lenient`, and a spent bucket still refuses before any database work.
- Verified the proxy itself is healthy in production — it returns real JPEGs; the problem was only its cost.

### DIFF
--- a/api/functions/__tests__/collection-art-ratelimit.test.ts
+++ b/api/functions/__tests__/collection-art-ratelimit.test.ts
@@ -1,0 +1,73 @@
+// Which rate-limit bucket the cover-art proxy spends.
+//
+// Separate file from collection-art.test.ts on purpose: this pins one property that is easy
+// to regress by copying another endpoint's boilerplate, and a whole page of a fan's
+// collection depends on it.
+//
+// A collection grid renders 15 tiles, and most have no stored art_url — only ~28% of a real
+// import matches an Unstream release — so one page view fires a dozen or more requests here.
+// On the 30/min 'standard' bucket that spends the person's entire budget in two page views
+// and then 429s their settings page, their release lists, and any artist page they open.
+// Reported in production 2026-08-16.
+
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+const mocks = vi.hoisted(() => ({
+  mockFrom: vi.fn(),
+  mockCreateClient: vi.fn(),
+  mockCheckRateLimit: vi.fn((): Promise<{ limited: boolean; response?: unknown }> =>
+    Promise.resolve({ limited: false })
+  ),
+  mockGetClientIp: vi.fn(() => '203.0.113.7'),
+}));
+
+vi.mock('../db', () => ({ getClient: () => ({ from: mocks.mockFrom }) }));
+vi.mock('@supabase/supabase-js', () => ({ createClient: mocks.mockCreateClient }));
+vi.mock('../ratelimit', () => ({
+  checkRateLimit: mocks.mockCheckRateLimit,
+  getClientIp: mocks.mockGetClientIp,
+}));
+
+import { handler } from '../collection-art';
+
+const ITEM_ID = '11111111-1111-4111-8111-111111111111';
+
+describe('collection-art rate limiting', () => {
+  beforeEach(() => {
+    vi.resetAllMocks();
+    mocks.mockCheckRateLimit.mockResolvedValue({ limited: false });
+    mocks.mockCreateClient.mockReturnValue({
+      auth: { getUser: vi.fn().mockResolvedValue({ data: { user: null }, error: 'no token' }) },
+    });
+    // Item lookup returns nothing; the request is refused after the rate-limit check, which
+    // is all this file cares about.
+    mocks.mockFrom.mockReturnValue({
+      select: vi.fn(() => ({
+        eq: vi.fn(() => ({ maybeSingle: vi.fn(() => Promise.resolve({ data: null, error: null })) })),
+      })),
+    });
+  });
+
+  it('spends the lenient bucket, not the standard one shared with account endpoints', async () => {
+    await handler({ httpMethod: 'GET', headers: {}, pathParameters: { id: ITEM_ID } });
+
+    expect(mocks.mockCheckRateLimit).toHaveBeenCalledWith(
+      '203.0.113.7',
+      'lenient',
+      expect.anything()
+    );
+  });
+
+  it('still refuses the request when the bucket is spent', async () => {
+    mocks.mockCheckRateLimit.mockResolvedValue({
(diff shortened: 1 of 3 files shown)


brandonlucasgreen (commented):

**Security:** No auth/authz, input-validation, secret-handling, or injection issues found. The rate-limit identifier (`getClientIp`, unchanged) and fail-open behavior on Redis outage are both untouched by this diff.

**Regression:**
- **plausible (low)** — `api/functions/collection-art.ts:96`: moving the proxy from `standard` to `lenient` raises the per-IP ceiling ~4x (30/min·1k/day → 120/min·5k/day). Each cache-miss request triggers a live Bandcamp-credential decrypt + authenticated Subsonic fetch, so this is a real increase in the abuse ceiling for that path, not just a UX fix — worth confirming that's an accepted trade-off.
- **needs-manual-verification (low)** — `api/functions/ratelimit.ts:36`: the `lenient` bucket key is IP-only with no per-endpoint discrimination, and is already shared with `search-suggest`, `feed-releases`, `release-detail`, `check-releases`. This relocates the co-tenancy collision from the `standard` bucket (the original bug) rather than isolating the art proxy — much less likely to starve neighbors at 120/min, but confirm that's intended rather than assumed isolation.

---

brandonlucasgreen:
@brandonlucasgreen

Reviewed the diff (`api/functions/collection-art.ts` line ~93, `api/tsconfig.json`, new test file) plus the full current `collection-art.ts` and `ratelimit.ts` for context, since the change itself is a single-line tier swap.

## Security

No issues found in the changed lines themselves. The things that matter for an image-proxy endpoint hold up:

- **SSRF — confirmed, not applicable here.** The fetched URL is never client-supplied. `itemId` is validated as a UUID (`/^[0-9a-f-]{36}$/i`, line ~98), used only to look up a DB row, and the actual upstream request goes to Bandcamp's Subsonic API using `item.external_id` plus a credential decrypted server-side from `bandcamp_connections`. There's no `fetch(userUrl)` path here, so `isUrlHostnameAllowed()` doesn't apply and its absence isn't a gap. Unaffected by this PR either way.
- **Authorization/IDOR — confirmed, unaffected by this diff.** Access requires `viewerId === item.user_id` OR (`provenance === 'purchased' && !hidden && owner.saved_artists_public === true`), all unchanged in this PR. Item IDs are UUIDs, so brute-force enumeration is infeasible at either 30/min or 120/min — the tier change doesn't meaningfully widen this.
- **Secrets/injection — confirmed n/a.** Credential decryption logic is untouched; the diff is a string-literal tier swap, a comment, and a new test file. Nothing else moved.
- **Plausible, non-blocking:** the rate limiter here is keyed by *requester IP* (`getClientIp`), not by the *item owner* being probed. Each request that isn't served from CDN cache is a real Bandcamp Subsonic call made with that owner's decrypted credential. Raising the ceiling 30→120/min gives a single IP up to 4x more real upstream requests per minute against any one public collection's owner before this PR's caching absorbs repeats. This exposure predates the PR (it existed at 30/min too) and requires the target collection to be public-shared, so I'm not blocking on it — but it's worth a follow-up: consider keying this specific limiter by the *item owner's* `user_id` instead of/in addition to IP, since an attacker spread across a few IPs, or targeting several different victims from one IP, isn't meaningfully throttled per-victim today. **Needs-manual-verification** for whether this is an accepted trade-off already.
- **Needs-manual-verification, out of scope of this diff:** `BASE_HEADERS` sets `Access-Control-Allow-Origin: '*'` (pre-existing, not touched here). Worth a separate look given this endpoint gates access by bearer auth / sharing state, but unrelated to the rate-limit change.

## Regression

- **Confirmed — 120/min is sized correctly for the stated worst case.** `COLLECTION_PAGE_SIZE = 15` is a hard cap (`apps/web/src/utils/collection-grid.ts`), not proportional to collection size, so a 50+ item collection still can't drive more than 15 proxy calls from one page load — pagination bounds the burst regardless of collection size. 120/min leaves ~8 page-loads/min of headroom.
- **Confirmed — misses are cheap and still correctly rate-limited.** The rate-limit check runs before any DB/Subsonic work (confirmed by the new test: `mockFrom` not called when limited), and the "no art" paths (missing `art_url`/connection, decrypt failure, fetch failure) all return 404 with the short 300s `MISS_CACHE_HEADERS` — misses aren't exempted from the budget, and aren't expensive when they do count against it.
- **Confirmed — no API contract drift.** `/api/collection/art` isn't documented in `docs/openapi.yaml` (no matches) and isn't under `/api/v1/`, so there's no public rate-limit doc to update.
- **Confirmed — test/typecheck hygiene.** New test file correctly pins `'lenient'` as the tier and is added to `api/tsconfig.json`'s narrow `include` list per repo convention.

No injection, secret-handling, or state-transition concerns — n/a as expected for a config-only change.

---

brandonlucasgreen:
Superseded by #458, which squash-merged this work into `main` together with #455.

Verified on `main` before closing: `collection-art.ts` is on the `lenient` tier and `collection-art-ratelimit.test.ts` is in place. Closing the PR only, leaving the branch.

PR #458, opened 2026-08-16
Fix the 429 incident: AuthContext request churn, account tier, and the art proxy's budget

brandonlucasgreen:
**Integration branch combining #455 and #457 so the two halves of today's rate-limit incident ship in one production deploy instead of two.** Both merged cleanly with no conflicts; nothing here is new work beyond the merge.

Merging this closes #455 and #457.

## The incident

Settings failing to load, releases not loading, the collection UI apparently missing — on one laptop, while the same account worked fine on a phone. That "works on my phone" detail is what identified it: an **IP-scoped 429**, not a deploy or a code removal. (I separately verified the deployed bundles still contain all the UI, and that no commit since #438 touched those pages.)

## Two independent causes, both needed

**From #455 — request churn and a shared bucket.** `AuthContext` set state on every Supabase session emission, and each emission carries a fresh Session object, so the session's *identity* changed and re-fired every `[session]` effect. `/settings` has six, so one page load fired **twelve** requests and every tab focus fired six more. Account endpoints also shared the 30/min `standard` bucket with artist page views and lookups, so ordinary browsing could 429 the settings page. Adds an `account` tier (60/min, 2000/day), and makes a 429 report as backpressure rather than five separate Sentry errors telling the user to "try again" — the one thing that makes it worse.

**From #457 — the art proxy's burst.** `/api/collection/art` was on `standard` (30/min per IP). A collection grid renders 15 tiles and most miss `art_url`, because only ~28% of a real import matches an Unstream release — so **one page view fires a dozen or more requests** and two spend the whole budget. Everything else on `standard` then 429s, `me-recent-releases` included, which is why releases stopped loading. It moves to `lenient` (120/min), the bucket whose own comment already describes this exact failure for typeahead.

Neither alone is sufficient: #455 doesn't touch `collection-art.ts`, and `me-recent-releases` stays on `standard`, so a collection page view could still 429 the release lists.

## Verification of the combination

Re-run against the merged tree, not just the two branches separately:

- **1,232 API tests** and **759 web unit tests** pass; typecheck clean; `npm run build` green; lint at the pre-existing 71-problem baseline.
- Confirmed both changes coexist: `collection-art.ts` on `lenient`, the `account` tier present in `ratelimit.ts`, and all three relevant test files intact — `collection-art.test.ts` (edited by #455) alongside `collection-art-ratelimit.test.ts` (added by #457), which were deliberately kept separate so they'd never collide.

## Why one PR

Deploy previews are off and the Mac PRs (#456, #439) are cancelled by the ignore script, so they're free. These two are the only outstanding changes that touch `api/` or `apps/web/`, and merging them separately would cost two production deploys at 15 credits each.

### DIFF
--- a/api/functions/__tests__/collection-art-ratelimit.test.ts
+++ b/api/functions/__tests__/collection-art-ratelimit.test.ts
@@ -1,0 +1,73 @@
+// Which rate-limit bucket the cover-art proxy spends.
+//
+// Separate file from collection-art.test.ts on purpose: this pins one property that is easy
+// to regress by copying another endpoint's boilerplate, and a whole page of a fan's
+// collection depends on it.
+//
+// A collection grid renders 15 tiles, and most have no stored art_url — only ~28% of a real
+// import matches an Unstream release — so one page view fires a dozen or more requests here.
+// On the 30/min 'standard' bucket that spends the person's entire budget in two page views
+// and then 429s their settings page, their release lists, and any artist page they open.
+// Reported in production 2026-08-16.
+
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+const mocks = vi.hoisted(() => ({
+  mockFrom: vi.fn(),
+  mockCreateClient: vi.fn(),
+  mockCheckRateLimit: vi.fn((): Promise<{ limited: boolean; response?: unknown }> =>
+    Promise.resolve({ limited: false })
+  ),
+  mockGetClientIp: vi.fn(() => '203.0.113.7'),
+}));
+
+vi.mock('../db', () => ({ getClient: () => ({ from: mocks.mockFrom }) }));
+vi.mock('@supabase/supabase-js', () => ({ createClient: mocks.mockCreateClient }));
+vi.mock('../ratelimit', () => ({
+  checkRateLimit: mocks.mockCheckRateLimit,
+  getClientIp: mocks.mockGetClientIp,
+}));
+
+import { handler } from '../collection-art';
+
+const ITEM_ID = '11111111-1111-4111-8111-111111111111';
+
+describe('collection-art rate limiting', () => {
+  beforeEach(() => {
+    vi.resetAllMocks();
+    mocks.mockCheckRateLimit.mockResolvedValue({ limited: false });
+    mocks.mockCreateClient.mockReturnValue({
+      auth: { getUser: vi.fn().mockResolvedValue({ data: { user: null }, error: 'no token' }) },
+    });
+    // Item lookup returns nothing; the request is refused after the rate-limit check, which
+    // is all this file cares about.
+    mocks.mockFrom.mockReturnValue({
+      select: vi.fn(() => ({
+        eq: vi.fn(() => ({ maybeSingle: vi.fn(() => Promise.resolve({ data: null, error: null })) })),
+      })),
+    });
+  });
+
+  it('spends the lenient bucket, not the standard one shared with account endpoints', async () => {
+    await handler({ httpMethod: 'GET', headers: {}, pathParameters: { id: ITEM_ID } });
+
+    expect(mocks.mockCheckRateLimit).toHaveBeenCalledWith(
+      '203.0.113.7',
+      'lenient',
+      expect.anything()
+    );
+  });
+
+  it('still refuses the request when the bucket is spent', async () => {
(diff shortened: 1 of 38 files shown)


brandonlucasgreen:
@brandonlucasgreen

Reviewed this specifically as a **merge-integration** check — #455 and #457 are being reviewed individually elsewhere, so I did not re-derive their standalone analyses. I confirmed the file sets are disjoint (455: 35 files / +824 -57; 457: 3 files / +81 -1; 458: 38 files / +905 -58, an exact sum), and diffed 458's base (`e472ddf`) against its head (`40666bb`) directly to check the composition claims.

## Security

**Bucket collision / double-application (confirmed, no issue found).** `checkRateLimit`'s tier resolution in `api/functions/ratelimit.ts` is a single mutually-exclusive ternary chain (`strict` → `lenient` → `account` → `standard`), each tier backed by its own `Ratelimit` instance with a distinct Redis prefix (`rl:strict`, `rl:lenient`, `rl:account`, `rl:standard`, plus `rl:daily:*` siblings). `collection-art.ts` calls `checkRateLimit(getClientIp(...), 'lenient', ...)`; the nine account endpoints call `checkRateLimit(await accountRateLimitKey(...), 'account', ...)`. These are separate call sites, separate tiers, separate Redis key namespaces — even if a raw IP identifier were reused across tiers, the per-tier prefix keeps the actual Redis keys disjoint. No overlap, no double-spend, no shared budget between the two fixes. Every handler in the diff calls `checkRateLimit` exactly once.

**No merge-conflict artifacts (confirmed).** #455 and #457 never touch the same file at the line level: 457 edits `collection-art.ts` (source) and adds `collection-art-ratelimit.test.ts`; 455 only adds two mock lines to the pre-existing `collection-art.test.ts`. `api/tsconfig.json` is touched only by 457 (one line, adding the new test to the typecheck include). The merge commit (`40666bb`) is a clean, non-conflicted merge (GitHub reports `mergeable_state: clean`), and the diff's additions/deletions sum exactly to the two source PRs with nothing left over — no duplicated blocks, no silently reverted lines.

No other integration-specific security findings.

## Regression

**Test files present (confirmed).** Comparing base vs. head via `git archive` on `api/functions/__tests__` and `apps/web/tests/unit`: 71→73 API test files, 58→60 web test files, with net +9 API test call-sites and +15 web test call-sites — exactly matching 455 (+7 API: `ratelimit-account-tier.test.ts` ×5, `me-bandcamp.test.ts` ×2; +15 web) plus 457 (+2 API: `collection-art-ratelimit.test.ts`). Nothing was dropped in the merge.

**"1,232 API tests" (plausible).** The +9 call-site delta above is consistent with #457's own reported total (1,225 = baseline 1,223 + 2) plus #455's +7, giving 1,223 + 9 = 1,232 — matches exactly. I didn't execute the suite myself, so this is arithmetic plausibility, not a confirmed run.

**"759 web unit tests" (needs-manual-verification — flagging a real inconsistency).** #457 touches **zero** `apps/web` files (verified via its file list: 3 files, all under `api/`). So the merged branch's web-unit-test total should be numerically identical to #455's own standalone total, since #457 cannot move that number. Static diffing confirms all +15 web call-sites (`AuthContext-session-identity.test.tsx` ×4, `LoginPageEmailValidation.test.tsx` ×5, `BandcampConnect.test.tsx` ×5, `NotificationPreferences.test.tsx` ×1) are attributable to #455's own commits, none to #457. Yet #455 reports **753** web unit tests and this PR reports **759** — a 6-test gap nothing in the diff explains. Worth confirming with an actual `npm run test:web` run against this branch before trusting the combined verification numbers at face value; it may just be a stale/copy-paste figure rather than a real functional gap, but it doesn't reconcile.

**Ordering dependency (confirmed: none).** The two fixes are file-disjoint aside from `api/tsconfig.json` (457-only), and the new `account` tier in `ratelimit.ts` is purely additive to the existing tier-resolution logic — independent of `collection-art.ts`'s `standard`→`lenient` reassignment. Matches #457's own "merge order doesn't matter" claim.

**"Neither alone is sufficient" (confirmed).** #455 never touches `collection-art.ts` (would stay on `standard` if #455 shipped alone), and #457 never touches `me-recent-releases.ts` or introduces the `account` tier (would stay on `standard` if #457 shipped alone). Both fixes are present, distinct, and both required in the merged tree, as claimed.

---

PR #459, opened 2026-08-17
Parallelize dashboard section loads, move saved artists to AuthContext

brandonlucasgreen:
## Summary

Refactors the dashboard to load all sections in parallel rather than serially, eliminating a performance bottleneck where the collection read (the heaviest request on the page) was blocked waiting for artist profile and saved artist fetches to complete. Also moves saved artist state management into `AuthContext` so mutations on the dashboard stay in sync with save buttons elsewhere in the app.

## Key changes

**Dashboard page structure:**
- Removed page-level `loading` flag that gated all three fetches (`/api/artist-auth`, `/api/saved-artists`, `/api/me/recent-releases`)
- Each section now fetches independently on mount and shows its own skeleton, allowing requests to start in parallel
- Extracted claimed artists into `ClaimedArtistsSection` component with its own fetch and error handling
- Extracted saved artists into `SavedArtistsSection` component that reads from `AuthContext` instead of fetching its own copy
- Recent releases fetch moved directly into the page component (no longer blocked by other sections)

**AuthContext enhancements:**
- Added `savedArtists`, `artistsLoaded`, `loadSavedArtists()`, `removeSavedArtist()`, and `setArtistSupported()` to the context
- `SavedArtist` interface now exported and includes `supported` and `supportedAt` fields
- `removeSavedArtist()` now returns a boolean indicating success (for undo UX)
- New `setArtistSupported()` method to toggle the supported mark on a saved artist

**API endpoint consolidation:**
- Introduced `resolveAccountRequest()` in `ratelimit.ts` to return both the rate-limit key and verified user, eliminating duplicate auth checks
- Updated all account endpoints (`me-bandcamp`, `me-recent-releases`, `me-feed-token`, `me-collection`, `me-location`, `me-notifications`, `me-username`, `user-sharing`, `saved-artists`, `saved-artists-sync`) to use `resolveAccountRequest()` instead of calling `authenticateBearerFast()` separately
- Removed redundant inline `authenticateRequest()` functions from each endpoint

**Tests:**
- Added `DashboardPageWaves.test.tsx` to verify that sections fetch in parallel (collection request starts before other sections resolve)
- Added `SavedArtistsSection.test.tsx` to verify mutations go through `AuthContext` rather than direct API calls
- Updated existing endpoint tests to mock `resolveAccountRequest()` instead of inline auth

## Notable implementation details

- The dashboard now renders immediately with skeletons for each section, rather than blocking on the slowest fetch. This is a wall-clock improvement from `max(t1, t2, t3) + t4` (collection blocked by others) to `max(t1, t2, t3, t4)`.
- `SavedArtistsSection` reads from `AuthContext` state rather than maintaining its own copy, ensuring that removing an artist here updates the save buttons on search results and artist pages in the same session.
- `ClaimedArtistsSection` silently hides itself if the user has no claimed profiles (no skeleton shown for empty state), since most users claim nothing.
- The undo toast for removing a saved artist only appears after the server confirms the removal, preventing misdescription if the removal fails and is rolled back.
- `resolveAccountRequest()` is narrower than the old pattern: it returns the verified user alongside the key, so callers don't re-verify the same token. The old `accountRateLimitKey()` remains for any future callers that genuinely only need a bucket name.

https://claude.ai/code/session_01UtrEqz1LBwwCGFVHkkKjKN

### DIFF
--- a/api/functions/__tests__/me-bandcamp.test.ts
+++ b/api/functions/__tests__/me-bandcamp.test.ts
@@ -14,7 +14,12 @@ vi.mock('../db', () => ({ getClient: () => ({ from: mocks.mockFrom }) }));
 vi.mock('@supabase/supabase-js', () => ({ createClient: mocks.mockCreateClient }));
 vi.mock('../ratelimit', () => ({
   // Only a bucket name; the endpoints' own auth is mocked separately.
-  accountRateLimitKey: async () => 'user:test-user',
+  // Mirrors the real helper: deriving the bucket verifies the token, so a request that
+  // carries one resolves to a user and one that doesn't resolves to null.
+  resolveAccountRequest: async (authHeader?: string) =>
+    authHeader
+      ? { key: 'user:test-user', user: { userId: 'user-1', email: 'test@example.com' } }
+      : { key: 'ip:127.0.0.1', user: null },
   checkRateLimit: mocks.mockCheckRateLimit,
   getClientIp: mocks.mockGetClientIp,
 }));
--- a/api/functions/__tests__/me-bandcamp.test.ts
+++ b/api/functions/__tests__/me-bandcamp.test.ts
@@ -12,12 +12,18 @@ const mocks = vi.hoisted(() => ({
 
 vi.mock('../db', () => ({ getClient: () => ({ from: mocks.mockFrom }) }));
 vi.mock('@supabase/supabase-js', () => ({ createClient: mocks.mockCreateClient }));
+/** An Authorization header whose token does not verify — see the ratelimit mock below. */
+const REJECTED_TOKEN = 'Bearer rejected-token';
+
 vi.mock('../ratelimit', () => ({
   // Only a bucket name; the endpoints' own auth is mocked separately.
   // Mirrors the real helper: deriving the bucket verifies the token, so a request that
-  // carries one resolves to a user and one that doesn't resolves to null.
+  // carries a good one resolves to a user. A missing header and the REJECTED_TOKEN sentinel
+  // both resolve to null — the second is how a test says "header present, signature bad",
+  // which is the case an auth regression would actually hide. Treating every truthy header
+  // as authenticated would let such a test pass without exercising anything.
   resolveAccountRequest: async (authHeader?: string) =>
-    authHeader
+    authHeader && authHeader !== REJECTED_TOKEN
       ? { key: 'user:test-user', user: { userId: 'user-1', email: 'test@example.com' } }
       : { key: 'ip:127.0.0.1', user: null },
   checkRateLimit: mocks.mockCheckRateLimit,
@@ -120,6 +126,11 @@ describe('me-bandcamp handler', () => {
       })),
     });
   }
+
+  it('returns 401 when the token was checked and rejected, not just absent', async () => {
+    const res = await handler({ httpMethod: 'GET', headers: { authorization: REJECTED_TOKEN }, body: null });
+    expect(res!.statusCode).toBe(401);
+  });
 
   it('returns 401 when not authenticated', async () => {
     const res = await handler({ httpMethod: 'GET', headers: {}, body: null });
--- a/api/functions/__tests__/me-collection.test.ts
+++ b/api/functions/__tests__/me-collection.test.ts
@@ -18,12 +18,18 @@ vi.mock('../db', async importOriginal => {
   };
 });
 vi.mock('@supabase/supabase-js', () => ({ createClient: mocks.mockCreateClient }));
+/** An Authorization header whose token does not verify — see the ratelimit mock below. */
+const REJECTED_TOKEN = 'Bearer rejected-token';
+
 vi.mock('../ratelimit', () => ({
   // Only a bucket name; the endpoints' own auth is mocked separately.
   // Mirrors the real helper: deriving the bucket verifies the token, so a request that
-  // carries one resolves to a user and one that doesn't resolves to null.
+  // carries a good one resolves to a user. A missing header and the REJECTED_TOKEN sentinel
+  // both resolve to null — the second is how a test says "header present, signature bad",
+  // which is the case an auth regression would actually hide. Treating every truthy header
+  // as authenticated would let such a test pass without exercising anything.
   resolveAccountRequest: async (authHeader?: string) =>
-    authHeader
+    authHeader && authHeader !== REJECTED_TOKEN
       ? { key: 'user:test-user', user: { userId: 'user-1', email: 'test@example.com' } }
       : { key: 'ip:127.0.0.1', user: null },
   checkRateLimit: mocks.mockCheckRateLimit,
@@ -54,6 +60,11 @@ describe('me-collection handler', () => {
         }),
       },
     });
+  });
+
+  it('returns 401 when the token was checked and rejected, not just absent', async () => {
+    const res = await handler({ httpMethod: 'GET', headers: { authorization: REJECTED_TOKEN }, body: null });
+    expect(res!.statusCode).toBe(401);
   });
 
   it('returns 401 when not authenticated', async () => {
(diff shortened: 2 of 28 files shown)


brandonlucasgreen:
**Automated PR Review (GLM-5.1)**

## Security

1. **`authenticateBearerFast` token extraction contract** — The old `authenticateRequest` explicitly checked `authHeader?.startsWith('Bearer ')` and extracted the token with `slice(7)` before calling Supabase. The new code passes the raw header to `authenticateBearerFast`. If that helper doesn't perform the same prefix validation and extraction, auth bypass or parse errors are possible.
   — **needs-manual-verification**

2. **POST `/me/bandcamp` rate-limits by IP before authenticating** — Authenticated users on shared IPs (NAT, VPN, office) now share a single rate-limit bucket for POST. Previously they'd authenticate first and get per-user buckets. The comment says this is intentional to block unauthenticated floods, but it does change the rate-limit contract for legitimate authenticated POSTs.
   — **confirmed**

## Regression Risk

3. **Test mocks treat any truthy `authHeader` as valid** — Several test files (`me-bandcamp`, `me-collection`, `me-location`, `me-notifications`, `me-username`, `user-sharing`) use an inline mock where any truthy `authHeader` resolves to a valid user. Tests that send an `Authorization` header to exercise a non-401 path will now always authenticate, even with a syntactically invalid or forged token. 401 test paths must explicitly override the mock to return `user: null`. If any 401 test was removed or not updated, it would silently pass. The `me-recent-releases` and `saved-artists-reporting` mocks delegate to a controllable `vi.fn()`, which is safer.
   — **plausible**

4. **`resolveAccountRequest` error path: `user: null` with an IP key** — When `authenticateBearerFast` throws (JWKS unreachable, network partition), the request falls through to `{ key: 'ip:…', user: null }`. Every endpoint checks `account.user` and returns 401, so the auth decision is safe. However, the old `accountRateLimitKey` + separate auth flow would also have produced a 401, so this is consistent. Noting for awareness: during a JWKS outage, all authenticated requests become 401s with no per-user rate-limit accounting — the same as before, but now combined into a single code path.
   — **confirmed** (consistent, not a regression — just calling out the single-point-of-failure consolidation)

5. **`SavedArtistsSection` depends on `artistsLoaded` from `AuthContext` instead of local loading state** — If `AuthContext` enters an error state where `artistsLoaded` never becomes `true` and `savedArtists` remains `[]`, the section shows a skeleton forever with no error UI. The old component had its own fetch and could display an error message. Whether this is a regression depends on AuthContext's error handling.
   — **needs-manual-verification**

6. **Undo now also restores `supported` status** — `SavedArtistsSection` calls both `saveArtist` and `setArtistSupported(id, true)` on undo. If `setArtistSupported` fails after `saveArtist` succeeds, the artist is re-saved but the support flag is lost. This is an improvement over the old behavior (which didn't restore support at all), but the two calls are not atomic.
   — **plausible**

cc @brandonlucasgreen

brandonlucasgreen:
Thanks — four of the six held and are fixed in 687e138. Two don't, and I want to say why rather than quietly skip them.

## Fixed

**3 — mocks treat any truthy header as valid.** Right, and the gap is the interesting half: the mock could only express "no header", not "header present, signature bad". A `REJECTED_TOKEN` sentinel now resolves to no user, and each of the six files asserts a 401 for a token that was *checked and refused*. For the record, none of the existing 401 tests was a false pass — all six send `headers: {}` — so this closes the hole rather than fixing a break.

**4 — the error path.** Held, but not quite where you put it. A JWKS outage does **not** produce 401s: `verifyJwtLocally` returns `'unavailable'` and `authenticateBearerFast` falls back to the auth server, so it degrades to the old round trip. The 401 needs `authenticateBearerFast` to *throw*, i.e. that fallback failed too — Supabase Auth unreachable. `resolveAccountRequest` was swallowing that, where the endpoints' own unguarded `getUser` surfaced it as a 500 that reached Sentry. It's now reported, so a wave of these reads as an outage rather than a wave of expired sessions. Same rule as "never cache uncertainty": "couldn't check" is not "not signed in".

**5 — indefinite skeleton.** Confirmed a regression I introduced. `loadSavedArtists` reports a failure and returns *without* setting `artistsLoaded` — correct, since an empty list must never render as "you have saved nobody" when we don't know — but the section read only `artistsLoaded`, so a failure shimmered forever where the old page-level error used to speak. `AuthContext` now exposes `savedArtistsError` and the section offers a retry.

**6 — non-atomic undo.** The two writes can't be atomic: they're two actions through an endpoint that takes one at a time. So what matters is that a half-done restore says *which* half. `saveArtist` now reports whether it stuck (matching `removeSavedArtist`), and undo checks that before touching the mark. The bug was worse than described — a *failed* save still ran `setArtistSupported` and could report "couldn't restore the supported mark" for a restore that never happened.

## Not changing

**1 — Bearer prefix.** Verified, as requested. `authenticateBearerFast` opens with exactly what the inline helpers had:

```ts
if (!authHeader?.startsWith('Bearer ')) return null;
const token = authHeader.slice(7);
```

Same validation, same extraction. No bypass and no parse path.

**2 — POST `/me/bandcamp` rate-limit contract.** This one's premise is wrong. POST has never been per-user: both before and after this PR it is `checkRateLimit(ip, 'strict', …)`. The diff moves *where* that call sits, not its key or its tier. It's IP-keyed on purpose, and the existing comment says why — keying a credential-checking oracle by user would let an attacker buy more guessing budget by holding more Unstream accounts. What did change is that GET/DELETE now resolve the user before their limit so they can keep their per-user bucket, while POST still limits *before* the token is examined at all, so an unauthenticated flood can't reach the auth server. That ordering split is the one place the two differ and it's commented as such.

`npm run verify` green: 1242 API + 772 web unit tests.

---

brandonlucasgreen:
## Re: the semantic-revert warning

Intentional cleanup — but the warning deserves a real answer rather than "ignore it", so here's the check I ran.

**The cause it suggests doesn't apply.** The branch isn't based on a pre-merge commit: its branch point is `4c4fc5a`, which is `origin/main`'s tip, and both flagged PRs are ancestors — #458 as `84e181b` (`git merge-base --is-ancestor 84e181b HEAD` passes) and #438 as `27c49ac`. Nothing here predates them.

**What the deleted lines are.** The *implementation* #458 and #438 introduced, replaced by a narrower one — the per-endpoint inline `authenticateRequest` helpers (`createClient` → `auth.getUser`), their `accountRateLimitKey` call sites, and the test mock lines that stood in for them. Not the fixes.

**Each of #458's fixes, checked individually:**

| #458 fix | Status |
|---|---|
| AuthContext ignores same-token emissions | `applySession` untouched — the diff doesn't contain `tokenUnchanged` or the `access_token ===` guard |
| Account endpoints get their own rate-limit bucket | 12 `'account'` literals across the 12 account files, byte-identical to main |
| Account limit keyed by user, not IP | Preserved by `resolveAccountRequest`; still asserted in `ratelimit-account-tier.test.ts` ("gives two users on one IP separate budgets", "falls back to the IP when the token does not verify") |
| Art proxy on the `'lenient'` tier | `collection-art.ts` unchanged |
| 429s stay out of Sentry in the settings panels | Those components untouched. AuthContext now *also* uses `RATE_LIMIT_MESSAGE` for a 429 on the saved-artists load — extending that fix, not undoing it |
| LoginPage email validation | Untouched |

`me-password.ts` is unchanged too, so the strict per-IP limit on the credential path is intact.

**Test coverage went up, not down.** The deleted test lines are the `accountRateLimitKey: async () => 'user:test-user'` mocks. Their replacements can express a token that was *checked and rejected*, not just absent — and six new 401 tests exercise exactly that, plus three for `resolveAccountRequest` and three for the review findings. 1242 API + 772 web unit tests, up from 1232 + 759 on main.

One note for whoever reads this thread next: the `check` job exits 0 whether or not it warns, so a green checkmark on it doesn't mean the warning above isn't there. I misread that myself earlier in this PR.

---

PR #465, opened 2026-08-21
Add the missing /api/me/listening endpoint for Apple Music library sync

brandonlucasgreen:
## Why

The Mac app's Apple Music library import (Support Loop Step 2, #439) POSTs `{source: "apple_music", signals: [{artistName, playCount}]}` to `/api/me/listening` — but the endpoint never shipped. There was no `me-listening` function, no redirect, and no reference to `listening_signals` anywhere in `api/`, so every sync failed with "Couldn't sync your library to Unstream (error 404)". The `listening_signals` table has existed since the bandcamp-collection migration (with `UNIQUE (user_id, source, artist_name)`), waiting for a writer.

## What

- **`api/functions/me-listening.ts`** — follows the `me-*` pattern (auth + rate limiting via `resolveAccountRequest`, local JWT verify, `account` tier).
  - **POST** validates the payload (source must be `apple_music` — the only writer today; malformed types 400, blank artist names are skipped so a library quirk can't brick the sync), dedupes repeated names keeping the highest count, then **diffs against stored play counts before writing**: a re-import of a library that gained 3 plays writes 3 rows, not thousands. This matters because rewriting a row dirties a page whether or not values changed — and the Supabase free tier is under a disk-IO budget warning. Writes go through a single chunked upsert (500 rows per request, same as `bandcamp-sync-background.ts`) on the unique key — never a per-signal loop.
  - The existing-signals read pages via `readAllPages` — a real library can exceed PostgREST's silent 1,000-row cap, and a truncated read would make page-two artists look new and get rewritten unchanged forever.
  - **DELETE** removes the user's `apple_music` rows in one statement (the app's "remove my data" action).
- **`netlify.toml`** — adds the `/api/me/listening` redirect.
- **`api/tsconfig.json`** — the function and its test are in the typecheck include, per the account-endpoint convention.
- **Test** (16 cases) covers the auth boundary (rejected token vs absent), validation, the changed-rows-only diff, 500-row chunking, the >1,000-row paged read, dedup, and both failure paths surfacing as 500s.

## Verification

`npm run verify` passes: typecheck + API function tests + web unit tests.

No client change needed — the Mac app already targets this URL and treats any 2xx as success.

### DIFF
--- a/api/functions/__tests__/me-listening.test.ts
+++ b/api/functions/__tests__/me-listening.test.ts
@@ -1,0 +1,263 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+const mocks = vi.hoisted(() => ({
+  mockFrom: vi.fn(),
+  mockCheckRateLimit: vi.fn(() => Promise.resolve({ limited: false })),
+  mockGetClientIp: vi.fn(() => '127.0.0.1'),
+}));
+
+// getClient is mocked; readAllPages stays real so the paged read of existing
+// signals is exercised, not stubbed (same pattern as bandcamp-sync.test.ts).
+vi.mock('../db', async importOriginal => {
+  const original = await importOriginal<typeof import('../db')>();
+  return { ...original, getClient: () => ({ from: mocks.mockFrom }) };
+});
+
+/** An Authorization header whose token does not verify — see the ratelimit mock below. */
+const REJECTED_TOKEN = 'Bearer rejected-token';
+
+vi.mock('../ratelimit', () => ({
+  // Mirrors the real helper: deriving the bucket verifies the token. A missing header and
+  // the REJECTED_TOKEN sentinel both resolve to null — the second is how a test says
+  // "header present, signature bad", the case an auth regression would actually hide.
+  resolveAccountRequest: async (authHeader?: string) =>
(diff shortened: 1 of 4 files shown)


brandonlucasgreen:
**Context discovered after opening this:** the endpoint was actually built once before, in #454 — but #454 was merged into the integration branch `batch/aug-16`, which never got merged to main. GitHub shows it "merged," so it looked shipped; none of it ever deployed. That's the real cause of the 5-day 404.

This PR intentionally does **not** resurrect the stranded implementation's POST/DELETE: that version predates the 429 fixes (#458) and the disk-IO work (#463/#464) — it uses the shared `standard` IP bucket, a `getUser` auth-server round trip, and rewrites the entire library on every sync with no diff. This one uses `resolveAccountRequest`, the `account` tier, and writes only changed rows.

Still stranded on `origin/batch/aug-16` (`c498b6d`) and worth a follow-up PR onto current main:
- the **GET gap report** (`/api/me/listening` GET — plays minus supported/collection artists), the product feature the import feeds
- the **OQ6 reversal**: imports stop writing `saved_artists` + the feed union in `db.ts`. Main's Bandcamp sync still auto-saves imported artists today, which was explicitly decided against.

PR #467, opened 2026-08-21
Rescue the rest of #454: the gap report, and collection ≠ saved artists

brandonlucasgreen:
#454 was merged into the `batch/aug-16` integration branch on 2026-08-16, and that branch never reached main. Its squash commit `c498b6d` is still stranded there. #465 has already rebuilt the POST/DELETE half of `/api/me/listening` on current patterns; this PR brings the other two pieces over, **reimplemented against current main** rather than cherry-picked — `batch/aug-16` is 9 commits behind and merging it would revert #458, #463 and #464.

## 1. `GET /api/me/listening` — the private gap report

Artists you play a lot and have never paid, ranked by plays, capped at 100. It subtracts artists you marked supported and artists you own a record by, comparing on the derived slug so the two halves can't disagree about who counts as covered.

- **Auth is #465's, not the stranded version's.** `resolveAccountRequest` (local JWT verification) and the `account` rate-limit tier, so the token is verified once and GET sits behind the same gate as POST and DELETE. The stranded version used a `getUser` round trip and the shared `standard` IP bucket, which is what caused the 429 incident.
- **Never public.** It's derived from one user's own rows and returned only to that authenticated user; `listening_signals` has no INSERT/UPDATE policy and the gap is deliberately absent from the collection page.
- **A failed signals read is a 500, not an empty gap.** An empty gap would read as "you already support everyone you listen to" — the most misleading answer this endpoint can give. The two exclusion reads degrade instead: losing them costs a row for someone already supported, where blanking the report is the bigger wrong.

One deliberate improvement on the stranded code: a supported artist is matched on `artistSlug(artist_name)` as well as the stored slug. A row saved from a search result carries a synthetic slug (`sufjanstevens`) that the canonical one never equals, so slug-only matching would tell you to go support someone you already support. These are the same two candidates `savedArtistIdsForUser` resolves in `db.ts`, for the same reason.

## 2. The OQ6 reversal

> "Collection Imports should always add to the collection NOT saved artists. However, you should get upcoming/recent release information from artists both in the collection and the saved artists list." — 2026-08-16

**Main has been auto-writing `saved_artists` on every import this whole time.** `markArtistsSupported` is gone from `bandcamp-sync-background.ts`; an import now writes `collection_items` and stops.

Their releases still reach the feed: `getFeedReleasesForUser` unions saved ∪ collection, resolving collection artists through a matched release's `artist_id` or the derived slug. ~72% of a real import matches no release, so the slug path is the common one, not a fallback. A failed collection read narrows the feed rather than blanking it.

Removing the mark also drops the cached artists select to `id, name` — matching an artist is now only how we find their releases.

## ⚠️ Expect `semantic-revert-check` to flag this

It should. Deleting `markArtistsSupported` removes #445's dedup fix (looking a saved row up by `artist_id` as well as slug, so an import couldn't duplicate a row or resurrect a tombstone). That fix is correctly deleted rather than reverted: it was protecting a write that no longer happens. Nothing else calls the function.

## Not in scope

The 35 `saved_artists` rows the live import already created — Brandon: *"don't worry about deleting the 35 rows, that was only in my own account. i can clean that up later myself."* This stops new ones; it doesn't clean up old ones.

## Testing

`npm run verify` green — 1300 API tests, 781 web tests, typecheck clean.

New coverage: the gap's ranking, both exclusion paths (including the synthetic-slug case), cross-source merging, the 100 cap, and the read-failure asymmetry; the sync asserting it never touches `saved_artists`; and the feed union, ported from the stranded `feed-collection-artists.test.ts`. Every new assertion was checked against a deliberately broken implementation — dropping the collection half of the union, treating a failed read as empty, and raising the cap each fail exactly the test that names them.

`saved-artist-resolution.test.ts`'s fake client gained a `range` method so the feed's new collection read can answer; without it every test in that file would have fallen into the catch block and returned `[]`.

### DIFF
--- a/api/functions/__tests__/bandcamp-sync.test.ts
+++ b/api/functions/__tests__/bandcamp-sync.test.ts
@@ -45,16 +45,15 @@ function internalEvent(body: unknown) {
 interface TableMocks {
   connectionRow?: Record<string, unknown> | null;
   releases?: Record<string, unknown>[];
-  savedRows?: Record<string, unknown>[];
 }
 
-// Routes supabase table calls; records connection updates, item upserts, and
-// saved_artists writes (the auto-mark-supported side effect).
-function setupDb({ connectionRow = null, releases = [], savedRows = [] }: TableMocks) {
+// Routes supabase table calls; records connection updates and item upserts.
+//
+// `saved_artists` deliberately has no branch: the import stopped writing it on 2026-08-16, so
+// reaching for that table now throws "unexpected table" rather than being quietly absorbed.
+function setupDb({ connectionRow = null, releases = [] }: TableMocks) {
   const connectionUpdates: Record<string, unknown>[] = [];
   const upsertBatches: Record<string, unknown>[][] = [];
-  const savedInserts: Record<string, unknown>[][] = [];
-  const savedUpdates: { patch: Record<string, unknown>; column: string; values: unknown }[] = [];
 
   mocks.mockFrom.mockImplementation((table: string) => {
     if (table === 'bandcamp_connections') {
@@ -87,43 +86,13 @@ function setupDb({ connectionRow = null, releases = [], savedRows = [] }: TableM
         })),
       };
     }
-    if (table === 'saved_artists') {
-      return {
-        select: vi.fn(() => ({
-          eq: vi.fn(() => ({
-            // Honours the column actually queried. Without this the mock answers the
-            // artist_id read and the artist_slug read identically, and a row filed under a
-            // synthetic slug would look findable by slug when in production it isn't — the
-            // exact blindness the dedup fix exists to close.
-            in: vi.fn((column: string, values: unknown[]) =>
-              Promise.resolve({
-                data: savedRows.filter(row => values.includes(row[column])),
-                error: null,
-              })
-            ),
-          })),
-        })),
-        upsert: vi.fn((rows: Record<string, unknown>[]) => {
-          savedInserts.push(rows);
-          return Promise.resolve({ error: null });
-        }),
-        update: vi.fn((patch: Record<string, unknown>) => ({
-          eq: vi.fn(() => ({
-            in: vi.fn((column: string, values: unknown) => {
-              savedUpdates.push({ patch, column, values });
-              return Promise.resolve({ error: null });
-            }),
-          })),
-        })),
-      };
-    }
     throw new Error(`unexpected table ${table}`);
   });
 
-  return { connectionUpdates, upsertBatches, savedInserts, savedUpdates };
+  return { connectionUpdates, upsertBatches };
 }
 
-const SUFJAN = { id: 'artist-1', slug: 'sufjan-stevens', name: 'Sufjan Stevens', image_url: 'https://img/s.jpg' };
+const SUFJAN = { id: 'artist-1', name: 'Sufjan Stevens' };
 
 describe('bandcamp-sync-background handler', () => {
   beforeEach(() => {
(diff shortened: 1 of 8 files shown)


brandonlucasgreen:
## Semantic-revert warnings: verified, one genuine deletion (intended), the rest are noise

Checked all four flagged files against the seven named PRs. **No accidental revert.** One deletion is real and deliberate — #445's — and I should have called it out louder in the description.

### Why it's so noisy

`api/functions/__tests__/bandcamp-sync.test.ts` is shared by **six of the seven** flagged PRs. Deleting one `describe` block from it blames all six at once. The bot counts deleted lines per file and blames whoever last touched them; it can't tell a removal from a revert.

The suggested cause doesn't apply here — the branch is current:

```
git rev-list --count HEAD..origin/main   ->  0
```

So this is the in-place-edit / deliberate-removal mode, not a stale base. Nothing to rebase.

### #463 and #464 (disk I/O) — fully intact

These were the ones worth worrying about, since I edited the exact block #464 added. Token counts, `origin/main` vs branch, in `bandcamp-sync-background.ts`:

| token | main | branch |
|---|---|---|
| `cacheGetOrFetch` | 2 | 2 |
| `collection:artist-name-map` | 1 | 1 |
| `3600` (cache TTL) | 1 | 1 |
| `read => read.ok` (never cache a failed read) | 1 | 1 |
| `sameInstant` | 2 | 2 |
| `toWrite` (the changed-rows diff) | 5 | 5 |
| `readAllPages` | 3 | 3 |

I narrowed the *columns* inside #464's cached read (`id, slug, name, image_url` → `id, name`) because the dropped fields only existed for `markArtistsSupported`. The caching, the TTL and the don't-cache-failures predicate are untouched — the change makes that read smaller, not less cached.

`re-sync write churn`, #463's six-test describe block, is untouched and still passing.

### #450, #449, #448, #438 — untouched

| token | main | branch |
|---|---|---|
| `resolveCollectionArtists` / `resolveArtistsSafely` | 2 / 3 | 2 / 3 |
| `SubsonicError` / `isAuthFailure` | 3 / 1 | 3 / 1 |
| `decryptCredential` | 2 | 2 |
| `releaseMatchKey` | 3 | 3 |
| `release_id` / `art_ref` | 7 / 5 | 7 / 5 |

Their tests survive by name: *"still resolves stored artists when Bandcamp refuses to hand over the collection"*, *"still reports a successful import when artist resolution fails"*, *"looks for the artists behind unmatched items only after the import is recorded"* (#450); *"tells the user to reconnect when the stored credential is rejected"*, *"marks the credential unusable when decryption fails"* (#448); *"never erases a release link the linker set after the last sync"* (#449).

### #445 — a real deletion, and the point of the PR

`markArtistsSupported` is gone: `markArtistsSupported` 2→0, `SAVED_SELECT` 3→0, `supported_at` 3→0.

#445 fixed that function to look a saved row up by `artist_id` as well as slug, so an import couldn't duplicate a saved artist or resurrect a tombstone. That fix was protecting a write that **no longer happens** — the OQ6 reversal is precisely "stop writing `saved_artists` on import." It's deleted rather than reverted, and nothing else called it:

```
git grep saved_artists -- api/functions/bandcamp-sync-background.ts \
                           api/functions/collection-matching.ts \
                           api/functions/me-bandcamp.ts     ->  no matches
```

The guarantee #445 protected is now structural instead of defensive: an import can't duplicate a saved row or resurrect a tombstone because it doesn't touch the table at all. `setupDb` in the test file deliberately has no `saved_artists` branch any more, so a regression that re-adds the write throws `unexpected table saved_artists` rather than passing quietly.

### Test counts

`bandcamp-sync.test.ts` goes 24 → 17, which is a real drop, so here are the removed names:

- saves a matched artist as supported when they have no saved row
- upgrades an existing unsupported row instead of inserting
- leaves an already-supported row untouched (original supported_at preserved)
- never resurrects a tombstoned row — permanent dismissal sticks across re-syncs
- upgrades a row saved under a synthetic slug rather than duplicating the artist
- never resurrects a tombstone filed under a synthetic slug
- supports the live row when a tombstone for the same artist sits beside it
- marks nothing for unmatched artists

All eight test `markArtistsSupported`. They're replaced by one test asserting the write never happens, plus the feed union tests that keep the other half of the promise.

Across the PR: `bandcamp-sync.test.ts` −7, `me-listening.test.ts` 16 → 30, `feed-collection-artists.test.ts` +5 new. **Net +12**, and `npm run verify` is green (1300 API, 781 web).

### The two 1-line flags against #465

Both in-place edits, both supersets: `'POST, DELETE, OPTIONS'` → `'GET, POST, DELETE, OPTIONS'` in the CORS header, and the "returns 404 for other methods" test switched from `GET` to `PUT` because GET is now a real route. #465's POST and DELETE paths are otherwise byte-identical.

